PR #7, opened 2019-12-17
Vendoring changes and changes pertaining to the etcd statefulset endpoint in probes

georgekuruvillak:
This PR contains changes that:
* fixed vendoring from Gardener perspective.
* changed the template for statefulset pod endpoint so that it can be used in tls server certificates.


Makefile
```diff
@@ -84,7 +84,7 @@ docker-push:
 # download controller-gen if necessary
 controller-gen:
 ifeq (, $(shell which controller-gen))
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.0-beta.2
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.4
```
ggaurav10: The CRD in path `./config/crd/bases/` doesn't seem to be generated using this version.
Please keep in mind that the CRD generated using this version will not be compatible with k8s versions < 1.15 mainly because of an extra added field `.spec.preserveUnknownFields`

PR #9, opened 2020-02-06
Error handling in case of secrets being absent

georgekuruvillak:
This PR adds the logic to handle error cases where the store secrets are absent as in case of etcd-events.


georgekuruvillak:
@amshuman-kr

controllers/etcd_controller.go
```diff
@@ -735,83 +727,110 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
+	if etcd.Spec.Backup.Store != nil {
+		storeSecret := corev1.Secret{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{
+			Name:      etcd.Spec.Backup.Store.SecretRef.Name,
+			Namespace: etcd.Namespace,
+		}, &storeSecret)
+		if err != nil && !errors.IsNotFound(err) {
 			return err
 		}
+		if err == nil {
```
amshuman-kr: What if the secret is not found?

controllers/etcd_controller.go
```diff
@@ -735,83 +727,110 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
-			logger.Infof("Adding finalizer (%s) for secret %s by etcd (%s)", FinalizerName, clientSecret.GetName(), etcd.Name)
-			clientSecretCopy := clientSecret.DeepCopy()
-			finalizers.Insert(FinalizerName)
-			clientSecretCopy.Finalizers = finalizers.UnsortedList()
-			if err := r.Update(context.TODO(), clientSecretCopy); err != nil {
-				return err
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		if err == nil {
```
amshuman-kr: Same as above.

controllers/etcd_controller.go
```diff
@@ -735,83 +727,110 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
-			logger.Infof("Adding finalizer (%s) for secret %s by etcd (%s)", FinalizerName, serverSecret.GetName(), etcd.Name)
-			serverSecretCopy := serverSecret.DeepCopy()
-			finalizers.Insert(FinalizerName)
-			serverSecretCopy.Finalizers = finalizers.UnsortedList()
-			if err := r.Update(context.TODO(), serverSecretCopy); err != nil {
-				return err
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		if err == nil {
```
amshuman-kr: Same as above.

controllers/etcd_controller.go
```diff
@@ -735,83 +727,110 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
+	if etcd.Spec.Backup.Store != nil {
+		storeSecret := corev1.Secret{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{
+			Name:      etcd.Spec.Backup.Store.SecretRef.Name,
+			Namespace: etcd.Namespace,
+		}, &storeSecret)
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		if err == nil {
```
amshuman-kr: Same as above.

controllers/etcd_controller.go
georgekuruvillak: It would fail at statefulset creation as pod will not be able to mount the secret. Shall I error out here?

api/v1alpha1/etcd_types.go
```diff
@@ -42,19 +42,18 @@ type MetricsLevel string
...
 // StorageProvider defines the type of object store provider for storing backups.
-// +kubebuilder:validation:Enum=aws;gcp;os;azure;alicloud
 type StorageProvider string
 
 // StoreSpec defines parameters related to ObjectStore persisting backups
 type StoreSpec struct {
-	// +optional
-	Container string `json:"container,omitempty"`
+	// +required
+	Container string `json:"container"`
```
swapnilgm: Why are we making them required?  They were optional considering the case on no backup i.e. empty provider and 'Local' provider for development case.

controllers/etcd_controller.go
```diff
@@ -811,41 +819,36 @@ func (r *EtcdReconciler) removeFinalizersToDependantSecrets(etcd *druidv1alpha1.
 	}
 	if etcd.Spec.Etcd.TLS != nil {
 		clientSecret := corev1.Secret{}
-		err = r.Client.Get(context.TODO(), types.NamespacedName{
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{
 			Name:      etcd.Spec.Etcd.TLS.ClientTLSSecretRef.Name,
 			Namespace: etcd.Namespace,
-		}, &clientSecret)
-		if err != nil && !errors.IsNotFound(err) {
+		}, &clientSecret); err != nil {
```
swapnilgm: It should return ` nil`  is err is err is 'NotFound.'

controllers/etcd_controller.go
```diff
@@ -789,16 +799,14 @@ func (r *EtcdReconciler) addFinalizersToDependantSecrets(etcd *druidv1alpha1.Etc
...
-	if err != nil && !errors.IsNotFound(err) {
-		return err
-	}
-	if err == nil {
+	if etcd.Spec.Backup.Store != nil {
+		storeSecret := corev1.Secret{}
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{
+			Name:      etcd.Spec.Backup.Store.SecretRef.Name,
+			Namespace: etcd.Namespace,
+		}, &storeSecret); err != nil {
```
swapnilgm: It should return ` nil`  is err is err is 'NotFound.'

controllers/etcd_controller.go
```diff
@@ -811,41 +819,36 @@ func (r *EtcdReconciler) removeFinalizersToDependantSecrets(etcd *druidv1alpha1.
...
-				if err := r.Update(context.TODO(), clientSecretCopy); err != nil {
-					return err
-				}
+		if finalizers := sets.NewString(clientSecret.Finalizers...); finalizers.Has(FinalizerName) {
+			logger.Infof("Removing finalizer (%s) from secret %s", FinalizerName, clientSecret.GetName())
+			clientSecretCopy := clientSecret.DeepCopy()
+			finalizers.Delete(FinalizerName)
+			clientSecretCopy.Finalizers = finalizers.UnsortedList()
+			if err := r.Update(context.TODO(), clientSecretCopy); err != nil {
+				return err
```
swapnilgm: It should return ` nil`  is err is err is 'NotFound.'

controllers/etcd_controller.go
```diff
@@ -811,41 +819,36 @@ func (r *EtcdReconciler) removeFinalizersToDependantSecrets(etcd *druidv1alpha1.
...
 		}
+
 		serverSecret := corev1.Secret{}
-		err = r.Client.Get(context.TODO(), types.NamespacedName{
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{
 			Name:      etcd.Spec.Etcd.TLS.ServerTLSSecretRef.Name,
 			Namespace: etcd.Namespace,
-		}, &serverSecret)
-		if err != nil && !errors.IsNotFound(err) {
+		}, &serverSecret); err != nil {
```
swapnilgm: It should return ` nil`  is err is err is 'NotFound.'

controllers/etcd_controller.go
```diff
@@ -811,41 +819,36 @@ func (r *EtcdReconciler) removeFinalizersToDependantSecrets(etcd *druidv1alpha1.
...
-				serverSecretCopy.Finalizers = finalizers.UnsortedList()
-				if err := r.Update(context.TODO(), serverSecretCopy); err != nil {
-					return err
-				}
+		if finalizers := sets.NewString(serverSecret.Finalizers...); finalizers.Has(FinalizerName) {
+			logger.Infof("Removing finalizer (%s) from secret %s", FinalizerName, serverSecret.GetName())
+			serverSecretCopy := serverSecret.DeepCopy()
+			finalizers.Delete(FinalizerName)
+			serverSecretCopy.Finalizers = finalizers.UnsortedList()
+			if err := r.Update(context.TODO(), serverSecretCopy); err != nil {
```
swapnilgm: It should return ` nil`  is err is err is 'NotFound.'

controllers/etcd_controller.go
georgekuruvillak: I cannot return nil here if its not found as it will not remove the remaining finalizers.

controllers/etcd_controller.go
georgekuruvillak: same here

controllers/etcd_controller.go
swapnilgm: Ahh. Right. But you will have to add check to ignore not found error.

controllers/etcd_controller.go
swapnilgm: Again, right but you will have to handle notFound case.

api/v1alpha1/etcd_types.go
georgekuruvillak: changed.

controllers/etcd_controller.go
georgekuruvillak: updated.

Issue #23, opened 2020-03-06
[Feature] Control etcd + backup-restore sidecar versions


rfranzke:
**Feature (What you would like to be added):**
The etcd-druid should control the both the versions for etcd + backup-restore sidecar.

**Motivation (Why is this needed?):**
It controls the manifests + configuration for the statefulset and the used versions must fit to it. Hence, it makes sense to control them.

**Approach/Hint to the implement solution (optional):**
Please use the image vector approach (https://github.com/gardener/gardener/blob/master/charts/images.yaml) with the use of https://github.com/gardener/gardener/tree/master/pkg/utils/imagevector.
It must be possible to overwrite the image vector during deployment time.

PR #24, opened 2020-03-12
Image vector change

georgekuruvillak:
This PR has changes related to three aspects:
* Restart pods for etcd statefulsets in crashloopbackoff once statefulsets are synced.
* Implement image vector support
* Adopt resource recommendations on statefulset from hvpa, if any.

Fixes #
#23

**Release note**:

```improvement operator

```


api/v1alpha1/etcd_types.go
```diff
@@ -113,8 +112,8 @@ type EtcdConfig struct {
 	// +optional
 	ClientPort *int `json:"clientPort,omitempty"`
 	// Image defines the etcd container image and tag
-	// +required
-	Image string `json:"image"`
+	// +optional
+	Image *string `json:"image"`
```
rfranzke:
```suggestion
	Image *string `json:"image,omitempty"`
```

api/v1alpha1/etcd_types.go
```diff
@@ -74,8 +73,8 @@ type BackupSpec struct {
 	// +optional
 	TLS *TLSConfig `json:"tls,omitempty"`
 	// Image defines the etcd container image and tag
-	// +required
-	Image string `json:"image"`
+	// +optional
+	Image *string `json:"image"`
```
rfranzke:
```suggestion
	Image *string `json:"image,omitempty"`
```

controllers/etcd_controller.go
```diff
@@ -577,6 +635,18 @@ func (r *EtcdReconciler) syncStatefulSetSpec(ss *appsv1.StatefulSet, cm *corev1.
 	ssCopy := ss.DeepCopy()
 	ssCopy.Spec.Replicas = decoded.Spec.Replicas
 	ssCopy.Spec.UpdateStrategy = decoded.Spec.UpdateStrategy
+
+	// Applying suggestions from
+	containers := getContainerMapFromPodTemplateSpec(ssCopy.Spec.Template.Spec)
```
swapnilgm: I didn't get this? what are we trying to achieve here? i assumed this was backward compatibility code first which is trying to sync `resource` requirement from existing sts to the sts which will be deployed as reconcile part of `etcd` crd resource.  But ideally that should be `etcd` crd deployers job and not of etcd-druid.  On the same line, since here etcd-druid is trying to do it on its own, what will happen to resource configured in etcd crd. Those are not propagated to etcd then.
I might have still missed something from druid working related to backward compatibility in regards to gardener. Please correct me in that case.

main.go
```diff
@@ -71,6 +71,11 @@ func main() {
 		setupLog.Error(err, "unable to initialize controller with chart renderer")
 		os.Exit(1)
 	}
+	ec, err = ec.InitializeControllerWithImageVector()
+	if err != nil {
+		setupLog.Error(err, "unable to initialize controller with image vector")
+		os.Exit(1)
+	}
```
swapnilgm: Will you please replace line65-to line74 with single call to `controller.NewReconcilerWithImageVector `?

swapnilgm:
In addition, I still feel, etcd crd should only allow to configure the `versions` and  not the image repository. Druid is clearly operator for etcd, which knows how to handle etcd only from repository configured in druid (usually upstream Etcd). But as a part of `spec.etcd.image` let's say i configured postgres (commonly nobody is gonna do this) image with current logic off course druid doesn't know how to deploy it and it will fail. So, the contract though `spec.etcd.image` field here is not explicit. Hence i would still prefer if we change the field to `spec.etcd.version`.  And keep image repository configurable only through imagevector to druid. 

This is more of design/api standard questions, so i'm not going hard on it. Whatever majority finds good i'm ok with it. I just wanted to record the thoughts here in case discussion kicks in.

rfranzke:
I don't understand your arguments @swapnilgm. Yes, it clearly controls etcd from upstream, and of course only an etcd image can be specified here. Though, if I want to use my own etcd image (maybe modified/customized/patched out vulnerable layers) then I have to state the repository. Why should this be forbidden?

swapnilgm:
> I don't understand your arguments @swapnilgm. Yes, it clearly controls etcd from upstream, and of course only an etcd image can be specified here. Though, if I want to use my own etcd image (maybe modified/customized/patched out vulnerable layers) then I have to state the repository. Why should this be forbidden?

No, this won't be forbidden. I'm just saying right location for configuring repository as per me should be  in imagevector. And not on etcd crd spec. Since, this way contract of druid being aware of etcd image and functioning is intact. Let's say your patched etcd image stored under your own repository break the usual etcd deployment knowledge that is known to druid. Since druid is unaware of the error in such case, will just fail to deploy your image but won't be able to tell what is the exact error. This clearly states the issue on user side and not on druid side. Of course sane user will not do this but IMO druid is not restricting it current way even though it should.  So, that's why i said it's more of software api standard question.

rfranzke:
OK, but then I'd remove the  `image` field completely from the `Etcd` resource. It doesn't make sense to mix the stuff (repo in image vector, tag in custom resource) unless you validate what is allowed to be put into the custom resource.

controllers/etcd_controller.go
```diff
@@ -530,11 +557,34 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
...
+		if err := r.Get(context.TODO(), types.NamespacedName{Name: filteredStatefulSets[0].Name, Namespace: filteredStatefulSets[0].Namespace}, ss); err != nil {
+			return nil, err
+		}
 		// Statefulset is claimed by for this etcd. Just sync the specs
 		if ss, err = r.syncStatefulSetSpec(ss, cm, svc, etcd); err != nil {
 			return nil, err
 		}
+
+		// restart etcd pods in crashloop backoff
+		selector, err := metav1.LabelSelectorAsSelector(ss.Spec.Selector)
```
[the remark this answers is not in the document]
amshuman-kr: `StatefulSet` controller doesn't seem to do rolling update if pods are on `CrashloopBackoff`? Hence , the deletion of such pods during update.

amshuman-kr:
> OK, but then I'd remove the image field completely from the Etcd resource. It doesn't make sense to mix the stuff (repo in image vector, tag in custom resource) unless you validate what is allowed to be put into the custom resource.

But still it might be a valid requirement (not now but perhaps in the future) for druid to support multiple versions of etcd. So, at least `version` should be part of the CRD. But then removing image but retaining version might be too big a change. For now, I am fine continuing the way things are in this PR.

controllers/etcd_controller.go
georgekuruvillak: This logic here is till etcd resouce can be updated by HVPA. Till then we will need druid to not override the changes applied by HVPA with the values in resources.

main.go
georgekuruvillak: done

rfranzke:
> But still it might be a valid requirement (not now but perhaps in the future) for druid to support multiple versions of etcd. So, at least version should be part of the CRD. But then removing image but retaining version might be too big a change. For now, I am fine continuing the way things are in this PR.

Sure, makes sense. I think this is similar to the `hyperkube` image in the g/g repo. We don't specify a tag in the image vector here but the tag comes from the `Shoot` resource.
If you'd follow the same approach, you wouldn't specify a tag for the `etcd` image in your image vector and would take the version from the `Etcd` resource. However, as said above, I guess it makes sense to validate which versions are possible in this case. If somebody wants to overwrite the versions then he/she has to make sure that all tags that were possible before are now possible as well.

PR #32, opened 2020-03-24
Fix wget in etcd bootstrap script

shreyas-s-rao:

This PR fixes bug in etcd bootstrap script which would continue to initialize backup sidecar even if updating wget fails (due to network errors). 

Fixes #

cc @georgekuruvillak @rfranzke 

**Release note**:

```improvement operator
Fix wget in etcd bootstrap script.
```

### DIFF
--- a/charts/etcd/templates/configmap-etcd-bootstrap.yaml
+++ b/charts/etcd/templates/configmap-etcd-bootstrap.yaml
@@ -26,7 +26,24 @@ data:
     # install wget from apk in order to pass --ca-certificate flag because
     # busybox wget only has bare minimum features, without --ca-certificate option
     apk update
+    if [ $? -ne 0 ]
+    then
+      echo "apk update failed"
+      exit 1
+    fi
     apk add wget
+    if [ $? -ne 0 ]
+    then
+      echo "failed to update wget"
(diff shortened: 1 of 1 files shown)


charts/etcd/templates/configmap-etcd-bootstrap.yaml
```diff
@@ -19,14 +19,21 @@ metadata:
     uid: {{ .Values.uid }}
 data:
   bootstrap.sh: |-
-    #!/bin/sh
+    #!/bin/sh -e
```
georgekuruvillak:
```suggestion
    #!/bin/sh
```
I tried this with `-e`. The VALIDATIONMARKER is not created. Lets remove `-e` as it will affect the validation flow.
```
#!/bin/sh -e

trap_and_propagate() {
        PID=$1
        shift
        for sig in "$@" ; do
            trap "kill -$sig $PID" "$sig"
        done
}

rm -rf $VALIDATION_MARKER
etcd &
ETCDPID=$!
trap_and_propagate $ETCDPID INT TERM
wait $ETCDPID
RET=$?
echo $RET > VALIDATION_MARKER
exit $RET
```

PR #33, opened 2020-03-25
Added predicates to the controller mgr with option reconcile only if annotation present.

georgekuruvillak:
This PR does the following:
* remove lastError after successful reconciliation
* fix the labels/selector issue in the statefulset manifest
* introduce `.status.observedGeneration`
* only reconcile if `gardener.cloud/operation=reconcile` and `oldMeta.generation != newMeta.generation`
Fixes #

**Release note**:

```improvement operator
* Removed .status.lastError after successful reconciliation
* Fixed the labels/selector issue in the statefulset manifest
* Introduced .status.observedGeneration
* Added logic to only reconcile if `gardener.cloud/operation=reconcil`e and `oldMeta.generation != newMeta.generation` or `.status.LastError != nil`
```


controllers/etcd_controller.go
```diff
@@ -179,6 +181,9 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if !reflect.DeepEqual(etcd.Spec, etcdCopy.Spec) {
 		etcdCopy.Spec = etcd.Spec
 	}
+
+	etcdCopy.Status.ObserverdGeneration = &etcdCopy.Generation
```
rfranzke: Where is the update-status call for this?

georgekuruvillak: The idea is to eventually update this in `updateEtcdErrorStatus` or `updateEtcdStatus`.

rfranzke: OK, I see. Thanks.

api/v1alpha1/etcd_types.go
```diff
@@ -258,6 +258,9 @@ type LastOperation struct {
 
 // EtcdStatus defines the observed state of Etcd
 type EtcdStatus struct {
+	// ObservedGeneration is the most recent generation observedfor this resource.
+	// +optional
+	ObserverdGeneration *int64 `json:"observedGeneration,omitempty"`
```
shreyas-s-rao:
```suggestion
	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
```

controllers/etcd_controller.go
shreyas-s-rao:
```suggestion
	etcdCopy.Status.ObservedGeneration = &etcdCopy.Generation
```

pkg/predicate/predicate.go
```diff
@@ -0,0 +1,176 @@
...
+	return true
+}
+
+// Create implements default UpdateEvent filter for validating generation change
+func (GenerationChangedPredicate) Create(e event.CreateEvent) bool {
+	etcd, ok := e.Object.(*druidv1alpha1.Etcd)
+	if !ok {
+		return true
+	}
+	if *etcd.Status.ObserverdGeneration == etcd.Generation {
```
shreyas-s-rao:
```suggestion
	if *etcd.Status.ObservedGeneration == etcd.Generation {
```

api/v1alpha1/etcd_types.go
georgekuruvillak: Thanks. Done.

controllers/etcd_controller.go
```diff
@@ -179,11 +180,29 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-	logger.Infof("Reconciling etcd: %s/%s", etcd.GetNamespace(), etcd.GetName())
+
+	etcdCopy.Status.ObservedGeneration = &etcdCopy.Generation
+	if _, ok := etcd.Annotations[common.GardenerOperation]; ok {
+		delete(etcd.Annotations, common.GardenerOperation)
+		if err := r.Update(context.TODO(), etcd); err != nil {
+			if err := r.updateEtcdErrorStatus(etcd, err); err != nil {
+				return ctrl.Result{
+					Requeue:      true,
+					RequeueAfter: time.Second * 5,
```
rfranzke: Why is this needed? It automatically requeues with exponential backoff on error.

controllers/etcd_controller.go
```diff
@@ -179,11 +180,29 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+				}, err
+			}
+			return ctrl.Result{
+				Requeue:      true,
+				RequeueAfter: time.Second * 5,
```
rfranzke: Why is this needed? It automatically requeues with exponential backoff on error.

controllers/etcd_controller.go
```diff
@@ -179,11 +180,29 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if !reflect.DeepEqual(etcd.Spec, etcdCopy.Spec) {
```
rfranzke: I don't get this code - How can it not be equal after 178? Remove it.

controllers/etcd_controller.go
```diff
@@ -179,11 +180,29 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 		etcdCopy.Spec = etcd.Spec
 	}
-	logger.Infof("Reconciling etcd: %s/%s", etcd.GetNamespace(), etcd.GetName())
+
+	etcdCopy.Status.ObservedGeneration = &etcdCopy.Generation
+	if _, ok := etcd.Annotations[common.GardenerOperation]; ok {
```
rfranzke: Shouldn't you work on the `etcdCopy` from here on?

controllers/etcd_controller.go
```diff
@@ -179,11 +180,29 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		delete(etcd.Annotations, common.GardenerOperation)
+		if err := r.Update(context.TODO(), etcd); err != nil {
```
rfranzke: Move the update call out of the `if operation annotation is set` condition. The observed generation should also we updated.

pkg/predicate/predicate.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// Or builds a logical OR gate of passed predicates.
+func Or(predicates ...predicate.Predicate) predicate.Predicate {
+	return &or{predicates}
+}
+
+// HasOperationAnnotation is a predicate for the operation annotation.
+func HasOperationAnnotation() predicate.Predicate {
+	return FromMapper(MapperFunc(func(e event.GenericEvent) bool {
```
rfranzke: Can't you just follow what was done for `LastOperationNotSuccessful` below? Then you don't need all this complicated mapper functions.

```golang
func HasOperationAnnotation() predicate.Predicate {
	f := func(obj runtime.Object) bool {
		etcd, ok := obj.(*druidv1alpha1.Etcd)
		if !ok {
			return false
		}
		return etcd.Annotations[common.GardenerOperation] == common.GardenerOperationReconcile
	}

	return predicate.Funcs{
		CreateFunc: func(event event.CreateEvent) bool {
			return f(event.Object)
		},
		UpdateFunc: func(event event.UpdateEvent) bool {
			return f(event.ObjectNew)
		},
		GenericFunc: func(event event.GenericEvent) bool {
			return f(event.Object)
		},
		DeleteFunc: func(event event.DeleteEvent) bool {
			return f(event.Object)
		},
	}
}
```

pkg/predicate/mapper.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
rfranzke: delete the whole file

controllers/etcd_controller.go
```diff
@@ -1045,3 +1055,24 @@ func (r *EtcdReconciler) claimConfigMaps(etcd *druidv1alpha1.Etcd, selector labe
...
+		druidpredicates.GenerationChangedPredicate{},
+		druidpredicates.LastOperationNotSuccessful(),
+	}
+	builder := ctrl.NewControllerManagedBy(mgr).WithOptions(controller.Options{
+		MaxConcurrentReconciles: workers,
+	})
+	if !ignoreOperationAnnotation {
+		predicates = append(predicates, druidpredicates.HasOperationAnnotation())
+	}
+	builder = builder.WithEventFilter(druidpredicates.Or(predicates...))
```
rfranzke: @timuthy - what about the case if the deletionTimestamp is set? Is a predicate needed for it?

amshuman-kr:
@georgekuruvillak Is this still WIP or are the changes from your side done?

controllers/etcd_controller.go
georgekuruvillak: Update does not update the `status` sub-resource. If update failed, it would still reconcile as part of retries. Do I need to explicitly update it here. I am updating the status eventually.

controllers/etcd_controller.go
georgekuruvillak: makes sense.

controllers/etcd_controller.go
georgekuruvillak: makes sense.

controllers/etcd_controller.go
georgekuruvillak: If I don't add a predicate for deletionTimestamp, it will still reconcile. Do I need to add it explicitly?

pkg/predicate/mapper.go
georgekuruvillak: makes sense.

controllers/etcd_controller.go
rfranzke: OK, I see. Why not setting the observedGeneration then in your updateStatus functions like you set all the other fields there?

controllers/etcd_controller.go
rfranzke: OK, it should actually reconcile if the deletion timestamp is set. How is this ensured?

pkg/predicate/predicate.go
georgekuruvillak: Makes sense. It would be fine to return true in case of `DeleteEvent` if annotation is not set too, ryt?

rfranzke: Hm yes, I guess.

controllers/etcd_controller.go
rfranzke: OK. we just figured out that the generation is increased when the  `Etcd` resource gets deleted. Hence, nothing to do.

timuthy: Can you check if the deletion case works properly?
Because here is what I think can happen:
- The object is deleted 
- `deletionTimestamp` is set, no `generation` change
- Update event is triggered (due to `finalizer` the deletion event is not triggered immediately)
- All predicates return `false`

Also check the following scenario:
- Delete the resource
- Controller starts deletion operation
- Interrupt/restart controller while deletion operation is running
- Make sure controller continues after restart (now it gets a `create` event for an object with a `deletionTimestamp`.)

controllers/etcd_controller.go
```diff
@@ -1045,3 +1055,24 @@ func (r *EtcdReconciler) claimConfigMaps(etcd *druidv1alpha1.Etcd, selector labe
 	cm := NewEtcdDruidRefManager(r, etcd, selector, etcdGVK, canAdoptFunc)
 	return cm.ClaimConfigMaps(ss)
 }
+
+// SetupWithManager sets up manager with a new controller and r as the reconcile.Reconciler
+func (r *EtcdReconciler) SetupWithManager(mgr ctrl.Manager, workers int, ignoreOperationAnnotation bool) error {
+	predicates := []predicate.Predicate{
+		druidpredicates.GenerationChangedPredicate{},
+		druidpredicates.LastOperationNotSuccessful(),
```
timuthy: What about shoots with `lastOperation: Failed`? In the extensions we stop reconciling related resources.
https://github.com/gardener/gardener-extensions/blob/master/pkg/controller/infrastructure/controller.go#L69

/cc @rfranzke

pkg/common/common.go
```diff
@@ -20,5 +20,7 @@ const (
 	// BackupRestore is the key for the etcd-backup-restore image in image vector
 	BackupRestore = "etcd-backup-restore"
 	// ChartPath is the directory that contains the default image vector file
-	ChartPath = "charts"
+	ChartPath                  = "charts"
+	GardenerOperation          = "gardener.cloud/operation"
+	GardenerOperationReconcile = "reconcile"
```
timuthy: You don't need to introduce these constants. Please use the ones from https://github.com/gardener/gardener/blob/master/pkg/apis/core/v1beta1/constants/types_constants.go instead.

controllers/etcd_controller.go
rfranzke: @timuthy see https://github.com/gardener/etcd-druid/pull/33#discussion_r399212827 - we figured out that the generation is bumped when the deletion timestamp gets set. Hence, the first scenario should be covered.

The second scenario should also be covered because the `observedGeneration` is only updated at the end of a successful reconciliation, or after an error occurred (but then the `lastError` will be set as well, allowing to pick the `Etcd` up again). WDYT?

controllers/etcd_controller.go
rfranzke: Nice catch, thanks. Though, it's not working for the regular extensions right now anyways, so I suggest to exclude it for now and fix it afterwards in one go for all components.

pkg/common/common.go
rfranzke: I told the same, there is a lot of duplicate code that is maintained in g/g or g/gext already (not only in this PR but also in the already merged code in `master`). @georgekuruvillak explained that he doesn't want to code-wise depend on g/g and g/gext. I'm not sure if this is a good idea, but I suggest to continue with the "copy" approach for now and address this after the PR. WDYT?

controllers/etcd_controller.go
timuthy: Okay thanks @rfranzke sounds good to me.

pkg/common/common.go
timuthy: Hm, I just saw that it already has a dependency to g/g and that's why I thought why not reuse it already.

controllers/etcd_controller.go
georgekuruvillak: done

controllers/etcd_controller.go
timuthy: Okay true. Does it at make sense (also for the extensions) to check the failed state at the beginning of a reconciliation and if it's failed we return w/o re-queuing?

rfranzke: Yep, sounds good :)

rfranzke: @timuthy We decided to postpone this for the next release - the druid is not yet doing anything with the `extensions.gardener.cloud/v1alpha1.Cluster` resource - but this is where the information where the shoot's last operation is `Failed`. As it doesn't work for the regular extension controllers yet let's proceed now without this check and then fix it altogether in the extensions + druid.

controllers/etcd_controller.go
```diff
@@ -670,12 +683,25 @@ func (r *EtcdReconciler) syncStatefulSetSpec(ss *appsv1.StatefulSet, cm *corev1.
...
 	if err != nil {
-		logger.Infof("patching statefulset failed for %s", ss.Name)
 		return nil, err
 	}
 	return ssCopy, err
 }
 
+func (r *EtcdReconciler) recreateStatefulset(ss *appsv1.StatefulSet) error {
+
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
amshuman-kr: Please merge this `RetryOnConflict` with the one [below](https://github.com/gardener/etcd-druid/pull/33/files#diff-8e5c816f8b2f1adeef36a64d81b3e0f6R699).

controllers/etcd_controller.go
```diff
@@ -670,12 +683,25 @@ func (r *EtcdReconciler) syncStatefulSetSpec(ss *appsv1.StatefulSet, cm *corev1.
...
+		return r.Delete(context.TODO(), ss)
+	})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
amshuman-kr: See [above](https://github.com/gardener/etcd-druid/pull/33/files#diff-8e5c816f8b2f1adeef36a64d81b3e0f6R693).

pkg/predicate/predicate.go
```diff
@@ -0,0 +1,193 @@
...
+		return false
+	}
+	if e.MetaNew == nil {
+		logger.Error("update event has no new metadata", "event", e)
+		return false
+	}
+	if e.MetaNew.GetGeneration() == e.MetaOld.GetGeneration() {
+		return false
+	}
+	return true
```
amshuman-kr: Shouldn't we add the `etcd.Status.ObservedGeneration != nil && *etcd.Status.ObservedGeneration == etcd.Generation` check here as well, just to be safe?

controllers/etcd_controller.go
```diff
@@ -174,42 +178,48 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			}
+			return ctrl.Result{
+				Requeue: true,
+			}, err
+		}
 	}
+
 	logger.Infof("Reconciling etcd: %s/%s", etcd.GetNamespace(), etcd.GetName())
-	if !etcdCopy.DeletionTimestamp.IsZero() {
+	if !etcd.DeletionTimestamp.IsZero() {
```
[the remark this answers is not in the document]
rfranzke: Just do `if etcd.DeletionTimestamp != nil` - that's enough

controllers/etcd_controller.go
```diff
@@ -174,42 +178,48 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 		return ctrl.Result{}, err
 	}
-	etcdCopy := etcd.DeepCopy()
-	// Update the found object and write the result back if there are any changes
-	if !reflect.DeepEqual(etcd.Spec, etcdCopy.Spec) {
-		etcdCopy.Spec = etcd.Spec
+
+	if _, ok := etcd.Annotations[v1beta1constants.GardenerOperation]; ok {
+		delete(etcd.Annotations, v1beta1constants.GardenerOperation)
+		if err := r.Update(context.TODO(), etcd); err != nil {
```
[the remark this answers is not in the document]
rfranzke: The annotation shall be removed at the beginning of the reconciliation as of https://github.com/gardener/gardener/blob/master/docs/extensions/reconcile-trigger.md 

> They have to remove this annotation as soon as they begin with their reconcile operation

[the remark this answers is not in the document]
amshuman-kr: 👍 I was just worried about the case where the controller might not get to complete the reconciliation due to whatever reasons, including the controller pod going down.

controllers/etcd_controller.go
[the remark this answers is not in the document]
amshuman-kr: Agreed.

controllers/etcd_controller.go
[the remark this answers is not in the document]
rfranzke: Hm, good point, indeed. @georgekuruvillak how do you make sure that druid picks up the reconciliation again in such case?

pkg/predicate/predicate.go
georgekuruvillak: In case of an update, the oldMeta and newMeta.Generation would be different. So we needn't compare with the status. In case of create, we dont have an oldMeta to compare against.  Create is also triggered when etcd-druid startup. So I am using the observedGeneration.

controllers/etcd_controller.go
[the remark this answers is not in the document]
georgekuruvillak: It checks for nil inside the IsZero() function.

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -29,9 +29,6 @@ spec:
     matchLabels:
       name: etcd
       instance: {{ .Values.name }}
-{{- if .Values.labels }}
```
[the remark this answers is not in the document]
georgekuruvillak: selector in statefulset has to be immutable, else it errors. In druid, we have implementation to recreate sts with the new selector. but to adhere to sts and deployment standards, its better that the selector is immutable once sts is created.

controllers/etcd_controller.go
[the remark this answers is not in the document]
georgekuruvillak: I'll remove this annotation at the end after the status is updated. That way, if the status is not updated for some reason, the annotation remains and it reconciles one restart. If status is updated but fails before annotation is removed, just one more reconciliation.

pkg/predicate/predicate.go
amshuman-kr: I know. Additional safety check won't hurt, no? What if there is a but in upstream libraries? We know client-go watch timeout implementation sucked.

controllers/etcd_controller.go
amshuman-kr: For the case where delete succeeded but create failed for some reason, you could save the info in a variable.

For example,
```go
func (r *EtcdReconciler) recreateStatefulset(ss *appsv1.StatefulSet) error {
        skipDelete := false
	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
                 if !skipDelete {
		        if err := r.Delete(context.TODO(), ss); err != nil && !errors.IsNotFound(err) {
			        return err
		        }
                 }

                 skipDelete = true

		return r.Create(context.TODO(), ss)
	})
	return err
}
```

PR #34, opened 2020-03-30
Stop reconciliation of etcd resources on sts changes

georgekuruvillak:
This PR is required to stop reconciliation of etcd resources on sts changes. It results in multiple reconciliations. Instead, etcd druid queries for sts readiness and updates the etcd status accordingly. If sts, not ready for a pre-determined time, the reconcile operation errors out.
Fixes #

**Release note**:

```improvement operator
Changed logic in etcd druid to query for statefulset readiness and update the etcd status accordingly
```

### DIFF
--- a/controllers/etcd_controller.go
+++ b/controllers/etcd_controller.go
@@ -48,6 +48,7 @@ import (
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 
 	"github.com/gardener/gardener/pkg/utils/imagevector"
+	gardenerretry "github.com/gardener/gardener/pkg/utils/retry"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -67,7 +68,12 @@ const (
 	FinalizerName = "druid.gardener.cloud/etcd-druid"
 	// DefaultImageVector is a constant for the path to the default image vector file.
 	DefaultImageVector = "images.yaml"
-	timeout            = time.Second * 30
+	// DefaultTimeout is the default timeout for retry operations.
+	DefaultTimeout = time.Minute * 1
+	// DefaultInterval is the default interval for retry operations.
+	DefaultInterval = 5 * time.Second
+	// EtcdReady implies that etcd is ready
+	EtcdReady = true
 )
 
 // EtcdReconciler reconciles a Etcd object
@@ -168,8 +174,8 @@ func (r *EtcdReconciler) InitializeControllerWithImageVector() (*EtcdReconciler,
 // Reconcile reconciles the etcd.
 func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
-	etcd := &druidv1alpha1.Etcd{}
-	if err := r.Get(context.TODO(), req.NamespacedName, etcd); err != nil {
+	etcdCopy := &druidv1alpha1.Etcd{}
+	if err := r.Get(context.TODO(), req.NamespacedName, etcdCopy); err != nil {
 		if errors.IsNotFound(err) {
 			// Object not found, return.  Created objects are automatically garbage collected.
 			// For additional cleanup logic use finalizers.
@@ -179,11 +185,18 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
+	etcd, err := r.updateEtcdStatusAsNotReady(etcdCopy)
+	if err != nil && !errors.IsNotFound(err) {
+		return ctrl.Result{
+			Requeue: true,
+		}, err
+	}
+
 	logger.Infof("Reconciling etcd: %s/%s", etcd.GetNamespace(), etcd.GetName())
 	if !etcd.DeletionTimestamp.IsZero() {
 		logger.Infof("Deletion timestamp set for etcd: %s", etcd.GetName())
 		if err := r.removeFinalizersToDependantSecrets(etcd); err != nil {
-			if err := r.updateEtcdErrorStatus(etcd, err); err != nil {
+			if err := r.updateEtcdErrorStatus(etcd, nil, err); err != nil {
 				return ctrl.Result{
 					Requeue: true,
 				}, err
@@ -198,16 +211,12 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			finalizers := sets.NewString(etcd.Finalizers...)
 			finalizers.Delete(FinalizerName)
 			etcd.Finalizers = finalizers.UnsortedList()
-			if err := r.Update(context.TODO(), etcd); err != nil {
-				if err := r.updateEtcdErrorStatus(etcd, err); err != nil {
-					return ctrl.Result{
-						Requeue: true,
-					}, err
-				}
+			if err := r.Update(context.TODO(), etcd); err != nil && !errors.IsConflict(err) {
 				return ctrl.Result{
 					Requeue: true,
 				}, err
 			}
+
 		}
 		logger.Infof("Deleted etcd %s successfully.", etcd.GetName())
 		return ctrl.Result{}, nil
@@ -219,7 +228,7 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		finalizers.Insert(FinalizerName)
 		etcd.Finalizers = finalizers.UnsortedList()
 		if err := r.Update(context.TODO(), etcd); err != nil {
-			if err := r.updateEtcdErrorStatus(etcd, err); err != nil {
+			if err := r.updateEtcdErrorStatus(etcd, nil, err); err != nil {
 				return ctrl.Result{
 					Requeue: true,
 				}, err
@@ -230,7 +239,7 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 	}
 	if err := r.addFinalizersToDependantSecrets(etcd); err != nil {
-		if err := r.updateEtcdErrorStatus(etcd, err); err != nil {
+		if err := r.updateEtcdErrorStatus(etcd, nil, err); err != nil {
 			return ctrl.Result{
 				Requeue: true,
 			}, err
@@ -239,7 +248,7 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	svc, ss, err := r.reconcileEtcd(etcd)
 	if err != nil {
-		if err := r.updateEtcdErrorStatus(etcd, err); err != nil {
+		if err := r.updateEtcdErrorStatus(etcd, ss, err); err != nil {
 			return ctrl.Result{
 				Requeue: true,
 			}, err
@@ -288,7 +297,7 @@ func (r *EtcdReconciler) reconcileServices(etcd *druidv1alpha1.Etcd, renderedCha
 	}
 
 	if len(filteredServices) > 0 {
-		// TODO: Sync spec and delete OR First delete and then sync spec?
+		logger.Infof("Claiming existing etcd services for etcd:%s in namespace:%s", etcd.Name, etcd.Namespace)
 
 		// Keep only 1 Service. Delete the rest
 		for i := 1; i < len(filteredServices); i++ {
@@ -321,9 +330,7 @@ func (r *EtcdReconciler) reconcileServices(etcd *druidv1alpha1.Etcd, renderedCha
 		return nil, err
 	}
 
-	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		return r.Create(context.TODO(), ss)
-	})
+	err = r.Create(context.TODO(), ss)
 
 	// Ignore the precondition violated error, this service is already updated
 	// with the desired label.
@@ -416,7 +423,7 @@ func (r *EtcdReconciler) reconcileConfigMaps(etcd *druidv1alpha1.Etcd, renderedC
 	}
 
 	if len(filteredCMs) > 0 {
-		// TODO: Sync spec and delete OR First delete and then sync spec?
+		logger.Infof("Claiming existing etcd configmaps for etcd:%s in namespace:%s", etcd.Name, etcd.Namespace)
 
 		// Keep only 1 Configmap. Delete the rest
 		for i := 1; i < len(filteredCMs); i++ {
@@ -449,9 +456,7 @@ func (r *EtcdReconciler) reconcileConfigMaps(etcd *druidv1alpha1.Etcd, renderedC
 		return nil, err
 	}
 
-	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		return r.Create(context.TODO(), cm)
-	})
+	err = r.Create(context.TODO(), cm)
 
 	// Ignore the precondition violated error, this machine is already updated
 	// with the desired label.
@@ -543,7 +548,7 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
 	}
 
 	if len(filteredStatefulSets) > 0 {
-		// TODO: Sync spec and delete OR First delete and then sync spec?
+		logger.Infof("Claiming existing etcd statefulsets for etcd:%s in namespace:%s", etcd.Name, etcd.Namespace)
 
 		// Keep only 1 statefulset. Delete the rest
 		for i := 1; i < len(filteredStatefulSets); i++ {
@@ -585,7 +590,7 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
 			}
 		}
 
-		return ss.DeepCopy(), err
+		return r.waitUntilStatefulSetReady(ss)
 	}
 
 	// Required statefulset doesn't exist. Create new
@@ -594,9 +599,7 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
 		return nil, err
 	}
 
-	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		return r.Create(context.TODO(), ss)
-	})
+	err = r.Create(context.TODO(), ss)
 
 	// Ignore the precondition violated error, this machine is already updated
 	// with the desired label.
@@ -611,7 +614,8 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
 	if err := controllerutil.SetControllerReference(etcd, ss, r.Scheme); err != nil {
 		return nil, err
 	}
-	return ss.DeepCopy(), err
+
+	return r.waitUntilStatefulSetReady(ss)
 }
 
 func getContainerMapFromPodTemplateSpec(spec v1.PodSpec) map[string]v1.Container {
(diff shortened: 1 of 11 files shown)


georgekuruvillak:
@rfranzke

controllers/etcd_controller.go
```diff
@@ -963,33 +961,53 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
 }
 
 func (r *EtcdReconciler) updateEtcdStatus(etcd *druidv1alpha1.Etcd, svc *corev1.Service, ss *appsv1.StatefulSet) error {
+
+	sts, err := r.waitUntilStatefulSetReady(ss)
```
rfranzke: I'd not do this in the `updateStatus` function but in the reconciliation function.

controllers/etcd_controller.go
```diff
@@ -963,33 +961,53 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
...
 }
 
+func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*appsv1.StatefulSet, error) {
+	ss := &appsv1.StatefulSet{}
+	err := gardenerretry.UntilTimeout(context.TODO(), DefaultInterval, DefaultTimeout, func(ctx context.Context) (bool, error) {
+		if err := r.Get(context.TODO(), types.NamespacedName{Name: sts.Name, Namespace: sts.Namespace}, ss); err != nil {
+			return false, err
+		}
+		if err := health.CheckStatefulSet(ss); err != nil {
+			return false, err
```
rfranzke:
```suggestion
			return false, nil
```

controllers/etcd_controller.go
```diff
@@ -1077,8 +1095,5 @@ func (r *EtcdReconciler) SetupWithManager(mgr ctrl.Manager, workers int, ignoreO
 	builder = builder.WithEventFilter(druidpredicates.Or(predicates...))
 	return builder.
 		For(&druidv1alpha1.Etcd{}).
-		Owns(&appsv1.StatefulSet{}).
```
rfranzke: Shouldn't you add these `Owns()` calls if the operation annotation is ignored?

rfranzke: Fine for me. In the Gardener context it's only important that these mapping to the `Etcd` resource are not happening as they bypass the predicates.

controllers/etcd_controller.go
rfranzke: If we return an error here then the whole `wait` function doesn't make sense. Then we can just return the error to the reconcile function and backoff when the STS is not ready yet, and try again in the next reconciliation.

swapnilgm: Ahh sorry my bad. I thought returning `err` will end the retries.

controllers/etcd_controller.go
```diff
@@ -963,33 +961,53 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
...
 		return err
 	}
 	return r.removeOperationAnnotation(etcd)
 }
 
+func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*appsv1.StatefulSet, error) {
+	ss := &appsv1.StatefulSet{}
+	err := gardenerretry.UntilTimeout(context.TODO(), DefaultInterval, DefaultTimeout, func(ctx context.Context) (bool, error) {
+		if err := r.Get(context.TODO(), types.NamespacedName{Name: sts.Name, Namespace: sts.Namespace}, ss); err != nil {
+			return false, err
```
rfranzke:
```suggestion
			return gardenerretry.SevereError(err)
```

controllers/etcd_controller.go
rfranzke:
```suggestion
			return gardenerretry.MinorError(err)
```

controllers/etcd_controller.go
```diff
@@ -963,33 +961,53 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
...
+		}
+		if err := health.CheckStatefulSet(ss); err != nil {
+			return false, err
+		}
+		return true, nil
```
rfranzke:
```suggestion
		return gardenerretry.Ok()
```

rfranzke:
Can we restart the `.status.ready` at the beginning of a reconciliation? Just to be sure that the value is not accidentally interpreted during a reconciliation

controllers/etcd_controller.go
amshuman-kr: I disagree. I don't think removing `Owns` on managed downstream resources (`StatefulSet`, `ConfigMap` and `Service`) is a good idea. Even if operation annotation is not ignored. It is the job of druid, as an etcd controller, to reconcile the `StatefulSet` if it is modified incompatibly and/or deleted, no?

Please do not see druid as only a gardener extension. It is also a controller that is in charge of managing etcd. I am increasingly regretting the decision not to introduce a separate etcd extension controller in addition to druid which ideally should manage only the etcd.

rfranzke: As said, the mappings introduced by `Owns()` function are bypassing the predicates. This does not fit to the Gardener approach. Also, in the Gardener context it is not necessary that it reconciles on `StatefulSet`/`ConfigMap`/`Service` changes. These resources live in the seed and no end-user tampers with it. Yes, this is not the mainstream controller convention, but it is what we need to ensure shoot reconciliations only happen in the maintenance time window, and especially only then when Gardener demands a reconciliation.

As said above - I'm fine if you add these `Owns()` calls when the operation annotation is ignored. When druid is used in Gardener this will anyways never be the situation, hence, I don't care.

From my point of of view you have two options now:

1. Remove the `Owns()` completely or add them conditionally if operation annotation is ignored.
2. Disagree with 1), go one step back and develop something that fits to the nature of Gardener. 

Druid is blocking a Gardener + extensions release for 2 weeks now. Please decide - either we get a first working druid version in now and you can improve the code in the iterations later on, or we don't do it. Then I'll go ahead and revert everything that is druid-related for now so that Gardener can be released. Please tell me how to proceed.

controllers/etcd_controller.go
amshuman-kr: We should do this at the beginning of reconciliation. Also, then we can avoid the wait and simply `ReconcileAfter` if the `StatefulSet` is not ready.

rfranzke: How does it make sense to check the statefulset status at the beginning if we already know that subsequent steps might UPDATE it, potentially fixing a broken state?

controllers/etcd_controller.go
```diff
@@ -963,33 +961,53 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
 }
 
 func (r *EtcdReconciler) updateEtcdStatus(etcd *druidv1alpha1.Etcd, svc *corev1.Service, ss *appsv1.StatefulSet) error {
+
```
amshuman-kr: Should we split the `updateXXXEtcdStatus` funcs into two parts?

1. A func to update the `EtcdStatus` (success or failure). This could/should be called at the beginning (as soon as potentially possible).
1. A func to actually update the status sub-resource (success or failure). This could be at the end before returning.

controllers/etcd_controller.go
```diff
@@ -963,33 +961,53 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
...
 
 	if err := r.Status().Update(context.TODO(), etcd); err != nil && !errors.IsNotFound(err) {
 		return err
 	}
 	return r.removeOperationAnnotation(etcd)
 }
 
+func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*appsv1.StatefulSet, error) {
+	ss := &appsv1.StatefulSet{}
+	err := gardenerretry.UntilTimeout(context.TODO(), DefaultInterval, DefaultTimeout, func(ctx context.Context) (bool, error) {
```
amshuman-kr: I don't like this waiting during reconciliation. We should `ReconcileAfter` instead. See [above](https://github.com/gardener/etcd-druid/pull/34/files#r400686316).

controllers/etcd_controller.go
```diff
@@ -963,33 +961,53 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
...
+		if err := r.Get(context.TODO(), types.NamespacedName{Name: sts.Name, Namespace: sts.Namespace}, ss); err != nil {
```
amshuman-kr: Shouldn't the controller create the `StatefulSet` if it is not found? Can we not shuffle this logic with the overall reconciliation logic to handle this smoothly?

controllers/etcd_controller.go
georgekuruvillak: Going with option 1 will get gardener unblocked. @amshuman-kr, I'll get it out this way and then ideate and work on the owns() functionality with druid. wdyt?

controllers/etcd_controller.go
amshuman-kr: 👍I misspoke about the beginning part. I meant whenever we do the check for statefulset in the overall reconciliation. The reconciling after part stands still.

controllers/etcd_controller.go
amshuman-kr: I am uncomfortable with 1. But even otherwise, considering the number of issues we have faced with the integration, how about the following choices?

1. Remove the Owns() completely or add them conditionally if operation annotation is ignored. But  put the whole etcd-druid integration under a feature gate and disable the feature gate in gardener in this release until we have resolved all the issues currently understood and added tests.
1. Revert the current etcd-druid integration, go one step back and develop something that fits to the nature of Gardener and then file a fresh integration.

rfranzke: I don't get it, we are working since two weeks on resolving the issues and are getting closer now. This PR is necessary for sure, then we will run another test and check if etcd-druid behaves correctly now. Only then we will release Gardener. Why do we now start talking about introducing a feature gate? It doesn't make sense at this point anymore. Also, we would have to touch all extensions as well (and they don't have feature gates yet). Then we can also go ahead and revert everything and wait for a cleaner version of druid.

I am open to try getting the druid ready for 1-2 more days, but as said, this PR would be needed for sure. So, please decide how you want to continue.

PS: I'm also somewhat confused why you said this:

>  I am increasingly regretting the decision not to introduce a separate etcd extension controller in addition to druid which ideally should manage only the etcd.

My understanding was that etcd-druid was MAINLY designed to fit to Gardener (as it's also developed in the `gardener` org). If this can be made optional to also work outside of Gardener environment then fine, no worries, but the prio 1 should be Gardener. Why else should we develop an etcd operator if we don't use it in Gardener?

georgekuruvillak: If we remove owns{}, it gets implemented how gardener was already doing it. if it goes in this way, now. it wouldnt break the existing gardener contract on the controlplane.
Next iteration I can add in the owns logic that is acceptable with gardener contract. That way release is unblocked too. I feel putting the feature under feature-gate now is similar to reverting effortwise.

amshuman-kr: @rfranzke I just aired my overall concern. I am just worried what else we have missed out.

> PS: I'm also somewhat confused why you said this:

>> I am increasingly regretting the decision not to introduce a separate etcd extension controller in addition to druid which ideally should manage only the etcd.

Sorry for not giving the background. Initially there was some discussion about whether just druid will be enough or we will need an additional gardener extension controller that would take care of gardener-specific stuff. We ended up deciding against the additional extension controller. I am not sure if an additional extension controller would have solved this particular problem.

My main concern with "Remove the Owns() completely or add them conditionally if operation annotation is ignored." is that earlier without druid, if something went wrong with etcd statefulset, gardener/gardenlet would reconcile it. Now with druid removing `Owns()`, nobody would unless it is maintenance window or forced reconciliation. Is this correct understanding? Am I missing something?

amshuman-kr: > My understanding was that etcd-druid was MAINLY designed to fit to Gardener (as it's also developed in the gardener org). If this can be made optional to also work outside of Gardener environment then fine, no worries, but the prio 1 should be Gardener. Why else should we develop an etcd operator if we don't use it in Gardener?

To be clear, my concern is not gardener-specific stuff in druid. That is fine. My concern is the correct behaviour if something goes wrong with the statefulset etc.Please see [above](https://github.com/gardener/etcd-druid/pull/34#discussion_r400739260).

controllers/etcd_controller.go
georgekuruvillak: The controller creates/reconciles the statefulset. Here, it checks if the statefulset is ready. it can be because of misconfiguration of etcd resource resulting in pods failing(eg. bucket name provided wrongly).

amshuman-kr: Yes. Why not do this check where the controller reconciles the statefulset?

georgekuruvillak: Yup. Makes sense. Ill move the logic into the reconcile cycle before updating the status.

controllers/etcd_controller.go
rfranzke: > is that earlier without druid, if something went wrong with etcd statefulset, gardener/gardenlet would reconcile it.

Gardener would have sent its desired state of the StatefulSet again - but only if the shoot reconciliation starts. If the statefulset fails in between then Gardener does nothing (though, this  never happened and basically also cannot happen unless a human intervenes).

> Now with druid removing Owns(), nobody would unless it is maintenance window or forced reconciliation. Is this correct understanding?

Exactly, as described, this is exactly how it was before. When Gardener decides to reconcile a shoot then it annotates the `Etcd` with the operation annotation which tells druid to start. If this annotation does not exist then druid shall not act (because Gardener didn't demand it). With `Owns()` druid would reconcile always if there is for example a status update in the statefulset. It wouldn't respect the predicates (operation annotation), leading to reconciliations outside of the maintenance time window and eventually to disruptions.

amshuman-kr: > If the statefulset fails in between then Gardener does nothing (though, this never happened and basically also cannot happen unless a human intervenes).

If I understood right then if statefulset is unhealthy then even Gardner would have done nothing until reconciliation is triggered (manually or in maintenance window). This addresses my concern. Then I am fine going with "1. Remove the Owns() completely or add them conditionally if operation annotation is ignored."

controllers/etcd_controller.go
```diff
@@ -998,6 +1030,16 @@ func (r *EtcdReconciler) removeOperationAnnotation(etcd *druidv1alpha1.Etcd) err
 	return nil
 }
 
+func (r *EtcdReconciler) updateEtcdStatusAsNotReady(etcd *druidv1alpha1.Etcd) (*druidv1alpha1.Etcd, error) {
+	etcdCopy := etcd.DeepCopy()
+	etcdCopy.Status.Ready = false
```
rfranzke: Should `Ready` be a `*bool` instead of `bool` so that you can set it to `nil` here? Setting it to `false` is a little weird - actually, the statefulset might be running fine and controllers listening on the `.status.ready` field might misinterpret this temporary `ready=false` status. WDYM?

controllers/etcd_controller.go
```diff
@@ -611,7 +621,13 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
 	if err := controllerutil.SetControllerReference(etcd, ss, r.Scheme); err != nil {
 		return nil, err
 	}
-	return ss.DeepCopy(), err
+
+	sts, err := r.waitUntilStatefulSetReady(ss)
```
rfranzke:
```suggestion
	return r.waitUntilStatefulSetReady(ss)
```

controllers/etcd_controller.go
georgekuruvillak: It does get requeued after the wait reaches the timeout and errors out. It would result in less reconcile loops triggered.

controllers/etcd_controller.go
```diff
@@ -585,7 +592,12 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
 			}
 		}
 
-		return ss.DeepCopy(), err
+		sts, err := r.waitUntilStatefulSetReady(ss)
```
rfranzke:
```suggestion
		return r.waitUntilStatefulSetReady(ss)
```

controllers/etcd_controller.go
```diff
@@ -956,40 +972,56 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
 	lastErrStr := fmt.Sprintf("%v", lastError)
 	etcd.Status.LastError = &lastErrStr
 	etcd.Status.ObservedGeneration = &etcd.Generation
+	etcd.Status.Ready = false
```
rfranzke: What is the purpose of the `ready` field? If it is to show whether the etcd statefulset is fine, then I suggest to change it to

```suggestion
	etcd.Status.Ready = health.CheckStatefulSet(ss) == nil
```

controllers/etcd_controller.go
```diff
@@ -963,33 +961,53 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
...
-	etcd.Status.ReadyReplicas = ss.Status.ReadyReplicas
-	etcd.Status.UpdatedReplicas = ss.Status.UpdatedReplicas
-	etcd.Status.Ready = (health.CheckStatefulSet(ss) == nil)
+	etcd.Status.CurrentReplicas = sts.Status.CurrentReplicas
+	etcd.Status.ReadyReplicas = sts.Status.ReadyReplicas
+	etcd.Status.UpdatedReplicas = sts.Status.UpdatedReplicas
 	etcd.Status.ServiceName = &svcName
 	etcd.Status.LastError = nil
 	etcd.Status.ObservedGeneration = &etcd.Generation
+	etcd.Status.Ready = true
```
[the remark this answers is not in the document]
rfranzke: See question above. If the purpose of `Ready` is to show the statefulset status then change it to 
```suggestion
	etcd.Status.Ready = health.CheckStatefulSet(ss) == nil
```

controllers/etcd_controller.go
amshuman-kr: A `ReconcileAfter` would avoid too many reconcile loops too, right?

controllers/etcd_controller.go
```diff
@@ -956,40 +964,57 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, lastErr
 	lastErrStr := fmt.Sprintf("%v", lastError)
 	etcd.Status.LastError = &lastErrStr
 	etcd.Status.ObservedGeneration = &etcd.Generation
+	etcd.Status.Ready = nil
```
rfranzke: Why is it set to `nil` here? You explained that the `Ready` field shall show the status of the etcd statefulset. Maybe the statefulset is ready but simply the reconciliation failed? I think it's cleaner to also call `health.CheckStatefulSet` here and set `Ready` accordingly.

controllers/etcd_controller.go
ggaurav10: If i understand right, we need `ignoreOperationAnnotation` so that etcd-druid's reconciliation can be controlled by gardener by adding/removing the "reconcile" annotation during maintenance window.
Will it be cleaner/easier if we instead add a `reconciliationTimeWindow` itself under `etcd`'s spec. etcd-controller would then reconcile based only on the field if set, otherwise all the time.

controllers/etcd_controller.go
```diff
@@ -585,7 +590,7 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
 			}
 		}
 
-		return ss.DeepCopy(), err
+		return r.waitUntilStatefulSetReady(ss)
```
amshuman-kr: I still do no like this. But will park it for a separate PR later in the interest of getting druid out and unblocking gardener release.

PR #35, opened 2020-04-02
Add priority class name, use same temp directory as before

rfranzke:
Added the `gardener-shoot-control-plane` priority class to the etcd pods and changed the temporary directory path for the snapstore from `/tmp` to `/var/etcd/data/temp`.
Also, changed back the value of `--etcd-connection-timeout` to `5m` (instead of `30s`) as it was before introduction of etcd-druid.

**Release note**:

```improvement operator
Added the `gardener-shoot-control-plane` priority class to the etcd pods and changed the temporary directory path for the snapstore from `/tmp` to `/var/etcd/data/temp`.
```

### DIFF
--- a/charts/etcd/templates/etcd-statefulset.yaml
+++ b/charts/etcd/templates/etcd-statefulset.yaml
@@ -48,6 +48,9 @@ spec:
       - ip: "127.0.0.1"
         hostnames:
         - {{ .Values.name }}-local
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
+      {{- end }}
       containers:
       - name: etcd
         image: {{ .Values.etcd.image }}
--- a/charts/etcd/values.yaml
+++ b/charts/etcd/values.yaml
@@ -1,9 +1,10 @@
 name: test
 uid: uuid-of-etcd-resource
-serviceName: test 
+serviceName: test
 configMapName: test
 
 replicas: 1
+#priorityClassName: foo
 
 tlsServerSecret: etcd-server-tls
 tlsClientSecret: etcd-client-tls
@@ -31,8 +32,8 @@ etcd:
 
 backup:
   pullPolicy: IfNotPresent
-  snapstoreTempDir: "/tmp"
-  etcdConnectionTimeout: 30
+  snapstoreTempDir: "/var/etcd/data/temp"
+  etcdConnectionTimeout: 5m
   etcdQuotaBytes: 8Gi
   garbageCollectionPolicy: LimitBased
   maxBackups: 7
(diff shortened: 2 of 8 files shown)


charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -48,6 +48,9 @@ spec:
       - ip: "127.0.0.1"
         hostnames:
         - {{ .Values.name }}-local
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
```
rfranzke, line 52: FYI: https://github.com/gardener/gardener/blob/v1.1.6/charts/seed-controlplane/charts/etcd/templates/etcd-statefulset.yaml#L35

charts/etcd/values.yaml
```diff
@@ -31,8 +32,8 @@ etcd:
 
 backup:
   pullPolicy: IfNotPresent
-  snapstoreTempDir: "/tmp"
-  etcdConnectionTimeout: 30
+  snapstoreTempDir: "/var/etcd/data/temp"
+  etcdConnectionTimeout: 5m
```
rfranzke, line 36: https://github.com/gardener/gardener-extension-provider-aws/blob/v1.4.0/vendor/github.com/gardener/gardener-extensions/pkg/webhook/controlplane/etcd.go#L65

charts/etcd/values.yaml
```diff
@@ -31,8 +32,8 @@ etcd:
 
 backup:
   pullPolicy: IfNotPresent
-  snapstoreTempDir: "/tmp"
-  etcdConnectionTimeout: 30
+  snapstoreTempDir: "/var/etcd/data/temp"
```
rfranzke, line 35: https://github.com/gardener/gardener-extension-provider-aws/blob/v1.4.0/vendor/github.com/gardener/gardener-extensions/pkg/webhook/controlplane/etcd.go#L69

go.mod
```diff
@@ -35,6 +35,7 @@ require (
 	k8s.io/klog v1.0.0 // indirect
 	k8s.io/utils v0.0.0-20191218082557-f07c713de883 // indirect
 	sigs.k8s.io/controller-runtime v0.4.0
+	sigs.k8s.io/controller-tools v0.2.4 // indirect
```
swapnilgm, line 38: Any idea from where this is coming?

controllers/etcd_controller.go
```diff
@@ -808,7 +808,7 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
 		etcdValues["image"] = etcd.Spec.Etcd.Image
 	}
 
-	var quota int64 = 2 * 1024 * 1024 * 1024 // 2Gib
+	var quota int64 = 8 * 1024 * 1024 * 1024 // 8Gi
```
swapnilgm, line 811: Suggestion for line below:
```suggestion
	var deltaSnapshotMemoryLimit int64 = 104857600 //100Mib
```

go.mod
rfranzke: I ran `make generate`

controllers/etcd_controller.go
rfranzke: done

Issue #47, opened 2020-04-21
[BUG] HVPA not able to scale etcd.


KesavanKing:
**Describe the bug:**
VPA recommender misses the permission to get scale subresource because of which VPA on etcd is not happening  

**Expected behavior:**
As load increases based on VPA recommendations , etcd should be scaled.

**How To Reproduce (as minimally and precisely as possible):**

**Logs:**

**Screenshots (if applicable):**

**Environment (please complete the following information):**
 - Etcd version/commit ID : 
 - Etcd-druid version/commit ID :
 - Cloud Provider [All/AWS/GCS/ABS/Swift/OSS]: All

**Anything else we need to know?**:

swapnilgm:
Fixed via #48

PR #48, opened 2020-04-23
Removed owner reference from etcd statefulsets

georgekuruvillak:
This PR removes the dependance on ownerReference in etcd statefulset. It has been observerd that ownerReference in statefulset prevents VPA from recommending values for vertical scaling. 
Fixes #
#47

**Release note**:

```improvement operator
Removed owner reference from etcd StatefulSet so that HVPA can recommend resource recommendations. VPA does not support StatefulSet having `ownerReferences` set to another top-level controller .
```


controllers/etcd_controller.go
```diff
@@ -262,10 +275,13 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
-		finalizers := sets.NewString(etcd.Finalizers...)
+		// Deep copy of etcd resource required here to patch the object. Update call results in
+		// StorageError. See also: https://github.com/kubernetes/kubernetes/issues/71139
+		etcdCopy := etcd.DeepCopy()
+		finalizers := sets.NewString(etcdCopy.Finalizers...)
 		finalizers.Delete(FinalizerName)
-		etcd.Finalizers = finalizers.UnsortedList()
-		if err := r.Update(ctx, etcd); client.IgnoreNotFound(err) != nil {
+		etcdCopy.Finalizers = finalizers.UnsortedList()
+		if err := r.Patch(ctx, etcdCopy, client.MergeFrom(etcd)); client.IgnoreNotFound(err) != nil {
```
amshuman-kr, line 284: Why not an an `Update` here?

georgekuruvillak: As mentioned in the comment, for a resource with deletiontimestamp set, in place patch is fine. Update results in error as it does a get first

controllers/etcd_controller.go
```diff
@@ -764,6 +775,20 @@ func checkForEtcdOwnerReference(refs []metav1.OwnerReference, etcd *druidv1alpha
 	return false
 }
 
+func checkForEtcdAnnotations(annotations map[string]string, etcd metav1.Object) bool {
+	var ownedBy, ownerType string
+	var ok bool
+	if ownedBy, ok = annotations[common.GardenerOwnedBy]; !ok {
```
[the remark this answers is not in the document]
georgekuruvillak: annotations is not a pointer so it cannot be nil, ryt?

controllers/controller_ref_manager.go
```diff
@@ -230,10 +303,25 @@ func (m *EtcdDruidRefManager) AdoptStatefulSet(ss *appsv1.StatefulSet) error {
...
+		}
+		ownersCopy = append(ownersCopy, *owner)
+	}
+	if len(ownersCopy) == 0 {
+		ssClone.OwnerReferences = nil
+	} else {
+		ssClone.OwnerReferences = ownersCopy
 	}
 
+	ssClone.Annotations[common.GardenerOwnedBy] = fmt.Sprintf("%s/%s", m.Controller.GetNamespace(), m.Controller.GetName())
```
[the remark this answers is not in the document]
georgekuruvillak: annotations is not a pointer so it cannot be nil, ryt?

controllers/etcd_controller_test.go
```diff
@@ -106,7 +105,7 @@ var _ = Describe("Druid", func() {
 			s := &appsv1.StatefulSet{}
 			getStatefulset(c, instance, s)
 			testLog.Info("fetched Statefulset", "statefulset", s.Name)
-			Eventually(len(s.OwnerReferences), timeout).ShouldNot(BeZero())
+			Eventually(checkForEtcdAnnotations(s.GetAnnotations(), instance), timeout).Should(BeTrue())
```
swapnilgm, line 108: Make `checkForEtcdAnnotations ` public in controllers package. And keep the package names here as `controllers_test`

controllers/suite_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package controllers_test
+package controllers
```
swapnilgm, line 15: See comment above.

controllers/etcd_controller_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package controllers_test
+package controllers
```
swapnilgm, line 15: See comment below.

controllers/controller_ref_manager.go
```diff
@@ -434,29 +346,74 @@ func (m *EtcdDruidRefManager) ClaimConfigMaps(cms *corev1.ConfigMapList, filters
...
+	default:
+		return fmt.Errorf("cannot adopt resource: %s", objType)
 	}
-
-	cmClone := cm.DeepCopy()
-	// Note that ValidateOwnerReferences() will reject this patch if another
-	// OwnerReference exists with controller=true.
-	if err := controllerutil.SetControllerReference(m.Controller, cmClone, m.reconciler.Scheme); err != nil {
-		return err
+	annotations := clone.GetAnnotations()
```
swapnilgm, line 391: Please move the annotation related code to only  statefulset case. And also put the reason behind this logic in comment.

controllers/etcd_controller_test.go
georgekuruvillak: I made this change so that I could unit test private functions as well. Ill mark the private functions in the test files with a `fortest` suffix. WDYT?

ggaurav10:
@georgekuruvillak and I have already tested working from VPA/HVPA side, and it worked as expected.

controllers/suite_test.go
georgekuruvillak: lets keep it the same package unless we cannot go ahead this way.

controllers/etcd_controller_test.go
georgekuruvillak: lets keep it the same package unless we cannot go ahead this way.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+
+			})
+			It("should adopt and delete statefulset", func() {
+				defer WithWd("..")()
+				err = c.Create(context.TODO(), instance)
+				Expect(err).NotTo(HaveOccurred())
+				err = getReconciledStatefulset(c, instance, &sts)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(sts.Name).ShouldNot(BeEmpty())
+				Expect(checkEtcdAnnotations(sts.GetAnnotations(), instance)).Should(BeTrue())
```
amshuman-kr, line 331: Please add a check for ownerReferences here (before and after).

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+
+	Context("when deleting etcd resources", func() {
 		var err error
 		var instance *druidv1alpha1.Etcd
 		var c client.Client
-		var ss *appsv1.StatefulSet
-		BeforeEach(func() {
-			instance = getEtcd("foo3", "default")
+		//Deletion of etcd resource deployment with existing statefulsets with owner reference.
+		Context("when  statefulset without ownerReference and without owner annotations", func() {
```
amshuman-kr, line 273:
```suggestion
		Context("when  statefulset with ownerReference and without owner annotations", func() {
```

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
-			c.Create(context.TODO(), ss)
+			})
+			It("should adopt and delete statefulset", func() {
+				defer WithWd("..")()
+				err = c.Create(context.TODO(), instance)
+				Expect(err).NotTo(HaveOccurred())
+				err = getReconciledStatefulset(c, instance, &sts)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(sts.Name).ShouldNot(BeEmpty())
+				Expect(checkEtcdAnnotations(sts.GetAnnotations(), instance)).Should(BeTrue())
```
amshuman-kr, line 293: Please add a check for ownerReferences here (before and after).

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				err = retryTillStatefulSetRemoved(c, &sts)
+				Expect(err).NotTo(HaveOccurred())
+				err = retryTillEtcdRemoved(c, instance)
+				Expect(err).NotTo(HaveOccurred())
+			})
 		})
-		AfterEach(func() {
-			c.Delete(context.TODO(), instance)
+		//Deletion of etcd resource with existing statefulsets without owner reference and with owner annotations.
+		Context("when  statefulset without ownerReference and without owner annotations", func() {
```
amshuman-kr, line 349:
```suggestion
		Context("when  statefulset without ownerReference and with owner annotations", func() {
```

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				storeSecret := instance.Spec.Backup.Store.SecretRef.Name
+				errors := createSecrets(c, instance.Namespace, storeSecret)
+				Expect(len(errors)).Should(BeZero())
+
+			})
+			It("should restart pod", func() {
+				defer WithWd("..")()
+				err = c.Create(context.TODO(), instance)
+				Expect(err).NotTo(HaveOccurred())
+				_, err = retryTillPodsDeleted(c, instance)
```
amshuman-kr, line 254: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				err = c.Status().Update(context.TODO(), &sts)
+				Expect(err).NotTo(HaveOccurred())
+
+			})
+			AfterEach(func() {
+				// sleep is required here to ensure that etcd is reconciled and finalizer added before it is deleted.
+				time.Sleep(2 * time.Second)
+				err = c.Delete(context.TODO(), instance)
+				Expect(err).NotTo(HaveOccurred())
+				err = retryTillStatefulSetRemoved(c, &sts)
```
amshuman-kr, line 304: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				err = retryTillEtcdRemoved(c, instance)
```
amshuman-kr, line 306: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
-			getStatefulset(c, instance, s)
-			testLog.Info("fetched Statefulset", "statefulset", s.Name)
-			Eventually(len(s.OwnerReferences), timeout).ShouldNot(BeZero())
+			})
+			AfterEach(func() {
+				// sleep is required here to ensure that etcd is reconciled and finalizer added before it is deleted.
+				time.Sleep(2 * time.Second)
+				err = c.Delete(context.TODO(), instance)
+				Expect(err).NotTo(HaveOccurred())
+				err = retryTillStatefulSetRemoved(c, &sts)
```
amshuman-kr, line 342: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				err = retryTillEtcdRemoved(c, instance)
```
amshuman-kr, line 344: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				err = c.Status().Update(context.TODO(), &sts)
+				Expect(err).NotTo(HaveOccurred())
+
+			})
+			AfterEach(func() {
+				// sleep is required here to ensure that etcd is reconciled and finalizer added before it is deleted.
+				time.Sleep(2 * time.Second)
+				err = c.Delete(context.TODO(), instance)
+				Expect(err).NotTo(HaveOccurred())
+				err = retryTillStatefulSetRemoved(c, &sts)
```
amshuman-kr, line 380: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				err = retryTillEtcdRemoved(c, instance)
```
amshuman-kr, line 382: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+			})
 		})
 	})
-
 })
 
-func getStatefulset(c client.Client, instance *druidv1alpha1.Etcd, ss *appsv1.StatefulSet) {
+func retryTillPodsDeleted(c client.Client, etcd *druidv1alpha1.Etcd) (*corev1.Pod, error) {
```
amshuman-kr, line 389: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+		}
+		return false, nil
+	}, ctx.Done())
+	if err != nil {
+		return nil, err
+	}
+	return pod, err
+}
+
+func retryTillEtcdRemoved(c client.Client, etcd *druidv1alpha1.Etcd) error {
```
amshuman-kr, line 414: Please use `Eventually` instead.

controllers/etcd_controller_test.go
```diff
@@ -54,71 +81,384 @@ var _ = Describe("Druid", func() {
...
+				return true, nil
+			}
+			return false, err
+		}
+		return false, nil
+	}, ctx.Done())
+	return err
+}
+
+func retryTillStatefulSetRemoved(c client.Client, ss *appsv1.StatefulSet) error {
```
amshuman-kr, line 436: Please use `Eventually` instead.

api/v1alpha1/etcd_types.go
```diff
@@ -133,8 +133,8 @@ type EtcdSpec struct {
 	// It must match the pod template's labels.
 	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
 	Selector *metav1.LabelSelector `json:"selector"`
-	// +optional
-	Labels map[string]string `json:"labels,omitempty"`
+	// +required
```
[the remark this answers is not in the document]
georgekuruvillak: This is required as the selector is mandatory. Labels cannot be made optional as a result. In fact, I will have to add a validating webhook to prevent creation of the resource if the selector and labels mismatch. This is present for sts in k8s. Ill adopt that logic here.

controllers/controller_ref_manager.go
```diff
@@ -487,6 +405,23 @@ func (m *EtcdDruidRefManager) ReleaseConfigMap(cm *corev1.ConfigMap) error {
 	return err
 }
 
+func (m *EtcdDruidRefManager) disown(obj metav1.Object) {
```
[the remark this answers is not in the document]
georgekuruvillak: The idea behind disown() is to remove the ownerreferences. The annotation changes are kept in the AdoptResource and ReleaseResource explicitly for clarity. When we have HVPA support for CRs we could remove them.

controllers/controller_ref_manager.go
```diff
@@ -434,44 +327,69 @@ func (m *EtcdDruidRefManager) ClaimConfigMaps(cms *corev1.ConfigMapList, filters
...
-	cmClone := cm.DeepCopy()
-	// Note that ValidateOwnerReferences() will reject this patch if another
-	// OwnerReference exists with controller=true.
-	if err := controllerutil.SetControllerReference(m.Controller, cmClone, m.reconciler.Scheme); err != nil {
-		return err
+	var clone metav1.Object
+	switch objType := obj.(type) {
+	case *appsv1.StatefulSet:
+		clone = obj.(*appsv1.StatefulSet).DeepCopy()
+		m.disown(clone)
```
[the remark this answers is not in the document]
georgekuruvillak: When druid adopts a statefulset, it removes the owner references and adds the annotations instead. disown() removes the owner references.

api/v1alpha1/etcd_types.go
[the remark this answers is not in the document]
amshuman-kr: 👍 But as discussed offline, just making it as `required` doesn't prevent people from setting it to an empty map. That can only be addressed using a validating webhook which we can add separately in a different PR.

georgekuruvillak:
@amshuman-kr I have changed the tests to table tests. PTAL

controllers/etcd_controller_test.go
```diff
@@ -54,85 +90,229 @@ var _ = Describe("Druid", func() {
...
+			instance := getEtcd(name, "default", false)
+			c := mgr.GetClient()
+			switch setupStatefulSet {
+			case WithoutOwnerReference:
+				ss = createStatefulset(name, "default", instance.Spec.Labels)
+				Expect(ss.OwnerReferences).Should(BeNil())
+			case WithOwnerReference:
+				ss = createStatefulsetWithOwnerReference(instance)
+				Expect(len(ss.OwnerReferences)).ShouldNot(BeZero())
+			default:
```
amshuman-kr, line 123: Should we have a case for pre-existing `StatefulSet` with annotations? I know this won't happen when we make the next release. But it would be good to cover the case which would happen for every reconciliation after that. WDYT?

controllers/etcd_controller_test.go
```diff
@@ -54,85 +90,229 @@ var _ = Describe("Druid", func() {
...
+			Eventually(func() error { return statefulsetIsCorrectlyReconcilced(c, instance, s) }, timeout, pollingInterval).Should(BeNil())
+			setStatefulSetReady(s)
+			err = c.Status().Update(context.TODO(), s)
+			Expect(err).NotTo(HaveOccurred())
+			err = c.Delete(context.TODO(), instance)
+			Expect(err).NotTo(HaveOccurred())
+		},
+		Entry("when statefulset not owned by etcd, druid should adopt the statefulset", "foo20", WithoutOwnerReference, false),
+		Entry("when statefulset owned by etcd with owner reference set, druid should remove ownerref and add annotations", "foo21", WithOwnerReference, false),
+		Entry("when etcd has the spec changed, druid should reconcile statefulset", "foo3", WithoutOwnerReference, false),
```
amshuman-kr, line 144: Should we have a case for pre-existing `StatefulSet` with annotations? I know this won't happen when we make the next release. But it would be good to cover the case which would happen for every reconciliation after that. WDYT?

controllers/etcd_controller_test.go
```diff
@@ -54,85 +90,229 @@ var _ = Describe("Druid", func() {
...
+			// Object not found, return.  Created objects are automatically garbage collected.
+			// For additional cleanup logic use finalizers
+			return nil
+		}
+		return err
+	}
+	return fmt.Errorf("statefulset not removed")
+}
+
+func statefulsetIsCorrectlyReconcilced(c client.Client, instance *druidv1alpha1.Etcd, ss *appsv1.StatefulSet) error {
```
amshuman-kr, line 298:
```suggestion
func statefulsetIsCorrectlyReconciled(c client.Client, instance *druidv1alpha1.Etcd, ss *appsv1.StatefulSet) error {
```

controllers/etcd_controller_test.go
```diff
@@ -54,85 +90,229 @@ var _ = Describe("Druid", func() {
...
 			defer WithWd("..")()
-
 			err = c.Create(context.TODO(), instance)
 			Expect(err).NotTo(HaveOccurred())
-
 			ss := appsv1.StatefulSet{}
-			getStatefulset(c, instance, &ss)
-			testLog.Info("fetched Statefulset", "statefulset", ss.Name)
-			Eventually(ss.Name, timeout).ShouldNot(BeEmpty())
+			Eventually(func() error { return statefulsetIsCorrectlyReconcilced(c, instance, &ss) }, timeout, pollingInterval).Should(BeNil())
```
amshuman-kr, line 101:
```suggestion
			Eventually(func() error { return statefulsetIsCorrectlyReconciled(c, instance, &ss) }, timeout, pollingInterval).Should(BeNil())
```

controllers/etcd_controller_test.go
```diff
@@ -54,85 +90,229 @@ var _ = Describe("Druid", func() {
...
 			Expect(err).NotTo(HaveOccurred())
-
 			s := &appsv1.StatefulSet{}
-			getStatefulset(c, instance, s)
-			testLog.Info("fetched Statefulset", "statefulset", s.Name)
-			Eventually(len(s.OwnerReferences), timeout).ShouldNot(BeZero())
-		})
-		AfterEach(func() {
-			c.Delete(context.TODO(), instance)
+			Eventually(func() error { return statefulsetIsCorrectlyReconcilced(c, instance, s) }, timeout, pollingInterval).Should(BeNil())
```
amshuman-kr, line 135:
```suggestion
			Eventually(func() error { return statefulsetIsCorrectlyReconciled(c, instance, s) }, timeout, pollingInterval).Should(BeNil())
```

controllers/etcd_controller_test.go
```diff
@@ -54,85 +90,229 @@ var _ = Describe("Druid", func() {
...
+			})
+			It("should restart pod", func() {
+				defer WithWd("..")()
+				err = c.Create(context.TODO(), instance)
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(func() error { return podDeleted(c, instance) }, timeout, pollingInterval).Should(BeNil())
+			})
+			AfterEach(func() {
+				s := &appsv1.StatefulSet{}
+				Eventually(func() error { return statefulsetIsCorrectlyReconcilced(c, instance, s) }, timeout, pollingInterval).Should(BeNil())
```
amshuman-kr, line 190:
```suggestion
				Eventually(func() error { return statefulsetIsCorrectlyReconciled(c, instance, s) }, timeout, pollingInterval).Should(BeNil())
```

controllers/etcd_controller_test.go
```diff
@@ -54,85 +90,229 @@ var _ = Describe("Druid", func() {
...
+			}
+
+			storeSecret := instance.Spec.Backup.Store.SecretRef.Name
+			errors := createSecrets(c, instance.Namespace, storeSecret)
+			Expect(len(errors)).Should(BeZero())
+			c.Create(context.TODO(), ss)
+			defer WithWd("..")()
+			err := c.Create(context.TODO(), instance)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(func() error { return statefulsetIsCorrectlyReconcilced(c, instance, &sts) }, timeout, pollingInterval).Should(BeNil())
```
amshuman-kr, line 225:
```suggestion
			Eventually(func() error { return statefulsetIsCorrectlyReconciled(c, instance, &sts) }, timeout, pollingInterval).Should(BeNil())
```

controllers/etcd_controller_test.go
```diff
@@ -54,85 +90,229 @@ var _ = Describe("Druid", func() {
...
+			setStatefulSetReady(&sts)
+			err = c.Status().Update(context.TODO(), &sts)
+			Expect(err).NotTo(HaveOccurred())
+			err = c.Delete(context.TODO(), instance)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(func() error { return statefulSetRemoved(c, &sts) }, timeout, pollingInterval).Should(BeNil())
+			Eventually(func() error { return etcdRemoved(c, instance) }, timeout, pollingInterval).Should(BeNil())
+		},
+		Entry("when  statefulset with ownerReference and without owner annotations, druid should adopt and delete statefulset", "foo61", WithOwnerReference, false),
```
amshuman-kr, line 234: Should we have a case of pre-existing `StatefulSet` with no ownerReference and no annotations in which case the it should not be removed?

georgekuruvillak: Actually, if there is no owner reference/ owner annotations, the druid will adopt it if the selector match. This is because, in the pre druid scenario, we didnt have both and druid went ahead adopting it. Should we not continue the behavior?

controllers/etcd_controller_test.go
georgekuruvillak: Pondering upon how the test would be like, what would be the expected behavior of the test. That the annotations still persist after the reconcile?

controllers/etcd_controller_test.go
georgekuruvillak: Pondering upon how the test would be like, what would be the expected behavior of the test. That the annotations still persist after the reconcile?

amshuman-kr: Yes. Basically, the same `StatefulSet` is adopted and left unmodified (or updated if spec changed). That would be the usual case, right?

controllers/etcd_controller_test.go
amshuman-kr: But it won't adopt if the etcd being deletes, no? Then the `StatefulSet` would be left untouched. A negative case.

controllers/etcd_controller_test.go
georgekuruvillak: Yup. Handled that.

amshuman-kr: This change looks good.

controllers/etcd_controller_test.go
amshuman-kr: Can you please make the change for the negative case, I mentioned above? I did not mean the adoption+deletion case.

amshuman-kr: Based on offline conversation. We can take it forward to another PR since this requires reconcile call to be brought in into the test func.

PR #53, opened 2020-05-05
Revendoring to latest gardener and ci related changes

### LINKED ISSUE #50: [Feature] Automatic Update-PR script
This repository should also benefit from automatic update PRs of dependent components. etcd-druid deploys etcd-backup-restore, hence, when a new version of it is released then automatic update PRs should be opened by CI, similar to https://github.com/gardener/gardener/pull/2260.

Less manual actions.

You need such a script: https://github.com/gardener/gardener/blob/master/hack/.ci/set_dependency_version
You don't need to copy it but can also call it like the extensions as you already vendor `gardener/gardener`: https://github.com/gardener/gardener-extension-provider-aws/blob/master/.ci/set_dependency_version, https://github.com/gardener/gardener-extension-provider-aws/blob/master/hack/tools.go#L23

georgekuruvillak:
This PR revendors etcd-druid to use g/g v1.3.1. It updates the ci scripts.
Fixes #50 

**Release note**:

```improvement operator
NONE
```


.ci/component_descriptor
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
ialidzhikov, line 1: File is not executable
```
chmod +x .ci/component_descriptor
```

.ci/set_dependency_version
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+"$(dirname $0)"/../vendor/github.com/gardener/gardener/hack/.ci/set_dependency_version
```
ialidzhikov, line 3:
```
$ ls -la vendor/github.com/gardener/gardener/hack/.ci
ls: cannot access 'vendor/github.com/gardener/gardener/hack/.ci': No such file or directory
```

You need to run `make revendor`.

georgekuruvillak: Done

.ci/component_descriptor
georgekuruvillak: Done

PR #55, opened 2020-05-08
Remove the cluster autooscaler safe-to-evict  annotation from etcd sts.

### LINKED ISSUE #54: [Enhancement] let's not hardcode "cluster-autoscaler.kubernetes.io/safe-to-evict=false" for all etcd
etcd-druid should not hardcode the annotation "cluster-autoscaler.kubernetes.io/safe-to-evict=false"  on etcd pods and let user configure it using annotation field in CRD.

Not  every etcd is critical to system. The above mentioned annotation is specific to cluster-autoscaler and not etcd. Depending on the use of etcd CRD creator should have choice to add this annotation.
From gardener's POV, this etcd-main is critical but etcd-events is not that critical. So, the annotation should be set for etcd-main but not etcd-events. In future, we thought of deploying etcd for cilium networking extension, there also probably this annotation might not be required.
 
Remove the annotation from https://github.com/gardener/etcd-druid/blob/master/charts/etcd/templates/etcd-statefulset.yaml#L30.

swapnilgm:
Please check https://github.com/gardener/etcd-druid/issues/54 for description
Fixes #54

**Release note**:

```user operator
:warning: Etcd-druid NO MORE adds the annotation `"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"` to etcd pods. Please make use of `.spec.annotations` to configure such annotation.
```

### DIFF
--- a/charts/etcd/templates/etcd-statefulset.yaml
+++ b/charts/etcd/templates/etcd-statefulset.yaml
@@ -27,7 +27,6 @@ spec:
   template:
     metadata:
       annotations:
-        "cluster-autoscaler.kubernetes.io/safe-to-evict": "false"
         checksum/etcd-bootstrap-configmap: {{ include (print $.Template.BasePath "/etcd-bootstrap-configmap.yaml") . | sha256sum }}
 {{- if .Values.annotations }}
 {{ toYaml .Values.annotations | indent 8 }}


swapnilgm:
> Looks fine. Gardener should be modified to pass the annotation in the `Etcd` spec, when this change is integrated, right?

Yes. I have created issue https://github.com/gardener/gardener/issues/2304 to track this.

PR #59, opened 2020-06-16
Enhance etcd scale down behavior

shreyas-s-rao:
This PR adds handling of etcd scale down and enhances etcd deletion, by triggering full snapshot of the etcd data and then deleting the associated etcd data volume. Also adds `volumeRetentionPolicy` field to etcd spec to define the policy which druid will respect while deleting the PVCs during etcd scale down and deletion. Default behavior is `Delete`.

Fixes partially #2 

**Release note**:

```noteworthy user
On etcd scale down and etcd deletion, trigger full snapshot of etcd data and delete etcd data volume, according to newly added `volumeRetentionPolicy`. Default behavior is `Delete`.
```


controllers/etcd_controller.go
```diff
@@ -237,6 +251,8 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
 		}, err
 	}
 
+	// TODO: respect VolumeRetentionPolicy field
```
swapnilgm, line 254: It makes sense to expose this. Will you please update the CRD spec in same PR?

controllers/etcd_controller.go
```diff
@@ -1123,11 +1150,177 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
 		}
+		if ss.Status.Replicas != *ss.Spec.Replicas {
+			return gardenerretry.MinorError(fmt.Errorf("ss not scaled"))
+		}
 		return gardenerretry.Ok()
 	})
 	return ss, err
 }
 
+func (r *EtcdReconciler) prepareEtcdForScaleDown(etcd *druidv1alpha1.Etcd, ss *appsv1.StatefulSet, svc *v1.Service) (*corev1.PersistentVolumeClaimList, error) {
```
swapnilgm, line 1161: Let's follow `ctx` as parameter for new function we add where we require context.
```suggestion
func (r *EtcdReconciler) prepareEtcdForScaleDown(ctx Context.context, etcd *druidv1alpha1.Etcd, ss *appsv1.StatefulSet, svc *v1.Service) (*corev1.PersistentVolumeClaimList, error) {
```

controllers/etcd_controller.go
```diff
@@ -1123,11 +1150,177 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+			Transport: tr,
+		}
+		httpScheme = "https"
+	}
+
+	serviceEndpoint := fmt.Sprintf("%s://%s:%d", httpScheme, endpoint, httpPort)
+	return httpClient, serviceEndpoint, nil
+}
+
+func (r *EtcdReconciler) getClientTLSConfig(etcd *druidv1alpha1.Etcd) (*tls.Config, error) {
```
swapnilgm, line 1217: Same as above. Pass `ctx`.

controllers/etcd_controller.go
```diff
@@ -1123,11 +1150,177 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+	tlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		RootCAs:      caCertPool,
+	}
+	tlsConfig.BuildNameToCertificate()
+
+	return tlsConfig, nil
+}
+
+func (r *EtcdReconciler) makeBackupRestoreHTTPCall(httpClient *http.Client, serviceEndpoint, path string) ([]byte, error) {
```
swapnilgm, line 1255: Use context http call. And configure timeout for this call.

controllers/etcd_controller.go
```diff
@@ -1123,11 +1150,177 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+}
+
+func (r *EtcdReconciler) removePVCs(etcd *druidv1alpha1.Etcd, ss *appsv1.StatefulSet, pvcList *corev1.PersistentVolumeClaimList) error {
+	for _, pvc := range pvcList.Items {
+		pvName := pvc.Spec.VolumeName
+		if err := r.Delete(context.TODO(), &pvc); err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		// TODO: don't use namespace since PV is cluster-scoped
+		if err := r.waitForPVDeletion(pvName, pvc.Namespace); err != nil {
```
swapnilgm, line 1302: We should wait for PVC deletion. PV is abstracted by PVC.

shreyas-s-rao:
@swapnilgm Thanks for your review. I've addressed the changes. PTAL.

controllers/etcd_controller.go
```diff
@@ -215,6 +215,8 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
 		}
 	}
 
+	handleEtcdScaleDown := etcd.Spec.Replicas == 0 && etcd.Status.Replicas > 0 && etcd.Status.ReadyReplicas > 0
```
swapnilgm, line 218: Why is it dependent on status field? Status filed might not represent current state of system. It simply represent last reconciled state. Specifically in case of this field, there might be disparity between actual system state and recorded state under status.
So, simply check for `spec.replicase`.

controllers/etcd_controller.go
```diff
@@ -1123,11 +1199,63 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
 }
 
+func (r *EtcdReconciler) prepareEtcdForScaleDown(etcd *druidv1alpha1.Etcd, ss *appsv1.StatefulSet, svc *v1.Service) (*corev1.PersistentVolumeClaimList, error) {
+	// TODO: wait for etcd health to be false so that etcd is not serving traffic
+
+	if err := r.takeFullSnapshot(etcd, svc); err != nil {
+		return nil, err
+	}
+
+	selector, err := metav1.LabelSelectorAsSelector(ss.Spec.Selector)
```
swapnilgm, line 1217: Isn't statefulset identifies PVC by name?

controllers/etcd_controller.go
shreyas-s-rao: Scaledown logic should only be applied when etcd spec replicas is 0 but actual current replicas is 1. We can get this value only from status field, or else we'll have to do a manual check on the existence of the PVC and then decide to execute scaledown logic

controllers/etcd_controller.go
shreyas-s-rao: Getting PVC by labels seemed safer than getting it by name.

controllers/etcd_controller.go
swapnilgm: But then, what happens if the druid fails to update the status may be because of network issue or similar. And the spec.replica is scaled down to zero: It won't delete the PVC.
So, to prevent it, it's fine to do those check every time we find spec.replicas to zero.

controllers/etcd_controller.go
```diff
@@ -251,6 +271,42 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
+			Requeue: true,
+		}, err
+	}
+
+	pvcList := &v1.PersistentVolumeClaimList{}
+	if ss != nil {
+		pvcList, err = r.listPersistentVolumeClaims(ss)
+		if err != nil {
+			return ctrl.Result{
+				Requeue: true,
```
amshuman-kr, line 295: Shouldn't we use `RequeueAfter` here to allow for the statefulset controller to take action?

controllers/etcd_controller.go
```diff
@@ -215,7 +215,27 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
+		return ctrl.Result{
+			Requeue: true,
+		}, err
+	}
+
+	pvcList := &v1.PersistentVolumeClaimList{}
+	if ss != nil {
+		pvcList, err = r.listPersistentVolumeClaims(ss)
+		if err != nil {
+			return ctrl.Result{
```
amshuman-kr, line 230: Shouldn't we use `RequeueAfter` here to allow for the statefulset controller to take action?

controllers/etcd_controller.go
```diff
@@ -251,6 +271,42 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
+			}, err
+		}
+	}
+
+	if svc != nil && len(pvcList.Items) > 0 {
+		snapshotCtx, cancel := context.WithTimeout(ctx, DefaultTimeout)
+		defer cancel()
+		if err := r.takeFullSnapshot(snapshotCtx, etcd, svc); err != nil {
+			return ctrl.Result{
+				Requeue: true,
```
amshuman-kr, line 305: Shouldn't we use `RequeueAfter` here to allow for whatever transient problem that caused the error to go away?

controllers/etcd_controller.go
```diff
@@ -251,6 +271,42 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
 
+	svc, err := r.getService(ctx, etcd)
+	if err != nil {
+		logger.Error(err, "error getting service")
+		return ctrl.Result{
+			Requeue: true,
+		}, err
+	}
+
+	ss, err := r.getStatefulSet(ctx, etcd)
```
amshuman-kr, line 282: Can we find a way to modularize the code to avoid repetition of [this](https://github.com/gardener/etcd-druid/pull/59/files#diff-8e5c816f8b2f1adeef36a64d81b3e0f6R218-R234) here?

controllers/etcd_controller.go
```diff
@@ -251,6 +271,42 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
+			return ctrl.Result{
+				Requeue: true,
+			}, err
+		}
+	}
+
+	if svc != nil && len(pvcList.Items) > 0 {
+		snapshotCtx, cancel := context.WithTimeout(ctx, DefaultTimeout)
+		defer cancel()
+		if err := r.takeFullSnapshot(snapshotCtx, etcd, svc); err != nil {
```
amshuman-kr, line 303: Are we cutting off the ingress traffic to etcd before this?

controllers/etcd_controller.go
```diff
@@ -587,6 +644,17 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
 			return nil, err
 		}
 
+		// delete PVC on etcd scale down based on volumeRetentionPolicy
+		if handleScaleDown {
```
amshuman-kr, line 648: Can we find a way to modularize to avoid repetition [here](https://github.com/gardener/etcd-druid/pull/59/files#diff-8e5c816f8b2f1adeef36a64d81b3e0f6R300-R308)?

controllers/etcd_controller.go
```diff
@@ -262,7 +318,15 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
 		}, err
 	}
 
-	if err := r.removeFinalizersToDependantSecrets(ctx, etcd); err != nil {
+	if etcd.Spec.VolumeRetentionPolicy == nil || *etcd.Spec.VolumeRetentionPolicy == druidv1alpha1.VolumeRetentionPolicyDelete {
```
amshuman-kr, line 321: Is [this](https://github.com/gardener/etcd-druid/pull/59/files#diff-8e5c816f8b2f1adeef36a64d81b3e0f6R650) not required here?

controllers/etcd_controller.go
```diff
@@ -587,6 +644,17 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
...
 		}
 
+		// delete PVC on etcd scale down based on volumeRetentionPolicy
+		if handleScaleDown {
+			pvcList := &corev1.PersistentVolumeClaimList{}
+			if pvcList, err = r.prepareEtcdForScaleDown(etcd, ss, svc); err != nil {
+				return nil, err
+			}
+			if etcd.Spec.VolumeRetentionPolicy == nil || *etcd.Spec.VolumeRetentionPolicy == druidv1alpha1.VolumeRetentionPolicyDelete {
+				defer r.removePVCs(pvcList.Items)
```
amshuman-kr, line 654: What if there are errors while deleting PVCs? Shouldn't we `RequeueAfter`?

controllers/etcd_controller.go
```diff
@@ -262,7 +318,15 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
 		}, err
 	}
 
-	if err := r.removeFinalizersToDependantSecrets(ctx, etcd); err != nil {
+	if etcd.Spec.VolumeRetentionPolicy == nil || *etcd.Spec.VolumeRetentionPolicy == druidv1alpha1.VolumeRetentionPolicyDelete {
+		if err := r.removePVCs(pvcList.Items); err != nil {
+			return ctrl.Result{
+				Requeue: true,
```
amshuman-kr, line 324: Shouldn't we `RequeueAfter` to give time to whatever it is that is preventing the PVC deletion?

controllers/etcd_controller.go
```diff
@@ -1123,11 +1197,111 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+}
+
+func (r *EtcdReconciler) removePVCs(pvcList []corev1.PersistentVolumeClaim) error {
+	for _, pvc := range pvcList {
+		if err := r.Delete(context.TODO(), &pvc); err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		ctx, cancel := context.WithTimeout(context.TODO(), DefaultTimeout)
+		defer cancel()
+		if err := r.waitForPVCDeletion(ctx, pvc.Name, pvc.Namespace); err != nil {
```
amshuman-kr, line 1284: Isn't it better to fire all deletes and then `RequeueAfter` for the errors?

controllers/etcd_controller.go
```diff
@@ -1123,11 +1197,111 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
 		}
+		if ss.Status.Replicas != *ss.Spec.Replicas {
+			return gardenerretry.MinorError(fmt.Errorf("statefulset not scaled"))
+		}
 		return gardenerretry.Ok()
 	})
 	return ss, err
 }
 
+func (r *EtcdReconciler) getStatefulSet(ctx context.Context, etcd *druidv1alpha1.Etcd) (*appsv1.StatefulSet, error) {
```
amshuman-kr, line 1208: How does this interact with `claimStatefulSets`?

controllers/etcd_controller.go
```diff
@@ -1123,11 +1197,111 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+		return nil, err
+	}
+
+	if len(statefulSets.Items) > 0 {
+		return &statefulSets.Items[0], nil
+	}
+	return nil, nil
+}
+
+func (r *EtcdReconciler) getService(ctx context.Context, etcd *druidv1alpha1.Etcd) (*v1.Service, error) {
```
amshuman-kr, line 1226: How does this interact with `claimServices`?

api/v1alpha1/etcd_types.go
```diff
@@ -156,6 +165,9 @@ type EtcdSpec struct {
 	// VolumeClaimTemplate defines the volume claim template to be created
 	// +optional
 	VolumeClaimTemplate *string `json:"volumeClaimTemplate,omitempty"`
+	// VolumeRetentionPolicy defines the policy for retaining volumes on etcd scale-down and deletion
+	// +optional
```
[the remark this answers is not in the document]
swapnilgm: Not a strong opinion. I was thinking on the following lines. 
If we make it mutable,  it might result in ambiguous state representation. Like, If somebody scale down it initially without any retention. Then change the field value. It will see that even though retention policy is now to retain but the volumes are missing. Said that I agree to that counter argument that is result of user action we can't do anything about it. Hence no strong opinion.

But probably now you are thinking of "RetainAll" and "RetainMaster" (I am assuming master is leader in multi-node cluster), what if user hibernate when policy set to RetainMaster and the mutate field to RetainAll. The ambiguous state here is count of PVC is only 1 even though policy is RetailAll. 

So, user has to remember its history of action of CRD to know the reasoning for correct state of system in such scenarios.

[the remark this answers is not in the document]
swapnilgm: Also, regarding "VolumeClaimTemplate" its the name for volume. Statefulset will create PVC with this prefix. It this is changed, then there is possibility of orphan volumes. Again here PVC claim logic will delete those orphan volumes so its fine again. No strong opinion. So, fine let's keep it mutable.

controllers/etcd_controller.go
swapnilgm: Not yet in this PR. From gardeners CP perspective, this will happen since service will be brought down first. From stand alone etcd-druid's perspective, i was suggesting to reuse healthz controller logic from multi-node etcd pr later.

api/v1alpha1/etcd_types.go
```diff
@@ -30,6 +30,11 @@ const (
 	Basic MetricsLevel = "basic"
 	// Extensive is a constant for metrics level extensive.
 	Extensive MetricsLevel = "extensive"
+
+	// VolumeRetentionPolicyDelete defines the delete policy for retaining old volumes
+	VolumeRetentionPolicyDelete VolumeRetentionPolicy = "Delete"
+	// VolumeRetentionPolicyRetain defines the retain policy for retaining old volumes
+	VolumeRetentionPolicyRetain VolumeRetentionPolicy = "Retain"
```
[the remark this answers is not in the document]
shreyas-s-rao: Maybe this is something we should discuss. With `RetainMaster`, what would scale up look like for, say, a 3-node etcd cluster? Would we allow the 2 other etcd's to come up empty and then sync the data? Or would we follow our current approach of letting backup sidecar restore data on the other two nodes? If it's case 2, then I'm not sure we have much use for `RetainMaster`, since restoration is happening anyway on the other two etcd nodes. WDYT?

controllers/etcd_controller.go
shreyas-s-rao: We do have it. Druid will take a full snapshot before removing the statefulset, by calling `takeFullSnapshot()`. As per your suggestion, I've replaced it with `prepareEtcdForScaleDown()` for efficient code reuse.

controllers/etcd_controller.go
shreyas-s-rao: Since error handling for getting sts and listing PVCs is different (`Requeue` and `RequeueAfter`), combining the two operations into a single method would require a new switch/if condition or some extra code to handle the common error from the method, which might be unnecessary.

controllers/etcd_controller.go
shreyas-s-rao: Nice catch. I'll enhance the error handling here

controllers/etcd_controller.go
shreyas-s-rao: Yes, makes sense. Let me make that change as well. Thanks.

api/v1alpha1/etcd_types.go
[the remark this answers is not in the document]
amshuman-kr: Agreed about the history part which can be solved by introducing the `volumeRetentionPolicy` in the status, I think.

Also, good point about `RetainMaster` meaning the leader. For better terminology, we should call it `RetainLeader`.

api/v1alpha1/etcd_types.go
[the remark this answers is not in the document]
amshuman-kr: I was thinking of case 1 (letting others nodes come up empty) but not a hard assumption. The main point was to rename `Retain` as `RetainAll` to make the behaviour clear. We can introduce `RetainMaster` (or better `RetainLeader`) later if it makes sense.

controllers/etcd_controller.go
amshuman-kr: I was thinking, maybe the common mention might return the `ctrl.Result` in addition to error and then in the reconcile/delete functions we could return whatever result was returned in case of error. To use `Requeue` or `RequeueAfter` is dependent on the statefulset or PVCs and not on whether it is called from reconcile or delete.

shreyas-s-rao:
@swapnilgm @amshuman-kr Thanks for your reviews. I've addressed the comments. PTAL.

controllers/etcd_controller.go
```diff
@@ -191,86 +200,123 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		if err != nil {
+			return reconcileResultRequeueAfter, err
 		}
-		return ctrl.Result{
-			Requeue: true,
-		}, err
 	}
 
-	svc, ss, err := r.reconcileEtcd(etcd)
+	handleEtcdScaleDown := etcd.Spec.Replicas == 0 && len(pvcList.Items) != 0
```
amshuman-kr, line 243: Can't this check be done inside `reconcileEtcd`? Also is the `len(pvcList.Items) != 0` check sound? Isn't a check if the backup is switched on (storage-provider?) a better check? There could be possible use-cases where we want to take snapshots even with `Local` storage provider. WDYT?

controllers/etcd_controller.go
```diff
@@ -191,86 +200,123 @@ func (r *EtcdReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+
+	// take full snapshot only if etcd was previously reconciled successfully
+	if etcd.Status.LastOperation.Type == druidv1alpha1.LastOperationTypeReconcile && etcd.Status.LastOperation.State == druidv1alpha1.LastOperationStateSucceeded {
+		svc, err := r.getService(ctx, etcd)
+		if err != nil {
+			logger.Error(err, "error getting service")
+			return reconcileResultRequeue, err
+		}
+
+		if svc != nil && len(pvcList.Items) > 0 {
```
amshuman-kr, line 293: Same comments as [above](https://github.com/gardener/etcd-druid/pull/59/files#diff-8e5c816f8b2f1adeef36a64d81b3e0f6R243)

controllers/etcd_controller.go
amshuman-kr: The deletion and scale down cases have not really been modularized. Only the check is moved out. But the real logic is still repeated here and in `delete`. What I meant in the comment about is to modularize both the check as well as the snapshot + PVC deletion part.

controllers/etcd_controller.go
```diff
@@ -576,64 +613,93 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
...
 			if utils.IsPodInCrashloopBackoff(pod.Status) {
 				if err := r.Delete(context.TODO(), &pod); err != nil {
 					logger.Error(err, fmt.Sprintf("error deleting etcd pod in crashloop: %s/%s", pod.Namespace, pod.Name))
-					return nil, err
+					return nil, &reconcileResultRequeueAfter, err
+				}
+			}
+		}
+
+		s, err := r.waitUntilStatefulSetReady(ss)
```
amshuman-kr, line 662: Why wait? Why not simply `reconcileResultRequeueAfter`? Also, statefulset scale down should trigger the watch event anyway.

controllers/etcd_controller.go
```diff
@@ -576,64 +613,93 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
...
 		}
 		podList := &v1.PodList{}
 		if err := r.List(context.TODO(), podList, client.InNamespace(etcd.Namespace), client.MatchingLabelsSelector{Selector: selector}); err != nil {
-			return nil, err
+			return nil, &reconcileResultRequeue, err
 		}
 
 		for _, pod := range podList.Items {
 			if utils.IsPodInCrashloopBackoff(pod.Status) {
 				if err := r.Delete(context.TODO(), &pod); err != nil {
```
amshuman-kr, line 655: It would be better to aggregate the errors in the loop and then `reconcileResultRequeueAfter` once after the loop. This way when multi-node supports gets in we can get deletes for all crashing pods in one reconciliation instead of potentially one deletion per reconciliation in the worst case.

controllers/etcd_controller.go
```diff
@@ -576,64 +613,93 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
...
+
+		s, err := r.waitUntilStatefulSetReady(ss)
+		if err != nil {
+			return s, &reconcileResultRequeue, err
+		}
+
+		// delete PVC on etcd scale down based on volumeRetentionPolicy
+		if handleScaleDown {
+			if etcd.Spec.VolumeRetentionPolicy == nil || *etcd.Spec.VolumeRetentionPolicy == druidv1alpha1.VolumeRetentionPolicyDeleteAll {
+				if err = r.removePVCs(pvcList.Items); err != nil {
```
amshuman-kr, line 670: It would be better to aggregate the errors in the loop and then `reconcileResultRequeueAfter` once after the loop. This way when multi-node supports gets in we can get deletes for all PVCs in one reconciliation instead of potentially one deletion per reconciliation in the worst case.

controllers/etcd_controller.go
```diff
@@ -576,64 +613,93 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
...
 	if err == errorsutil.ErrPreconditionViolated {
 		logger.Infof("Statefulset %s precondition doesn't hold, skip updating it.", ss.Name)
 		err = nil
 	}
 	if err != nil {
-		return nil, err
+		return nil, &reconcileResultRequeue, err
+	}
+
+	sts, err := r.waitUntilStatefulSetReady(ss)
```
amshuman-kr, line 697: Is there a way to skip this wait?

controllers/etcd_controller.go
```diff
@@ -1066,23 +1137,29 @@ func canDeleteStatefulset(sts *appsv1.StatefulSet, etcd *druidv1alpha1.Etcd) boo
 
 }
 
-func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, sts *appsv1.StatefulSet, lastError error) error {
+func (r *EtcdReconciler) updateEtcdErrorStatus(etcd *druidv1alpha1.Etcd, sts *appsv1.StatefulSet, lastError error, description string, operationType druidv1alpha1.LastOperationType) error {
```
amshuman-kr, line 1140: Are there multiple possible values for `operationType`? If no, can be bake it into this func and avoid the extra parameter?

controllers/etcd_controller.go
```diff
@@ -1066,23 +1137,29 @@ func canDeleteStatefulset(sts *appsv1.StatefulSet, etcd *druidv1alpha1.Etcd) boo
...
 
 	if err := r.Status().Update(context.TODO(), etcd); err != nil && !errors.IsNotFound(err) {
 		return err
 	}
 	return r.removeOperationAnnotation(etcd)
 }
 
-func (r *EtcdReconciler) updateEtcdStatus(etcd *druidv1alpha1.Etcd, svc *corev1.Service, sts *appsv1.StatefulSet) error {
-
+func (r *EtcdReconciler) updateEtcdStatus(etcd *druidv1alpha1.Etcd, svc *corev1.Service, sts *appsv1.StatefulSet, description string, operationType druidv1alpha1.LastOperationType, operationProgress int) error {
```
amshuman-kr, line 1162: Same s [above](https://github.com/gardener/etcd-druid/pull/59/files#diff-8e5c816f8b2f1adeef36a64d81b3e0f6R1140).

controllers/etcd_controller.go
amshuman-kr: Can we not avoid this func? Ideally, resequencing the reconcile/delete flow can get us `reconcileStatefulSet` before PVC clean up and we could get the statetulset from there. For snapshot we do not need the statefulset or PVCs anyway as mentioned above.

controllers/etcd_controller.go
```diff
@@ -1123,11 +1208,142 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+	}
+
+	return pvcList, nil
+}
+
+func (r *EtcdReconciler) removePVCs(pvcList []corev1.PersistentVolumeClaim) error {
+	resultCh := make(chan utils.PVCDeletionResult, len(pvcList))
+
+	for _, pvc := range pvcList {
+		go func(resultCh chan<- utils.PVCDeletionResult, pvc v1.PersistentVolumeClaim) {
```
amshuman-kr, line 1295: I didn't mean that we should delete the PVCs in goroutines. Just that we should fire the deletes and aggregate the errors in a loop and then outside the loop check only once if there are errors and _enqueueAfter_.

controllers/etcd_controller.go
```diff
@@ -1123,11 +1208,142 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+		if result.Error != nil {
+			logger.Errorf("error in deletion pvc %s: %v", result.PVC.Name, result.Error)
+			err = fmt.Errorf("failed to delete pvc %s: %v", result.PVC.Name, result.Error)
+		}
+	}
+
+	return err
+}
+
+func (r *EtcdReconciler) waitForPVCDeletion(ctx context.Context, name, namespace string) error {
```
amshuman-kr, line 1333: Can we avoid this?

controllers/etcd_controller.go
```diff
@@ -1123,11 +1208,142 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+	return persistentVolumeClaimList, nil
+}
+
+func (r *EtcdReconciler) prepareEtcdForScaleDown(parentCtx context.Context, etcd *druidv1alpha1.Etcd, ss *appsv1.StatefulSet, svc *v1.Service) (*corev1.PersistentVolumeClaimList, error) {
+	// TODO: wait for etcd health to be false so that etcd is not serving traffic
+
+	if parentCtx == nil {
+		parentCtx = context.TODO()
+	}
+	ctx, cancel := context.WithTimeout(parentCtx, DefaultTimeout)
```
amshuman-kr, line 1277: Don't we need longer timeout here?

controllers/etcd_controller.go
```diff
@@ -576,64 +587,112 @@ func (r *EtcdReconciler) reconcileStatefulSet(cm *corev1.ConfigMap, svc *corev1.
...
+		for i, pod := range podList.Items {
 			if utils.IsPodInCrashloopBackoff(pod.Status) {
 				if err := r.Delete(context.TODO(), &pod); err != nil {
 					logger.Error(err, fmt.Sprintf("error deleting etcd pod in crashloop: %s/%s", pod.Namespace, pod.Name))
-					return nil, err
+					errs[i] = err
 				}
 			}
 		}
+		for _, err := range errs {
```
amshuman-kr, line 651: Why not use something [errors aggregate](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/errors/errors.go#L35)?

amshuman-kr (changes requested):
Thanks for change, @shreyas-s-rao!

The PR looks good. Could you please add tests cases to cover the volume deletion (both `Etcd` scale down and deletion scenarios)? Integration or unit tests. Whichever is easier for you to add.

amshuman-kr (approved):
LGTM. The below pending issues can be addressed in other PRs. But we should not make a release without the integration tests.

1. Integration tests to cover the scale down and deletion scenarios.
1. Avoid the `waitUntilStatefulSetReady` in favour of `reconcileAfter`.
1. Streamline `getStatefulSet` with the rest of the claim logic.

ishan16696:
Given the `multi-node etcd` scenarios the scaling down in `multi-node etcd` will be very different from this scaling down.
I guess we can close this PR.

timuthy:
@shreyas-s-rao what do we do with this PR? It's getting dusty 😅

ishan16696:
Given the `multi-node etcd` the feature of scaling down the multi-node etcd will be different from the scaling down of a single node etcd and design of scale-down feature of multi-node etcd hasn’t been discussed yet.
Hence, closing this PR. Please feel free to open it you feel otherwise.
/close

PR #60, opened 2020-06-18
Fix the deployment target

swapnilgm:
Signed-off-by: Swapnil Mhamane <swapnil.mhamane@sap.com>

`make deploy` was failing to deploy druid because wrongly configured kustomization.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
-namespace: etcd-druid-system
+namespace: garden
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
@@ -12,21 +12,20 @@ namePrefix: etcd-druid-
 #commonLabels:
 #  someName: someValue
 
-bases:
-- ../crd
-- ../rbac
-- ../manager
+resources:
+  - ../crd
+  - ../rbac
+  - ../manager
 
 patches:
-- manager_image_patch.yaml
+  - manager_image_patch.yaml
   # Protect the /metrics endpoint by putting it behind auth.
   # Only one of manager_auth_proxy_patch.yaml and
   # manager_prometheus_metrics_patch.yaml should be enabled.
-#- manager_auth_proxy_patch.yaml
+  #- manager_auth_proxy_patch.yaml
   # If you want your controller-manager to expose the /metrics
   # endpoint w/o any authn/z, uncomment the following line and
   # comment manager_auth_proxy_patch.yaml.
   # Only one of manager_auth_proxy_patch.yaml and
   # manager_prometheus_metrics_patch.yaml should be enabled.
-- manager_prometheus_metrics_patch.yaml
-
+  - manager_prometheus_metrics_patch.yaml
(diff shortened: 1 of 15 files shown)


config/default/kustomization.yaml
```diff
@@ -12,21 +12,20 @@ namePrefix: etcd-druid-
 #commonLabels:
 #  someName: someValue
 
-bases:
-- ../crd
-- ../rbac
-- ../manager
+resources:
```
georgekuruvillak, line 15: Each entry in this list must resolve to an existing resource definition in YAML.  These are the resource files that kustomize reads, modifies and emits as a YAML string, with resources separated by document markers ("---"). For bases, thats not the case.

Checkout: https://kubectl.docs.kubernetes.io/pages/examples/kustomize.html

config/default/kustomization.yaml
```diff
@@ -12,21 +12,20 @@ namePrefix: etcd-druid-
...
+  - ../crd
+  - ../rbac
+  - ../manager
 
 patches:
-- manager_image_patch.yaml
+  - manager_image_patch.yaml
```
georgekuruvillak, line 21: Shouldn't this be a list in patches. Why do we need this intendation?

config/default/kustomization.yaml
```diff
@@ -12,21 +12,20 @@ namePrefix: etcd-druid-
...
   # Protect the /metrics endpoint by putting it behind auth.
   # Only one of manager_auth_proxy_patch.yaml and
   # manager_prometheus_metrics_patch.yaml should be enabled.
-#- manager_auth_proxy_patch.yaml
+  #- manager_auth_proxy_patch.yaml
```
georgekuruvillak, line 25: same as above

config/default/manager_image_patch.yaml
```diff
@@ -2,11 +2,11 @@ apiVersion: apps/v1
...
+  namespace: system
 spec:
   template:
     spec:
       containers:
-      # Change the value of image field below to your controller image URL
-      - image: eu.gcr.io/gardener-project/gardener/etcd-druid:v0.2.0-dev
-        name: druid
+        # Change the value of image field below to your controller image URL
+        - image: eu.gcr.io/gardener-project/gardener/etcd-druid:v0.4.0-dev
```
georgekuruvillak, line 11: same as above

georgekuruvillak:
Please check the indentations in the yaml files.

config/default/kustomization.yaml
swapnilgm: Bases is deprecated. https://kubernetes-sigs.github.io/kustomize/api-reference/kustomization/bases/

config/default/kustomization.yaml
swapnilgm: Thats was default indentation for yaml configured in vscode. And we use same accross repos. I didn't modified it explicitly. And irrespective of that, i do prefer list to be indent, though yaml allows unindented list.

config/default/kustomization.yaml
swapnilgm: https://github.com/gardener/etcd-druid/pull/60#discussion_r442632871

config/default/manager_image_patch.yaml
swapnilgm: https://github.com/gardener/etcd-druid/pull/60#discussion_r442632871

PR #132, opened 2021-02-04
Added the design doc for backup compation support.

abdasgupta:
This is the design doc for adding compaction support to ETCD Backup Restore.

Fixes #

**Release note**:

### DIFF
--- a/docs/compactor.md
+++ b/docs/compactor.md
@@ -1,0 +1,124 @@
+# Backup Compaction for ETCD
+
+## Current Problem
+To ensure recoverability of ETCD, backups of the database are taken at regular interval.
+Backups are of two types: Full Snapshots and Incremental Snapshots. 
+
+### Full Snapshots
+Full snapshot is a snapshot of the complete database at given point in time.The size of the database keeps changing with time and typically the size is relatively large (measured in 100s of megabytes or even in gigabytes. For this reason, full snapshots are taken after some large intervals.
+
+### Incremental Snapshots
+Incremental Snapshots are collection of events on ETCD database, obtained through running WATCH API Call on ETCD. After some short intervals, all the events that are accumulated through WATCH API Call are saved in a file and named as Incremental Snapshots at relatively short time intervals.
+
+### Recovery from the Snapshots
+
+#### Recovery from Full Snapshots
+As the full snapshots are snapshots of the complete database, the whole database can be recovered from a full snapshot in one go. ETCD provides API Call to restore the database from a full snapshot file.
+
+#### Recovery from Incremental Snapshots
+Delta snapshots are collection of retrospective ETCD events. So, to restore from Incremental snapshot file, the events from the file are needed to be applied sequentially on ETCD database through ETCD Put/Delete API calls. As it is heavily dependent on ETCD calls sequentially, restoring from Incremental Snapshot files can take long if there are numerous commands captured in Incremental Snapshot files.
+
+Delta snapshots are applied on top of running ETCD database. So, if there is inconsistency between the state of database at the point of applying and the state of the database when the delta snapshot commands were captured, restoration will fail.
+
+Currently, in Gardener setup, ETCD is restored from the last full snapshot and then the delta snapshots, which were captured after the last full snapshot.
+
+The main problem with this is that the complete restoration time can be unacceptably large if the rate of change coming into the etcd database is quite high because there are large number of events in the delta snapshots to be applied sequentially.
+A secondary problem is that, though auto-compaction is enabled for etcd, it is not quick enough to compact all the changes from the incremental snapshots being re-applied during the relatively short period of time of restoration (as compared to the actual period of time when the incremental snapshots were accumulated). This may lead to the etcd pod (the backup-restore sidecar container, to be precise) to run out of memory and/or storage space even if it is sufficient for normal operations.
+
+## Solution
+### Compaction command
+To help with the problem mentioned earlier, our proposal is to introduce `compact` subcommand with `etcdbrctl`. On execution of `compact` command, A separate embedded ETCD process will be started  where the ETCD data will be restored from the snapstore (exactly as in the restoration scenario today). Then the new ETCD database will be compacted and defragmented using ETCD API calls. The compaction will strip off the ETCD database of old revisions as per the ETCD auto-compaction configuration. The defragmentation will free up the unused fragment memory space released after compaction. Then a full snapshot of the compacted database will be saved in snapstore which then can be used as the base snapshot during any subsequent restoration (or backup compaction).
+
+### How the solution works
+The newly introduced compact command does not disturb the running ETCD while compacting the backup snapshots. The command is designed to run potentially separately (from the main ETCD process/container/pod). ETCD Druid can be configured to periodically schedule the newly introduced compact command as a separate job (scheduled periodically) based on a parameter which is defined in ETCD CRD as  `backupCompactionSchedule`.
+
+### Example ETCD CRD: 
+```yaml
+apiVersion: druid.gardener.cloud/v1alpha1
+kind: Etcd
+...
+spec:
+  ...
+  backup:
+    ...
+    backupCompactionSchedule: "0/30 * * * *" # Backup compaction job is executed every 30m on the clock
+    ...
+  ...
+```
+
+### Example Cron Job:
+An example can be found [here](https://github.com/gardener/etcd-druid/blob/master/charts/etcd/templates/etcd-compaction-cronjob.yaml)
+
+### **Points to take care while saving the compacted snapshot:**
+As compacted snapshot and the existing periodic full snapshots are taken by different processes running in different pods but accessing same store to save the snapshots, some problems may arise:
+1. When uploading the compacted snapshot to the snapstore, there is the problem of how does the restorer know when to start using the newly compacted snapshot. This communication needs to be atomic.
+2. With a regular schedule for compaction that happens potentially separately from the main etcd pod, is there a need for regular scheduled full snapshots anymore?
+3. We are planning to introduce new directory structure, under v2 prefix, for saving the snapshots (compacted and full), as mentioned in details below. But for backward compatibility, we also need to consider the older directory, which is currently under v1 prefix, during accessing snapshots.
+
+#### **How to swap full snapshot with compacted snapshot atomically**
+
+Currently, full snapshots and the subsequent delta snapshots are grouped under same prefix path in the snapstore. When a full snapshot is created, it is placed under a prefix/directory with the name comprising of timestamp. Then subsequent delta snapshots are also pushed into the same directory. Thus each prefix/directory contains a single full snapshot and the subsequent delta snapshots. So far, it is the job of ETCDBR to start main ETCD process and snapshotter process which takes full snapshot and delta snapshot periodically. But as per our proposal, compaction will be running as parallel process to main ETCD process and snapshotter process. So we can't reliably co-ordinate between the processes to achieve switching to the compacted snapshot as the base snapshot atomically.
+
+##### **Current Directory Structure**
+```yaml
+- Backup-192345
+    - Full-Snapshot-0-1-192345
+    - Incremental-Snapshot-1-100-192355
+    - Incremental-Snapshot-100-200-192365
+    - Incremental-Snapshot-200-300-192375
+- Backup-192789
+    - Full-Snapshot-0-300-192789
+    - Incremental-Snapshot-300-400-192799
+    - Incremental-Snapshot-400-500-192809
+    - Incremental-Snapshot-500-600-192819
+```
+
+To solve the problem, proposal is:
+1. ETCDBR will take the first full snapshot after it starts main ETCD Process and snapshotter process. After taking the first full snapshot, snapshotter will continue taking full snapshots. On the other hand, ETCDBR compactor command will be run as periodic job in a separate pod and use the existing full or compacted snapshots to produce further compacted snapshots. Full snapshots and compacted snapshots will be named after same fashion. So, there is no need of any mechanism to choose which snapshots(among full and compacted snapshot) to consider as base snapshots. 
+2. Flatten the directory structure of backup folder. Save all the full snapshots, delta snapshots and compacted snapshots under same directory/prefix. Restorer will restore from full/compacted snapshots and delta snapshots sorted based on the revision numbers in name (or timestamp if the revision numbers are equal).
+
+##### **Proposed Directory Structure**
+```yaml
+Backup :
+    - Full-Snapshot-0-1-192355 (Taken by snapshotter)
+    - Incremental-Snapshot-revision-1-100-192365
+    - Incremental-Snapshot-revision-100-200-192375
+    - Full-Snapshot-revision-0-200-192379 (Taken by snapshotter)
+    - Incremental-Snapshot-revision-200-300-192385
+    - Full-Snapshot-revision-0-300-192386 (Taken by compaction job)
+    - Incremental-Snapshot-revision-300-400-192396
+    - Incremental-Snapshot-revision-400-500-192406
+    - Incremental-Snapshot-revision-500-600-192416
+    - Full-Snapshot-revision-0-600-192419 (Taken by snapshotter)
+    - Full-Snapshot-revision-0-600-192420 (Taken by compaction job)
+```
+
+##### What happens to the delta snapshots that were compacted?
+The proposed `compaction` sub-command in `etcdbrctl` (and hence,  the `CronJob` provisioned by `etcd-druid` that will schedule it at a regular interval) would only upload the compacted full snapshot.
+It will not delete the snapshots (delta or full snapshots) that were compacted.
+These snapshots which were superseded by a freshly uploaded compacted snapshot would follow the same life-cycle as other older snapshots.
+I.e. they will be garbage collected according to the configured backup snapshot retention policy.
+For example, if an `exponential` retention policy is configured and if compaction is done every `30m` then there might be at most `48` additional (compacted) full snapshots (`24h * 2`) in the backup for the latest day. As time rolls forward to the next day, these additional compacted snapshots (along with the delta snapshots that were compacted into them) will get garbage collected retaining only one full snapshot for the day before according to the retention policy.
+
+##### **Future work**
+In future, we have plan to stop the snapshotter just after taking the first full snapshot. Then, the compaction job will be solely responsible for taking subsequent full snapshots. The directory structure would be looking like following:
+
+```yaml
+Backup :
+    - Full-Snapshot-0-1-192355 (Taken by snapshotter)
+    - Incremental-Snapshot-revision-1-100-192365
+    - Incremental-Snapshot-revision-100-200-192375
+    - Incremental-Snapshot-revision-200-300-192385
+    - Full-Snapshot-revision-0-300-192386 (Taken by compaction job)
+    - Incremental-Snapshot-revision-300-400-192396
+    - Incremental-Snapshot-revision-400-500-192406
+    - Incremental-Snapshot-revision-500-600-192416
+    - Full-Snapshot-revision-0-600-192420 (Taken by compaction job)
+```
+
+#### Backward Compatibility
+1. **Restoration** : The changes to handle the newly proposed backup directory structure must be backward compatible with older structures at least for restoration because we need have to restore from backups in the older structure. This includes the support for restoring from a backup without a metadata file if that is used in the actual implementation.
+2. **Backup** : For new snapshots (even on a backup containing the older structure), the new structure may be used. The new structure must be setup automatically including creating the base full snapshot.
+3. **Garbage collection** : The existing functionality of garbage collection of snapshots (full and incremental) according to the backup retention policy must be compatible with both old and new backup folder structure. I.e. the snapshots in the older backup structure must be retained in their own structure and the snapshots in the proposed backup structure should be retained in the proposed structure. Once all the snapshots in the older backup structure go out of the retention policy and are garbage collected, we can think of removing the support for older backup folder structure.
+
+**Note:** Compactor will run parallel to current snapshotter process and work only if there is any full snapshot already present in the store. By current design, a full snapshot will be taken if there is already no full snapshot or the existing full snapshot is older than 24 hours. It is not limitation but a design choice. As per proposed design, the backup storage will contain both periodic full snapshots as well as periodic compacted snapshot. Restorer will pickup the base snapshot whichever is latest one.


docs/compactor.md
```diff
@@ -0,0 +1,36 @@
+# Compaction for ETCD
+
+## Current Problem
+To ensure recoverability of ETCD, backups of the database are taken at regular interval.
+Backups are of two types: Full Snapshots and Incremental Snapshots. 
+
+**Full Snapshots:** Full snapshots are memory snapshots of the database. ETCD is an in-memory database, so the size of the database keeps growing with time. As a result, the size of the memory snapshots grows with time. Full Snapshots are taken after some large intervals.
```
amshuman-kr, line 7:
```suggestion
**Full Snapshots:** Full snapshot is a snapshot of the complete database at given point in time.The size of the database keeps changing with time and typically the size is relatively large (measured in 100s of megabytes or even in gigabytes. For this reason, full snapshots are taken after some large intervals.
```

docs/compactor.md
```diff
@@ -0,0 +1,36 @@
...
+
+**Incremental Snapshots:** Incremental Snapshots are collection of commands on ETCD database, obtained through running WATCH API Call on ETCD. After some short intervals, all the commands that are accumulated through WATCH API Call are saved in a file and named as Incremental Snapshots.
+
+### **Recovery from the Snapshots:**
+**Recovery from Full Snapshots:** As the full snapshots are just the snapshot of memory, it takes less time to recover from a full snapshot. Recovery only needs to copy the provided full snapshot file from disk to memory. ETCD provides API Call to restore the database from a full snapshot file.
```
amshuman-kr, line 12:
```suggestion
**Recovery from Full Snapshots:** As the full snapshots are snapshots of the complete database, the whole database can recoved form a full snapshot in one go. ETCD provides API Call to restore the database from a full snapshot file.
```

docs/compactor.md
```diff
@@ -0,0 +1,36 @@
...
+
+Delta snapshots are applied on top of running ETCD database. So, if there is inconsistency between the state of database at the point of applying and the state of the database when the delta snapshot commands were captured, restoration will fail.
+
+Currently, in Gardener setup, ETCD is restored from the last full snapshot and then the delta snapshots, which were captured after the last full snapshot.
+
+Now, the problem is that as times go by the size of the ETCD database increases. And the problem becomes more troubling as in the case of ETCD every change in ETCD database creates a new revision of the database. So, the Full Snapshot size also increases. But without compaction, the older revisions of ETCD database are also included in the snapshots. So even during restoration, ETCD is restored with older uncompacted huge database. Eventually, ETCD gets Out of Memory.
```
amshuman-kr, line 20:
```suggestion

The main problem with this is that the complete restoration time can be unacceptably long if the rate of change coming into the etcd database is quite high because there are large amount delta snapshots to be applied sequentially.

A secondary problem is that, though auto-compaction is enabled for etcd, the auto-compaction period it is not quick enough to compact all the changes from the incremental snapshots being re-applied during the relatively short period of time of restoration (as compared to the actual period of time when the incremental snapshots were accumulated). This may lead to the etcd pod (the backup-restore sidecar container, to be precise) to run out of memory and/or storage space even if it is sufficient for normal operations.
```

docs/compactor.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Solution
+**Compaction command:** To help with the problem mentioned earlier, the PR https://github.com/gardener/etcd-backup-restore/pull/301 introduces compaction command that reduces the size of full snapshot. So, after restoration from the full snapshot, ETCD database is stripped off of old revisions.
```
amshuman-kr, line 23: Can you please explain the solution without referring to the PR? Let's keep the documentation relevant even after the PR has been merged.

amshuman-kr, line 23: I am not sure if we should completely compact all revisions. Just honour the auto-compaction configuration.

docs/compactor.md
```diff
@@ -0,0 +1,36 @@
...
+
+**How the PR works:** The compact command added through PR does not disturb the running ETCD while captuing the compacted snapshot. The command is run separately alongside main ETCD process.
```
amshuman-kr, line 25:
```suggestion
**How the PR works:** The compact command added through PR does not disturb the running ETCD while captuing the compacted snapshot. The command is designed to run potentially separately (from the main ETCD process/container/pod). 
```

docs/compactor.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Solution
+**Compaction command:** To help with the problem mentioned earlier, the PR https://github.com/gardener/etcd-backup-restore/pull/301 introduces compaction command that reduces the size of full snapshot. So, after restoration from the full snapshot, ETCD database is stripped off of old revisions.
+
+**How the PR works:** The compact command added through PR does not disturb the running ETCD while captuing the compacted snapshot. The command is run separately alongside main ETCD process. 
+When the compact command is run, it fetches the latest full snapshot and delta snapshots from the store. Then it restores an embedded etcd server from the snapshots. Then it runs compact and defragment operation on embedded ETCD database. Then it takes the snapshot of the embedded ETCD database and save it into the store.
```
amshuman-kr, line 26:
```suggestion
When the compact command is run, it fetches the latest full snapshot and delta snapshots from the store. Then it restores an embedded etcd server from the snapshots (exactly the way the current restoration process does). Then it runs compact and defragment operation on embedded ETCD database. Then it takes the snapshot of the embedded ETCD database and save it into the store.
```

docs/compactor.md
```diff
@@ -0,0 +1,36 @@
...
+
+#### **Points to take care while saving the compacted snapshot:**
+As compacted snapshot and full snapshot are taken by different processes running in different pods but accessing same store to save the snapshots, some problems may arise:
+1.	If there are both full snapshots and compacted snapshots, a pointer is needed to let the restorer know which is what. We can come up with a metadata file that would point to correct snapshot file that can be used to restore.
+2.	If frequency of taking full snapshots is more than taking compacted snapshots, there might be situations when compacted snapshots be totally ignored while restoring as full snapshots might have latest timestamp.
+3.	Uploading the compacted snapshot to the store should be atomic.
```
amshuman-kr, line 32:
```suggestion
1.	When uploading the compacted snapshot to the snapstore, there is the problem of how does the restorer know when to start using the newly compacted snapshot. This communication needs to be atomic.
2.	With a regular schedule for compaction that happens potentially separately from the main etcd pod, is there a need for regular scheduled full snapshots anymore?
```

docs/compactor.md
```diff
@@ -0,0 +1,36 @@
...
+
+Our proposed approach is to remove the responsibility of taking full snapshot from ETCD Backup Restore. Only compactor (a job that will run compact command at regular interval) will take compacted snapshot and save to store. Currently full snapshots and the subsequent delta snapshots are grouped under same prefix. We are generating multiple prefixes named after the timestamps of full snapshots. Our proposal is to flatten the prefixes and save all compacted snapshots and delta snapshots under same prefix. Also, name snapshots after revision number instead of timestamps. Name a delta snapshot after the revision number of first command captured in that particular delta snapshot. Restorer will restore from compacted snapshots that have highest revision number in the name and delta snapshots that have higher revision numbers in name than the comapcted full snapshot.
+
+**Note:** ETCD Backup Restore needs to take one full snapshot just after starting. Compactor will work only if there is any full snapshot already present in the store. It is not limitation but a design choice
```
amshuman-kr, line 36: I have already suggested to limit the section above only to the _description_ of the problems arising out of compaction. It might be better to dedicate the rest of the section to the solutions for these problems. It might also be a good idea to split this into two separate sections. 
1. Folder structure and metadata change and rationale behind it. Preferably, with examples for existing and proposed structure/metadata and backward-compatibility. This can be referred to directly in the PR for folder structure change which can be release independent of compaction.
2. How the proposed folder structure/metadata will be used in compaction and what other changes are applicable (no scheduled full snapshot, first full snapshot still being required, backward compatibility for scheduled full snapshot for at least a release or two etc.)

amshuman-kr (changes requested):
Thanks for the write up @abdasgupta! Some comments/suggestions below.

docs/compactor.md
abdasgupta: One request: Can you point me to current auto-compaction configuration that we are using in ETCDBR ?

amshuman-kr: https://github.com/gardener/etcd-druid/blob/8307a62ad71ec0de3bf78de6d0e2bbc07700042c/charts/etcd/templates/etcd-bootstrap-configmap.yaml#L129-L130

But we should make this configurable or considerably reduce the retentions period.
cc @shreyas-s-rao

abdasgupta: Yes. Then we should keep the provision in PR. Right?

amshuman-kr: Do you mean the provision to compact?

abdasgupta: Should keep the provision to accept `auto-compaction-retention` value in PR https://github.com/gardener/etcd-backup-restore/pull/301 ?

abdasgupta:
@amshuman-kr made the necessary changes. Would you check?

docs/compactor.md
```diff
@@ -0,0 +1,86 @@
...
+2.	With a regular schedule for compaction that happens potentially separately from the main etcd pod, is there a need for regular scheduled full snapshots anymore?
+3.	Uploading the compacted snapshot to the store should be atomic.
+
+But new problems arise with the introduction of compaction - 
+
+Currently full snapshots and the subsequent delta snapshots are grouped under same prefix. When a full snapshot is created, it is placed under a prefix/directory with the name comprising of timestamp. Then subsequent delta snapshots are also pushed into the same directory. Thus each prefix/directory contains a full snapshot and subsequent delta snapshots. So far, it is the job of ETCDBR to start main ETCD process and snapshotter process which takes full snapshot and delta snapshot periodically. But as per our proposal, compaction will be running as parallel process to main ETCD process and snapshotter process. So we can't reliably sync compacted snapshots and subsequent delta snapshots under valid directory/prefix.
```
amshuman-kr, line 36:
```suggestion
Currently, full snapshots and the subsequent delta snapshots are grouped under same prefix path in the snapstore. When a full snapshot is created, it is placed under a prefix/directory with the name comprising of timestamp. Then subsequent delta snapshots are also pushed into the same directory. Thus each prefix/directory contains a single full snapshot and the subsequent delta snapshots. So far, it is the job of ETCDBR to start main ETCD process and snapshotter process which takes full snapshot and delta snapshot periodically. But as per our proposal, compaction will be running as parallel process to main ETCD process and snapshotter process. So we can't reliably co-ordinate between the processes to achieve switching to the compacted snapshot as the base snapshot atomically.```

File: docs/compactor.md [hunk 10]
```diff
@@ -0,0 +1,86 @@
...
+    * Incremental-Snapshot-192375
+* Backup-192789
+    * Full-Snapshot-192789
+    * Incremental-Snapshot-192799
+    * Incremental-Snapshot-192809
+    * Incremental-Snapshot-192819
+
+To solve the problem, proposal is:
+1. ETCDBR will take the first full snapshot after it starts main ETCD Process and snapshotter process. But after taking the first full snapshot, snapshotter will only continue taking delta snapshots but no further full snapshots.
+2. Flatten the directory structure of backup folder. Save all the full snapshots, delta snapshots and compacted snapshots under same directory/prefix. Also, name snapshots after revision number instead of timestamps. Restorer will restore from compacted snapshots that have highest revision number in the name and delta snapshots that have higher revision numbers in name than the comapcted full snapshot.
```
amshuman-kr, line 52: From the example below, I think you are using the revision of the first change in the incremental snapshot name. I think it might be more useful to use the revision of the last change. Even better, why not use both first and last revision in the incremental snapshot name? Redundant information is sometimes good :-)

docs/compactor.md
```diff
@@ -0,0 +1,86 @@
...
+    * Incremental-Snapshot-revision-23
+    * Incremental-Snapshot-revision-28
+    * Incremental-Snapshot-revision-33 (It containes command upto revision 38)
+    * Full-Snapshot-revision-38 (Compacted)
+    * Incremental-Snapshot-revision-39
+    * Incremental-Snapshot-revision-42
+    * Incremental-Snapshot-revision-43
+    * Incremental-Snapshot-revision-49
+
+3. A meta-data file might need to be added in the Backup directory to mark which compacted snapshot is ready to be used for restoration. Restorer would study the meta-data file and pickup the ready full snapshot from there and apply all the delta snapshots having revision number greater than that. So, with the backup directory as above, if the content of metadata file is like following:
```
amshuman-kr, line 72:
```suggestion
3. A meta-data file might need to be added in the Backup directory to mark which compacted (or full) snapshot is ready to be used as the base for restoration. Restorer would study the meta-data file and pickup the ready full snapshot from there and apply all the delta snapshots having revision number greater than that. So, with the backup directory as above, if the content of metadata file is like following:
```

docs/compactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+* Full-Snapshot-revision-1
+* Full-Snapshot-revision-18
```
amshuman-kr, line 75: Should we go for a json or yaml structure here?

docs/compactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+Then, during restoration Full-Snapshot-revision-18 and all the delta snaphots from revision number 19 to 49 will be applied. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.** As Full-Snapshot-revision-38 is not present in the metadata file, it would not be considered as ready yet.
```
amshuman-kr, line 77:
```suggestion
Then, during restoration Full-Snapshot-revision-18, and all the delta snaphots from revision number 19 to 49 will be applied. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.** As Full-Snapshot-revision-38 is not present in the metadata file, it would not be considered as ready yet.
```

docs/compactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+Another approach to make sure that restorer only picks up snapshots which are ready, full snapshots can be named with extension `.part` until the snapshot is ready. When the snapshot is ready, `.part` extension can be removed from it's name. Restorer would avoid any file having extension as `.part`. 
+
+So, if Full-Snapshot-revision-38 is not ready in the above example, it can be named as `Full-Snapshot-revision-38.part`. Restorer will apply Full-Snapshot-revision-18 and all the delta snaphots from revision number 19 to 49. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.**
```
amshuman-kr, line 81:
```suggestion
Another approach to make sure that restorer only picks up snapshots which are ready, full snapshots can be named with extension `.part` until the snapshot is ready. When the snapshot is ready, `.part` extension can be removed from it's name. Restorer would avoid any file having extension as `.part`. 

So, if Full-Snapshot-revision-38 is not ready in the above example, it can be named as `Full-Snapshot-revision-38.part`. Restorer will apply Full-Snapshot-revision-18 and all the delta snaphots from revision number 19 to 49. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.**

This option is preferred as it avoids introducing a metadata file. But it would work only if renaming to remove the `.part` extension (upon completion of upload) is an atomic operation in all the supported snapstores. This needs to be evaluated.
```

docs/compactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+4. Keep provisions for backward compatibility so that ETCDBR can work with both old and new type of backup store. We may want to introduce a metadata file in object store for new type of backup structure. If ETCDBR encounter metadata file suggesting a backup store contains new type of directory structure, it switches to compact mode.
+5. **Garbage collection** : Garbage collection is very much important for this compaction story. As the restorer have to look for latest compacted full snapshots from among the numerous full snapshots already present in a directory, it is essential to reduce the number of search space. Garbage collecter should work based on some policy and regularly clean up the directory from old unwanted snapshots. 
+
+**Note:** ETCD Backup Restore needs to take one full snapshot just after starting. Compactor will work only if there is any full snapshot already present in the store. It is not limitation but a design choice
```
amshuman-kr, line 86:
```suggestion
#### Backward Compatibility

1. **Restoration** : The changes to handle the newly proposed backup directory structure must be backward compatible with older structures at least for restoration because we need have to restore from backups in the older structure. This includes the support for restoring from a backup without a metadata file if that is used in the actual implementation.
2. **Backup** : For new snapshots (even on a backup containing the older structure), the new structure may be used. The new structure must be setup automatically including creating the base full snapshot (and also the creation of a metadata file if necessary).
3. **Garbage collection** : The existing functionality of garbage collection of snapshots (full and incremental) that fall out of the backup retention policy must be compatible with both old and new backup folder structure. I.e. the snapshots in the older backup structure must be retained in their own structure and the snapshots in the proposed backup structure should be retained in the proposed structure. Once all the snapshots in the older backup structure go out of the retention policy and are garbage collected, we can think of removing the support for older backup folder structure.  

**Note:** ETCD Backup Restore needs to take one full snapshot just after starting. Compactor will work only if there is any full snapshot already present in the store. It is not limitation but a design choice
```

amshuman-kr (changes requested):
Thanks @abdasgupta for the detailed write up. It is a lot more understandable to a lay person.  I have suggested a few rewordings and formatting to ease the reading flow and to also make some sub-sections directly linkable.

docs/compactor.md
abdasgupta: There is one problem. During parsing the file names for restoration, it might be more difficult to parse and compare with two revision numbers. Do we really need two revision numbers? Yes, we can name with revision of last change. But I am not completely sure if it's more useful. From user readability perspective, I think naming with first revision is more useful. WDYT?

docs/compactor.md
abdasgupta: Used yaml

docs/compactor.md
```diff
@@ -0,0 +1,86 @@
...
+
+**How the solution works:** The newly introduced compact command does not disturb the running ETCD while captuing the compacted snapshot. The command is designed to run potentially separately (from the main ETCD process/container/pod). ETCD Druid may take care of starting the newly introduced compact command as a separate job based on some user defined parameter.
+
+#### **Points to take care while saving the compacted snapshot:**
+As compacted snapshot and full snapshot are taken by different processes running in different pods but accessing same store to save the snapshots, some problems may arise:
+1.	When uploading the compacted snapshot to the snapstore, there is the problem of how does the restorer know when to start using the newly compacted snapshot. This communication needs to be atomic.
+2.	With a regular schedule for compaction that happens potentially separately from the main etcd pod, is there a need for regular scheduled full snapshots anymore?
+3.	Uploading the compacted snapshot to the store should be atomic.
```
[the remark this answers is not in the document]
abdasgupta: Yes. Rectified

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+* Full-Snapshot-revision-1
+* Full-Snapshot-revision-18
+
+Then, during restoration Full-Snapshot-revision-18, and all the delta snaphots from Incremental-Snapshot-revision-22-18 to Incremental-Snapshot-revision-52-48 will be applied. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.** As Full-Snapshot-revision-38 is not present in the metadata file, it would not be considered as ready yet.
+
+Another approach to make sure that restorer only picks up snapshots which are ready, full snapshots can be named with extension `.part` until the snapshot is ready. When the snapshot is ready, `.part` extension can be removed from it's name. Restorer would avoid any file having extension as `.part`.  
+
+So, if Full-Snapshot-revision-38 is not ready in the above example, it can be named as `Full-Snapshot-revision-38.part`. Restorer will apply Full-Snapshot-revision-18 and all the delta snaphots from Incremental-Snapshot-revision-22-18 to Incremental-Snapshot-revision-52-48. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.**
```
vlerenc, line 84: Same as above, both, full and delta snapshots need to be **ready**. That's just a nitpick, because you have written `ready` with full snapshot, but not with the delta snapshots.

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+2. **Backup** : For new snapshots (even on a backup containing the older structure), the new structure may be used. The new structure must be setup automatically including creating the base full snapshot (and also the creation of a metadata file if necessary).
+3. **Garbage collection** : The existing functionality of garbage collection of snapshots (full and incremental) that fall out of the backup retention policy must be compatible with both old and new backup folder structure. I.e. the snapshots in the older backup structure must be retained in their own structure and the snapshots in the proposed backup structure should be retained in the proposed structure. Once all the snapshots in the older backup structure go out of the retention policy and are garbage collected, we can think of removing the support for older backup folder structure.  
+
+**Note:** ETCD Backup Restore needs to take one full snapshot just after starting. Compactor will work only if there is any full snapshot already present in the store. It is not limitation but a design choice
```
vlerenc, line 93: Hmm... I know we do that, but hasn't that already caused trouble?

/cc @amshuman-kr @shreyas-s-rao 

I don't really see why this is necessary/beneficial. In the contrary, this seems problematic. Right when we need ETCD the most, i.e. when it starts and all the API server start hammering it, we also do a full snapshot. What was the motivation?

Wouldn't it make more sense to only do that if:
- There is no full snapshot at all (we need at least one for the delta snapshots to make sense, but not a new one every time ETCD starts)
- The last one is "too old"

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+Another approach to make sure that restorer only picks up snapshots which are ready, full snapshots can be named with extension `.part` until the snapshot is ready. When the snapshot is ready, `.part` extension can be removed from it's name. Restorer would avoid any file having extension as `.part`.  
+
+So, if Full-Snapshot-revision-38 is not ready in the above example, it can be named as `Full-Snapshot-revision-38.part`. Restorer will apply Full-Snapshot-revision-18 and all the delta snaphots from Incremental-Snapshot-revision-22-18 to Incremental-Snapshot-revision-52-48. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.**
+
+This option is preferred as it avoids introducing a metadata file. But it would work only if renaming to remove the `.part` extension (upon completion of upload) is an atomic operation in all the supported snapstores. This needs to be evaluated.
+
+#### Backward Compatibility
```
vlerenc, line 88: I am wary of all the effort with backward compatibility. Isn't there a simple and robust way around that by detecting and rewriting the backup structure?

I assume, the main problem are the revisions we are now encoding into the names? How about "faking it"?
- We leave the old structure as is until a future release for human operators / emergency measures
- We take the latest full snapshot and then all its delta snapshots and copy them over (with the marker file approach from above, just to be on the safe side)
- As revisions, we fake it by starting at 0 and incrementing by 1, i.e. the full snapshot is 0, the first delta is 1, the next 2, etc.

The resulting revisions in the filenames will be far too small, but who cares? The algorithm picks the highest full snapshot and then all consecutive deltas and they would all be in order. The next full/incremental snapshots would carry the proper revisions.

Note: We discussed the need for revisions only in the context of clustered ETCD. The migration of the old backup structure will be completed long before we can ship clustered ETCD. Only then we need true revisions, before that only ordering.

WDYT? This way, even if ETCD dies on us, migration can be executed before restore and restore can work only from new backup structure.

vlerenc (changes requested):
Thanks @abdasgupta, I made a few comments and proposals (especially in regards to backward compatibility, which I would prefer to avoid, if safely possible) and would drop the metadata file and the move/rename option and replace it with another solution proposed in the comments.

docs/compactor.md
abdasgupta: The main problem is not only the naming parts. Actually that is not much of problem. We are currently saving full snapshots and subsequent delta snapshots till the next full snapshots together into separate directory. The main problem is that we are planning to scrap this directory structure and bring all the full snapshots and delta snapshots under one directory

vlerenc: @abdasgupta Can you please respond also to the other comments and also expand your explanation above?

docs/compactor.md
vlerenc: On a call out-of-band, it was said we don't take full snapshot always, but only if we miss one or it's older than 24h.

Let's then add this info to the "note" above.

shreyas-s-rao:
@abdasgupta what's the status here? Any pending updates to the PR or review comments?

abdasgupta:
As we are just completed with compaction PR, I will update the doc accordingly and we can merge this as well

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+    - Incremental-Snapshot-192375
+- Backup-192789
+    - Full-Snapshot-192789
+    - Incremental-Snapshot-192799
+    - Incremental-Snapshot-192809
+    - Incremental-Snapshot-192819
+```
+
+To solve the problem, proposal is:
+1. ETCDBR will take the first full snapshot after it starts main ETCD Process and snapshotter process. But after taking the first full snapshot, snapshotter will only continue taking delta snapshots but no further full snapshots.
```
amshuman-kr, line 52: This is not a central point of the design and is not yet implemented either (we decided to postpone this implementation to after observing the backup compaction at work in the landscapes). Can you please move this to a later sections (say, something titled `Future Work`, for example)?

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+2. Flatten the directory structure of backup folder. Save all the full snapshots, delta snapshots and compacted snapshots under same directory/prefix. Also, name snapshots after revision number instead of timestamps. Restorer will restore from compacted snapshots that have highest revision number in the name and delta snapshots that have higher revision numbers in name than the comapcted full snapshot.
```
amshuman-kr, line 53: > Also, name snapshots after revision number instead of timestamps

I guess this is now out-of-date. We didn't change the snapshot file names but only the sorting criteria, right?

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+    - Incremental-Snapshot-revision-32-27
+    - Incremental-Snapshot-revision-38-32
+    - Full-Snapshot-revision-38 (Compacted)
+    - Incremental-Snapshot-revision-41-38
+    - Incremental-Snapshot-revision-43-41
+    - Incremental-Snapshot-revision-48-43
+    - Incremental-Snapshot-revision-52-48
+```
+
+3. A meta-data file might need to be added in the Backup directory to mark which compacted (or full) snapshot is ready to be used as the base for restoration. Restorer would study the meta-data file and pickup the ready full snapshot from there and apply all the delta snapshots having revision number greater than that. So, with the backup directory as above, if the content of metadata file is like following:
```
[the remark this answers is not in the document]
amshuman-kr: I think we have not used any metadata file in the implementation. So, I guess section is now out-dated and needs a rewrite.

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+* Full-Snapshot-revision-1
+* Full-Snapshot-revision-18
+
+Then, during restoration Full-Snapshot-revision-18, and all the delta snaphots from Incremental-Snapshot-revision-22-18 to Incremental-Snapshot-revision-52-48 will be applied. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.** As Full-Snapshot-revision-38 is not present in the metadata file, it would not be considered as ready yet.
```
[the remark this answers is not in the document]
amshuman-kr: I guess this part is also now out-dated and needs a rewrite.

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+Another approach to make sure that restorer only picks up snapshots which are ready, full snapshots can be named with extension `.part` until the snapshot is ready. When the snapshot is ready, `.part` extension can be removed from it's name. Restorer would avoid any file having extension as `.part`.
```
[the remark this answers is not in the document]
amshuman-kr: I guess this part is also now out-dated and needs a rewrite.

docs/compactor.md
amshuman-kr, line 84:
```suggestion
So, if Full-Snapshot-revision-38 is not ready in the above example, it can be named as `Full-Snapshot-revision-38.part`. Restorer will apply Full-Snapshot-revision-18 and all the delta snaphots from Incremental-Snapshot-revision-22-18 to Incremental-Snapshot-revision-52-48.

**Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.**
```

> So, if Full-Snapshot-revision-38 is not ready in the above example, it can be named as `Full-Snapshot-revision-38.part`. Restorer will apply Full-Snapshot-revision-18 and all the delta snaphots from Incremental-Snapshot-revision-22-18 to Incremental-Snapshot-revision-52-48.

Also, I guess this part is now out-dated and needs a rewrite.

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+So, if Full-Snapshot-revision-38 is not ready in the above example, it can be named as `Full-Snapshot-revision-38.part`. Restorer will apply Full-Snapshot-revision-18 and all the delta snaphots from Incremental-Snapshot-revision-22-18 to Incremental-Snapshot-revision-52-48. **Restorer would restore from latest ready full snapshot and all subsequent delta snapshots.**
+
+This option is preferred as it avoids introducing a metadata file. But it would work only if renaming to remove the `.part` extension (upon completion of upload) is an atomic operation in all the supported snapstores. This needs to be evaluated.
+
+#### Backward Compatibility
+1. **Restoration** : The changes to handle the newly proposed backup directory structure must be backward compatible with older structures at least for restoration because we need have to restore from backups in the older structure. This includes the support for restoring from a backup without a metadata file if that is used in the actual implementation.
```
amshuman-kr, line 89: Can you please remove the reference to the metadata file and mention all the three cases for restoration (latest snapshots in v1 only, in v2 only and in both v1 and v2)?

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+2. **Backup** : For new snapshots (even on a backup containing the older structure), the new structure may be used. The new structure must be setup automatically including creating the base full snapshot (and also the creation of a metadata file if necessary).
```
amshuman-kr, line 90:
```suggestion
2. **Backup** : For new snapshots (even on a backup containing the older structure), the new structure may be used. The new structure must be setup automatically including creating the base full snapshot.
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+3. **Garbage collection** : The existing functionality of garbage collection of snapshots (full and incremental) that fall out of the backup retention policy must be compatible with both old and new backup folder structure. I.e. the snapshots in the older backup structure must be retained in their own structure and the snapshots in the proposed backup structure should be retained in the proposed structure. Once all the snapshots in the older backup structure go out of the retention policy and are garbage collected, we can think of removing the support for older backup folder structure.
```
amshuman-kr, line 91:
```suggestion
3. **Garbage collection** : The existing functionality of garbage collection of snapshots (full and incremental) according to the backup retention policy must be compatible with both old and new backup folder structure. I.e. the snapshots in the older backup structure must be retained in their own structure and the snapshots in the proposed backup structure should be retained in the proposed structure. Once all the snapshots in the older backup structure go out of the retention policy and are garbage collected, we can think of removing the support for older backup folder structure.  
```

docs/compactor.md
amshuman-kr: @abdasgupta Can you please add the information to the doc?

amshuman-kr (changes requested):
Thanks for the improvement in the docs @abdasgupta! Some comments below. Could you please also add a section giving an example of how the snapshots would look like with scheduled full snapshot on and scheduled backup compaction also on (at a shorter interval than full snapshots)?

docs/compactor.md
abdasgupta: Yes, but sorting criteria will be based on revision number present in the name.

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+**How the solution works:** The newly introduced compact command does not disturb the running ETCD while compacting the backup snapshots. The command is designed to run potentially separately (from the main ETCD process/container/pod). ETCD Druid may take care of starting the newly introduced compact command as a separate job (scheduled periodically and also on-demand if required) based on some user defined parameter.
```
amshuman-kr, line 26:
```suggestion
###How the solution works
The newly introduced compact command does not disturb the running ETCD while compacting the backup snapshots. The command is designed to run potentially separately (from the main ETCD process/container/pod). ETCD Druid may take care of starting the newly introduced compact command as a separate job (scheduled periodically and also on-demand if required) based on some user defined parameter.
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+### **Points to take care while saving the compacted snapshot:**
+As compacted snapshot and the existing periodic full snapshots are taken by different processes running in different pods but accessing same store to save the snapshots, some problems may arise:
+1.	When uploading the compacted snapshot to the snapstore, there is the problem of how does the restorer know when to start using the newly compacted snapshot. This communication needs to be atomic.
```
amshuman-kr, line 30:
```suggestion
1. When uploading the compacted snapshot to the snapstore, there is the problem of how does the restorer know when to start using the newly compacted snapshot. This communication needs to be atomic.
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+2.	With a regular schedule for compaction that happens potentially separately from the main etcd pod, is there a need for regular scheduled full snapshots anymore?
```
amshuman-kr, line 31:
```suggestion
2. With a regular schedule for compaction that happens potentially separately from the main etcd pod, is there a need for regular scheduled full snapshots anymore?
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+3.  We are planning to introduce new directory structure, under v2 prefix, for saving the snapshots (compacted and full), as mentioned in details below. But for backward compatibility, we also need to consider the older directory, which is currently under v1 prefix, during accessing snapshots.
```
amshuman-kr, line 32:
```suggestion
3. We are planning to introduce new directory structure, under v2 prefix, for saving the snapshots (compacted and full), as mentioned in details below. But for backward compatibility, we also need to consider the older directory, which is currently under v1 prefix, during accessing snapshots.
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### **How to swap full snapshot with compacted snapshot atomically**
+
+Currently, full snapshots and the subsequent delta snapshots are grouped under same prefix path in the snapstore. When a full snapshot is created, it is placed under a prefix/directory with the name comprising of timestamp. Then subsequent delta snapshots are also pushed into the same directory. Thus each prefix/directory contains a single full snapshot and the subsequent delta snapshots. So far, it is the job of ETCDBR to start main ETCD process and snapshotter process which takes full snapshot and delta snapshot periodically. But as per our proposal, compaction will be running as parallel process to main ETCD process and snapshotter process. So we can't reliably co-ordinate between the processes to achieve switching to the compacted snapshot as the base snapshot atomically.```
```
amshuman-kr, line 36:
```suggestion
Currently, full snapshots and the subsequent delta snapshots are grouped under same prefix path in the snapstore. When a full snapshot is created, it is placed under a prefix/directory with the name comprising of timestamp. Then subsequent delta snapshots are also pushed into the same directory. Thus each prefix/directory contains a single full snapshot and the subsequent delta snapshots. So far, it is the job of ETCDBR to start main ETCD process and snapshotter process which takes full snapshot and delta snapshot periodically. But as per our proposal, compaction will be running as parallel process to main ETCD process and snapshotter process. So we can't reliably co-ordinate between the processes to achieve switching to the compacted snapshot as the base snapshot atomically.
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+
+**Current Directory Structure**
```
amshuman-kr, line 38:
```suggestion
##### **Current Directory Structure**
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+    - Incremental-Snapshot-300-400-192799
+    - Incremental-Snapshot-400-500-192809
+    - Incremental-Snapshot-500-600-192819
+```
+
+To solve the problem, proposal is:
+1. ETCDBR will take the first full snapshot after it starts main ETCD Process and snapshotter process. After taking the first full snapshot, snapshotter will continue taking full snapshots. On the other hand, ETCDBR compactor command will be run as periodic job in a separate pod and use the existing full or compacted snapshots to produce further compacted snapshots. Full snapshots and compacted snapshots will be named after same fashion. So, there is no need of any mechanism to choose which snapshots(among full and compacted snapshot) to consider as base snapshots. 
+2. Flatten the directory structure of backup folder. Save all the full snapshots, delta snapshots and compacted snapshots under same directory/prefix. Restorer will restore from full/compacted snapshots and delta snapshots sorted based on the revision numbers in name (or timestamp if the revision numbers are equal).
+
+**Proposed Directory Structure**
```
amshuman-kr, line 56:
```suggestion
##### **Proposed Directory Structure**
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+3. **Garbage collection** : The existing functionality of garbage collection of snapshots (full and incremental) according to the backup retention policy must be compatible with both old and new backup folder structure. I.e. the snapshots in the older backup structure must be retained in their own structure and the snapshots in the proposed backup structure should be retained in the proposed structure. Once all the snapshots in the older backup structure go out of the retention policy and are garbage collected, we can think of removing the support for older backup folder structure.
+
+**Note:** Compactor will run paralley to current snapshotter process and work only if there is any full snapshot already present in the store. By current design, a full snapshot will be taken if there is already no full snapshot or the existing full snapshot is older than 24 hours. It is not limitation but a design choice. As per proposed design, the backup storage will contain both periodic full snapshots as well as periodic compacted snapshot. Restorer will pickup the base snapshot whichever is latest one.
```
amshuman-kr, line 93:
```suggestion
**Note:** Compactor will run parallel to current snapshotter process and work only if there is any full snapshot already present in the store. By current design, a full snapshot will be taken if there is already no full snapshot or the existing full snapshot is older than 24 hours. It is not limitation but a design choice. As per proposed design, the backup storage will contain both periodic full snapshots as well as periodic compacted snapshot. Restorer will pickup the base snapshot whichever is latest one.
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+    - Incremental-Snapshot-revision-200-300-192385
+    - Full-Snapshot-revision-0-300-192386 (Taken by compaction job)
+    - Incremental-Snapshot-revision-300-400-192396
+    - Incremental-Snapshot-revision-400-500-192406
+    - Incremental-Snapshot-revision-500-600-192416
+    - Full-Snapshot-revision-0-600-192419 (Taken by snapshotter)
+    - Full-Snapshot-revision-0-600-192420 (Taken by compaction job)
+```
+
+**Future work**
```
amshuman-kr, line 72:
```suggestion
##### **Future work**
```

docs/compactor.md
```diff
@@ -0,0 +1,93 @@
...
+    - Full-Snapshot-revision-0-200-192379 (Taken by snapshotter)
+    - Incremental-Snapshot-revision-200-300-192385
+    - Full-Snapshot-revision-0-300-192386 (Taken by compaction job)
+    - Incremental-Snapshot-revision-300-400-192396
+    - Incremental-Snapshot-revision-400-500-192406
+    - Incremental-Snapshot-revision-500-600-192416
+    - Full-Snapshot-revision-0-600-192419 (Taken by snapshotter)
+    - Full-Snapshot-revision-0-600-192420 (Taken by compaction job)
+```
+
```
amshuman-kr, line 71:
```suggestion

##### What happens to the delta snapshots that were compacted?

The proposed `compaction` sub-command in `etcdbrctl` (and hence,  the `CronJob` provisioned by `etcd-druid` that will schedule it at a regular interval) would only upload the compacted full snapshot.
It will not delete the snapshots (delta or full snapshots) that were compacted.
These snapshots which were superseded by a freshly uploaded compacted snapshot would follow the same life-cycle as other older snapshots.
I.e. they will be garbage collected according to the configured backup snapshot retention policy.

For example, if an `exponential` retention policy is configured and if compaction is done every `30m` then there might be at most `48` additional (compacted) full snapshots (`24h * 2`) in the backup for the latest day. As time rolls forward to the next day, these additional compacted snapshots (along with the delta snapshots that were compacted into them) will get get garbage collected retaining only one full snapshot for the day before according to the retention policy.

```

amshuman-kr (changes requested):
Thanks for the changes @abdasgupta. I have suggested some styling and typo fixes as well as a clarification for garbage collection for older snapshots and how many new snapshots are likely to be added for the latest day. If you find them reasonably, you can apply them.

docs/compactor.md
amshuman-kr: > ETCD Druid may take care of starting the newly introduced compact command as a separate job (scheduled periodically and also on-demand if required) based on some user defined parameter. 

This is a proposal document on `etcd-druid`. So, it would be good to mention the suggested changes to the `Etcd` CRD.

docs/compactor.md
```diff
@@ -0,0 +1,168 @@
...
+{{- end }}
+{{- if .Values.store.storageContainer }}
+            - --store-container={{ .Values.store.storageContainer }}
+{{- end }}
+            ports:
+            - containerPort: {{ .Values.backup.port }}
+              name: server
+              protocol: TCP
+            resources:
+{{ toYaml .Values.backup.resources | indent 14 }}
```
amshuman-kr, line 91: Giving helm template content in the example doesn't make sense. Could you please clean the variable substitution part up. Or alternatively, just hyperlink to the corresponding template file in the helm chart in the repo. Then you can skip giving the `CronJob` example. The `Etcd` CRD example looks good.

amshuman-kr (changes requested):
Thanks @abdasgupta. Almost there. A couple of more suggestions on the changes.

amshuman-kr (approved):
LGTM

PR #141, opened 2021-02-25
Maintain project dependencies and images

timuthy:
**What this PR does / why we need it**:
This PR updates dependency of the druid component.

Along the way, it improves several things:
- [Gardener's check script](https://github.com/gardener/gardener/blob/master/hack/check.sh) is now executed as part of `make check` in favor of `go vet` only (Make target `vet` has been dropped)
- `make revendor` now updates `.github` templates from `gardener/gardener`
- Several golang-ci findings have been resolved
- The `controller-gen` from the `vendor` directory is now considered to generate assets in order to generate a stable output
- Curl has been removed from the Druid image for security reasons
- New leader election options have been added to eventually migrate to `lease.coordination.k8s.io` based leader election in the future (/cc @timebertt)

⚠️ PR #139 must be merged first!

/cc @shreyas-s-rao @amshuman-kr 

I split the PR into different commits to make reviewing easier. HTH.

**Release note**:

```improvement developer
Several Go and image dependencies have been updated.
```
```improvement developer
- [Gardener's check script](https://github.com/gardener/gardener/blob/master/hack/check.sh) is now executed as part of `make check` in favor of `go vet` only (Make target `vet` has been dropped)
- `make revendor` now updates `.github` templates from `gardener/gardener`
- Several golang-ci findings have been resolved
- The `controller-gen` from the `vendor` directory is now considered to generate assets in order to get a stable output.
```
```action operator
etcd-druid images won't be delivered with `curl` anymore.
```
```action operator
etcd-druid now needs CRUD permissions for `leaese.coordination.k8s.io` object in order to perform leader election.
```

PR #146, opened 2021-03-01
Read PVC warning events

timuthy:
This PR makes Druid read events for PVCs if they are still unbound after a `StatefulSet` does not get ready. Respective warnings will appear in the `.status.lastError` description which makes it easier for operators to spot the root cause.

⚠️ This PR is based on #141. Only commit 924a361dcdf12cb4fc10092ece7791aadb5a9d92 is relevant for this PR.

**Release note**:

```improvement operator
If an etcd `StatefulSet` remains pending, warning events of unbound `PVC`s are now added to the `.status.lastError` of the `etcd` resource. This makes it easier for operators to spot potential issues.
```

### DIFF
--- a/controllers/etcd_controller.go
+++ b/controllers/etcd_controller.go
@@ -25,13 +25,14 @@ import (
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 	"github.com/gardener/etcd-druid/pkg/chartrenderer"
-	kubernetes "github.com/gardener/etcd-druid/pkg/client/kubernetes"
+	"github.com/gardener/etcd-druid/pkg/client/kubernetes"
 	"github.com/gardener/etcd-druid/pkg/common"
 	druidpredicates "github.com/gardener/etcd-druid/pkg/predicate"
 	"github.com/gardener/etcd-druid/pkg/utils"
 
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/kubernetes/health"
 	gardenerretry "github.com/gardener/gardener/pkg/utils/retry"
 
@@ -40,7 +41,9 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	eventsv1 "k8s.io/api/events/v1"
+	eventsv1beta1 "k8s.io/api/events/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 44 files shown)


timuthy:
ping @gardener/etcd-druid-maintainers

amshuman-kr:
@timuthy Even if I used rebase while merging #141, there are still conflicts. I am sorry, I have to request you to rebase this PR also.

timuthy:
Thanks for taking care and no worries @amshuman-kr. I noticed that due to the update to a newer Go version, I'll also have to adjust the CI/CD `check` step which is currently failing.

amshuman-kr:
Thanks @timuthy . I have merged #150. Please proceed with the rebase/resolving conflicts.

timuthy:
Thanks @amshuman-kr, I've resolved the conflicts in this PR.

amshuman-kr (approved):
Thanks for rebasing @timuthy! LGTM

controllers/etcd_controller.go
```diff
@@ -1124,23 +1138,63 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+		}
+	}
+
 	return ss, err
 }
 
-func (r *EtcdReconciler) removeOperationAnnotation(etcd *druidv1alpha1.Etcd) error {
+func (r *EtcdReconciler) fetchPVCEventsFor(ctx context.Context, ss *appsv1.StatefulSet) (string, error) {
+	pvcs := &corev1.PersistentVolumeClaimList{}
+	if err := r.List(ctx, pvcs, client.InNamespace(ss.GetNamespace())); err != nil {
```
shreyas-s-rao, line 1159: Can you use a label selector here, like `client.MatchingLabelsSelector{Selector: selector}` to filter the PVCs?

controllers/etcd_controller.go
```diff
@@ -1124,23 +1138,63 @@ func (r *EtcdReconciler) waitUntilStatefulSetReady(sts *appsv1.StatefulSet) (*ap
...
+		return "", err
+	}
+
+	var (
+		pvcMessages  string
+		volumeClaims = ss.Spec.VolumeClaimTemplates
+	)
+	for _, volumeClaim := range volumeClaims {
+		for _, pvc := range pvcs.Items {
+			if pvc.Status.Phase != corev1.ClaimBound && !strings.HasPrefix(pvc.GetName(), fmt.Sprintf("%s-%s", volumeClaim.Name, ss.Name)) {
```
shreyas-s-rao, line 1169: If you're using label selector above, you won't need to have the PVC name check condition here.

shreyas-s-rao (changes requested):
@timuthy thanks for your PR. I was somehow not able to reproduce the desired working of the functionality of the PR. To reproduce, I deliberately failed the PVC from binding by specifying a random storage class, and the PVC is in pending state, but druid doesn't print the PVC unbound warning as expected. Instead, I get `error 'non-exact field matches are not supported by the cache' occurred while fetching more details`. Do you have any idea about this?

timuthy:
Thanks @shreyas-s-rao for your review. I fixed the bug you experienced in 8f14dc9e77335d95b74f8aac2fc524c2393d4056. Unfortunately, the cache exclusion for events only happened for the Env tests. Can you try it again?

controllers/etcd_controller.go
timuthy: After looking at our landscapes and discussed, selecting the PVCs via a label selector is unclear because the labels on the PVCs can be different. Please let me know if you have another idea or if we should stick to the name filtering in the lines below.

shreyas-s-rao (approved):
@timuthy Thanks for making the changes. Tested the PR now, and it works as expected. LGTM

PR #163, opened 2021-04-21
Added mutex lock before every update of ETCD instance.

abdasgupta:
Sometimes the following error is coming while running the controllers:
```
Expected success, but got an error:
      <*errors.StatusError | 0xc0002314a0>: {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "Operation cannot be fulfilled on statefulsets.apps \"foo3\": the object has been modified; please apply your changes to the latest version and try again",
              Reason: "Conflict",
              Details: {Name: "foo3", Group: "apps", Kind: "statefulsets", UID: "", Causes: nil, RetryAfterSeconds: 0},
              Code: 409,
          },
      }
      Operation cannot be fulfilled on statefulsets.apps "foo3": the object has been modified; please apply your changes to the latest version and try again
```

Though this error is harmless as reconciler will reque after the error, but it is due to the fact that both the controllers(ETCD controller, custodian controller) are trying to update same ETCD resource. So to avoid data race condition, I added mutex lock/unlock before ETCD update operation.

### DIFF
--- a/controllers/etcd_controller.go
+++ b/controllers/etcd_controller.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
@@ -70,6 +71,9 @@ var (
 		&eventsv1beta1.Event{},
 		&eventsv1.Event{},
 	}
+
+	once  sync.Once
+	mutex *sync.Mutex
 )
 
 const (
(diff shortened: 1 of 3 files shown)


shreyas-s-rao (approved):
@abdasgupta thanks for fixing the race conditions in the tests. LGTM

PR #174, opened 2021-05-19
Removed synchronisation before updating ETCD status.

### LINKED ISSUE #171: [Feature] Avoid locks during reconciliation
 
#163 introduced locking the main and custodian controller for every update to the `Etcd` resource and its status.

This should be avoided and the race conditions in the tests should be solved in a different way.

Such synchronisation will lead to performance bottlenecks.

 Use `StatusWriter.Patch()`?
 
 Credit: @timuthy

abdasgupta:

 /area control-plane

Locking mechanism was introduced before updating the ETCD status by two controllers of druid. It may have performance issue for large clusters. This PR removed the locks.
Fixes #171 

**Release note**:

```other operator
Removed synchronisation before updating ETCD status.
```


amshuman-kr (approved):
Thanks @abdasgupta for the fix. LGTM and worked locally for me.

/invite @timuthy Especially, could you please check if the use of `retry.RetryOnConflict` for `Status().Update()` is idiomatic?

controllers/etcd_controller.go
```diff
@@ -1150,16 +1129,18 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(ctx context.Context, etcd *druidv
 		etcd.Status.Ready = &ready
 	}
 
-	if err := r.Status().Update(ctx, etcd); err != nil && !apierrors.IsNotFound(err) {
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		return r.Status().Update(ctx, etcd)
```
timuthy, line 1133: This will unfortunately not work if there is a conflict because you need to re-fetch the current `etcd` resource from the cluster (to get up-to-date `ResourceVersion`), transform it and call `Update` again.

The Gardener utility functions [TryUpdate](https://github.com/gardener/gardener/blob/07a56785cb69aff4a1529943dfc8f5d6ec97f015/pkg/utils/kubernetes/update.go#L29) and [TryUpdateStatus](https://github.com/gardener/gardener/blob/07a56785cb69aff4a1529943dfc8f5d6ec97f015/pkg/utils/kubernetes/update.go#L35) handle this automatically for you (also `DeepCopy()` is not required before). You can see an example usage [here](https://github.com/gardener/gardener/blob/77e09dd522d7bfb682cc35d0be437201b606061b/extensions/pkg/controller/worker/state_actuator.go#L71).

The same applies for other status updates in this PR.

timuthy (changes requested):
Thanks for the fix @abdasgupta. I left two comments in the review.

controllers/etcd_controller.go
```diff
@@ -275,10 +263,6 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
 
 // updateWithLock is wrapper function which helps in synchronization between controller threads(etcd-druid controller and custodian controller).
 func (r *EtcdReconciler) updateWithLock(ctx context.Context, etcd *druidv1alpha1.Etcd) error {
```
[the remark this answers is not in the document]
abdasgupta: Yes. Addressed. Thanks.

controllers/etcd_controller.go
abdasgupta: I did. Thanks.

amshuman-kr (changes requested):
Thanks for the changes @abdasgupta. I see that `kutil "github.com/gardener/gardener/pkg/utils/kubernetes"` is already imported in the controller files. Can you please use `kutil.TryUpdateStatus` instead of `extensioncontroller.TryUpdateStatus`? This way, you can avoid the vendoring changes and keep the PR changes minimal.

amshuman-kr (approved):
Thanks for the change! LGTM

amshuman-kr:
@timuthy Could you PTAL? Merge would be unblocked after that.

timuthy (approved):
/lgtm

PR #176, opened 2021-05-20
Removed synchronisation before updating ETCD status.

### LINKED ISSUE #171: [Feature] Avoid locks during reconciliation

amshuman-kr:
Cherry picked from #174.

Fixes #171 

**Release note**:

```other operator
Removed synchronisation before updating ETCD status.
```


timuthy (approved):
/lgtm

Issue #151, opened 2021-03-31
[Feature] Modularize and enhance status management according to multi-node ETCD proposal


amshuman-kr:
**Feature (What you would like to be added):**

1. Enhance the `Etcd` resource status structure according to the changes proposed [here](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#status)
2. Enhance etcd-backup-restore and etcd-druid to update the corresponding member's [health status](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#health-check) in the `Etcd` resource status. This need not include the task of [cutting off traffic in case of backup failure](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#cutting-off-client-requests) yet as the evaluation/decision on that is pending for the scenario of multi-node ETCD with ephemeral persistence. The `etcd-backup-restore` needs to consider the following scenarios in the implementation.
   - `etcd-druid` can maintain the `AllMembersReady` and `Ready` [conditions](AllMembersReady) as well as the following transitions for the member status where the member's `etcd-backup-restore` is unable to update its own status. Probably this has to be done in `etcd-druid` by enhancing the etcd status controller ([custodian](https://github.com/gardener/etcd-druid/pull/133)).
     1. [From stale health to `Unknown`](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#6-some-members-have-not-updated-their-status-for-a-while)
     2. [From `Unknown` to `NotReady`](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#7-some-members-have-been-in-unknown-status-for-a-while)
     3. [From not ready pod to `NotReady` status](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#8-some-member-pods-are-not-ready-but-have-not-had-the-chance-to-update-their-status).

**Motivation (Why is this needed?):**

Pick individually executable pieces of the [multi-node proposal](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md).

**Approach/Hint to the implement solution (optional):**

Also, it would be preferable to use `StatusWriter.Patch()` to avoid race-conditions.

amshuman-kr:
Split the updating of the `Etcd` resource status structure out to https://github.com/gardener/etcd-druid/issues/168.

amshuman-kr:
[...]

PR #180, opened 2021-05-26
Custodian controller improvements

timuthy:

/area robustness
/kind enhancement

This PR improves several aspects of the custodian and etcd controllers as a preparation for #151:
- Adds a separate worker count flag for the custodian controller
- Controllers now make use of the Controller-Runtime provided logger instead of instantiating another instance via `logrus`
- Improves event triggers for the custodian controller, i.e., reconciliation happens on `create/update/delete` events for `etcd` and `create/delete` events for `statefulset`, as well as `status update` for `statefulsets`. Additionally, a mapper has been introduced which maps the `statefulset` for which an event was triggered to the corresponding `etcd` resource. Earlier, this has only worked because in most of the cases `etcd` and `statefulset` have the same `namespace/name`. 
- Custodian controller does not try to claim a `statefulset` anymore, since this is the job of the `etcd controller`.
- The `statefulset` condition sync has been removed as I didn't find any `statefulsets` which `conditions` in our system and further more, we should not try to sync and information from `statefulset.status` to `etcd` by default. If a `statusfulset.status.condition` is really of interest, then it can be read directly from the `statefulset`.

I created dedicated commits to simply the review.

Fixes parts of #151

Previously, the custodian controller was setup by the following instruction:
`builder.For(&appsv1.StatefulSet{}).Owns(&druidv1alpha1.Etcd{})`

This caused event triggers for **all** `statefulset`s ([ref](https://github.com/kubernetes-sigs/controller-runtime/blob/72fc2c61d2b34cfee62910d22ae7546f46b1893c/pkg/builder/controller.go#L228)) (-> we don't need to enqueue for events of all `statefulset`s in the cluster) **and** `etcd`s with an `ownerReference` ([ref](https://github.com/kubernetes-sigs/controller-runtime/blob/72fc2c61d2b34cfee62910d22ae7546f46b1893c/pkg/builder/controller.go#L234)) (-> `etcd`s don't have an owner reference).

**Release note**:

```feature operator
It is now possible to configure the count of custodian controller workers by `--custodian-workers`.
```
```other operator
The reconciliation loop count of the custodian controller has been improved. Now, reconciliation only happens at relevant `create/update/delete` events.
```


amshuman-kr (approved):
Thanks a lot for the changes and some much needed corrections! Not the least, the long-pending use of `gomock` ❤️ 

/lgtm

amshuman-kr:
@abdasgupta Can you please give your comments? Based on that I will merge.

abdasgupta:
I have went through the PR. I am fine with the changes. I also tested it in local cluster. It works as it should. Thanks @timuthy for your PR.

abdasgupta (approved):
LGTM

Issue #158, opened 2021-04-12
[Feature] Enhance reconciliation to handle multi-node scenario in etcd-druid


amshuman-kr:
**Feature (What you would like to be added):**

Enhance reconciliation to handle [multi-node scenario](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md) in `etcd-druid`. This should include the following topics.

- [x] Maintain the `Ready` and `AllMembersReady` [conditions](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#conditions) based on the contents of the [`members`](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#members) section of the `Etcd` resource [status](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#status).
- [x] Create member `Lease` objects for every member pod.
- [x] Create separate `Service`s for client and etcd peer ([ref](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#separating-peer-and-client-traffic)) -> TBD (#147)

**Motivation (Why is this needed?):**

Pick individually executable pieces of the [multi-node proposal](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md).

**Approach/Hint to the implement solution (optional):**

timuthy:
/status on-hold
Waiting for completion of https://github.com/gardener/etcd-backup-restore/issues/322 and #193 in order to decide on the actions that need to be performed and how these must be implemented.

abdasgupta:
Scope of this issue is completed. #354 is created to track further.

PR #188, opened 2021-06-02
Modularization & Status Enhancements

### LINKED ISSUE #151: [Feature] Modularize and enhance status management according to multi-node ETCD proposal

timuthy:

/area control-plane
/kind enhancement

This PR adds the following status checks for the `etcd` resource:
- Conditions:
  - Ready check: Checks if resource has enough `Ready` members in `status.members` to fulfill the quorum.
  - AllMembersReady check: Checks if all members in `status.members` are `Ready`.
- Members:
  - Ready check: Treats the `LastUpdateTime` as a heartbeat and checks if it is within the expected time range (configurable via `--etcd-member-threshold`).

Fixes #151

The PR does not handle cases in which an etcd member is lost and replaced and thus the corresponding `status.member` has to be removed. This needs to be done by the component which removes the lost member from the etcd cluster.

**Release note**:

```feature operator
Various `condition` and etcd `member` checks have been added to Etcd-Druid. The results of those checks will be reflected in the `etcd.status` sub-resource.
- Conditions:
  - Ready check: Checks if resource has enough `Ready` members in `status.members` to fulfill the quorum.
  - AllMembersReady check: Checks if all members in `status.members` are `Ready`.
- Members:
  - Ready check: Treats the `LastUpdateTime` as a heartbeat and checks if it is within the expected time range (configurable via `--etcd-member-threshold`).
```
```feature operator
A re-sync mechanism has been added for the Custodian controller. The new flag `--custodian-sync-period (default 30s)` controls the duration after which the Custodian controller re-enqueues `etcd` resources for reconciliation. This can be considered as a health check interval.
```


pkg/health/condition/check_ready.go
```diff
@@ -0,0 +1,62 @@
...
+		return &result{
+			conType: druidv1alpha1.ConditionTypeReady,
+			status:  druidv1alpha1.ConditionUnknown,
+			reason:  "NoMembersInStatus",
+			message: "Cannot determine readiness of cluster since status has no members",
+		}
+	}
+
+	var (
+		quorum       = len(status.Members)/2 + 1
```
amshuman-kr, line 32:
```suggestion
		quorum       = status.Replicas/2 + 1
```

pkg/health/condition/check_ready.go
```diff
@@ -0,0 +1,62 @@
...
+			readyMembers++
+		}
+	}
+
+	if readyMembers < quorum {
+		return &result{
+			conType: druidv1alpha1.ConditionTypeReady,
+			status:  druidv1alpha1.ConditionFalse,
+			reason:  "QuorumLost",
+			message: "The majority of ETCD nodes are not ready",
```
amshuman-kr, line 47:
```suggestion
			message: "The majority of ETCD members are not ready",
```

pkg/health/condition/check_ready.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	}
+
+	return &result{
+		conType: druidv1alpha1.ConditionTypeReady,
+		status:  druidv1alpha1.ConditionTrue,
+		reason:  "Quorate",
+		message: "The majority of ETCD nodes is ready",
```
amshuman-kr, line 55:
```suggestion
		message: "The majority of ETCD members is ready",
```

pkg/health/condition/condition_suite_test.go
```diff
@@ -0,0 +1,52 @@
...
+	return r.ConStatus
+}
+
+func (r *result) Reason() string {
+	return r.ConReason
+}
+
+func (r *result) Message() string {
+	return r.ConMessage
+}
```
amshuman-kr, line 52: Perhaps I missed it, but where is this used?

pkg/health/status/check.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// Execute condition checks after the etcd member checks because we need their result here.
+	if err := c.executeConditionChecks(status); err != nil {
+		return err
+	}
+	return nil
+}
+
+// executeConditionChecks runs all registered condition checks **in parallel**.
```
amshuman-kr, line 78: There will not be too many conditions. Is there a need to parallel execution? Maybe we can introduce it later if needed?

amshuman-kr (commented):
Thanks a lot for the PR @timuthy! Some very minor comments/questions below.

Besides that, I see that you have addressed the transition to `Unknown` state for the members but not the further transition [from `Unknown` to `NotReady`](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#7-some-members-have-been-in-unknown-status-for-a-while) and also the separate transition [for not ready pod to `NotReady` status](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#8-some-member-pods-are-not-ready-but-have-not-had-the-chance-to-update-their-status). Would you address that in a separate PR or do you think those cases should be addressed differently?

timuthy:
> Besides that, I see that you have addressed the transition to Unknown state for the members but not the further transition from Unknown to NotReady and also the separate transition for not ready pod to NotReady status. Would you address that in a separate PR or do you think those cases should be addressed differently?

Thanks for mentioning it @amshuman-kr. I will update this PR as written in the proposal and as discussed via Slack.

pkg/health/status/check.go
timuthy: Yes your right, we don't benefit much from parallelization today as the checks are quite small and it was more a consideration to make it ready for future checks, i.e., devs will only need to add the check but won't need to bother with optimization. Do you see any disadvantage of having this parallelization already today?

pkg/health/condition/condition_suite_test.go
timuthy: It's used for the tests in `builder_test.go`, see [here](https://github.com/gardener/etcd-druid/pull/188/files#diff-84d0ce038daaa8eb3c3f532d288d2e5eca44e33a74c148519ca5af8461b8400bR87). I will move it to this file to make it clearer.

pkg/health/condition/check_ready.go
timuthy: Good point, I deliberately chose `status.Member` because it reflects the current state of the cluster. Let's imagine you change `spec.replicas` from `1 -> 3`, but the two new pods will remain in `Pending` for some time.
1. Using `status.Members`, the condition will still be `Status: "True"` because it underlying etcd cluster still only exists of one member, i.e., quorum is given.
2. Using `status.Replicas` then it'll be `Status: "False"`.

The question is if we want to rather reflect the cluster state (1.) or refer to what's given in the `.spec` (2.).

What do you think?

pkg/health/status/check.go
amshuman-kr: I am fine keeping the parallel checks 👍

pkg/health/condition/check_ready.go
amshuman-kr: To summarise my understanding of quorum in an etcd cluster, `quorum = cluster_size/2 + 1` where `cluster_size` is the current number of members (excluding learners) in the cluster. While scaling `n -> m` where `n > 1 and m > 1` this is quite straightforward. So, using `len(status.members)` or `status.Replicas` may not make much of a difference provided scaling is always done on fully healthy clusters (`AllMembersReady`). There might be a difference if we scale even when all members are not ready (but the cluster is quorate).

But it gets a bit more complicated while bootstrapping an etcd cluster (empty or from backup) because then the `cluster_size` is determined by the number of peer URLs passed as `ETCD_INITIAL_CLUSTER`. The case of `1 -> n` is a special case because it is really a bootstrap scenario. In that case (as well as the usual cluster bootstrap case), I thought `status.Replicas` may reflect the contents of `ETCD_INITIAL_CLUSTER` better than `len(status.members)` because it will take time for the pods to come up and their sidecars to populate `status.members`. During that time the cluster is really not quorate.

In the case of scaling `n -> m` where `n > 1 and m > 1` when all members are not ready, using `status.Replicas` might be more accurate as well.

WDYT?

amshuman-kr: > In the case of scaling n -> m where n > 1 and m > 1 when all members are not ready, using status.Replicas might be more accurate as well.

On second thought, `len(status.members)` might be a better measure of quorum in this case. So, in summary, the bootstrap case might require `status.Replicas` to be considered.

amshuman-kr: No sure if we can detect the bootstrap case with current fields in the status. Might we need to introduce a `Bootstrapping` condition type?

amshuman-kr: cc @shreyas-s-rao @abdasgupta @ishan16696

amshuman-kr: I gave it some more thought. According to the [Kubernetes API reference](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/stateful-set-v1/#StatefulSetStatus)...

> replicas is the number of Pods created by the StatefulSet controller.

So, it indicates the number of pods that are present at any point in time. I think calculating quorum based on this is equally problematic during bootstrap (e.g., if initial cluster size is 5, we might wrongly declare that the cluster is quorate already when 2 replicas are ready).

Using `spec.Replicas` instead would lead to wrong quorum checks for the the case of `n -> m` where `n > 1 and m > 1` (e.g. when scaling `3  -> 7`, we might wrongly declare that the cluster is not quorate until there are 4 ready replicas). 

Adding a `Bootstrapping` condition type can help, but will make the logic complicated to both calculate the condition as well as to consider it in the `Ready` condition checker to use `spec.Replicas` for cluster size in that case.

Another alternative is to introduce a new field `status.initialReplicas` which can be set to `spec.Replicas` by `etcd-druid` whenever bootstrapping triggered. Then the `Ready` condition checker can consider `max(status.IntialReplicas, status.Replicas)` as the cluster size checking calculating quorum.

WDYT?

timuthy: > Using spec.Replicas instead would lead to wrong quorum checks for the the case of n -> m where n > 1 and m > 1 (e.g. when scaling 3 -> 7, we might wrongly declare that the cluster is not quorate until there are 4 ready replicas).

I think the same is true for `status.replicas` because as mentioned in the example of my previous comment, the count might be updated quite fast (i.e. Pod created) but then depending on different factors, it might take longer for the pod to be started (e.g. long disk attachment time) and the member to eventually to join the cluster the first time. I don't think that we want to report a quorum loss in the meantime because (a) there is still a quorum, (b) potential remediation (e.g. killing the pod) can end in a worse situation.

From what I can see:
- Druid does know very little about a running etcd cluster
- It should rely on information provided by the etcd-br side car through `etcd.status.members[]`
- Thus, the etcd-br side car should update the corresponding member as soon as possible
- For bootstrapping I either like your `status.intialReplicas` or we follow the [status.lastOperation](https://github.com/gardener/gardener/blob/0968f39752e046235ed1bdf4618781b6b57268c9/pkg/apis/core/v1beta1/types_common.go#L101) approach which we already use for the most of our Gardener resources. The `status.lastOperation.type` will identify if the cluster is bootstrapping (`Create`/`Restore`) or not.

amshuman-kr: > I think the same is true for status.replicas

Agreed about the problem with using `status.Replicas`. We should use `len(status.Members)` like you have done in the PR 👍 

> For bootstrapping I either like your status.intialReplicas or we follow the status.lastOperation approach which we already use for the most of our Gardener resources. The status.lastOperation.type will identify if the cluster is bootstrapping (Create/Restore) or not.

I am thinking that we might need both. Only using `status.lastOperation` would mean that we have to use `spec.Replicas` as initial replicas in the bootstrapping case. But this might be incorrect, if `spec.replicas` is updated while bootstrapping is on-going. So, it might be safer for druid to copy `spec.replicas` to `status.initialReplicas` at the beginning of triggering bootstrapping. Though strictly not necessary `status.lastOperation` might be more useful for human users to understand what is going on.

timuthy: > But this might be incorrect, if spec.replicas is updated while bootstrapping is on-going

You have point there which at the same time brings me to my next question: Should we even exit the reconciliation loop as long as not all pods are ready after a spec change (e.g. `spec.replicas: 1` -> `spec.replicas: 3`)? I like the short cycle that the Druid has today but OTHO waiting until everything is up-and-running also has its benefits, e.g., a `Processing` operation signals that the cluster is still "under construction" and thus conditions can flap in the meantime.

timuthy:
> Besides that, I see that you have addressed the transition to Unknown state for the members but not the further transition from Unknown to NotReady and also the separate transition for not ready pod to NotReady status

Added via 134958db8e8cbbb9f6cec30fdbeaa05aa78587e7

pkg/health/condition/check_ready.go
amshuman-kr: > waiting until everything is up-and-running also has its benefits, e.g., a Processing operation signals that the cluster is still "under construction" and thus conditions can flap in the meantime.

I find long reconciliations against the _spirit_ of control-loops, especially, if they are reliant on such long reconciliation loops to not be interrupted. E.g. while waiting for the cluster to bootstrap the druid pod might restart and then it will need a way to compute the current state during the next reconciliation. Using `spec.replicas` at that time could be just as wrong. So, something like `status.intialReplicas` will be still required (but only in rarer cases). I think solutions that work for short reconciliation loops can be simpler because of lack of in-memory state and work better in general.

But this is in no way an argument against `Processing` or `Bootstrapping` operation type in `status.lastOperation`. In fact, it is an argument that we should have it but along with `status.initialReplicas` and short reconcilations.

pkg/health/etcdmember/check_ready.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return results
+}
+
+func (r *readyCheck) checkPodIsRunning(ctx context.Context, namespace string, member druidv1alpha1.EtcdMemberStatus) (bool, error) {
+	pod := &corev1.Pod{}
+	if err := r.cl.Get(ctx, kutil.Key(namespace, member.PodRef.Name), pod); err != nil {
+		return false, err
+	}
+	return pod.Status.Phase == corev1.PodRunning, nil
```
amshuman-kr, line 88: Should we check the [ContainersReady](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions) condition here?

amshuman-kr (commented):
Thanks for the change @timuthy! They look good. Do you think adding `status.initialReplicas` and using `max(status.initialReplicas, len(status.Members))` make sense? The maintenance of `status.initialReplicas` could be postponed to later issues where the custodian is enhanced for other reconciliation actions like triggering bootstrapping.

WDYT?

pkg/health/condition/check_ready.go
amshuman-kr: I have updated the solution in the proposal in #189 based on our out-of-band discussion.

timuthy:
Thanks for the discussion @amshuman-kr. I added the missing pieces in the last 3 commits (separated it for an easier review). We can squash merge the PR once we are ready.

controllers/etcd_custodian_controller.go
```diff
@@ -131,43 +133,47 @@ func (ec *EtcdCustodian) updateEtcdStatus(ctx context.Context, logger logr.Logge
...
-		etcd.Status.UpdatedReplicas = sts.Status.UpdatedReplicas
-		etcd.Status.Ready = &ready
-		logger.Info(fmt.Sprintf("ETCD status updated for statefulset current replicas: %v, ready replicas: %v, updated replicas: %v", sts.Status.CurrentReplicas, sts.Status.ReadyReplicas, sts.Status.UpdatedReplicas))
-		return nil
-	})
-}
+		// Bootstrap is a special case which is handled by the etcd controller.
+		if !inBootstrap(etcd) && len(members) != 0 {
+			etcd.Status.ClusterSize = pointer.Int32Ptr(int32(len(members)))
+		}
```
amshuman-kr, line 142: Strictly speaking, this is incorrect. Only the initial bootstrapping is triggered by the etcd controller. Later bootstrapping due to [quorum failure](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#10-quorum-lost-with-a-majority-of-members-notready) is to be [done by the custodian controller](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#recommended-action-9). But this change could be done in https://github.com/gardener/etcd-druid/issues/158.

pkg/health/condition/check_ready.go
```diff
@@ -14,22 +14,26 @@
...
-			reason:  "NoMembersInStatus",
-			message: "Cannot determine readiness of cluster since status has no members",
+			reason:  "ClusterSizeUnknown",
+			message: "Cannot determine readiness of cluster since no cluster size has been calculated",
 		}
 	}
 
 	var (
-		quorum       = len(status.Members)/2 + 1
+		size         = utils.Max(int(*status.ClusterSize), len(status.Members))
```
amshuman-kr, line 35: This could be now simply `int(*status.ClusterSize)`, right?

amshuman-kr (commented):
Thanks a lot for the changes @timuthy! The changes look good. Just a couple of comments below (the first one is merely FYI).

controllers/etcd_custodian_controller.go
timuthy: I added this comment as this reflects the current state.

Tbh, I'm not convinced yet that the Custodian controller performs self healing in the form of re-bootstrapping because the Etcd controller can in the meantime still reconcile `etcd` resources which can lead to races. In order to avoid a race, you need to introduce some kind of lock which is also not too nice, IMO.

Maybe I didn't check all cases in detail but at the moment I think an alternative would be easier to implement:
1. Custodian sees that cluster is not healthy and self healing is required.
2. Custodian triggers a reconciliation for the affected `etcd` resource.
3. Etcd controller kicks in and takes necessary action to bring cluster up again (=> desired state). In the meantime any changes to the very same `etcd` resource cannot lead to a race.

We don't have to decide in this PR of course, but still wanted to give an alternative approach to think about.

pkg/health/condition/check_ready.go
timuthy: Not if the cluster size changes at the same time. `status.ClusterSize` is only calculated in the status update [here](https://github.com/gardener/etcd-druid/pull/188/files#diff-a2ef4cdc737ef6465c303ae9670522738c19205fb68a412813727558230c43a3R141), after the conditions have been calculated.

controllers/etcd_custodian_controller.go
amshuman-kr: > In order to avoid a race, you need to introduce some kind of lock which is also not too nice, IMO.

Even if only the etcd controller is responsible for bootstrapping non-quorate clusters, some sort of race avoidance is still required because the `Etcd` spec might get updated while bootstrapping is still on-going at which case it should either wait until the on-going bootstrapping completed before applying the latest spec changes or should abort the current bootstrapping (which might unnecessarily prolong the downtime).

> 2. Custodian triggers a reconciliation for the affected etcd resource.

This cannot be the regular etcd controller reconciliation with which might apply `Etcd` resource spec changes along with triggering bootstrapping. As a part of the contract of a gardener extension, `etcd-druid` not support apply `Etcd` resource spec changes unless `reconcile` `operation` annotation is present.

On the other hand, it makes no sense for the `etcd-druid` to wait for the next shoot reconciliation to fix a non-quorate ETCD cluster.

But let's take this discussion out of this PR.

amshuman-kr (approved):
Thanks @timuthy for the changes. /lgtm

PR #190, opened 2021-06-11
Enable profiling of etcd-backup-restore-sidecar

majst01:
  /area ops-productivity
  /area robustness
 /kind enhancement

We face currently a memory leak in the etcd-backup-restore-sidecar, but profiling is not configurable with the etcd-druid.

### DIFF
--- a/config/crd/bases/druid.gardener.cloud_etcds.yaml
+++ b/config/crd/bases/druid.gardener.cloud_etcds.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.5.0
   creationTimestamp: null
   name: etcds.druid.gardener.cloud
 spec:
--- a/config/crd/bases/druid.gardener.cloud_etcds.yaml
+++ b/config/crd/bases/druid.gardener.cloud_etcds.yaml
@@ -78,6 +78,10 @@ spec:
                     description: DeltaSnapshotPeriod defines the period after which
                       delta snapshots will be taken
                     type: string
+                  enableProfiling:
+                    description: EnableProfiling defines if profiling should be enabled
+                      for the etcd-backup-restore-sidecar
+                    type: boolean
                   fullSnapshotSchedule:
                     description: FullSnapshotSchedule defines the cron standard schedule
                       for full snapshots.
@@ -229,10 +233,6 @@ spec:
                     description: DefragmentationSchedule defines the cron standard
                       schedule for defragmentation of etcd.
                     type: string
-                  enableProfiling:
-                    description: EnableProfiling defines if profiling should be enabled
-                      for the etcd-backup-restore-sidecar
-                    type: boolean
                   image:
                     description: Image defines the etcd container image and tag
                     type: string
--- a/config/crd/bases/druid.gardener.cloud_etcds.yaml
+++ b/config/crd/bases/druid.gardener.cloud_etcds.yaml
@@ -229,6 +229,10 @@ spec:
                     description: DefragmentationSchedule defines the cron standard
                       schedule for defragmentation of etcd.
                     type: string
+                  enableProfiling:
+                    description: EnableProfiling defines if profiling should be enabled
+                      for the etcd-backup-restore-sidecar
+                    type: boolean
                   image:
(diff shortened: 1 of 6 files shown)


amshuman-kr (commented):
Thanks a lot for your contribution @majst01!

Since enabling profiling might have performance implications, enabling profiling by default may not be desirable. Wouldn't making this controller by either a CLI flag to `etcd-druid` or via field in `Etcd` resource spec make more sense?

majst01:
Hi @amshuman-kr 

set it false by default and made it configurable by the etcdconfig resource, unsure if this is the correct way to do

amshuman-kr (commented):
Thanks for the quick changes @majst01! /lgtm

/second-opinion
/invite @abdasgupta @shreyas-s-rao Does the flag look good under `EtcdConfig`? Or is better under `EtcdSpec` or `SharedConfig`?

config/crd/bases/druid.gardener.cloud_etcds.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
```
timuthy, line 4: Something went wrong with the CRD generation because I don't see the new field here. Can you run `make install-requirements` and `make manifests` again?

majst01: Ha, good catch, added. Dunno why this was not added in the first run.

shreyas-s-rao:
@majst01 Thanks a lot for your contribution to etcd-druid. It would be great to know where and how exactly this memory leak in backup sidecar is occurring.
Regarding where the `EnableProfiling` should be kept, since the profiling is only for the backup sidecar and not the etcd itself, it would make more sense to keep it under `BackupSpec` rather than `EtcdConfig`. @amshuman-kr @abdasgupta WDYT?

majst01:
The memory leak occurs in the backup sidecar, we guess it occurs since we enabled full-backups-only.
You are right, putting it into BackupSpec makes more sense then

shreyas-s-rao (approved):
@majst01 Thanks for making the requested changes. LGTM

amshuman-kr (approved):
Thanks a lot @shreyas-s-rao @majst01 ! /lgtm

majst01:
Thanks guys !
is a container image created from master ?

amshuman-kr:
> is a container image created from master ?

The master build failed due to a failing `make check-generate`. I have raised #192  to fix it. There should be a master build image after that is merged. Sorry, for not noticing this in this PR.

shreyas-s-rao:
@amshuman-kr thanks for fixing the check.

@majst01 the container image with your change has been created here -> `eu.gcr.io/gardener-project/gardener/etcd-druid:v0.6.0-dev-a47e74fd1756951720f8f309cfa34060c677c021`.

shreyas-s-rao:
@majst01 Are you still seeing the memory leak in backup sidecar? FYI, I recently fixed some potential memory leak in backup-restore code in https://github.com/gardener/etcd-backup-restore/pull/357 and will be released as part of etcd-backup-restore `v0.13.0`. This might possibly fix the memory leak you saw as well.

majst01:
Hi @shreyas-s-rao,

yes we still see memory-leaks, some of our loaded clusters need a etcd-backup-restore sidecar restart every 3-4 days. Sadly we where not able to grab a useful pprof dump to see what happens there.
Is v0.13.0 a simple drop in replacement ? or is an update of g/g required ?

majst01:
Do you have an ETA of `v0.13.0` ? Or is a container built from this PR which can be used ?

shreyas-s-rao:
@majst01 Normally, etcd-backup-restore release is packaged as part of etcd-druid release, and etcd-druid release is packed into a g/g release, so it might take atleast 2-3 more weeks before you can consume this out of the box. You can use the new container image `eu.gcr.io/gardener-project/gardener/etcdbrctl:v0.13.0-dev-8edc33593b4706d09c3ef95d40107b872ce19391` within the `etcd` resource spec (by setting the field `spec.backup.image`), but most likely the etcd spec will be reconciled by gardener to the original version and end up reversing your changes, so you might have to wait for a g/g release.

majst01:
OK, but the imagevector of etcdbrctl can be overridden by the gardener imagevectors if i remember correct. Will check.

amshuman-kr:
I wonder if the fix https://github.com/gardener/etcd-backup-restore/pull/357 is related to the memory leak mentioned here.

PR #193, opened 2021-06-17
Updated configmap mountpoints based on ETCD replicas.

abdasgupta:

/area control-plane
/kind enhancement

This PR changes the volume mount points of configmap onto the ETCD pods based on the size of clusters.

Solves 1st point of https://github.com/gardener/etcd-backup-restore/issues/322

**Release note**:

```other operator
None
```


controllers/etcd_controller.go
```diff
@@ -365,6 +365,10 @@ func (r *EtcdReconciler) reconcileServices(ctx context.Context, logger logr.Logg
 		return nil, err
 	}
 
+	if etcd.Spec.Replicas > 1 {
+		svc.Spec.ClusterIP = "None"
+	}
```
amshuman-kr, line 370: Wouldn't it work without this change? The pods would be scaled down and the service's `podSelector` will no longer match.

controllers/etcd_controller.go
```diff
@@ -412,6 +416,10 @@ func (r *EtcdReconciler) syncServiceSpec(ctx context.Context, logger logr.Logger
 	if err != nil {
 		return nil, err
 	}
+
+	if etcd.Spec.Replicas > 1 {
+		svcCopy.Spec.ClusterIP = "None"
+	}
```
amshuman-kr, line 422: Same as [above](https://github.com/gardener/etcd-druid/pull/193/files#diff-3480d2ce46715d0b85c7f54bb4ae498a20da29ad632c61b1127c32254c0ad73cR368-R370).

controllers/etcd_controller.go
```diff
@@ -554,6 +561,12 @@ func (r *EtcdReconciler) getConfigMapFromEtcd(etcd *druidv1alpha1.Etcd, rendered
 	if err = decoder.Decode(&decoded); err != nil {
 		return nil, err
 	}
+
+	decoded.Name = fmt.Sprintf("etcd-bootstrap-%s", string(etcd.UID[:6]))
+	config := decoded.Data["etcd.conf.yaml"]
+	config = fmt.Sprintf("%s\ninitial-cluster: %s", config, values["initialCluster"])
+
+	decoded.Data["etcd.conf.yaml"] = config
```
amshuman-kr, line 569: Here we are modifying the rendered template. Why not make the change in the configmap template directly? Secondary changes after rendering will get quickly become confusing.

controllers/etcd_controller.go
```diff
@@ -852,8 +865,11 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
 	}
 
 	var statefulsetReplicas int
-	if etcd.Spec.Replicas != 0 {
-		statefulsetReplicas = 1
+	// Check if Spec.Replicas is odd or even
+	if etcd.Spec.Replicas&1 == 1 {
+		statefulsetReplicas = etcd.Spec.Replicas
+	} else {
+		return map[string]interface{}{}, fmt.Errorf("Spec.Replicas should not be even number: %d", etcd.Spec.Replicas)
```
amshuman-kr, line 872: Should we set this error on the `Etcd` resource status somewhere? Otherwise, this message will be lost in the logs.

controllers/etcd_controller.go
```diff
@@ -978,6 +994,18 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
+	if statefulsetReplicas > 1 {
+		etcdConfigMountPath = "/var/etcd/template/config/"
+		// form initial cluster
+		initialCluster = ""
+		for i := 0; i < statefulsetReplicas; i++ {
+			initialCluster = initialCluster + fmt.Sprintf("%s=http://%s-%d.%s:2380,", etcd.Name, etcd.Name, i, etcd.Name)
+		}
+	}
+
+	initialCluster = strings.Trim(initialCluster, ",")
```
amshuman-kr, line 1008: Can you please make this into a helper function which then can be unit tested to see if the initial cluster string is being generated properly in all the cases?

controllers/etcd_controller.go
```diff
@@ -978,6 +994,18 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
 	}
 
+	etcdConfigMountPath := "/var/etcd/config/"
+	initialCluster := fmt.Sprintf("%s=http://0.0.0.0:2380", etcd.Name)
+	if statefulsetReplicas > 1 {
+		etcdConfigMountPath = "/var/etcd/template/config/"
+		// form initial cluster
+		initialCluster = ""
+		for i := 0; i < statefulsetReplicas; i++ {
+			initialCluster = initialCluster + fmt.Sprintf("%s=http://%s-%d.%s:2380,", etcd.Name, etcd.Name, i, etcd.Name)
```
amshuman-kr, line 1004:
```suggestion
			initialCluster = initialCluster + fmt.Sprintf("%s=http://%s-%d.%s:2380,", etcd.Name, etcd.Name, i, svc.Name)
```
Technically, the trailing name in the DNS should be service name (though it is the same as the `Etcd` resource name in this case).

controllers/etcd_controller_test.go
```diff
@@ -602,6 +605,214 @@ var _ = Describe("Druid", func() {
...
+				if sts.Spec.Template.Spec.Containers[0].VolumeMounts[1].MountPath != "/var/etcd/config/" {
+					return fmt.Errorf("Etcd config for ETCD container is not properly updated")
+				}
+
+				if sts.Spec.Template.Spec.Containers[1].VolumeMounts[1].MountPath != "/var/etcd/config/" {
+					return fmt.Errorf("Etcd config ETCD backup container is not properly updated")
+				}
+
+				return nil
+			}, timeout, pollingInterval).Should(BeNil())
```
amshuman-kr, line 714:
```suggestion
			}, timeout, pollingInterval).Should(Succeed())
```
Reads better this way.

controllers/etcd_controller_test.go
```diff
@@ -602,6 +605,214 @@ var _ = Describe("Druid", func() {
...
+
+			Eventually(func() error {
+				return c.Get(context.TODO(), types.NamespacedName{
+					Name:      instance.Name,
+					Namespace: instance.Namespace,
+				}, instance)
+			}, timeout, pollingInterval).Should(BeNil())
```
amshuman-kr, line 721:
```suggestion
			}, timeout, pollingInterval).Should(Succeed())
```

controllers/etcd_controller_test.go
```diff
@@ -602,6 +605,214 @@ var _ = Describe("Druid", func() {
...
+
+			// No Service has been created by controller as even number of replicas are not allowed
+			svc = &corev1.Service{}
+			Expect(c.Get(context.TODO(), types.NamespacedName{
+				Name:      fmt.Sprintf("%s-client", instance.Name),
+				Namespace: instance.Namespace,
+			}, svc)).Should(matchers.BeNotFoundError())
+			svc = nil
+		})
+		It("configmaps are mounted properly when ETCD replicas are odd number", func() {
```
amshuman-kr, line 685: Would it be possible to re-use the common code between the case for `replicas: 1` and `replicas: 3`?

amshuman-kr (dismissed):
Thanks for the changes @abdasgupta! Looks good overall. Some small changes and suggestions below.

controllers/etcd_controller.go
abdasgupta: I didn't understand.. Eventually, pods would be scaled up and clusterIP is for then

controllers/etcd_controller.go
abdasgupta: Okay. You mean in configmap of chart directory?

controllers/etcd_controller.go
abdasgupta: The error returned this way is already set as error status [https://github.com/gardener/etcd-druid/blob/048277792bde4401801303aa62d177d10fcaa243/controllers/etcd_controller.go#L257](url)

controllers/etcd_controller.go
abdasgupta: svc name is not available there.. So I improved the readability to let know that etcd name is being used as pod name and svc name

controllers/etcd_controller_test.go
abdasgupta: Yes. Used describetable

config/crd/bases/druid.gardener.cloud_etcds.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.5.0
+    controller-gen.kubebuilder.io/version: v0.4.1
```
aaronfern, line 7: Just a clarification. Is this an intended change? 
I imagine this happened when running `make test` locally before pushing the changes and hence is not intended.

aaronfern (commented):
Hi @abdasgupta 
I just have one thing I wanted to clarify

config/crd/bases/druid.gardener.cloud_etcds.yaml
abdasgupta: I don't know how this change happened!!

abdasgupta: I think it was done automatically. I didn't make any changes here. @amshuman-kr , Any Idea?

aaronfern: I've seen this before
From what I gather, if the `controller-gen` version installed locally is of a different version, it updated this file to it's version before running `make test`
Why it tries to have the entire description in one line I'm not too sure though

amshuman-kr: Good catch @aaronfern! I think if you update your `controller-gen` and `make generate` again, it would go away.

amshuman-kr: However, I tried to freshly install `controller-gen` from the currently vendored package, ran `make manifests` and got exactly the same as I see in this PR. So, at least it is consistent with the vendoring. 

I think in the previous update the controller-gen was locally updated but the vendoring was not updated.

For consistent environment, we should ideally update `controller-gen` by updating vendoring first. Maybe we should pick this in a separate PR?

aaronfern (commented):
Even though this PR introduces a component for configmap, it is still deployed using the chart
Could we also introduce the changes to deploy the configmap using the component deployer in this PR?

ishan16696 (dismissed):
Hi @abdasgupta ,
I have few concerns regarding that I have made few comments.

pkg/component/etcd/configmap/values.go
```diff
@@ -0,0 +1,79 @@
...
+	svcName := etcd.Name
+	podName := fmt.Sprintf("%s-%d", etcd.Name, 0)
+	initialCluster := fmt.Sprintf("%s=%s://%s.%s:2380", podName, protocol, podName, svcName)
+	if statefulsetReplicas > 1 {
+		etcdConfigMountPath = "/var/etcd/template/config/"
+		// form initial cluster
+		initialCluster = ""
+		for i := 0; i < statefulsetReplicas; i++ {
+			podName = fmt.Sprintf("%s-%d", etcd.Name, i)
+			initialCluster = initialCluster + fmt.Sprintf("%s=%s://%s.%s:2380,", podName, protocol, podName, svcName)
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/component/etcd/configmap/values.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+type Values struct {
+	// EtcdName is the name of the etcd resource.
+	EtcdName string
+	// EtcdName is the UID of the etcd resource.
+	EtcdUID types.UID
+	// InitialCluster is the initial cluster value to bootstrap ETCD.
+	InitialCluster string
+	// EtcdConfigMountPath is the mountpath for ETCD config.
```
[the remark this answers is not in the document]
abdasgupta: Done

abdasgupta:
> Even though this PR introduces a component for configmap, it is still deployed using the chart
> Could we also introduce the changes to deploy the configmap using the component deployer in this PR?

Done

controllers/etcd_controller.go
```diff
@@ -1544,29 +1439,3 @@ func (r *EtcdReconciler) claimConfigMaps(ctx context.Context, etcd *druidv1alpha
 	cm := NewEtcdDruidRefManager(r.Client, r.Scheme, etcd, selector, etcdGVK, canAdoptFunc)
```
aaronfern, line 1439: This comment refers to line #1425:
The function `claimConfigMaps` is unused. Can you remove this?

controllers/etcd_controller_test.go
```diff
@@ -824,11 +970,12 @@ func podDeleted(c client.Client, etcd *druidv1alpha1.Etcd) error {
 
 }
 
-func validateEtcdWithDefaults(instance *druidv1alpha1.Etcd, s *appsv1.StatefulSet, cm *corev1.ConfigMap, svc *corev1.Service) {
+func validateEtcdWithDefaults(instance *druidv1alpha1.Etcd, s *appsv1.StatefulSet, cm *corev1.ConfigMap, clSvc *corev1.Service, prSvc *corev1.Service) {
 	// Validate Quota
 	configYML := cm.Data[etcdConfig]
 	config := map[string]string{}
 	err := yaml.Unmarshal([]byte(configYML), &config)
+	fmt.Println(err)
```
aaronfern, line 978: Can you please remove this?

controllers/etcd_controller_test.go
```diff
@@ -2462,6 +2623,37 @@ func getEtcd(name, namespace string, tlsEnabled bool) *druidv1alpha1.Etcd {
 	return instance
 }
 
+func getEtcdWithReplicas(name, namespace string, replicas int) *druidv1alpha1.Etcd {
```
aaronfern, line 2626: Would it be easier to just call `getEtcdWithDefault` and update the replicas here?

controllers/etcd_controller.go
```diff
@@ -1035,17 +935,15 @@ func checkEtcdAnnotations(annotations map[string]string, etcd metav1.Object) boo
...
-		statefulsetReplicas = 1
-	}
+	statefulsetReplicas := int(etcd.Spec.Replicas)
 
 	etcdValues := map[string]interface{}{
 		"clientPort":              val.Service.ClientPort,
 		"defragmentationSchedule": etcd.Spec.Etcd.DefragmentationSchedule,
 		"enableTLS":               (etcd.Spec.Etcd.TLS != nil),
-		"pullPolicy":              corev1.PullIfNotPresent,
+		"pullPolicy":              corev1.PullAlways,
```
aaronfern, line 944: Why are we using `PullAlways` over `PullIfNotPresent`?

controllers/etcd_controller.go
```diff
@@ -1091,7 +989,7 @@ func getMapFromEtcd(im imagevector.ImageVector, etcd *druidv1alpha1.Etcd, val co
 	}
 
 	backupValues := map[string]interface{}{
-		"pullPolicy":               corev1.PullIfNotPresent,
+		"pullPolicy":               corev1.PullAlways,
```
aaronfern, line 992: Same as above

charts/etcd/values.yaml
```diff
@@ -18,8 +18,9 @@ labels: {}
 etcd:
   initialClusterToken: initial
   initialClusterState: new
+  initialCluster: test
   enableTLS: false
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
aaronfern, line 23: I think these were changes done for local testing. Can you please revert?

charts/etcd/values.yaml
```diff
@@ -33,8 +34,10 @@ etcd:
   #password: password
   heartbeatDuration: 10s
 
+etcdConfigMountPath: "/var/etcd/config/"
 backup:
-  pullPolicy: IfNotPresent
+  port: 8080
+  pullPolicy: Always
```
aaronfern, line 40: Same as above

charts/images.yaml
```diff
@@ -2,8 +2,8 @@ images:
 - name: etcd-backup-restore
   sourceRepository: github.com/gardener/etcd-backup-restore
   repository: eu.gcr.io/gardener-project/gardener/etcdbrctl
-  tag: "v0.14.0"
+  tag: "v0.15.0"
```
aaronfern, line 5: Same as above

charts/images.yaml
```diff
@@ -2,8 +2,8 @@ images:
...
 - name: etcd
   sourceRepository: github.com/gardener/etcd-custom-image
   repository: eu.gcr.io/gardener-project/gardener/etcd
-  tag: "v3.4.13-bootstrap-1"
+  tag: "v3.4.13-bootstrap-2"
```
aaronfern, line 9: Same as above

config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
         # Change the value of image field below to your controller image URL
-        - image: eu.gcr.io/gardener-project/gardener/etcd-druid:v0.7.0-dev
+        - image: abdasgupta/etcd-druid:0.8.0-dev
```
aaronfern, line 11: Same as above

aaronfern (changes requested):
Just some minor changes and suggestions

Also, there is still some residual claim logic for config maps left in `controller_ref_manager.go`. As we would be removing the claim logic for config maps in this PR, please remove the unneeded logic there.
Could you also mention this removal of claim logic for config maps in the release notes?

charts/etcd/templates/etcd-configmap.yaml
```diff
@@ -53,6 +53,9 @@ data:
     # Initial cluster state ('new' or 'existing').
     initial-cluster-state: {{ .Values.etcd.initialClusterState }}
 
+    # Initial cluster
```
timuthy, line 56: I think you can delete this and the entire file as you have switched to a `component`, right?

controllers/etcd_controller.go
```diff
@@ -941,21 +844,15 @@ func (r *EtcdReconciler) reconcileRoleBinding(ctx context.Context, logger logr.L
...
-		Service: componentservice.GenerateValues(etcd),
-	}
-
-	values, err := getMapFromEtcd(r.ImageVector, etcd, val, r.disableEtcdServiceAccountAutomount)
-
-	if err != nil {
-		return nil, nil, err
+	// Check if Spec.Replicas is odd or even.
+	if etcd.Spec.Replicas&1 == 0 {
+		return nil, nil, fmt.Errorf("Spec.Replicas should not be even number: %d", etcd.Spec.Replicas)
```
timuthy, line 849: Can you please add a **breaking** release note for this?

controllers/etcd_controller_test.go
```diff
@@ -1609,7 +1752,7 @@ func validateEtcd(instance *druidv1alpha1.Etcd, s *appsv1.StatefulSet, cm *corev
 	}))
 }
 
-func validateStoreGCP(instance *druidv1alpha1.Etcd, s *appsv1.StatefulSet, cm *corev1.ConfigMap, svc *corev1.Service) {
+func validateStoreGCP(instance *druidv1alpha1.Etcd, s *appsv1.StatefulSet, cm *corev1.ConfigMap, clSvc *corev1.Service, prSvc *corev1.Service) {
```
timuthy, line 1755: Please double check the signature changes of these functions.

pkg/component/etcd/configmap/configmap.go
```diff
@@ -0,0 +1,208 @@
...
+			Name:      name,
+			Namespace: c.namespace,
+		},
+	}
+}
+
+func getObjectMeta(val Values) metav1.ObjectMeta {
+	labels := map[string]string{"name": "etcd", "instance": val.EtcdName}
+	return metav1.ObjectMeta{
+		Name:      fmt.Sprintf("etcd-bootstrap-%s", string(val.EtcdUID[:6])),
```
timuthy, line 191: Can you extract this to a function as done for the other `Get*Name` functions?

pkg/component/etcd/configmap/values_helper.go
```diff
@@ -0,0 +1,76 @@
...
+	protocol := "http"
+	if etcd.Spec.Etcd.TLS != nil {
+		protocol = "https"
+	}
+
+	statefulsetReplicas := int(etcd.Spec.Replicas)
+
+	etcdConfigMountPath := "/var/etcd/config/"
+	// Form the service name and pod name for mutinode cluster with the help of ETCD name
+	podName := fmt.Sprintf("%s-%d", etcd.Name, 0)
```
timuthy, line 60: Can you extract this to a function?

pkg/component/etcd/configmap/configmap.go
```diff
@@ -0,0 +1,208 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/gardener/gardener/pkg/controllerutils"
+	gardenercomponent "github.com/gardener/gardener/pkg/operation/botanist/component"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type component struct {
```
timuthy, line 32: Can you add rudimentary unit tests for this component?

timuthy (changes requested):

charts/images.yaml
abdasgupta: Why? This value should be updated, right? IMHO, We are not expecting this release of Druid to work with `v0.14.0` of `backup-restore`.

charts/images.yaml
abdasgupta: Why? This value should be updated, right? We are not expecting this release of Druid to work with `v3.4.13-bootstrap-1` of `custom-image`

config/default/manager_image_patch.yaml
abdasgupta: I would revert back this change. This change is automated

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -72,9 +73,9 @@ spec:
             - --cert=/var/etcd/ssl/client/tls.crt
             - --key=/var/etcd/ssl/client/tls.key
             - --cacert=/var/etcd/ssl/ca/ca.crt
-            - --endpoints=https://{{ .Values.name }}-local:{{ .Values.etcd.clientPort }}
+            - --endpoints=https://0.0.0.0:{{ .Values.etcd.clientPort }}
```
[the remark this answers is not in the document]
abdasgupta: This way, the livenessProbe is querying itself. Do you have any other suggestion here?

charts/images.yaml
timuthy: Simply because:
a) There will be a separate PR for `v0.15.0`.
b) `v0.15.0` is not released yet.
c) If we want to merge this PR before, the `main` branch is for sure broken (mainly for single node etcd).

charts/etcd/templates/etcd-statefulset.yaml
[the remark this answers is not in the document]
timuthy: Fine for me 👍

charts/etcd/values.yaml
```diff
@@ -18,8 +18,9 @@ labels: {}
 etcd:
   initialClusterToken: initial
   initialClusterState: new
+  initialCluster: test
```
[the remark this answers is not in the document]
timuthy: The `values.yaml` file is supposed to contain decent default values and for me it occurs like there is no good default value at all for `initialCluster` (esp. `test` does not represent a valid value). Hence, I suggest to remove it here and check in the chart that the value is set ([required example](https://github.com/gardener/gardener/blob/29a0c4433430ed735e9209c26b106c6817ad45d8/charts/gardener/gardenlet/charts/runtime/templates/deployment.yaml#L14)). WDYT?

timuthy:
@abdasgupta can you please rebase and check the failing tests? Thanks.

charts/etcd/values.yaml
```diff
@@ -33,8 +34,10 @@ etcd:
   #password: password
   heartbeatDuration: 10s
 
+etcdConfigMountPath: "/var/etcd/config/"
```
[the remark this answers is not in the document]
abdasgupta: Removed

controllers/compaction_lease_controller.go
```diff
@@ -112,7 +112,9 @@ func (lc *CompactionLeaseController) Reconcile(ctx context.Context, req ctrl.Req
 	}
 
 	if etcd.Spec.Backup.Store == nil {
-		return ctrl.Result{}, nil
+		return ctrl.Result{
+			Requeue: false,
+		}, nil
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/utils/namebuilder.go
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2022 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/utils/namebuilder.go
```diff
@@ -0,0 +1,46 @@
...
+
+package utils
+
+import (
+	"fmt"
+
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+)
+
+func GetPeerServiceName(etcd *druidv1alpha1.Etcd) string {
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/component/etcd/configmap/values_helper.go
```diff
@@ -0,0 +1,76 @@
...
+	domaiName := fmt.Sprintf("%s.%s.%s", utils.GetPeerServiceName(etcd), etcd.Namespace, "svc.cluster.local")
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/component/etcd/configmap/configmap.go
abdasgupta: Done

pkg/component/etcd/configmap/values_helper.go
abdasgupta: Done

pkg/component/etcd/configmap/configmap.go
abdasgupta: Done

config/default/manager_image_patch.yaml
abdasgupta: Done

charts/images.yaml
abdasgupta: Done

charts/images.yaml
abdasgupta: Done

controllers/etcd_controller_test.go
abdasgupta: Done

controllers/etcd_controller_test.go
abdasgupta: Done

controllers/etcd_controller.go
abdasgupta: Removed

pkg/component/etcd/configmap/configmap.go
```diff
@@ -0,0 +1,208 @@
...
+
+		serverPort := strconv.Itoa(int(pointer.Int32Deref(c.values.ServerPort, defaultServerPort)))
+
+		etcdConfigYaml = etcdConfigYaml + `
+    # List of comma separated URLs to listen on for client traffic.
+    listen-client-urls: ` + protocol + `://0.0.0.0:` + clientPort + `
+
+    # List of this member's client URLs to advertise to the public.
+    # The URLs needed to be a comma-separated list.
+    advertise-client-urls: ` + c.values.PeerServiceName + `@` + c.values.EtcdNameSpace + `@` + clientPort + `
```
timuthy, line 104: Values like `c.values.PeerServiceName` are never acutally used in the final (downloaded) ConfigMap because they are [replaced](https://github.com/gardener/etcd-backup-restore/pull/419/files#diff-3a74c5f616b5783bc51c10a48dcde0da7aa4d0e3b814070bf3d71eb7f05f6e44R436), right? Can we use some placeholder signal instead, e.g. `<Value>`, so that when an operator gets the ConfigMap it's obvious that this is only a placeholder?
The same goes for the other values as well.

timuthy (changes requested):

controllers/etcd_controller_test.go
timuthy: Ping

pkg/component/etcd/configmap/configmap.go
abdasgupta: Hmm.. So are you suggesting the ConfigMap to have `advertise-client-urls` value like `<etcd-peer-service@default@2379>` ?

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -27,7 +28,6 @@ spec:
   template:
     metadata:
       annotations:
-        checksum/etcd-configmap: {{ include (print $.Template.BasePath "/etcd-configmap.yaml") . | sha256sum }}
```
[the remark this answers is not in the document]
abdasgupta: How do I calculate the checksum now? Any suggestions?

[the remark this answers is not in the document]
timuthy: See https://github.com/gardener/gardener/blob/744e0ff5db4f7f58d70a2f43546d379303f8a2cf/pkg/operation/seed/seed.go#L993

pkg/component/etcd/configmap/configmap.go
timuthy: No, like `<Service>@<Namespace>@<Port>`

timuthy: Please forget about my comments above. I just realized the logic behind it and that you actually need those values. Sorry for the confusion.

However, you need to pass `c.values.ClientServiceName` here.

charts/etcd/templates/etcd-statefulset.yaml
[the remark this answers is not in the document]
abdasgupta: While implementing, I have a doubt. Should we take the checksum of just the data (means `etcd.conf.yaml`) of the configmap or the whole configmap. If we want to take the checksum of whole configmap, it may be very difficult to test and I would also want to know why that is necessary. If we want to take checksum of just `etcd.conf.yaml` residing in data of the configmap, then I understand the reason

charts/etcd/templates/etcd-statefulset.yaml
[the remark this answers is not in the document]
timuthy: What I forgot to mention: The server certificate for etcd needs to be replaced with this change since at least for Gardener `0.0.0.0` was not part of the `SAN` list. You should at least mention that explicitly in the release notes. Even better: Just change the endpoint in case of a multi-node etcd cluster to not break exiting single node clusters at all.

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -150,14 +153,14 @@ spec:
         - --cacert=/var/etcd/ssl/ca/ca.crt
         - --insecure-transport=false
         - --insecure-skip-tls-verify=false
-        - --endpoints=https://{{ .Values.name }}-local:{{ .Values.etcd.clientPort }}
+        - --endpoints=https://0.0.0.0:{{ .Values.etcd.clientPort }}
```
timuthy, line 156: Please revert this. `{{ .Values.name }}-local` is resolved to `127.0.0.1`.

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -150,14 +153,14 @@ spec:
...
         # enable TLS on backup-restore server reusing etcd cert bundle
         - --server-cert=/var/etcd/ssl/server/tls.crt
         - --server-key=/var/etcd/ssl/server/tls.key
 {{ else }}
         - --insecure-transport=true
         - --insecure-skip-tls-verify=true
-        - --endpoints=http://{{ .Values.name }}-local:{{ .Values.etcd.clientPort }}
+        - --endpoints=http://0.0.0.0:{{ .Values.etcd.clientPort }}
```
timuthy, line 163: Please revert this. `{{ .Values.name }}-local` is resolved to `127.0.0.1`.

pkg/component/etcd/configmap/configmap.go
```diff
@@ -0,0 +1,237 @@
...
+      key-file: /var/etcd/ssl/server/tls.key
+
+      # Enable peer cert authentication.
+      client-cert-auth: true
+
+      # Path to the peer server TLS trusted CA cert file.
+      trusted-ca-file: /var/etcd/ssl/ca/ca.crt
+
+      # Peer TLS using generated certificates
+      auto-tls: false`
```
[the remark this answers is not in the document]
timuthy: As discussed, we should enhance the `etcd` resource by

```
spec:
  etcd:
    tls:
      peerTLSSecretRef:
        name:
        namespace:
      tlsPeerCASecretRef:
        name:
        namespace:
```

This configuration should be considered to configure the `peer-transport-security:` section of this etcd configuration.

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -53,12 +53,17 @@ spec:
         command:
         - /var/etcd/bin/bootstrap.sh
         readinessProbe:
-          httpGet:
+          exec:
+            command:
+            - /usr/bin/curl
```
timuthy, line 58: See https://github.com/gardener/etcd-druid/pull/193#discussion_r817762008.

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -53,12 +53,17 @@ spec:
...
 {{- if .Values.etcd.enableTLS }}
-            scheme: HTTPS
-{{- end }}
-            path: /healthz
-            port: {{ .Values.backup.port }}
+            - --key=/var/etcd/ssl/client/tls.key
+            - --cacert=/var/etcd/ssl/ca/ca.crt
+            - --cert=/var/etcd/ssl/client/tls.crt
+            - https://0.0.0.0:{{ .Values.etcd.clientPort }}/health
```
timuthy, line 63:
```suggestion
            - https://{{ .Values.name }}-local:{{ .Values.etcd.clientPort }}/health
```
Otherwise `0.0.0.0` must be part of the server cert.

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -53,12 +53,17 @@ spec:
...
+{{ else }}
+            - http://0.0.0.0:{{ .Values.etcd.clientPort }}/health
```
timuthy, line 65:
```suggestion
            - http://{{ .Values.name }}-local:{{ .Values.etcd.clientPort }}/health
```

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -53,12 +53,17 @@ spec:
...
         - /var/etcd/bin/bootstrap.sh
         readinessProbe:
-          httpGet:
+          exec:
+            command:
+            - /usr/bin/curl
 {{- if .Values.etcd.enableTLS }}
-            scheme: HTTPS
-{{- end }}
-            path: /healthz
```
timuthy, line 60: Please use the `/healthz` for single node etcd clusters because this endpoint also considers owner checks.

timuthy (dismissed):
Last comments from my side. Some comments are still unresolved. @shreyas-s-rao please take over and dismiss my review as soon as comments are resolved. Thanks.

aaronfern (changes requested):
There are a few unnecessary functions relating to claiming configmaps in `controller_ref_manager.go` that are not needed now
[here](https://github.com/gardener/etcd-druid/blob/fefe1cf9c3e00aa571020f9ddb8652af7691c5cc/controllers/controller_ref_manager.go#L308), [here](https://github.com/gardener/etcd-druid/blob/fefe1cf9c3e00aa571020f9ddb8652af7691c5cc/controllers/controller_ref_manager.go#L365), and [here](https://github.com/gardener/etcd-druid/blob/fefe1cf9c3e00aa571020f9ddb8652af7691c5cc/controllers/controller_ref_manager.go#L426)
Can you please remove these?

controllers/etcd_controller.go
```diff
@@ -959,21 +862,15 @@ func (r *EtcdReconciler) reconcileRoleBinding(ctx context.Context, logger logr.L
 }
 
 func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd) (*string, *appsv1.StatefulSet, error) {
-	val := componentetcd.Values{
-		Lease:   componentlease.GenerateValues(etcd),
-		Service: componentservice.GenerateValues(etcd),
+	// Check if Spec.Replicas is odd or even.
+	if etcd.Spec.Replicas&1 == 0 {
```
aaronfern, line 866: This condition will turn out to be true if replicas is 0, which would prevent the cluster from being hibernated
Can you modify this condition to allow Replicas to be 0?

charts/etcd/templates/etcd-statefulset.yaml
abdasgupta: Done

charts/etcd/templates/etcd-statefulset.yaml
abdasgupta: Done

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -53,12 +53,17 @@ spec:
...
-            port: {{ .Values.backup.port }}
+            - --cert=/var/etcd/ssl/client/tls.crt
+            - --key=/var/etcd/ssl/client/tls.key
+            - --cacert=/var/etcd/ssl/ca/ca.crt
+            - https://0.0.0.0:{{ .Values.etcd.clientPort }}/health
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/component/etcd/configmap/configmap.go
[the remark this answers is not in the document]
abdasgupta: I changed the code to use separate certs for client and peer

charts/etcd/templates/etcd-statefulset.yaml
abdasgupta: Changed it. We need to modify both client and peer certificates anyway. Earlier, client certs were not considering multinode scenario

charts/etcd/templates/etcd-statefulset.yaml
abdasgupta: Done

charts/etcd/templates/etcd-statefulset.yaml
abdasgupta: Done

controllers/etcd_controller.go

api/v1alpha1/etcd_types_test.go
```diff
@@ -104,15 +104,24 @@ func getEtcd(name, namespace string) *Etcd {
...
+
+	peerTlsConfig := &TLSConfig{
 		TLSCASecretRef: corev1.SecretReference{
-			Name: "ca-etcd",
+			Name: "peer-url-ca-etcd",
+		},
+		ServerTLSSecretRef: corev1.SecretReference{
+			Name: "peer-url-etcd-server-tls",
 		},
 	}
```
aaronfern, line 126: Does peerTlsConfig not seed a `ClientTLSSecretRef`?

pkg/component/etcd/configmap/configmap_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+				Expect(cl.Get(ctx, kutil.Key(namespace, values.ConfigMapName), cm)).To(Succeed())
+				checkConfigmap(cm, values)
+			})
+		})
+	})
+
+	Describe("#Destroy", func() {
+		Context("when services do not exist", func() {
+			It("should destroy successfully", func() {
```
aaronfern, line 169: I don't understand the contexts of the tests in the `#Destroy` section
How is service in the picture?

pkg/component/etcd/configmap/values_helper.go
```diff
@@ -0,0 +1,75 @@
...
+		PeerServiceName:         utils.GetPeerServiceName(etcd),
+		ServerPort:              etcd.Spec.Etcd.ServerPort,
+		AutoCompactionMode:      etcd.Spec.Common.AutoCompactionMode,
+		AutoCompactionRetention: etcd.Spec.Common.AutoCompactionRetention,
+		ConfigMapName:           utils.GetConfigmapName(etcd),
+	}
+	return values
+}
+
+func prepareMultiNodeCluster(etcd *druidv1alpha1.Etcd) string {
```
aaronfern, line 50:
```suggestion
func prepareInitialCluster(etcd *druidv1alpha1.Etcd) string {
```

Suggesting a change as this will be used to create `initial-cluster` for a single node setup as well as a multi node setup

aaronfern (commented):
In general looks good to me. I did not test it out though
just a few questions and clarifications

api/v1alpha1/etcd_types_test.go
abdasgupta: Right. Currently, there is no need for client certificates for peer connection. Let me add a comment there as well.

pkg/component/etcd/configmap/configmap_test.go
abdasgupta: Mistake from my side

config/default/manager_image_patch.yaml
aaronfern, line 11: Please revert this before you merge

aaronfern (approved):
Thanks for the changes!
/lgtm

Just be sure to revert the image tag before you merge

config/default/manager_image_patch.yaml
abdasgupta: Done

PR #202, opened 2021-07-07
Fix reconciliation predicates.

### LINKED ISSUE #196: [BUG] Fix reconciliation predicates to fully comply with gardener extension contract
**Describe the bug:**
The main controller reconciles changes to `Etcd` resource `spec` even if `gardener.cloud/reconcile` annotation is not added. This is against the [gardener extension contract](https://github.com/gardener/gardener/blob/master/docs/extensions/reconcile-trigger.md).

**Expected behavior:**
The main controller should use the predicates in such a way that changes to the `Etcd` resource `spec` are reconciled only when the resource is also annotated appropriately. For example, see [here](https://github.com/gardener/gardener/blob/609107f2b19584315cc639b52668f73820dfc44a/extensions/pkg/controller/backupbucket/controller.go#L60-L72).

**Logs:**

 - Etcd version/commit ID : v0.5.2

amshuman-kr:

/area control-plane
/kind bug

If operation annotation is to be honoured, the reconciliation predicates now match only if either operation annotation is present, or last operation didn't succeed or if the resource is undergoing deletion. No other change (if not accompanied by any of these conditions) will trigger reconciliation.

Fixes #196 

**Release note**:

```bugfix operator
If operation annotation is to be honoured, the reconciliation predicates now match only if either operation annotation is present, or last operation didn't succeed or if the resource is undergoing deletion. No other change (if not accompanied by any of these conditions) will trigger reconciliation.
```

### DIFF
--- a/pkg/predicate/predicate_test.go
+++ b/pkg/predicate/predicate_test.go
@@ -26,6 +26,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
 	. "github.com/gardener/etcd-druid/pkg/predicate"
+
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 var _ = Describe("Druid Predicate", func() {
@@ -55,6 +57,12 @@ var _ = Describe("Druid Predicate", func() {
 	})
 
 	Describe("#StatefulSet", func() {
+		var pred predicate.Predicate
+
+		JustBeforeEach(func() {
+			pred = StatefulSetStatusChange()
+		})
+
 		Context("when status matches", func() {
 			BeforeEach(func() {
 				obj = &appsv1.StatefulSet{
@@ -70,12 +78,10 @@ var _ = Describe("Druid Predicate", func() {
 			})
 
 			It("should return false", func() {
-				predicate := StatefulSetStatusChange()
-
-				gomega.Expect(predicate.Create(createEvent)).To(gomega.BeTrue())
-				gomega.Expect(predicate.Update(updateEvent)).To(gomega.BeFalse())
-				gomega.Expect(predicate.Delete(deleteEvent)).To(gomega.BeTrue())
-				gomega.Expect(predicate.Generic(genericEvent)).To(gomega.BeTrue())
+				gomega.Expect(pred.Create(createEvent)).To(gomega.BeTrue())
+				gomega.Expect(pred.Update(updateEvent)).To(gomega.BeFalse())
+				gomega.Expect(pred.Delete(deleteEvent)).To(gomega.BeTrue())
+				gomega.Expect(pred.Generic(genericEvent)).To(gomega.BeTrue())
 			})
 		})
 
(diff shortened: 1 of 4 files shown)


pkg/predicate/predicate_test.go
```diff
@@ -104,172 +104,126 @@ var _ = Describe("Druid Predicate", func() {
...
-							Annotations: map[string]string{
-								v1beta1constants.GardenerOperation: v1beta1constants.GardenerOperationReconcile,
-							},
-						},
-					}
-				})
-
-				It("should return false", func() {
-					predicate := HasOperationAnnotation()
+			It("should return false", func() {
```
shreyas-s-rao, line 178: Typo: `should return true`?

pkg/predicate/predicate_test.go
```diff
@@ -104,172 +104,126 @@ var _ = Describe("Druid Predicate", func() {
...
 						},
-					}
-				})
+					},
+				}
+			})
 
-				It("should return false", func() {
-					predicate := HasOperationAnnotation()
+			It("should return false", func() {
```
shreyas-s-rao, line 220: Same as above

pkg/predicate/predicate_test.go
```diff
@@ -104,172 +104,126 @@ var _ = Describe("Druid Predicate", func() {
...
-		Describe("#OR", func() {
-			Context("when has no operation annotation", func() {
-				BeforeEach(func() {
-					obj = &druidv1alpha1.Etcd{
-						ObjectMeta: metav1.ObjectMeta{
-							Annotations: make(map[string]string),
-						},
-					}
-				})
+	Describe("#OR", func() {
```
shreyas-s-rao, line 189: I did not understand the use of this `Describe` block, since it's the same as the one above (`#HasOperationAnnotation`). Or am I missing something here?

shreyas-s-rao (changes requested):
@amshuman-kr Thanks a lot for the changes to the predicates and the addition of tests as well! I have a few questions. Please address them.

pkg/predicate/predicate_test.go
amshuman-kr: You are right. This was an existing section and I did not pay attention to the fact that it was identical to the other section.

amshuman-kr:
Thanks for the review @shreyas-s-rao! I have addressed your comments. Can you PTAL?

pkg/predicate/predicate_test.go
```diff
@@ -94,182 +100,185 @@ var _ = Describe("Druid Predicate", func() {
...
+		Context("when has neither operation annotation nor last error", func() {
+			BeforeEach(func() {
+				obj = &druidv1alpha1.Etcd{
+					ObjectMeta: metav1.ObjectMeta{
+						Annotations: make(map[string]string),
+					},
+				}
+			})
+
+			It("should return true", func() {
```
shreyas-s-rao, line 216: @amshuman-kr Thanks for making the changes and the rebase too! Just one small typo here to fix, otherwise LGTM

amshuman-kr: 😅

amshuman-kr:
@shreyas-s-rao I have addressed your comments. Can you PTAL?

shreyas-s-rao (approved):
@amshuman-kr Thanks for making the requested changes. LGTM

Issue #186, opened 2021-06-01
[Feature] Simplify reconciliation by simplifying claim logic for StatefulSet, Service and ConfigMap based on Etcd resource name


amshuman-kr:
**Feature (What you would like to be added):**

The reconciliation flow of `etcd-druid` includes claiming from potentially multiple pre-existing `StatefulSet`, `Service` and `ConfigMap` objects if they exist. This is done by [selecting the objects based on `spec.selector` in the `Etcd` resource](https://github.com/gardener/etcd-druid/blob/6030a7e67d726a665db690380fa2fa81adae3822/controllers/controller_ref_manager.go#L192), [claiming one of the matching objects (if any)](https://github.com/gardener/etcd-druid/blob/6030a7e67d726a665db690380fa2fa81adae3822/controllers/controller_ref_manager.go#L82-L153) and [deleting the rest of the objects (if any)](https://github.com/gardener/etcd-druid/blob/6030a7e67d726a665db690380fa2fa81adae3822/controllers/etcd_controller.go#L600). If no matching objects are found then a [new object is created](https://github.com/gardener/etcd-druid/blob/6030a7e67d726a665db690380fa2fa81adae3822/controllers/etcd_controller.go#L646).

The logic of claiming from multiple pre-existing objects objects based on`spec.selector` was done because of the following reasons.
1. Migration from the time before `etcd-druid` was introduced. I.e. adopting objects created from the time before `etcd-druid` was introduced minimised and simplified clean up.
2. Keeping options open before multi-node ETCD design was finalised to use a [single `StatefulSet` for all the members of an ETCD cluster](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#kubernetes-context). Another alternative of using [one `StatefulSet` for each member of an ETCD cluster](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#alternative-1) was still open at that time.

Now that the migration scenario as well as the multi-node design don't need the functionality of claiming from multiple pre-existing objects, we can simplify the claim logic to just pick the object to be claimed by the same name as the `Etcd` resource . We will still need the [claim functionality](https://github.com/gardener/etcd-druid/blob/6030a7e67d726a665db690380fa2fa81adae3822/controllers/controller_ref_manager.go#L82-L153) to mark it as claimed, of course.

**Motivation (Why is this needed?):**

**Approach/Hint to the implement solution (optional):**

abdasgupta:
[...]

PR #203, opened 2021-07-13
Migrate unit tests to use gomock.

### LINKED ISSUE #97: [Feature] Unit tests for etcd-druid reconciliation cycle
Unit tests for etcd-druid reconciliation cycle.

We should have both positive and negative scenarios covered in the unit tests to improve our own productivity and to avoid regression.

Replace the kubebuilder way of tests (running `kube-apiserver` and `etcd`) with mock APIs.

amshuman-kr:

/area control-plane
/kind technical-debt

Migrate all controller unit tests to use gomock framework. I have also tried to make the test flow easier to understand, to contribute and maintain.

Fixes #97 

This is in preparation for #186 so that any regressions can be caught early in the unit tests.
Apologies for large changes to `etcd_controller_test.go` but it made sense to rationalise test flow. I have also added additional tests in the deletion use-case as well many other permutations and combinations for regular reconciliation.

**Release note**:

```other operator
NONE
```


amshuman-kr:
The controller suite test execution time has come down from `2m35s` to `2.79s`.

amshuman-kr:
I have [rebased](https://github.com/gardener/etcd-druid/pull/203#event-5048414495) after the merge of #202 which conflicted with this PR.

amshuman-kr:
> I have [rebased](https://github.com/gardener/etcd-druid/pull/203#event-5048414495) after the merge of #202 which conflicted with this PR.

Something went wrong in the revendoring in the last rebase. Fixed [now](https://github.com/gardener/etcd-druid/pull/203#event-5048603863).

controllers/etcd_controller_test.go
```diff
@@ -147,654 +150,861 @@ func cmdIterator(element interface{}) string {
...
+				Describe("create or update", func() {
+					AfterEach(func() {
+						Expect(instance.Finalizers).To(ContainElement(FinalizerName), "etcd resource should have finalizer added")
+
+						if instance.Spec.Backup.Store != nil {
+							Expect(backupSecret.Finalizers).To(ContainElement(FinalizerName), "backup secret should have finalizer added")
+						}
+					})
+
+					Describe("without services and configmaps", func() {
```
timuthy, line 466: Can you add tests for "with services and configmaps" as well to verify the Druid's update mechanism? (-> `syncServiceSpec`, `syncConfigMapData`)

controllers/etcd_controller_test.go
```diff
@@ -147,654 +150,861 @@ func cmdIterator(element interface{}) string {
...
+											},
+										)
+									}
+								)
+
+								BeforeEach(func() {
+									sts = createStatefulset(instance.Name, "default", instance.Spec.Labels)
+									mockListGetAndUpdateStatefulSet(cl, labelSelector, sts).AnyTimes()
+
+									/*
```
timuthy, line 865: Is this still needed?

controllers/etcd_controller_test.go
```diff
@@ -2780,27 +2771,364 @@ func parseQuantity(q string) resource.Quantity {
...
+				src.Status.DeepCopyInto(&etcd.Status)
+				return nil
 			},
+		),
+	)
+
+	return calls(ret)
+}
+
+func mockListGetAndUpdateStatefulSet(cl *mockclient.MockClient, labelSelector labels.Selector, ss ...*appsv1.StatefulSet) calls {
```
timuthy, line 2872: `labelSelector` is unused.

timuthy (changes requested):
Kudos for your efforts to migrate.
I added a few inline comments.
Do you also want to remove the `envtest` tool from `.ci/test`?

One comment: The test cover many different use-cases which is also the reason there is a lot of nesting. As a reviewer it's not trivial to understand or as a developer not easy to maintain this with all the different mocks and `BeforeEach()`/`AfterEach`s being involved. But I can imagine that these tests will help us to extract functionality into smaller junks, especially from `etcd_controller.go` with a lower risk of introducing regressions while doing so. I really hope that we're going to go that way with the Druid :)

controllers/etcd_controller_test.go
amshuman-kr: I was in two minds if I should add that case or not :-) I decided against adding it in this PR considering that that case wasn't tested in the existing test cases. I have changed the claim logic considerably (simplified it to `CreateOrUpdate` or re-create)  in implementing #186 and will cover this case as part of that PR. I hope that is ok.

controllers/etcd_controller_test.go
amshuman-kr: No. Thanks for pointing out.

controllers/etcd_controller_test.go
amshuman-kr: Removed

amshuman-kr:
> Do you also want to remove the envtest tool from .ci/test?

It is [still used](https://github.com/gardener/etcd-druid/blob/master/api_tests/v1alpha1/suite_test.go#L65) in `api_tests`. I retained it just to verify if the CRDs are generated and working properly.

> The test cover many different use-cases which is also the reason there is a lot of nesting. As a reviewer it's not trivial to understand or as a developer not easy to maintain this with all the different mocks and BeforeEach()/AfterEachs being involved.

The existing tests covered whole reconciliation. Wanting to be sure that I still cover all the existing test cases after the migration, I retained the migrated tests also to cover whole of reconciliation. But I used nesting to introduce more variations in the test cases by piecemeal nesting and also to avoid too much duplication of the test setup.

> But I can imagine that these tests will help us to extract functionality into smaller junks, especially from etcd_controller.go with a lower risk of introducing regressions while doing so. I really hope that we're going to go that way with the Druid :)

Indeed, the goal is to modularise the reconciliation flow and to be able to then test these smaller functions independently. This way we should also be able to reduce the nesting of test blocks into independent test descriptions. I think this is best done over a few more PRs than in one single PR, if only to make sure we are not regressing or worse, losing test cases that were already covered.

amshuman-kr:
@timuthy Thanks for the review! I have addressed your comments. Can you PTAL?

controllers/etcd_controller_test.go
timuthy: Okay sure, thank you :)

timuthy (approved):
/lgtm
Thank you 👍

amshuman-kr:
[Rebased](https://github.com/gardener/etcd-druid/pull/203#event-5198488955) and resolved conflicts. Made the gmock based tests to work for the `lease` listing changes introduced in #214.

amshuman-kr:
[Rebased](https://github.com/gardener/etcd-druid/pull/203#event-5232839617) after the [`v0.6.0`](https://github.com/gardener/etcd-druid/releases/tag/v0.6.0) release.

timuthy:
Let's close this in favor of https://github.com/gardener/etcd-druid/issues/278 which will allow us to have sufficient coverage for single components (+ easy maintenance as opposed to maintaining plenty of Gomock assertions) and only a handful of envtest like tests which test the controller from end-to-end.
/close

Issue #206, opened 2021-07-20
☂️ [Feature] Add EtcdMember resource


timuthy:
**Feature (What you would like to be added):**
A new resource `EtcdMemeber` should be added to the `druid.gardener.cloud/v1alpha1` API group.

Example:
```yaml
apiVersion: druid.gardener.cloud/v1alpha1
kind: EtcdMember
metadata:
  labels:
    gardener.cloud/owned-by: etcd-test
  name: etcd-test-0 # pod name
  namespace: default
  ownerReferences:
  - apiVersion: druid.gardener.cloud/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: etcd
    name: etcd-test
    uid: <UID>
status:
  id: "1"
  lastTransitionTime: "2021-07-20T10:34:04Z"
  lastUpdateTime: "2021-07-20T10:34:04Z"
  name: member1
  reason: up and running
  role: Member
  status: Ready
```

Every etcd member in a cluster should have a corresponding `EtcdMember` resource which contains the shown status information. The `EtcdMember` resource ought to be created and maintained by the backup-restore side car. Etcd-Druid may set `status: Unkown` after `heartbeatGracePeriod` ([ref](https://github.com/gardener/etcd-druid/tree/master/docs/proposals/multi-node#6-some-members-have-not-updated-their-status-for-a-while)).

**Motivation (Why is this needed?):**
The original [proposal](https://github.com/gardener/etcd-druid/tree/master/docs/proposals/multi-node) intended the status information of each etcd member to be part of a `[]members` list in the `etcd.status` resource. However, this will lead to update conflicts as multiple clients try to update the same resource at nearly the same time and we cannot use any adequate patch technique (SSA failed for K8s versions <= 1.21, strategic-merge not supported for CRDs) to prevent that.

Subtasks
- [ ] #595 

/cc @shreyas-s-rao @amshuman-kr

shreyas-s-rao:
Thanks @timuthy for summarizing the issue.

timuthy:
@amshuman-kr, @shreyas-s-rao and I iterated again over the approach and agreed that we find the `EtcdMember` design confusing because the resource itself only exists of a `status` and is not meant to be reconciled by any controller.
[...]

Issue #179, opened 2021-05-21
[Feature] Prepare Druid for Server-Side Apply


timuthy:
**Feature (What you would like to be added):**
Druid should use and support [Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) where applicable once Gardener dropped the support for seed clusters with K8s <= 1.17 (https://github.com/gardener/gardener/issues/4083).

**Motivation (Why is this needed?):**
[Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) makes working with the `etcd` resource more efficiently when there will be more then one actor (motivated [here](https://github.com/gardener/etcd-backup-restore/issues/333#issuecomment-845035513)).

Tasks to be done:
- [ ] Add [markers](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy) for merge strategy to `etcd` resource
- [ ] Use SSA for `etcd` status updates

timuthy:
We experienced some major issues with SSA in our controllers:
- `x-kubernetes-list-map-keys` does not work properly with any version `< 1.21` which makes it impossible to use SSA for arrays, e.g. `status.conditions`.
- Further issues are mentioned in [gardener/gardener#4122](https://github.com/gardener/gardener/issues/4122)

Therefore, we close this issue and will work with alternatives (like https://github.com/gardener/etcd-druid/issues/206) to avoid conflict errors.

PR #207, opened 2021-07-26
Introduced member leases to work-around limitations in Server-Side-Apply

amshuman-kr:

/area control-plane
/kind enhancement

Introduced member leases to work-around limitations in Server-Side-Apply.

Alternative design to #206.

Limitations in Server-Side Apply are documented in #179. I have updated the proposal with alternative design that was discussed out-of-band of using `leases` per member instead of introducing a new `EtcdMember` resource (as described in #206).
cc @shreyas-s-rao @timuthy @abdasgupta

**Release note**:

```other operator
NONE
```


docs/proposals/multi-node/README.md
```diff
@@ -561,30 +564,38 @@ status:
...
 ### Members
 
-The `members` section of the status is intended to be updated by the `etcd-backup-restore` sidecar container corresponding to the member for the most part.
-Especially, members can be marked with `status: Ready` only by their `etcd-backup-restore` sidecar container.
-However, they can be marked with `status: NotReady` either by their `etcd-backup-restore` sidecar container (with `reason: HeartbeatFailed`) or by `etcd-druid` (as explained [below](#decision-table-for-etcd-druid-based-on-the-status)).
+ The `members` section of the status is intended to be maintained by `etcd-druid` (preferraby, the `custodian` controller of `etcd-druid`) based on the [`leases` of the individual members](#member-leases).
+
+#### Note
+
+An earlier design in this proposal was for the individual `etcd-backup-restore` sidecars to update the corresponding `status.members` entries themselves. But this was redesigned to use [member `leases`](#member-leases) to avoid conflicts rising from frequent updates and the limitations in the support for [Server-Side Apply]() in some versions of Kubernetes.
```
timuthy, line 590: link for Server-Side Apply is missing

docs/proposals/multi-node/README.md
```diff
@@ -716,7 +775,7 @@ Mark the status condition type `AllMembersReady` to `true`.
 
 Mark the status condition type `AllMembersReady` to `false`.
 
-### 4. Majority members are `Ready` but `Ready` condition is stale
+### 5. Majority members are `Ready` but `Ready` condition is stale
```
timuthy, line 778: Can you elaborate how that can happen? The `Ready` condition is deduced right after the `members` are checked. Thus, the `Ready` condition should never be stale.

docs/proposals/multi-node/README.md
```diff
@@ -744,7 +803,7 @@ Mark the status condition type `AllMembersReady` to `false`.
 
 Mark the status condition type `Ready` to `true`.
 
-### 5. Majority members are `NotReady` but `Ready` condition is stale
+### 6. Majority members are `NotReady` but `Ready` condition is stale
```
timuthy, line 806: See previous question.

docs/proposals/multi-node/README.md
```diff
@@ -866,15 +899,15 @@ Mark the `n - s` members (corresponding to the pods that are not `Ready`) as `No
...
+    - Members with expired `lease`: N/A
   - conditions:
     - Ready: true
     - AllMembersReady: false
     - BackupReady: true
 
 #### Recommended Action
 
-Delete the `f` `NotReady` member pods to force restart of the pods if they do not automatically restart via failed `livenessProbe`. The expectation is that they will either re-join the cluster as an existing member or remove themselves and join as new members on restart of the container or pod.
+Delete the `f` `NotReady` member pods to force restart of the pods if they do not automatically restart via failed `livenessProbe`. The expectation is that they will either re-join the cluster as an existing member or remove themselves and join as new members on restart of the container or pod and [renew their `leases`](#member-leases).
```
timuthy, line 910: Should Druid reclaim the identity of the member `lease` when we delete the pod? A restart usually takes longer and otherwise during the next health check, we wouldn't know if the restart already finished.

docs/proposals/multi-node/README.md
```diff
@@ -982,6 +1019,7 @@ The superfluous members in the ETCD cluster will be cleaned up by the [leading `
 #### Recommended Action
 
 Remove the superfluous `m - n` member entries from `Etcd` status (numbered `n`, `n+1` ... `m`).
+Remove the superflous `m - n` [member `leases`](#member-leases) if they exist.
```
timuthy, line 1022:
```suggestion
Remove the superfluous `m - n` [member `leases`](#member-leases) if they exist.
```

timuthy (changes requested):
Looks good, thanks for writing the latest decisions down. I only have a few questions/suggestions.

docs/proposals/multi-node/README.md
amshuman-kr: In the proposal, I didn't make any assumptions about the order in which `members` and `conditions` are updated and left it for implementation. Hence, retained this section. But you are right, if `conditions` are computed after `members` are computed then this case would be redundant (also the the _Majority members are NotReady but Ready condition is stale_ case)?

Is it OK to retain these two cases just to allow for different implementations?

docs/proposals/multi-node/README.md
amshuman-kr: Same as [above](https://github.com/gardener/etcd-druid/pull/207#discussion_r676597905).

docs/proposals/multi-node/README.md
amshuman-kr: By _reclaiming the identity_ if you meant recreating the `lease` object (or maybe resetting `spec.holderIdentity`), I was thinking about this while writing the change but couldn't conclude either way 😅 

When the pod is restarted, member ID (and hence, `spec.holderIdentity`)  would change only if the data and/or WAL files are lost. Otherwise, the member would simply adopt the older identity (probably, most frequent case while using PVCs for the member pods). We could think of resetting `spec.holderIdentity` before restarting the pod and adding an additional check in `etcd-backup-restore` to renew the lease only if `spec.holderIdentity` is blank or the same as the ETCD member id. This might strengthen some edge-cases where the previous member pod is still somehow running. But I thought `StatefulSet` controller has this already under controller and so, was not sure if we need the additional check. WDYT?

amshuman-kr:
@timuthy Thanks for the review! I have addressed your comments. Can you please check if this fine or if more changes are required?

docs/proposals/multi-node/README.md
timuthy: > By reclaiming the identity if you meant recreating the lease object (or maybe resetting spec.holderIdentity)

Yes, that's what I meant. But the reason for this was not that a previous pod could still run, rather the following:
1. Lease is expired & not renewed
2. Status turns `UNKNOWN`
3. Status turns `FALSE` after `etcd-member-notready-threshold`
4. Custodian deletes the pod
5. Lease from (1.) is still expired
6. Custodian checks lease again **before** new pod renews the lease
7. Custodian deletes the pod again

docs/proposals/multi-node/README.md
```diff
@@ -732,7 +791,7 @@ Mark the status condition type `AllMembersReady` to `false`.
     - Ready: `r` where `r > n/2`
     - Members `NotReady` for long enough to be evicted, i.e. `lastTransitionTime > notReadyGracePeriod`: `nr` where `0 < nr < n/2`
     - Members with readiness status `Unknown` long enough to be considered `NotReady`, i.e. `lastTransitionTime > unknownGracePeriod`: `u` where `0 < u < n/2`
-    - Members with heartbeat stale enough to be considered as of `Unknown` readiness status, i.e. `lastHeartbeatTime > heartbeatGracePeriod`: `h` where `0 < h < n/2`
+    - Members with expired `lease`: N/A
```
timuthy, line 794: Hm, why not as before?
```
`h` where `0 < h < n/2` 
```

docs/proposals/multi-node/README.md
timuthy: Fair enough, let's keep it. Thanks for explaining.

docs/proposals/multi-node/README.md
amshuman-kr: The `lease` and member entries being out of sync was already addressed in [2. Member status is out of sync with their lease](https://github.com/gardener/etcd-druid/pull/207/files#diff-7918fb156cb729ac55570b3318ff8e792f008e5fe087f0a7c7a36db71374411bR694-R722). So, I thought adding more of the same here might be repetitive.

docs/proposals/multi-node/README.md
amshuman-kr: > Custodian checks lease again before new pod renews the lease
> Custodian deletes the pod again

I thought this could be avoided by checking if `creationTimestamp` of the pod is after the `lastTransitionTime` for its member (or `renewTime` of its `lease`). Or do you see the `Get` call on pod a unnecessary?

timuthy: I'm fine with both ways, just wanted to clarify how we're going to tackle this issue 👍

docs/proposals/multi-node/README.md
timuthy: Okay thanks. Thought, I missed something because of that change.

timuthy (approved):
/lgtm
Thanks for addressing the feedback

docs/proposals/multi-node/README.md
```diff
@@ -598,6 +609,24 @@ For example, if an ETCD cluster is scaled up from `3` to `5` and the new members
...
+This enhancement should keep `etcd-backup-restore` backward compatible.
+But it should be possible to use `etcd-backup-restore` Kubernetes-unaware as before this proposal. This is possible either by auto-detecting the existence of kubeconfig or by an explicit command-line flag (such as `--enable-etcd-lease-renewal` which can be defaulted to `false` for backward compatibility).
+
+A `member` entry in the `Etcd` resource [`status`](#status) would be marked as `Ready` (with `reason: LeaseSucceeded`) if the corresponding `pod` is ready  and the corresponding `lease` has not yet expired.
+The `member` entry would be marked as `NotReady` if the corresponding `pod` is not ready (with reason `PodNotReady`) or as `Unknown` if the corresponding `lease` has expired (with `reason: LeaseExpired`).
+
+While renewing the lease, the `etcd-backup-restore` sidecars also maintain the ETCD member id in the `spec.holderIdentity` field of the corresponding `lease` object since this information is only available to the `ETCD` member processes and the `etcd-backup-restore` sidecars. When the `lease` objects are created by `etcd-druid`, the `spec.holderIdentity` field would be empty.
```
shreyas-s-rao, line 626: I was under the assumption that the `role` field makes more sense to druid than the `id` field. For a human operator of the cluster, seeing the etcd members' roles in the `Etcd` status (like `Leader` and `Follower`) would make more sense from a debugging perspective, rather than seeing the etcd member ID. WDYT?

shreyas-s-rao (commented):
@amshuman-kr Thanks for updating the proposal with the changes from our discussion. I just had a doubt regarding the utility of `id` field over `role` field. Please address this. Thanks.

docs/proposals/multi-node/README.md
amshuman-kr: Populating member `id` in `spec.holderIdentity` made natural sense. But populating role (`Leader` or `Follower`) didn't make sense because it is not identity as such.

I did think about populating both in `spec.holderIdentity`. For example,
```
spec:
  holderIdentity: 272e204152:Leader
```
If it is not a leader then the value might be either just the `id` or `id:` or `id:Follower`.

What do you think of that? Then we can support both `id` and `role`. To be honest, it looks kludgy 😅

shreyas-s-rao: Yes, the whole concept of using `holderIdentity` for this seems a but weird. But then again, we would need _some_ way for backup sidecars to give this information to druid so that it may populate the status with information enough for a human operator to debug issues quickly. So in terms of utility, I'm fine with passing just the `Role` info (Leader or Follower), since the etcd member ID might not really be of much relevance at that point of time, especially since we're using the member name (ie, pod name) as the real "identifier" for the member etcds.

amshuman-kr: 👍  I don't like putting only `Leader` or `Follower` in `holderIdentity`. I will modify the proposal to include both member `id` and `role` in the `holderIdentity` (as in `272e204152:Leader`).

amshuman-kr: Done in https://github.com/gardener/etcd-druid/pull/207/commits/cfe150ff0a2b76825b80893f84c153abba85b444.

amshuman-kr:
@shreyas-s-rao Thanks for the review. I have changed the proposal as [discussed](https://github.com/gardener/etcd-druid/pull/207#discussion_r678155759). Now both `id` and `role` fields are present in `status.members` and populated via `:` separated value (like `272e204152:Leader`) in `spec.holderIdentity` of  the corresponding `leases`. Can you PTAL?

shreyas-s-rao (approved):
@amshuman-kr Thanks for addressing my concern. LGTM

PR #214, opened 2021-08-03
Use Leases for etcd member health checks

### LINKED ISSUE #206: ☂️ [Feature] Add EtcdMember resource

### LINKED ISSUE #219: [Feature] Disable cache for lease

#214 introduces listing of `leases` which would be automatically cached by the controller-runtime. It might be good to disable caching for `leases` until #215 has been addressed.

While `leases` are small objects and may not cause too much increase in memory usage but they get updated very frequently and hence might increase network traffic by quite a bit.

timuthy:

/area scalability
/kind api-change
/kind enhancement

This PR changes Druid's health checks in a way that it considers `Lease` resources in order to find out the readiness of a etcd cluster member.

Fixes #206 
Fixes #219 

Please have a look at the separate commits for an easier review process.
The creation/deletion of `Lease` objects through Druid will follow later in a separate PR.
/cc @amshuman-kr @shreyas-s-rao 

**Release note**:

```other operator
Druid now fetches `Lease` resources in order to derive the readiness state of an etcd cluster member. This serves as a preparation for the etcd multi-node feature.
```


pkg/health/condition/check_ready.go
```diff
@@ -31,16 +31,27 @@ func (r *readyCheck) Check(status druidv1alpha1.EtcdStatus) Result {
...
 		size         = utils.Max(int(*status.ClusterSize), len(status.Members))
 		quorum       = size/2 + 1
 		readyMembers = 0
 	)
 
 	for _, member := range status.Members {
-		if member.Status == druidv1alpha1.EtcdMemeberStatusReady {
-			readyMembers++
+		if member.Status == druidv1alpha1.EtcdMemeberStatusNotReady {
+			continue
```
amshuman-kr, line 52: This change would mean unknown members will still be counted un `readyMembers` and hence into readiness check. This is a deviation from the previous implementation but seems to be correct. 👍

pkg/health/etcdmember/check_ready.go
```diff
@@ -41,48 +48,96 @@ func (r *readyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) []Resul
...
-		if (err == nil && !ready) || apierrors.IsNotFound(err) {
-			results = append(results, &result{
-				id:     member.ID,
-				status: druidv1alpha1.EtcdMemeberStatusNotReady,
-				reason: "ContainersNotReady",
-			})
+		// Check if member state must be considered as not ready
+		if renew.Add(time.Duration(*leaseDurationSeconds) * time.Second).Add(r.memberConfig.EtcdMemberNotReadyThreshold).Before(checkTime) {
+			res.status = druidv1alpha1.EtcdMemeberStatusNotReady
+			res.reason = "LeaseExpired"
```
amshuman-kr, line 84: Shouldn't this be `UnknownGracePeriodExceeded`?

pkg/health/etcdmember/check_ready.go
```diff
@@ -41,48 +48,96 @@ func (r *readyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) []Resul
...
+				continue
+			}
+
+			res.status = druidv1alpha1.EtcdMemeberStatusUnknown
+			res.reason = "LeaseExpired"
+			results = append(results, res)
+			continue
+		}
+
+		res.status = druidv1alpha1.EtcdMemeberStatusReady
```
amshuman-kr, line 106: Typo in the constants `EtcdMemeber...` should be `EtcdMember...`. Sorry, I didn't catch it the first time around.

pkg/health/etcdmember/builder_test.go
```diff
@@ -98,54 +95,90 @@ var _ = Describe("Builder", func() {
 				conditions := builder.Build()
 
 				Expect(conditions).To(ConsistOf(
-					oldMembers["1"],
-					oldMembers["2"],
```
amshuman-kr, line 102: Is it intentional that the unmodified members should be not present in the build result?

amshuman-kr (commented):
@timuthy Thanks for the changes to include handling of `leases` in status! Looks good already. Just a few minor questions below.

Also, one additional question. Is there any explicit configuration required during startup to make sure that the leases are cached? Or is that postponed to #215 ?

pkg/health/condition/check_ready.go
timuthy: I noticed it while testing when UNKNOWN members caused the `Ready` condition to become `false` very early. Tolerating UNKNOWN might make sense as I see it today but we can still change it in the future if it turns out to be too tolerant.

pkg/health/etcdmember/builder_test.go
timuthy: The members that are part of the result slice are taken into account, see [here](https://github.com/gardener/etcd-druid/blob/550a0b6a0f132419190043970fd34ae1a3df53ba/pkg/health/etcdmember/builder_test.go#L86). The rational behind this is that there will be one result per `Lease`. So when the cluster is scaled down from 3 -> 1, druid removes two member `Leases` and the health check will overwrite the members array with now only having one member because there is only one `Lease` left.

amshuman-kr: Understood. Then it is intentional that the merge case is now no longer applicable, right? I.e. the builder always return as many members in the result slice as there are `leases`.

timuthy:
Thanks for the fast feedback @amshuman-kr.

> Also, one additional question. Is there any explicit configuration required during startup to make sure that the leases are cached? Or is that postponed to #215 ?

No, there is no additional configuration required. Leases will be cached automatically. #215 is an optimization, so that we don't cache all `Lease` objects in the cluster. We can also make caching available as a command line flag, so that one can turn it off in case there are too many `Lease` objects in a cluster. WDYT?

pkg/health/etcdmember/builder_test.go
timuthy: Yes, it only needs the old members in order to set the `LastTransitionTime`.

amshuman-kr:
>No, there is no additional configuration required. Leases will be cached automatically. #215 is an optimization, so that we don't cache all Lease objects in the cluster. We can also make caching available as a command line flag, so that one can turn it off in case there are too many Lease objects in a cluster. WDYT?

I now remember that controller-runtime now automatically caches if there one list call. Sounds good 👍

amshuman-kr (approved):
Thanks for the changes @timuthy! LGTM

timuthy:
@amshuman-kr as discussed, I added a command line option to disable the lease cache.

amshuman-kr (approved):
Thanks for disabling the `lease` cache! LGTM

amshuman-kr:
/invite @shreyas-s-rao @abdasgupta @ishan16696 @aaronfern 

The changes look good from my side. Shall we merge this?

config/crd/bases/druid.gardener.cloud_etcds.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.5.0
+    controller-gen.kubebuilder.io/version: v0.4.1
```
aaronfern, line 7: Just some clarification needed.
Is this an intended change?

amshuman-kr: https://github.com/gardener/etcd-druid/pull/193#discussion_r685698113

abdasgupta (approved):
LGTM

PR #222, opened 2021-08-23
Remove finalizers after waiting for statefulset.

amshuman-kr:

/area control-plane cost testing
/kind technical-debt

Remove finalizers from `Etcd` resources after waiting for statefulset.

Fixes #

Using the propagation policy `Foreground` failed some tests due to deletion timing out. So, I used `Background` and [returned error](https://github.com/gardener/etcd-druid/pull/222/files#diff-3480d2ce46715d0b85c7f54bb4ae498a20da29ad632c61b1127c32254c0ad73cR1305) to [force re-queueing](https://github.com/gardener/etcd-druid/blob/b8f86e9147e79ef2f15dde1a16468d81258621d6/controllers/etcd_controller.go#L286-L288) to a check to see if the statefulset is gone.

**Release note**:

```other operator
Remove finalizers from `Etcd` resources after waiting for statefulset.
```

### DIFF
--- a/controllers/etcd_controller.go
+++ b/controllers/etcd_controller.go
@@ -281,8 +281,8 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
 	logger := r.logger.WithValues("etcd", kutil.Key(etcd.Namespace, etcd.Name).String(), "operation", "delete")
 	logger.Info("Starting operation")
 
-	if err := r.removeDependantStatefulset(ctx, logger, etcd); err != nil {
-		if err := r.updateEtcdErrorStatus(ctx, deleteOp, etcd, nil, err); err != nil {
+	if waitForStatefulSetCleanup, err := r.removeDependantStatefulset(ctx, logger, etcd); err != nil {
+		if err = r.updateEtcdErrorStatus(ctx, deleteOp, etcd, nil, err); err != nil {
 			return ctrl.Result{
 				Requeue: true,
 			}, err
@@ -290,6 +290,10 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
 		return ctrl.Result{
 			Requeue: true,
 		}, err
+	} else if waitForStatefulSetCleanup {
+		return ctrl.Result{
+			RequeueAfter: 30 * time.Second,
+		}, nil
 	}
 
 	if err := r.removeFinalizersToDependantSecrets(ctx, logger, etcd); err != nil {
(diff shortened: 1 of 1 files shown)


controllers/etcd_controller.go
```diff
@@ -1293,15 +1293,20 @@ func (r *EtcdReconciler) removeDependantStatefulset(ctx context.Context, logger
...
 		if canDeleteStatefulset(&sts, etcd) {
-			logger.Info("Deleting statefulset", "statefulset", kutil.Key(sts.GetNamespace(), sts.GetName()).String())
-			if err := r.Delete(ctx, &sts); err != nil {
-				return err
+			var key = kutil.Key(sts.GetNamespace(), sts.GetName()).String()
+			logger.Info("Deleting statefulset", "statefulset", key)
+			if err := r.Delete(ctx, &sts, client.PropagationPolicy(metav1.DeletePropagationBackground)); err != nil {
+				errs = append(errs, fmt.Errorf("Error deleting statefulset %q: %s", key, err))
+			} else {
+				errs = append(errs, fmt.Errorf("Waiting for statefulset %q to be deleted", key))
```
timuthy, line 1305: I think it's safe to directly return here w/o error aggregation. There should usually only be one statefulset, right?

It's also a little bit unfortunate that after this error, we will re-queue very quickly and often due to the nature of the rate limiting queue, i.e. a lot of errors will appear while the pod is not gone.

amshuman-kr: > I think it's safe to directly return here w/o error aggregation. There should usually only be one statefulset, right?

Sure. I can simplify that.

> It's also a little bit unfortunate that after this error, we will re-queue very quickly and often due to the nature of the rate limiting queue, i.e. a lot of errors will appear while the pod is not gone.

I thought about it. One way to solve this would be to return a typed error if and check in the caller and use `RequeueAfter` for the specific type. WDYT?

amshuman-kr: > I thought about it. One way to solve this would be to return a typed error if and check in the caller and use RequeueAfter for the specific type. Wasn't sure if it was worth the complexity, WDYT?

I alternatively, I can make `removeDependantStatefulset` return an additional boolean flag to indicate if deletion went through and requeue to check if the statefulset is gone should be after some time. This change might be simpler. WDYT?

shreyas-s-rao: @amshuman-kr I feel the second option of using an additional boolean return value seems simpler for this problem, since it doesn't involve introducing a new error type just for this one occurrence of the error.

amshuman-kr: Done in https://github.com/gardener/etcd-druid/pull/222/commits/5f8e801bc2f1e5dd670cc1edf968225bb7327293

amshuman-kr:
I have implemented the change to `RequeueAfter` to wait for statefulset cleanup in https://github.com/gardener/etcd-druid/pull/222/commits/5f8e801bc2f1e5dd670cc1edf968225bb7327293. Can you PTAL?

cc @timuthy @shreyas-s-rao

abdasgupta (approved):
LGTM

timuthy (approved):
Thanks @amshuman-kr.
/lgtm

shreyas-s-rao (approved):
@amshuman-kr Thanks for making the changes. LGTM

PR #227, opened 2021-08-30
Updated label in cronjob job spec to allow it to be uniquely identifi…

aaronfern:

/area backup
/kind bug

Updated label in cronjob job template spec so that it can be identified

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/etcd/templates/etcd-compaction-cronjob.yaml
+++ b/charts/etcd/templates/etcd-compaction-cronjob.yaml
@@ -23,7 +23,7 @@ metadata:
     name: {{ .Values.name }}
     uid: {{ .Values.uid }}
 spec:
-  schedule: {{ .Values.backup.backupCompactionSchedule }} 
+  schedule: {{ .Values.backup.backupCompactionSchedule | quote }} 
   concurrencyPolicy: Forbid
 {{- if eq (int $.Values.statefulsetReplicas) 0 }}
   suspend: true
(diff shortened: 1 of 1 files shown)


charts/etcd/templates/etcd-compaction-cronjob.yaml
```diff
@@ -23,7 +23,7 @@ metadata:
     name: {{ .Values.name }}
     uid: {{ .Values.uid }}
 spec:
-  schedule: {{ .Values.backup.backupCompactionSchedule }} 
+  schedule: "{{ .Values.backup.backupCompactionSchedule }}"
```
amshuman-kr, line 26:
```suggestion
  schedule: {{ .Values.backup.backupCompactionSchedule | quote }} 
```

charts/etcd/templates/etcd-compaction-cronjob.yaml
```diff
@@ -41,7 +41,7 @@ spec:
 {{ toYaml .Values.annotations | indent 12 }}
 {{- end }}
           labels:
-            name: etcd
+            name: etcd-compaction-cronjob
```
amshuman-kr, line 44:
```suggestion
            name: etcd-backup-compaction
```
Let's use the `name` label to indicate what process is actually running in the pod. Technically, it is not a cronjob, just a pod. Pedantic, I know, but still 😅

amshuman-kr (changes requested):
Thanks for the quick change @aaronfern! Looks good. Just a couple of pedantic suggestions 🙂

amshuman-kr (approved):
LGTM!

timuthy:
@aaronfern can this be closed as cronjobs won't be used anymore?

aaronfern:
Yes, this isn't needed anymore. Closing 
/close

PR #230, opened 2021-09-13
Add Basic ETCD validation

### LINKED ISSUE #213: [Feature] Make Etcd CRD's `spec.backup.store` section immutable
Make Etcd CRD's `spec.backup.store` section immutable.

Make Etcd CRD's `spec.backup.store` section immutable so that the storage container location isn't allowed to change mid-usage of an etcd, due to potential mismatch of snapshotting and restoration locations, allowing restorations to happen from a different etcd's backup and rendering the shoot cluster unusable. Refer https://github.com/gardener/gardener/pull/4454 for a fix already made on Gardener, although we still want druid to be resilient to potential undesirable changes to the Etcd resource.

Since CRD immutability is yet to be support (refer https://github.com/kubernetes/kubernetes/issues/65973), it might make more sense to use something like a validating webhook on the Etcd resource updates.

/cc @amshuman-kr

### LINKED ISSUE #217: [Feature] Add validation for etcd resources

Please add validation code for [etcd resources](https://github.com/gardener/etcd-druid/blob/master/api/v1alpha1/etcd_types.go), similarly to the validation code that already exists for [other Gardener extension resources](https://github.com/gardener/gardener/tree/master/pkg/apis/extensions/validation), even though this is technically still dead code.

We are currently working on a new validating webhook in `seed-admission-controller` for such extension resources, see https://github.com/gardener/gardener/issues/4293, I think we could include the validation of `etcd` resources there as well. Alternatively, `etcd-druid` could introduce its own validating webhook if for whatever reason the above option is not good enough.

We recently had a rather severe issue that could have been prevented if we had such validation in place, see https://github.com/gardener/gardener-extension-provider-azure/issues/328#issuecomment-879824536. In this particular case, `gardenlet` was generating an `etcd` resource with a `spec.backup.store.prefix` set to `--` due to a data race. With validation in place, we could have detected `--` as an invalid `spec.backup.store.prefix` and prevented the reconciliation from continuing. This particular issue is already fixed in `gardenlet` (see https://github.com/gardener/gardener/pull/4459 and https://github.com/gardener/gardener/pull/4454), but similar issues may occur in the future.

vanjiii:

/area control-plane
/kind enhancement

It adds a basic etcd validation. 
- onCreate it validates for correct spec.backup.store.prefix name (this means whether the prefix consists of the etcd's name and etcd's namespace);
- onUpdate it validates for immutable spec.backup.store.

Fixes #217 
Fixes #213 

Note that the validation tests are located in the submodule `github.com/gardener/etcd-driud/api` and **not** in the main module (where the rest lives - `github.com/gardener/etcd-druid/api_tests`). 

/cc @stoyanr 
/cc @amshuman-kr


api/validation/etcd.go
```diff
@@ -0,0 +1,53 @@
...
+func ValidateEtcdUpdate(new, old *v1alpha1.Etcd) field.ErrorList {
+	return validateEtcdSpecUpdate(&new.Spec, &old.Spec, field.NewPath("spec"))
+}
+
+func validateEtcdSpec(etcd *v1alpha1.Etcd, path *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	prefix := etcd.Spec.Backup.Store.Prefix
+	if !validatePrefixName(prefix, etcd.Name, etcd.Namespace) {
+		allErrs = append(allErrs, field.Invalid(path.Child("backup.store.prefix"), prefix, "field is required"))
```
stoyanr, line 41: The error message "field is required" is wrong, and it can't be right since at this point you've lost the information what is actually wrong. Instead, you should write this as:
```suggestion
	allErrs = append(allErrs, validateBackupStorePrefix(prefix, etcd.Name, etcd.Namespace, path.Child("backup.store.prefix"))...)
```
And take care to properly report all validation errors via e.g. `field.Invalid` (there might be more than one) in `validateBackupStorePrefix`. This is a general pattern to be followed in all validation code.

api/validation/etcd.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return allErrs
+}
+
+func validatePrefixName(prefix, name, ns string) bool {
+	return strings.Contains(prefix, ns) && strings.Contains(prefix, name)
+}
+
+func validateEtcdSpecUpdate(new, old *v1alpha1.EtcdSpec, path *field.Path) field.ErrorList {
+	return apivalidation.ValidateImmutableField(new.Backup.Store, old.Backup.Store, path)
```
stoyanr, line 52:
```suggestion
	return apivalidation.ValidateImmutableField(new.Backup.Store, old.Backup.Store, path.Child("backup.store))
```

api/validation/etcd.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/gardener/etcd-druid/api/v1alpha1"
+
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateEtcd validates Etcd object.
+func ValidateEtcd(etcd *v1alpha1.Etcd) field.ErrorList {
+	return validateEtcdSpec(etcd, field.NewPath("spec"))
```
stoyanr, line 28: Validating resources should always validate the object meta, see e.g. https://github.com/gardener/gardener/blob/0212d5b5d90674b693e6110420b189a308ecc288/pkg/apis/extensions/validation/backupentry.go#L27-L31.

```suggestion
	allErrs := field.ErrorList{}
	allErrs = append(allErrs, apivalidation.ValidateObjectMeta(&etcd.ObjectMeta, false, apivalidation.NameIsDNSSubdomain, field.NewPath("metadata"))...)
	allErrs = append(allErrs, validateEtcdSpec(etcd, field.NewPath("spec"))...)
	return allErrs
```

api/validation/etcd.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// ValidateEtcdUpdate validates a Etcd object before an update.
+func ValidateEtcdUpdate(new, old *v1alpha1.Etcd) field.ErrorList {
+	return validateEtcdSpecUpdate(&new.Spec, &old.Spec, field.NewPath("spec"))
```
stoyanr, line 33: Validating resource updates should always validate the object meta, and include a normal validation of the new resource, see e.g. https://github.com/gardener/gardener/blob/0212d5b5d90674b693e6110420b189a308ecc288/pkg/apis/extensions/validation/backupentry.go#L36-L42

stoyanr (changes requested):
Please follow the general patterns of writing validation code, even though this validation is not yet intended to cover the entire `Etcd` resource properly.

stoyanr (approved):
Thanks for addressing my feedback!
/lgtm

stoyanr:
Since I am not a member of `etcd-druid-maintainers`, @shreyas-s-rao, @timuthy Could somebody of you take a look?

api/validation/etcd.go
```diff
@@ -0,0 +1,82 @@
...
+// ValidateEtcdSpecUpdate validates the specification of an Etcd before an update.
+func ValidateEtcdSpecUpdate(new, old *v1alpha1.EtcdSpec, deletionTimestampSet bool, path *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	if deletionTimestampSet && !apiequality.Semantic.DeepEqual(new, old) {
+		allErrs = append(allErrs, apivalidation.ValidateImmutableField(new, old, path)...)
+		return allErrs
+	}
+
+	allErrs = append(allErrs, apivalidation.ValidateImmutableField(new.Backup.Store, old.Backup.Store, path.Child("backup.store"))...)
```
timuthy, line 69: Generally, I don't see a reason to make the entire `.Backup.Store` structure immutable. Especially changing `.Backup.Store.SecretRef` should be allowed to roll credentials.

timuthy (changes requested):
Thanks for the PR. Just one comment from my side.

api/validation/etcd.go
```diff
@@ -66,8 +66,6 @@ func ValidateEtcdSpecUpdate(new, old *v1alpha1.EtcdSpec, deletionTimestampSet bo
 		return allErrs
 	}
 
-	allErrs = append(allErrs, apivalidation.ValidateImmutableField(new.Backup.Store, old.Backup.Store, path.Child("backup.store"))...)
-
```
stoyanr: This was clearly not what @timebertt and myself had in mind. We still need the prefix (`backup.store.prefix`) to be immutable, as was requested in the original issue - just not the entire `backup.store`.

stoyanr (changes requested):

stoyanr (approved):
/lgtm

timuthy (approved):
/lgtm

PR #233, opened 2021-09-20
Added role, rolebinding, serviceaccount, and environment variables.

aaronfern:
Added role, rolebinding, serviceaccount, and environment variables to allow etcd-backup-restore sidecar to become kubernetes aware

/area backup
/area disaster-recovery
/area high-availability
/kind enhancement

As part of making the backup-restore sidecar kubernetes aware,  this PR does the following
- Allows druid to now create `serviceaccount`, `role`, and `rolebinding` and associates them to the etcd pods, to authenticate it to the API server. 
- Passes the pod name and current namespace to the `backup-restore` pod as environment variables
- Removes logic for cronjob as it is to be replaced with trigger based compaction jobs instead

Partially fixes gardener/etcd-backup-restore#333

**Release note**:

```feature user
`druid` will now also reconcile a `serviceaccount`, a `role`, and a `rolebinding` as part of it's `etcd` reconcile flow and associate it with the etcd pod
```


charts/etcd/templates/etcd-role.yaml
```diff
@@ -0,0 +1,36 @@
...
+metadata:
+  name: {{ .Values.roleName }}
+  namespace: {{ .Release.Namespace }}
+  ownerReferences:
+  - apiVersion: druid.gardener.cloud/v1alpha1
+    blockOwnerDeletion: true
+    controller: true
+    kind: Etcd
+    name: {{ .Values.name }}
+    uid: {{ .Values.uid }}
```
timuthy, line 12: I suggest to not add a `ownerReference` here because depending on the propagation policy owned resources could be deleted too early.
It's not guaranteed that the deletion of the `etcd` resource happens in a way that this role is only removed when the finalizer is patched out from the `etcd` resource.

charts/etcd/templates/etcd-role.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - list
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - list
+  - get
+  - update
+  - delete
```
timuthy, line 35: `delete` should not be required since only the Druid deletes `Lease`s. See  https://github.com/gardener/etcd-druid/tree/master/docs/proposals/multi-node#member-leases:

> The lease objects are created and deleted only by etcd-druid but are continually renewed within the leaseDurationSeconds by the individual etcd-backup-restore sidecars (corresponding to their members) if the the corresponding ETCD member is ready and is part of the ETCD cluster.

charts/etcd/templates/etcd-rolebinding.yaml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: {{ .Values.roleBindingName }}
+  namespace: {{ .Release.Namespace }}
+  ownerReferences:
+  - apiVersion: druid.gardener.cloud/v1alpha1
+    blockOwnerDeletion: true
+    controller: true
+    kind: Etcd
+    name: {{ .Values.name }}
+    uid: {{ .Values.uid }}
```
timuthy, line 12: See previous comment about owner ref.

charts/etcd/templates/etcd-serviceaccount.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: {{ .Values.serviceAccountName }}
+  namespace: {{ .Release.Namespace }}
+  ownerReferences:
+  - apiVersion: druid.gardener.cloud/v1alpha1
+    blockOwnerDeletion: true
+    controller: true
+    kind: Etcd
+    name: {{ .Values.name }}
+    uid: {{ .Values.uid }}
```
timuthy, line 12: See previous comment about owner ref.

controllers/controller_ref_manager.go
```diff
@@ -288,6 +289,97 @@ func (m *EtcdDruidRefManager) ClaimCronJob(ctx context.Context, cronJob *batchv1
 	return claimed, utilerrors.NewAggregate(errlist)
 }
 
+func (m *EtcdDruidRefManager) ClaimServiceAccount(ctx context.Context, sa *corev1.ServiceAccount, filters ...func(*corev1.ServiceAccount) bool) (*corev1.ServiceAccount, error) {
```
timuthy, line 292: Why do we need to claim the `ServiceAccount` here? Isn't it sufficient to just create it via the chart? Same goes for `Role` and `RoleBinding`.

timuthy (changes requested):
Thanks for the PR. I left a few comments and questions.

controllers/controller_ref_manager.go
shreyas-s-rao: @timuthy you're right. @aaronfern I think this requires some history of why the claim logic was even necessary in the first place. Initially (3 years ago), there was only the etcd statefulset (that contains the etcd container and backup-restore sidecar) which was directly deployed by gardener. Later, we realised the need for a controller to perform tasks that a sidecar can't, and hence built etcd-druid. Druid could create new etcds with no issues, but from a backward-compatibility perspective, landscapes were already running multiple etcd statefulsets, so druid now had to "claim" these.

For the new resources you're introducing here (serviceaccount, role, rolebinding for the backup sidecar), you need not claim them, as you're sure that these resources are created fresh by druid.

shreyas-s-rao (changes requested):
@aaronfern Thanks for the PR. This PR will now need to be modified to remove cronjob, after out-of-band discussions to introduce trigger-based compaction jobs. Also, please remove the claim logic for the three resources you've introduced, as this is not a necessity, as mentioned [here](https://github.com/gardener/etcd-druid/pull/233#discussion_r718610791). Thanks.

charts/etcd/templates/etcd-role.yaml
shreyas-s-rao: > It's not guaranteed that the deletion of the etcd resource happens in a way that this role is only removed when the finalizer is patched out from the etcd resource

@timuthy I didn't exactly get this. [Etcd deletion flow](https://github.com/gardener/etcd-druid/blob/0828abe1338b0654a33cb5b118a2d5eeeb9bcda5/controllers/etcd_controller.go#L280) first deletes the etcd statefulset and then removes the finalizers on the dependent secrets (backup secret and TLS secrets) and finally removes finalizer from the etcd resource. That will then ensure the deletion of the etcd and subsequently the etcd bootstrap configmap and service (as well as serviceaccount, role, rolebinding from this PR).

timuthy: >  That will then ensure the deletion of the etcd and subsequently the etcd bootstrap configmap and service (as well as serviceaccount, role, rolebinding from this PR).

Exactly this order you described depends on the `propagationPolicy`. Try to execute the following and you will see that all owned objects are gone immediately even though the Druid did not remove the finalizer from the `etcd` resource.

```
kubectl delete etcd etcd-test --cascade="foreground"
```

charts/etcd/templates/etcd-role.yaml
aaronfern: Done

charts/etcd/templates/etcd-role.yaml
timuthy: In an out-of-band discussion we agreed with continuing with the `ownerReference` approach since we already took this for `ConfigMap` and `Service` resources and deleting the resources too early be accident is not very critical.

controllers/etcd_controller.go
```diff
@@ -1174,8 +1168,10 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
 		"statefulsetReplicas":     statefulsetReplicas,
 		"serviceName":             fmt.Sprintf("%s-client", etcd.Name),
 		"configMapName":           fmt.Sprintf("etcd-bootstrap-%s", string(etcd.UID[:6])),
-		"cronJobName":             getCronJobName(etcd),
 		"volumeClaimTemplateName": volumeClaimTemplateName,
+		"serviceAccountName":      fmt.Sprintf("%s-serviceaccount", etcd.Name),
+		"roleName":                fmt.Sprintf("%s-role", etcd.Name),
+		"roleBindingName":         fmt.Sprintf("%s-rolebinding", etcd.Name),
```
timuthy, line 1174: Should we include something like `-br-` in the name, so that the relation to backup-store is clear?

controllers/etcd_controller.go
```diff
@@ -791,142 +797,129 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
+	serviceAccountPath := getChartPathForServiceAccount()
+	chartPath := getChartPath()
+	renderedChart, err := r.chartApplier.Render(chartPath, etcd.Name, etcd.Namespace, values)
+	if err != nil {
+		return err
 	}
-
-	if !backupCompactionScheduleFound {
-		if err == nil {
-			err = r.Delete(ctx, &cronJob, client.PropagationPolicy(metav1.DeletePropagationForeground))
```
timuthy, line 810: I don't think that we can simply remove the deletion of cron jobs because people might have already enabled that feature, so these objects will remain in the cluster after switching to `Job` in https://github.com/gardener/etcd-druid/pull/235.
In general I have to admit, I don't like that this PR tries to include both: Adding new resource for Backup-Restore and removing the `CronJob` behavior. Why?
- It makes it depended on https://github.com/gardener/etcd-druid/pull/235.
- It breaks the main branch after merging.
- It makes reviews harder to understand.
- It makes the change history complex.
Not sure why that decision was made https://github.com/gardener/etcd-druid/pull/233#pullrequestreview-766861481 (/cc @shreyas-s-rao).

api/v1alpha1/etcd_types.go
```diff
@@ -141,9 +141,6 @@ type BackupSpec struct {
 	// EnableProfiling defines if profiling should be enabled for the etcd-backup-restore-sidecar
 	// +optional
 	EnableProfiling *bool `json:"enableProfiling,omitempty"`
-	// BackupCompactionSchedule defines the cron standard for compacting the snapstore
-	// +optional
-	BackupCompactionSchedule *string `json:"backupCompactionSchedule,omitempty"`
```
timuthy, line 146: We should have a breaking change release note for this.

config/crd/bases/druid.gardener.cloud_etcds.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.5.0
```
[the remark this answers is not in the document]
aaronfern: Thanks for catching this.
Updated

controllers/etcd_controller.go
abdasgupta: Yes, I already took care of cron job related codes in my PR. Would add a release note. Thanks.

shreyas-s-rao (dismissed):
@aaronfern Please remove the cronjob-removal related changes from your PR, since @abdasgupta will be including those changes in https://github.com/gardener/etcd-druid/pull/235 with a breaking change release note. Let's keep this PR only for adding RBAC for backup sidecar to renew snapshot and member leases.

Please also add the relevant tests for sa, role, rb, like ensuring that the sa has read/write permissions on leases but not on other objects, and also ensuring that if the role/rb are not deployed, then sa cannot access leases as well.

timuthy (approved):
/lgtm

PR #234, opened 2021-09-21
Adding functionality to start copy job

kris94:

/area control-plane-migration
/kind enhancement

This PR adds the following to the `etcd-druid` functionality:
1. A new Kubernetes object `EtcdCopyBackupsTask` is added. Its purpose is to hold the necessary information that is required for starting a ETCD copy backups operation.
2. A new controller is added to take care of the `EtcdCopyBackupsTask` object.

Based on the `EtcdCopyBackupsTask` configuration the new controller will start a `Job` that will call `etcdbctrl copy` command, that will copy all backups according to the configuration from the source bucket to the destination bucket.

Fixes #

Ref PR: [#369](https://github.com/gardener/etcd-backup-restore/pull/369/files)
**Release note**:

```other operator
NONE
```


api/v1alpha1/etcd_types.go
```diff
@@ -69,6 +69,19 @@ type CompressionPolicy string
...
 type CompactionMode string
 
+type CopySpec struct {
+	// SourceStore defines the specification of the source object store provider for copying backups.
+	// +optional
+	SourceStore *StoreSpec `json:"sourceStore,omitempty"`
+	// MaxAge defines the maximum age in days that one snapshot must be in order to be copied.
+	// By default all snapshots will be copied.
+	// +optional
+	MaxAge *uint32 `json:"maxAge,omitempty"`
```
stoyanr, line 79: Let's make this consistent with the naming in `etcd-backup-restore`
```suggestion
	MaxSnapshotAge *uint32 `json:"maxSnapshotAge,omitempty"`
```

api/v1alpha1/etcd_types.go
```diff
@@ -69,6 +69,19 @@ type CompressionPolicy string
...
+	// SnapshotsToCopy defines how many snapshots will be copied starting with the most recent ones.
+	// +optional
+	SnapshotsToCopy *uint32 `json:"snapshotsToCopy,omitempty"`
```
stoyanr, line 82: Let's make this consistent with the naming in `etcd-backup-restore`
```suggestion
	MaxSnapshots *uint32 `json:"maxSnapshots,omitempty"`
```

api/v1alpha1/etcd_types.go
```diff
@@ -116,6 +129,9 @@ type BackupSpec struct {
 	// Store defines the specification of object store provider for storing backups.
 	// +optional
 	Store *StoreSpec `json:"store,omitempty"`
+	// CopySpec defines the specification of copy job to be deployed for copying backups.
+	// +optional
+	CopySpec *CopySpec `json:"copySpec,omitempty"`
```
stoyanr, line 134:
```suggestion
	Copy *CopySpec `json:"copy,omitempty"`
```

charts/etcd/templates/etcd-copy-job.yaml
```diff
@@ -0,0 +1,272 @@
...
+{{- end }}
+{{- if .Values.sourceStore.storageContainer }}
+        - --source-store-container={{ .Values.sourceStore.storageContainer }}
+{{- end }}
+{{- if .Values.copySpec.maxAge }}
+        - --max-age-days={{ .Values.copySpec.maxAge }}
+{{- end }}
+{{- if .Values.sourceStore.snapshotsToCopy }}
+        - --last-n-snapshots={{ .Values.copySpec.snapshotsToCopy }}
+{{- end }}
```
stoyanr, line 66: Let's adapt the naming of values and command line parameters to whatever we agree on in `etcd-backup-restore` (see also above).

charts/images.yaml
```diff
@@ -2,7 +2,7 @@ images:
 - name: etcd-backup-restore
   sourceRepository: github.com/gardener/etcd-backup-restore
   repository: eu.gcr.io/gardener-project/gardener/etcdbrctl
-  tag: "v0.13.0"
+  tag: "TODO"
```
stoyanr, line 5: Is this `TODO` supposed to be here?

controllers/etcd_controller.go
```diff
@@ -254,6 +258,20 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
 		}, err
 	}
 
+	if etcd.Spec.Backup.CopySpec != nil {
+		values, err := r.getMapFromEtcd(etcd)
+		if err != nil {
+			return ctrl.Result{Requeue: true}, err
```
stoyanr, line 264: Let's match the style of other returns
```suggestion
			return ctrl.Result{
				Requeue: true,
			}, err
```

controllers/etcd_controller.go
```diff
@@ -254,6 +258,20 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
+		}
+		fmt.Printf("### Starting Copy Job\n")
+		if err := r.runCopyJob(ctx, etcd, values, logger); err != nil {
+			return ctrl.Result{
+				Requeue: true,
+			}, err
+		}
+		return ctrl.Result{Requeue: false}, nil
```
stoyanr, line 272:
```suggestion
		return ctrl.Result{}, nil
```

controllers/etcd_controller.go
```diff
@@ -929,6 +946,44 @@ func (r *EtcdReconciler) getCronJobFromEtcd(etcd *druidv1alpha1.Etcd, values map
...
+	job, err := r.getCopyJobFromEtcd(etcd, values, logger)
+	if err != nil {
+		return err
+	}
+	if err := r.Get(ctx, kutil.Key(job.Namespace, job.Name), &batchv1.Job{}); err == nil {
+		logger.Info(fmt.Sprintf("Job %s/%s is already running", job.Namespace, job.Name))
+		return nil
+	}
+
+	return r.Create(ctx, job)
```
stoyanr, line 976: If `err` above is non-nil and not `NotFound`, you should return it, instead of proceeding with `Create`.

controllers/etcd_controller.go
```diff
@@ -929,6 +946,44 @@ func (r *EtcdReconciler) getCronJobFromEtcd(etcd *druidv1alpha1.Etcd, values map
...
+}
+
+func decodeObject(renderedChart *chartrenderer.RenderedChart, path string, object interface{}) error {
```
stoyanr, line 979: Since we have introduced this function, how about using it also in other places with similar logic, e.g. line 938 above.

controllers/etcd_controller.go
```diff
@@ -929,6 +946,44 @@ func (r *EtcdReconciler) getCronJobFromEtcd(etcd *druidv1alpha1.Etcd, values map
...
+	if content, ok := renderedChart.Files()[path]; ok {
+		decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader([]byte(content)), 1024)
+		return decoder.Decode(&object)
+	}
+	return fmt.Errorf("missing template file in the charts: %s", path)
```
stoyanr, line 984:
```suggestion
	return fmt.Errorf("missing file %s in the rendered chart", path)
```

controllers/etcd_controller.go
```diff
@@ -1215,9 +1271,45 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
 
+	if etcd.Spec.Backup.CopySpec != nil {
+		sourceStoreValues, err := getStoreValues(etcd.Spec.Backup.CopySpec.SourceStore)
+		if err != nil {
+			return nil, err
+		}
+		values["sourceStore"] = sourceStoreValues
+		if etcd.Spec.Backup.CopySpec.MaxAge != nil {
+			values["copySpec"] = map[string]interface{}{
+				"maxAge": etcd.Spec.Backup.CopySpec.MaxAge,
```
stoyanr, line 1282:
```suggestion
				"maxAge": *etcd.Spec.Backup.CopySpec.MaxAge,
```

controllers/etcd_controller.go
```diff
@@ -1215,9 +1271,45 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
+			}
+		}
+		if etcd.Spec.Backup.CopySpec.SnapshotsToCopy != nil {
+			values["copySpec"] = map[string]interface{}{
+				"snapshotsToCopy": etcd.Spec.Backup.CopySpec.SnapshotsToCopy,
```
stoyanr, line 1287:
```suggestion
				"snapshotsToCopy": *etcd.Spec.Backup.CopySpec.SnapshotsToCopy,
```

controllers/etcd_controller.go
```diff
@@ -1215,9 +1271,45 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
+		if err != nil {
+			return nil, err
+		}
+		values["sourceStore"] = sourceStoreValues
+		if etcd.Spec.Backup.CopySpec.MaxAge != nil {
+			values["copySpec"] = map[string]interface{}{
+				"maxAge": etcd.Spec.Backup.CopySpec.MaxAge,
+			}
+		}
+		if etcd.Spec.Backup.CopySpec.SnapshotsToCopy != nil {
```
stoyanr, line 1285: I find it a bit confusing that if the second condition is also true, `values["copySpec"]` would be overwritten. I would suggest adding an else or using a switch to make it clear that the 2 conditions are supposed to be mutually exclusive.

controllers/etcd_controller.go
```diff
@@ -1215,9 +1271,45 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
 	}
 
+	if etcd.Spec.Backup.CopySpec != nil {
+		sourceStoreValues, err := getStoreValues(etcd.Spec.Backup.CopySpec.SourceStore)
+		if err != nil {
+			return nil, err
+		}
+		values["sourceStore"] = sourceStoreValues
+		if etcd.Spec.Backup.CopySpec.MaxAge != nil {
+			values["copySpec"] = map[string]interface{}{
```
stoyanr, line 1281:
```suggestion
			values["copy"] = map[string]interface{}{
```

controllers/etcd_controller.go
```diff
@@ -1215,9 +1271,45 @@ func (r *EtcdReconciler) getMapFromEtcd(etcd *druidv1alpha1.Etcd) (map[string]in
...
+				"maxAge": etcd.Spec.Backup.CopySpec.MaxAge,
+			}
+		}
+		if etcd.Spec.Backup.CopySpec.SnapshotsToCopy != nil {
+			values["copySpec"] = map[string]interface{}{
```
stoyanr, line 1286:
```suggestion
			values["copy"] = map[string]interface{}{
```

charts/etcd/templates/etcd-copy-job.yaml
```diff
@@ -0,0 +1,272 @@
...
+        - --source-storage-provider={{ .Values.sourceStore.storageProvider }}
+{{- end }}
+{{- if .Values.sourceStore.storePrefix }}
+        - --source-store-prefix={{ .Values.sourceStore.storePrefix }}
+{{- end }}
+{{- if .Values.sourceStore.storageContainer }}
+        - --source-store-container={{ .Values.sourceStore.storageContainer }}
+{{- end }}
+{{- if .Values.copySpec.maxAge }}
+        - --max-age-days={{ .Values.copySpec.maxAge }}
```
stoyanr, line 62:
```suggestion
        - --max-age-days={{ .Values.copy.maxAge }}
```

charts/etcd/templates/etcd-copy-job.yaml
```diff
@@ -0,0 +1,272 @@
...
+{{- end }}
+{{- if .Values.sourceStore.snapshotsToCopy }}
+        - --last-n-snapshots={{ .Values.copySpec.snapshotsToCopy }}
```
stoyanr, line 65:
```suggestion
        - --last-n-snapshots={{ .Values.copy.snapshotsToCopy }}
```

config/crd/bases/druid.gardener.cloud_etcds.yaml
```diff
@@ -65,6 +62,43 @@ spec:
                         - zlib
                         type: string
                     type: object
+                  copySpec:
+                    description: CopySpec defines the specification of copy job to be deployed for copying backups.
```
stoyanr, line 66:
```suggestion
                  copy:
                    description: Copy defines the specification of copy job to be deployed for copying backups.
```

stoyanr (changes requested):
Have you tested actually starting the copy job, especially with some of the additional params (e.g. `maxAge`)?

charts/images.yaml
kris94: Yes, this is just a reminder to switch to the proper version once we release the copy operation in the etcd backup restore.

controllers/etcd_controller.go
kris94: Before the last changes in the copy operation it was only possible to set one of the those properties, so this is kind of left over

api/v1alpha1/etcd_types.go
```diff
@@ -69,6 +69,19 @@ type CompressionPolicy string
 // +kubebuilder:validation:Enum=periodic;revision
 type CompactionMode string
 
+type CopySpec struct {
+	// SourceStore defines the specification of the source object store provider for copying backups.
+	// +optional
+	SourceStore *StoreSpec `json:"sourceStore,omitempty"`
+	// MaxAge defines the maximum age in days that one snapshot must be in order to be copied.
```
stoyanr, line 76:
```suggestion
	// MaxSnapshotAge defines the maximum age in days that one snapshot must be in order to be copied.
```

api/v1alpha1/etcd_types.go
```diff
@@ -69,6 +69,19 @@ type CompressionPolicy string
...
+	// By default all snapshots will be copied.
+	// +optional
+	MaxSnapshotAge *uint32 `json:"maxSnapshotAge,omitempty"`
+	// SnapshotsToCopy defines how many snapshots will be copied starting with the most recent ones.
```
stoyanr, line 80:
```suggestion
	// MaxSnapshots defines how many snapshots will be copied starting with the most recent ones.
```

stoyanr (changes requested):
I only spotted a few mistakes in comments. Also let's perhaps adapt the naming of copy parameters from "snapshots" to "backups" (if we change the naming in https://github.com/gardener/etcd-backup-restore/pull/369).

controllers/etcd_controller.go
```diff
@@ -254,6 +258,21 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
 	}
 
+	if etcd.Spec.Backup.Copy != nil {
+		values, err := r.getMapFromEtcd(etcd)
+		if err != nil {
+			return ctrl.Result{
+				Requeue: true,
+			}, err
+		}
+		if err := r.runCopyJob(ctx, etcd, values, logger); err != nil {
```
plkokanov, line 268: Hmm, I don't see where we update the status of the etcd resource to indicate that the job has finished.

I think that the flow should be the following:
1. Trigger the copy job if it has not been triggered.
2. Check if copy job is successfull or not
3. If it is not successful, and this is due to an error, then return an error (the requeue should happen automatically) and update the status of the ETCD resource as errored.
4. If it is not successfull, but is still running normally, return nil error but use Requeue
5. If the job has finished successfully, return nil and update the status of the etcd resource to Successful.

plkokanov (changes requested):

kris94:
In a meeting with @timuthy @stoyanr @plkokanov we discussed the current approach and we decided to change it in the following way:
1. A new resource will be developed called `etcd-copy-task` that will contain all required parameters for the copy job.
2. A new controller will be implemented that will reconcile that resource and will start the copy operation and will update the status of the `etcd-copy-task`

stoyanr:
> 1. A new resource will be developed called `EtcdCopyTask` that will contain all required parameters for the copy job.

I think we agreed to name it rather `EtcdCopyBackupsTask`. I think mentioning "backups" is important here.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,446 @@
...
+	if err := r.removeOperationAnnotation(ctx, logger, copyTask); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{
+			Requeue: true,
+		}, err
+	}
+
+	if !copyTask.Status.TaskDeployed {
```
stoyanr, line 127: As already mentioned, I don't think you need the `TaskDeployed` field, you should rather use the reference field.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,446 @@
...
+	if err := r.updateEtcdCopyBackupsTaskStatus(ctx, copyTask); err != nil {
+		return ctrl.Result{
+			Requeue: true,
+		}, err
+	}
+
+	return ctrl.Result{
+		Requeue:      true,
+		RequeueAfter: 30 * time.Second,
+	}, nil
```
stoyanr, line 172: Why always requeue? I think the controller should only requeue until the job has completed, either successfully or with an error.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,446 @@
...
+		if copyTask.Status.LastOperation != nil && copyTask.Status.LastOperation.State == druidv1alpha1.LastOperationStateActive {
+			return ctrl.Result{
+				Requeue: true,
+			}, errors.New("The resource can not be deleted while still processing")
+		}
+		if !reflect.DeepEqual(copyTask.Status.TaskReference.Kind, "") && !reflect.DeepEqual(copyTask.Status.TaskReference.Kind, "Job") {
+			return ctrl.Result{
+				Requeue: true,
+			}, fmt.Errorf("Unsupported task kind %s", copyTask.Status.TaskReference.Kind)
+		}
```
stoyanr, line 194: I believe this check is redundant, since your logic will not put anything else but `Job` in `Kind`.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,446 @@
...
+			if err := r.Delete(ctx, job); client.IgnoreNotFound(err) != nil {
+				return ctrl.Result{
+					Requeue: true,
+				}, err
+			}
+		} else if client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{
+				Requeue: true,
+			}, err
+		}
```
stoyanr, line 212: You should rather do the normal `clinet.IgnoreNotFound(err) != nil` already in the if above, and move the finalizer removal and the deletion after the if.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,446 @@
...
+		return ctrl.Result{
+			Requeue: true,
+		}, err
+	}
+
+	logger.Info(fmt.Sprintf("Deleted ETCDCopyBackupTask %s/%s successfully.", copyTask.Namespace, copyTask.Name))
+	return ctrl.Result{}, nil
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) updateEtcdCopyTaskStatusWithTaskDescription(ctx context.Context, copyTask *druidv1alpha1.EtcdCopyBackupsTask, name, kind, apiGroup string) error {
```
stoyanr, line 224: I think there are too many different status update methods. Could you try to combine at least `updateEtcdCopyTaskStatusWithTaskDescription` and `updateEtcdCopyBackupsTaskStatus` into one method? There is no need to pass `name`, `kind`, and `apiGroup` as parameters here from my perspective, you can hardcode / compute these already in the method.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,446 @@
...
+
+func (r *EtcdCopyBackupsTaskReconciler) updateEtcdCopyBackupsTaskStatus(ctx context.Context, copyTask *druidv1alpha1.EtcdCopyBackupsTask) error {
+	return kutil.TryUpdateStatus(ctx, retry.DefaultBackoff, r.Client, copyTask, func() error {
+		if copyTask.Status.TaskDeployed && copyTask.Status.TaskReference != nil {
+			lastOperation, err := r.getCopyJobLastOperation(ctx, copyTask)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					if copyTask.Status.TaskDeployed {
+						copyTask.Status.LastError = pointer.StringPtr(fmt.Sprintf("Task %s was forcefully deleted", copyTask.Status.TaskReference.Name))
+					}
```
stoyanr, line 253: This check is redundant from my perspective, if someone manages to delete the job it's not something we care too much about. We should simply set the reference field to `nil` here as we know it doesn't exist anymore and that's it.

main.go
```diff
@@ -131,6 +131,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	copyTaskController, err := controllers.NewEtcdEtcdCopyBackupsTaskReconcilerWithImageVector(mgr)
```
stoyanr, line 134:
```suggestion
	etcdCopyBackupsTaskController, err := controllers.NewEtcdEtcdCopyBackupsTaskReconcilerWithImageVector(mgr)
```
(or, `ecbtController`)

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,446 @@
...
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) reconcile(ctx context.Context, copyTask *druidv1alpha1.EtcdCopyBackupsTask) (ctrl.Result, error) {
+	logger := r.logger.WithValues("etcd", kutil.Key(copyTask.Namespace, copyTask.Name).String(), "operation", "reconcile")
+	logger.Info("Starting operation")
+
+	if err := r.addFinalizerToObject(ctx, copyTask, logger); err != nil {
+		return ctrl.Result{
+			Requeue: true,
+		}, err
```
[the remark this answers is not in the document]
kris94: https://github.com/gardener/etcd-druid/pull/234#discussion_r714916480 We should either match the stile in the project or not.
Anyway, I don't mind to make all `return Results{}` to be in  one line one the whole project, because I don't like this either.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,446 @@
...
+	taskList := druidv1alpha1.EtcdCopyBackupsTaskList{}
+
+	if err := r.List(ctx, &taskList, client.InNamespace(copyTask.Namespace)); err != nil {
+		return err
+	}
+	for _, task := range taskList.Items {
+		if !reflect.DeepEqual(task.Name, copyTask.Name) &&
+			reflect.DeepEqual(task.Spec.TargetStore.Provider, copyTask.Spec.TargetStore.Provider) &&
+			reflect.DeepEqual(task.Spec.TargetStore.Container, copyTask.Spec.TargetStore.Container) &&
+			reflect.DeepEqual(task.Spec.TargetStore.Prefix, copyTask.Spec.TargetStore.Prefix) {
```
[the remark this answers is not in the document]
stoyanr: OK, I see, thanks for the explanation!

api/v1alpha1/etcd_types.go
```diff
@@ -393,3 +393,95 @@ type EtcdList struct {
...
+type EtcdCopyBackupsTaskStatus struct {
+	// ObservedGeneration is the most recent generation observed for this resource.
+	// +optional
+	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
+	// LastOperation represents the latest available observations of an etcd backup copy task current state.
+	// +optional
+	LastOperation *LastOperation `json:"lastOperation,omitempty"`
+	// TaskReference represents the details of the task that is currently deployed to preform the copy operation.
+	// +optional
+	TaskReference *corev1.TypedLocalObjectReference `json:"taskReference,omitempty"`
```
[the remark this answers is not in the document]
stoyanr: I see this is still not addressed so let me justify my proposal. I think the name `TaskReference` is misleading since it's not a reference of a `Task`, but to a different object, e.g. `Job`, that has been created in order to perform whatever the task needs to perform.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,450 @@
...
+		if !ok {
+			return nil, errors.New("Unable to find etcdbrctl image")
+		}
+		values["image"] = val.String()
+	}
+
+	return values, nil
+}
+
+func getStoreValues(store *druidv1alpha1.StoreSpec) (map[string]interface{}, error) {
```
stoyanr, line 361: I still don't see this function used in `etcd_controller.go` (line 1203).

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,450 @@
...
+				copyTask.Status.TaskReference = nil
+				copyTask.Status.LastOperation = nil
+				return nil
+			} else {
+				copyTask.Status.LastError = pointer.StringPtr(err.Error())
+				return err
+			}
+		}
+		copyTask.Status.LastOperation = lastOperation
+		copyTask.Status.ObservedGeneration = &copyTask.Generation
```
stoyanr, line 249: I find the logic above somewhat hard to follow, I would propose something similar to:
```suggestion
		lastOperation, err := r.getLastOperation(ctx, copyTask)
		if err != nil {
			if apierrors.IsNotFound(err) {
				copyTask.Status.TaskReference = nil
				copyTask.Status.LastOperation = nil
				copyTask.Status.LastError = nil
			} else {
				copyTask.Status.LastError = pointer.StringPtr(err.Error())
			}
		} else {
			copyTask.Status.TaskReference = &corev1.TypedLocalObjectReference{
				Name:     copyTask.Name + workerSuffix,
				Kind:     "Job",
				APIGroup: pointer.StringPtr(batchv1.SchemeGroupVersion.String()),
			}
			copyTask.Status.LastOperation = lastOperation
		}
		copyTask.Status.ObservedGeneration = &copyTask.Generation
```

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,450 @@
...
+	if err != nil {
+		return nil, err
+	}
+	targetStoreValues, err := getStoreValues(copyTask.Spec.TargetStore)
+	if err != nil {
+		return nil, err
+	}
+	values["sourceStore"] = sourceStoreValues
+	values["targetStore"] = targetStoreValues
+	if copyTask.Spec.MaxBackupAge != nil || copyTask.Spec.MaxBackups != nil {
```
stoyanr, line 335: This if is not needed anymore.

api/v1alpha1/etcd_types.go
```diff
@@ -393,3 +393,92 @@ type EtcdList struct {
...
+	State LastOperationState `json:"state" protobuf:"bytes,4,opt,name=state,casttype=LastOperationState"`
+}
+
+// EtcdCopyBackupsTaskSpec defines the parameters for the copy operation
+type EtcdCopyBackupsTaskSpec struct {
+	// SourceStore defines the specification of object store provider for storing backups.
+	// +required
+	SourceStore *StoreSpec `json:"sourceStore,omitempty"`
+	// TargetStore defines the specification of object store provider for storing backups.
+	// +required
```
timuthy, line 426: Why is `SourceStore` optional?

api/v1alpha1/etcd_types.go
```diff
@@ -393,3 +393,92 @@ type EtcdList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []Etcd `json:"items"`
 }
+
+// LastOperationState is a string alias.
+type LastOperationState string
+
+const (
+	// LastOperationStateActive indicates that an operation is ongoing.
+	LastOperationStateActive LastOperationState = "Active"
```
timuthy, line 402:
```suggestion
	LastOperationStateProcessing LastOperationState = "Processing"
```
To be consistent with `g/g`.

api/v1alpha1/etcd_types.go
```diff
@@ -393,3 +393,92 @@ type EtcdList struct {
...
+type EtcdCopyBackupsTaskStatus struct {
+	// ObservedGeneration is the most recent generation observed for this resource.
+	// +optional
+	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
+	// LastOperation represents the latest available observations of an etcd backup copy task current state.
+	// +optional
+	LastOperation *LastOperation `json:"lastOperation,omitempty"`
+	// TaskReference represents the details of the task that is currently deployed to preform the copy operation.
+	// +optional
+	ObjectRef *corev1.TypedLocalObjectReference `json:"objectRef,omitempty"`
```
timuthy, line 450:
```suggestion
	TaskReference *corev1.TypedLocalObjectReference `json:"taskReference,omitempty"`
```

api/v1alpha1/etcd_types.go
```diff
@@ -393,3 +393,92 @@ type EtcdList struct {
...
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.ready`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// EtcdCopyBackupsTask is the Schema for the EtcdCopyBackupsTask object that
+// is responsible for declaring the parameters for etcd backup copy operation.
+type EtcdCopyBackupsTask struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   *EtcdCopyBackupsTaskSpec  `json:"spec,omitempty"`
```
timuthy, line 470: Why is it a pointer?

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,409 @@
...
+	if err := r.removeFinalizerFromObject(ctx, etcdCopyBackupsTask, logger); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	logger.Info(fmt.Sprintf("Deleted ETCDCopyBackupTask %s/%s successfully.", etcdCopyBackupsTask.Namespace, etcdCopyBackupsTask.Name))
+	return ctrl.Result{}, nil
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) updateEtcdCopyTaskStatusLastError(ctx context.Context, etcdCopyBackupsTask *druidv1alpha1.EtcdCopyBackupsTask, err error) error {
+	return kutil.TryUpdateStatus(ctx, retry.DefaultBackoff, r.Client, etcdCopyBackupsTask, func() error {
```
timuthy, line 201: Let's get rid of the try update logic and rather return an error. See https://github.com/gardener/gardener/blob/master/docs/development/kubernetes-clients.md#dont-retry-on-conflict.
I anyway don't expect conflicts today as there is likely just one controller updating the status.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,409 @@
...
+		etcdCopyBackupsTask.Status.LastError = pointer.StringPtr(err.Error())
+		etcdCopyBackupsTask.Status.ObservedGeneration = &etcdCopyBackupsTask.Generation
+		return nil
+	})
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) updateEtcdCopyBackupsTaskStatus(ctx context.Context, etcdCopyBackupsTask *druidv1alpha1.EtcdCopyBackupsTask) error {
+	return kutil.TryUpdateStatus(ctx, retry.DefaultBackoff, r.Client, etcdCopyBackupsTask, func() error {
```
timuthy, line 209: Same as previous comment about try update.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,409 @@
...
+
+func (r *EtcdCopyBackupsTaskReconciler) runCopyJob(ctx context.Context, etcdCopyBackupsTask *druidv1alpha1.EtcdCopyBackupsTask, values map[string]interface{}, logger logr.Logger) (*batchv1.Job, error) {
+	job, err := r.getCopyJob(etcdCopyBackupsTask, values, logger)
+	if err != nil {
+		return nil, err
+	}
+	return job, r.Create(ctx, job)
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) getValuesMap(etcdCopyBackupsTask *druidv1alpha1.EtcdCopyBackupsTask) (map[string]interface{}, error) {
```
timuthy, line 296: The only think you need from `EtcdCopyBackupsTaskReconciler` is the `imageVector`. My suggestion: Remove the pointer receiver, pass `imageVector` as a function argument. Then you can easily add unit tests.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,409 @@
...
+		if !ok {
+			return nil, errors.New("Unable to find etcdbrctl image")
+		}
+		values["image"] = val.String()
+	}
+
+	return values, nil
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) addFinalizerToObject(ctx context.Context, object client.Object, logger logr.Logger) error {
```
timuthy, line 338: This can be a reusable and tested util function if you remove the pointer receiver and pass the client.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,409 @@
...
+		finalizers.Insert(FinalizerName)
+		objectCopy.SetFinalizers(finalizers.UnsortedList())
+		if err := r.Patch(ctx, objectCopy, client.MergeFrom(object)); client.IgnoreNotFound(err) != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) removeFinalizerFromObject(ctx context.Context, object client.Object, logger logr.Logger) error {
```
timuthy, line 351: This can be a reusable and tested util function if you remove the pointer receiver and pass the client.

api/v1alpha1/etcd_types.go
timuthy: I rather addressed it because the doc string does not match the renamed field. Can you change the doc string then?

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,409 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if !etcdCopyBackupsTask.DeletionTimestamp.IsZero() {
+		return r.delete(ctx, etcdCopyBackupsTask)
+	}
+	return r.reconcile(ctx, etcdCopyBackupsTask)
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) reconcile(ctx context.Context, etcdCopyBackupsTask *druidv1alpha1.EtcdCopyBackupsTask) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
timuthy: Hm, what do you mean?

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,409 @@
...
+
+	return nil
+}
+
+// SetupWithManager sets up manager with a new controller and ec as the reconcile.Reconciler
+func (r *EtcdCopyBackupsTaskReconciler) SetupWithManager(mgr ctrl.Manager, ignoreOperationAnnotation bool) error {
+	builder := ctrl.NewControllerManagedBy(mgr)
+	builder = builder.WithEventFilter(buildPredicate(ignoreOperationAnnotation)).For(&druidv1alpha1.EtcdCopyBackupsTask{})
+	if ignoreOperationAnnotation {
+		builder = builder.Owns(&batchv1.Job{})
```
timuthy, line 387: Can you explain why`Job`s are only watched when `ignoreOperationAnnotation`?
Shouldn't we either decide to automatically enqueue after `30s` or to watch `Job`s, but not both?

controllers/etcd_copy_backups_controller.go
stoyanr: Also here, in addition to getting rid of the try update logic, shouldn't we rather update the status via a `Patch`?

controllers/etcd_copy_backups_controller.go
stoyanr: I tend to disagree with this suggestion. I would rather not make a method that naturally belongs to the type a standalone function just because it's a bit easier to add unit tests. Also, it's a unexported method that from my perspective shouldn't be tested on its own. I see we already have unit tests for `etcd_controller.go`, shouldn't we adopt a similar approach for this controller?

charts/etcd/templates/etcd-copy-backups-job.yaml
```diff
@@ -0,0 +1,270 @@
+{{- if  .Values.sourceStore}}
```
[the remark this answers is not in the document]
stoyanr: The existing templates in the `etcd` chart are all rendered independently from each other, similarly to how this one is rendered. Of course, the difference is that for the other charts this is done in the same controller, but is this strong enough reason to have 2 different charts?

api/v1alpha1/etcd_types.go
```diff
@@ -393,3 +393,92 @@ type EtcdList struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.ready`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// EtcdCopyBackupsTask is the Schema for the EtcdCopyBackupsTask object that
+// is responsible for declaring the parameters for etcd backup copy operation.
+type EtcdCopyBackupsTask struct {
```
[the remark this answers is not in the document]
stoyanr: We don't even have proper validation for `Etcd` yet. Since there are a few shared types, perhaps this could be added with a different PR?

api/v1alpha1/etcd_types.go
```diff
@@ -393,3 +393,92 @@ type EtcdList struct {
...
+	// LastUpdateTime is the last time the operation state transitioned from one to another.
+	LastUpdateTime metav1.Time `json:"lastUpdateTime" protobuf:"bytes,2,opt,name=lastUpdateTime"`
+	// State is the status of the last operation, one of Aborted, Processing, Succeeded, Error, Failed.
+	State LastOperationState `json:"state" protobuf:"bytes,4,opt,name=state,casttype=LastOperationState"`
+}
+
+// EtcdCopyBackupsTaskSpec defines the parameters for the copy operation
+type EtcdCopyBackupsTaskSpec struct {
+	// SourceStore defines the specification of object store provider for storing backups.
+	// +required
```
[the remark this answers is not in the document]
timuthy: Ups, sorry. Then, why is `SourceStore *StoreSpec` are pointer?

controllers/etcd_copy_backups_controller.go
timuthy: Also fine for me 👍

api/v1alpha1/etcd_types.go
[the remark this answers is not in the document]
stoyanr: Yes, this is what I meant, it should not be a pointer.

controllers/etcd_copy_backups_controller.go
```diff
@@ -0,0 +1,409 @@
...
+	return values, nil
+}
+
+func (r *EtcdCopyBackupsTaskReconciler) addFinalizerToObject(ctx context.Context, object client.Object, logger logr.Logger) error {
+	if !sets.NewString(object.GetFinalizers()...).Has(FinalizerName) {
+		objectCopy := object.DeepCopyObject().(client.Object)
+		finalizers := sets.NewString(objectCopy.GetFinalizers()...)
+		finalizers.Insert(FinalizerName)
+		objectCopy.SetFinalizers(finalizers.UnsortedList())
+		if err := r.Patch(ctx, objectCopy, client.MergeFrom(object)); client.IgnoreNotFound(err) != nil {
```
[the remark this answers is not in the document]
stoyanr: OK, thanks for the explanation!

api/v1alpha1/etcd_types.go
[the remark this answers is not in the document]
stoyanr: OK, makes sense.

stoyanr:
@timuthy Rebased on top of latest master, using the latest `etcd-backup-restore` release `v0.14.0`, and switched the pull policy to `IfNotPresent`. I consider this PR ready to be merged now.
/cc @kris94 @plkokanov

stoyanr:
Rebased on top of latest master

charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
```diff
@@ -0,0 +1,246 @@
...
+  name: {{ .Values.name }}
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    gardener.cloud/owned-by: "{{ .Release.Namespace }}/{{ .Values.ownerName }}"
+    gardener.cloud/owner-type: "etcdcopybackupstask"
+{{- if .Values.annotations }}
+{{ toYaml .Values.annotations | indent 4 }}
+{{- end }}
+  labels:
+    networking.gardener.cloud/to-public-networks: allowed
```
timuthy, line 13: Do we need this label?

controllers/etcd_controller.go
```diff
@@ -35,13 +35,10 @@ import (
 	"github.com/gardener/gardener/pkg/utils/imagevector"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	gardenerretry "github.com/gardener/gardener/pkg/utils/retry"
-
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
-
-	batchv1 "k8s.io/api/batch/v1beta1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
```
timuthy, line 40: Can you remove the `batchv1beta1` changes? This import will entirely be removed by https://github.com/gardener/etcd-druid/pull/235 and thus will cause unnecessary merge conflicts.

timuthy (commented):
Looks good so far, only 2 comments left from my side.

charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
stoyanr: We do, since the actual copying of backups by `etcd-backup-restore` needs to access the backup containers on the cloud providers. Or am I missing something?

controllers/etcd_controller.go
stoyanr: Done

charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
timuthy: To my understanding this is set via `spec.template.metadata.labels`, right?

stoyanr: Right :-( Not sure what this label was doing here, it looks like another leftover from before that I didn't clean. Thanks for noticing!

timuthy (approved):
/lgtm

PR #235, opened 2021-09-25
Enabled jobs instead of cronjobs to take care of running compaction.

abdasgupta:

/area performance
/kind enhancement

Earlier cronjobs are scheduled through druid for backup compaction of ETCD. But that incurs huge cost as we have to predefine schedule and resources for compaction jobs. This PR enabled to run jobs instead of cronjobs to run the backup compaction. Jobs are triggered based on the number of events that accumulated as delta snapshot after last full snapshot.

cc @aaronfern 
Fixes #

**Release note**:

```breaking operator
CronJob is no more used to schedule compaction job at regular interval. Instead, we are using Job.
Therefore, `BackupCompactionSchedule` field is removed from ETCD backup spec, as it was only necessary for scheduling CronJob.
```
```feature operator
A new controller named lease controller has been introduced. Lease controller will be responsible for creating compaction job based on the delta event lease.
For this, two new `Lease`s are introduced: One to hold the value of the latest full snapshot revision and one for the last delta revision.
```


controllers/etcd_custodian_controller.go
```diff
@@ -80,12 +125,183 @@ func (ec *EtcdCustodian) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+			}
+
+			chartPath := getChartPath()
+			renderedChart, err := ec.chartApplier.Render(chartPath, etcd.Name, etcd.Namespace, values)
+			if err != nil {
+				logger.Error(err, "Error while rendering chart")
+				return ctrl.Result{}, err
+			}
+
+			j, err := ec.reconcileJob(ctx, logger, etcd, renderedChart)
```
timuthy, line 297: I suggest to add a new reconciler/controller which is solely responsible for creating/reconciling the job. It's a clear separation of concern and not really related to what the custodian does today (mainly health checks). This controller can then watch the `Leases` and create the jobs whenever necessary.

controllers/etcd_custodian_controller.go
```diff
@@ -80,12 +125,183 @@ func (ec *EtcdCustodian) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
-		return ctrl.Result{
-			RequeueAfter: 30 * time.Second,
-		}, nil
+	// Get or Create full_snapshot_revisions lease object that will help to set BackupReady condition
+	nsName := types.NamespacedName{
+		Name:      "full-snapshot-revisions",
+		Namespace: etcd.Namespace,
+	}
+
+	fullLease := &coordinationv1.Lease{}
```
timuthy, line 134: I prefer to create the `Lease` objects in the `etcd-controller` because the objects themselves are bound to the lifecycle of the etcd cluster which is created/updated/deleted by the `etcd-controller`, not by the `custodian`.

timuthy (changes requested):
Thanks for the PR. I have a few suggestions and questions which I added as a comment to the affected line(s).

controllers/etcd_custodian_controller.go
```diff
@@ -168,6 +384,78 @@ func (ec *EtcdCustodian) updateEtcdStatus(ctx context.Context, logger logr.Logge
...
 
+func (ec *EtcdCustodian) reconcileJob(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd, renderedChart *chartrenderer.RenderedChart) (*batchv1.Job, error) {
+	logger.Info("Reconcile etcd compaction job")
+
+	var job batchv1.Job
+	err := ec.Get(ctx, types.NamespacedName{Name: getJobName(etcd), Namespace: etcd.Namespace}, &job)
+
+	if err == nil {
+		// Delete job and reque if the last job failed
+		if job.Status.Failed > 0 {
```
timuthy, line 395: Are you sure that this condition is sufficient to delete the entire job? What about retries? Then there will be `Failed` executions but you probably interrupt an active compaction process.

controllers/etcd_custodian_controller.go
```diff
@@ -168,6 +384,78 @@ func (ec *EtcdCustodian) updateEtcdStatus(ctx context.Context, logger logr.Logge
...
+	}
+
+	// Required job doesn't exist. Create new
+	j, err := ec.getJobFromEtcd(etcd, renderedChart, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	logger.Info("Creating job", "job", kutil.Key(j.Namespace, j.Name).String())
+	err = ec.Create(ctx, j)
```
timuthy, line 423: I think this will fail if the previous compaction was successful because nothing deleted the previous `Job` right?

controllers/etcd_custodian_controller.go
```diff
@@ -80,12 +125,183 @@ func (ec *EtcdCustodian) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+		if err != nil {
+			logger.Error(err, "Delta snap lease found but error occured during claiming it")
+			return ctrl.Result{
+				RequeueAfter: 5 * time.Second,
+			}, nil
+		}
+	}
+	// Run compaction job
+	if etcd.Spec.Backup.Store != nil {
+		hi, err := strconv.ParseInt(*deltaLease.Spec.HolderIdentity, 10, 32)
```
timuthy, line 275: Can you elaborate what component resets this count after the compaction was successful?

timuthy (changes requested):

charts/etcd/templates/etcd-compaction-job.yaml
```diff
@@ -0,0 +1,222 @@
+apiVersion: batch/v1
```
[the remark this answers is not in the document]
abdasgupta: Nice! I saw the comment just at the nick of time. I already had lease created in code. I can create Job like that. Thanks.

controllers/etcd_custodian_controller.go
```diff
@@ -80,12 +125,183 @@ func (ec *EtcdCustodian) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+	deltaLease := &coordinationv1.Lease{}
+	err1 = ec.Get(ctx, nsName, deltaLease)
+	if err1 != nil {
+		logger.Info("Couldn't fetch delta snap lease delta-snapshot-revisions because: " + err1.Error())
+		logger.Info("Creating the delta snap lease delta-snapshot-revisions")
+
+		renewTime := metav1.NewMicroTime(time.Now())
+		deltaLease = &coordinationv1.Lease{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "delta-snapshot-revisions",
```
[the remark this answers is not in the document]
abdasgupta: Hmm... I think it ll be better to append etcd name to the string delta-snapshot. Do you have any suggestion?

controllers/etcd_custodian_controller.go
```diff
@@ -80,12 +125,183 @@ func (ec *EtcdCustodian) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+		fullLease = &coordinationv1.Lease{}
+		err2 = ec.Get(ctx, nsName, fullLease)
+		if err2 != nil {
+			logger.Error(err2, "Full snap lease full-snapshot-revisions couldn't be fetched just after creation")
+			return ctrl.Result{
+				RequeueAfter: 5 * time.Second,
+			}, nil
+		}
+	}
+	if !checkEtcdOwnerReference(fullLease.GetOwnerReferences(), etcd) {
```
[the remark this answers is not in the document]
abdasgupta: I just did a safety check here. owner reference should not ideally be absent. Is it redundant ?

controllers/etcd_custodian_controller.go
abdasgupta: Okay.. Makes sense.

controllers/etcd_custodian_controller.go
abdasgupta: Agreed

controllers/etcd_custodian_controller.go
abdasgupta: I think, I can mention the number of retry in job spec and I kept it as default which is 1. Let me check and confirm

controllers/etcd_custodian_controller.go
abdasgupta: Job will be immediately deleted after successful completion. It can be mentioned in job spec

controllers/etcd_custodian_controller.go
abdasgupta: That component is present in etcd br. After etcd br executes compaction, this value will be reset

controllers/etcd_custodian_controller.go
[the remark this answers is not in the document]
timuthy: Okay. I was wondering if this is rather a theoretical use-case because there is no other component or process which removes the ownerRef on purpose.

controllers/etcd_custodian_controller.go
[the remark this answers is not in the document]
shreyas-s-rao: @abdasgupta you can add a short suffix, similar to how the bootstrap configmap is named [here](https://github.com/gardener/etcd-druid/blob/1d427e9167adac1476d1847c0e265c2c09d6bc62/controllers/etcd_controller.go#L1176)

controllers/etcd_custodian_controller.go
shreyas-s-rao: @timuthy so a new controller like `compaction_controller`? But if I remember right, we wanted `custodian_controller` to be taking care of etcd operations that had to be done outside the shoot maintenance window, so compaction would still fit into the responsibilities of `custodian_controller`.

timuthy: > we wanted custodian_controller to be taking care of etcd operations

Hm, I think we should restrict that responsibility to the etcd cluster only. Compaction isn't really related to the cluster directly, IMHO. It's rather a separate feature with a separate reconcile trigger. So why not clearly isolating it with a dedicated controller instead of squeezing it into the Custodian and worsen code maintenance in the future?

controllers/lease_controller.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	// Delete job and return if the job succeeded
+	if job.Status.Succeeded > 0 {
+		err = lc.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground))
+		if err != nil {
+			return nil, err
+		}
+		return job, nil
+	}
```
aaronfern, line 166: Should the controller not requeue after a successful job has been deleted? Won't it be possible that we find the previously run job, delete that, and wait for the next delta snap lease update to create a new job?

controllers/lease_controller.go
```diff
@@ -0,0 +1,456 @@
...
+		logger.Info("Couldn't fetch delta snap lease because: " + err.Error())
+
+		return ctrl.Result{
+			RequeueAfter: 10 * time.Second,
+		}, err
+	}
+
+	// Run compaction job
+	if etcd.Spec.Backup.Store != nil {
+		hi, err := strconv.ParseInt(*deltaLease.Spec.HolderIdentity, 10, 32)
```
aaronfern, line 105: Would suggest using a 64 bit int here instead of a 32 bit int as `Snapshot.LastRevision`s are `int64` type

controllers/etcd_controller.go
```diff
@@ -791,146 +804,167 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
+	// compaction job will be scheduled
+	nsName := types.NamespacedName{
+		Name:      getDeltaLease(etcd),
+		Namespace: etcd.Namespace,
+	}
+
+	deltaLease := &coordinationv1.Lease{}
+	err1 := r.Get(ctx, nsName, deltaLease)
+	if err1 != nil {
+		logger.Info("Couldn't fetch delta snap lease delta-snapshot-revisions because: " + err1.Error())
```
aaronfern, line 896: Replace `delta-snapshot-revisions` with actual delta snap lease name here and other occurrences as well

controllers/etcd_controller.go
```diff
@@ -791,146 +804,167 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
+	// Get or Create full_snapshot_revisions lease object that will help to set BackupReady condition
+	nsName := types.NamespacedName{
+		Name:      getFullLease(etcd),
+		Namespace: etcd.Namespace,
+	}
+
+	fullLease := &coordinationv1.Lease{}
+	err1 := r.Get(ctx, nsName, fullLease)
+	if err1 != nil {
+		logger.Info("Couldn't fetch full snap lease full-snapshot-revisions because: " + err1.Error())
```
aaronfern, line 817: Replace `full-snapshot-revisions` with actual full snap name here and other occurrences as well

controllers/controller_ref_manager.go
```diff
@@ -529,3 +531,218 @@ func CheckStatefulSet(etcd *druidv1alpha1.Etcd, statefulSet *appsv1.StatefulSet)
...
+		"etcd":                    etcdValues,
+		"backup":                  backupValues,
+		"sharedConfig":            sharedConfigValues,
+		"replicas":                etcd.Spec.Replicas,
+		"statefulsetReplicas":     statefulsetReplicas,
+		"serviceName":             fmt.Sprintf("%s-client", etcd.Name),
+		"configMapName":           fmt.Sprintf("etcd-bootstrap-%s", string(etcd.UID[:6])),
+		"jobName":                 getJobName(etcd),
+		"volumeClaimTemplateName": volumeClaimTemplateName,
+	}
```
aaronfern, line 708: Just an FYI
will have to append `roleName`, `roleBindingName`, and `serviceAccountName` here after [233](https://github.com/gardener/etcd-druid/pull/233) is merged

aaronfern (changes requested):
Thanks @abdasgupta for the well written PR!
Just one query from me and some nitpicks 
Thanks!

controllers/lease_controller.go
```diff
@@ -0,0 +1,456 @@
...
+		Spec: batchv1.JobSpec{
+			ActiveDeadlineSeconds: pointer.Int64Ptr(int64(activeDeadlineSeconds.Seconds())),
+			BackoffLimit:          pointer.Int32Ptr(0),
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: etcd.Spec.Annotations,
+					Labels:      getLabels(etcd),
+				},
+				Spec: v1.PodSpec{
+					//ServiceAccountName: fmt.Sprint(values["serviceAccountName"]),
```
aaronfern, line 218: Needs to be uncommented before merging to add a serviceaccount to the compaction job
This is needed for backup sidecar to be able to update the snapshot leases

controllers/lease_controller.go
```diff
@@ -0,0 +1,456 @@
...
+func getCmpctJobEnvVar(etcd *druidv1alpha1.Etcd) []v1.EnvVar {
+	var env []v1.EnvVar
+	if etcd.Spec.Backup.Store == nil {
+		return env
+	}
+
+	storeValues := etcd.Spec.Backup.Store
+
+	env = append(env, getEnvVarFromValues("STORAGE_CONTAINER", *storeValues.Container))
+	//env = append(env, getEnvVarFromFields("POD_NAMESPACE", "metadata.namespace"))
```
aaronfern, line 328: Needs to be uncommented before merging to add pod namespace as an env var to the compaction job
This is needed by backup sidecar to update the snapshot leases

controllers/lease_controller.go
```diff
@@ -0,0 +1,456 @@
...
+				Key: secretKey,
+			},
+		},
+	}
+}
+
+func getCompactJobCommands(etcd *druidv1alpha1.Etcd) []string {
+	command := []string{"" + "etcdbrctl" + " " + "compact"}
+	command = append(command, "--data-dir=/var/etcd/data")
+	command = append(command, "--snapstore-temp-directory=/var/etcd/data/tmp")
```
aaronfern, line 412: 3 new flags need to be added for backup sidecar to be able to be update snapshot leases as part of compaction
```
--enable-etcd-lease-renewal=true
--full-snapshot-lease-name=<full_snap_name>
--delta-snapshot-lease-name<delta_snap_name>
```

aaronfern (commented):
Hey @abdasgupta 
Just adding below some changes to the job that needs to be made before this PR is merged
This is to ensure it correctly works with the etcdbr version

main.go
```diff
@@ -71,6 +72,8 @@ func main() {
 	flag.IntVar(&etcdWorkers, "workers", 3, "Number of worker threads of the etcd controller.")
 	flag.IntVar(&custodianWorkers, "custodian-workers", 3, "Number of worker threads of the custodian controller.")
 	flag.DurationVar(&custodianSyncPeriod, "custodian-sync-period", 30*time.Second, "Sync period of the custodian controller.")
+	flag.Int64Var(&eventsThreshold, "events-threshold", 1000000, "Total number of events that can be allowed before a compaction job is triggered")
+	flag.StringVar(&activeDeadlineDuration, "active-deadline-duration", "3h", "Duration after which a running compaction job will be killed (Ex: \"300ms\", \"20s\", \"-1.5h\" or \"2h45m\")")
```
timuthy, line 76: Why not `DurationVar`?

main.go
```diff
@@ -50,16 +50,17 @@ func init() {
...
+		metricsAddr                string
+		enableLeaderElection       bool
+		leaderElectionID           string
+		leaderElectionResourceLock string
+		etcdWorkers                int
+		custodianWorkers           int
+		custodianSyncPeriod        time.Duration
+		disableLeaseCache          bool
+		eventsThreshold            int64
+		activeDeadlineDuration     string
```
timuthy, line 62: Why not `time.Duration`?

controllers/etcd_controller.go
```diff
@@ -791,146 +804,169 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
+			err2 := r.Create(ctx, fullLease)
+			if err2 != nil {
+				logger.Error(err2, "Full snap lease "+fullSnapshotRevisions+" couldn't be created")
+				return nil, err2
 			}
 
-			if foundEtcd.GetDeletionTimestamp() != nil {
-				return nil, fmt.Errorf("%v/%v etcd is marked for deletion", etcd.Namespace, etcd.Name)
+			fullLease = &coordinationv1.Lease{}
+			err2 = r.Get(ctx, nsName, fullLease)
```
timuthy, line 851: Why do we need to `Get` the object here again?

controllers/config/compaction.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package config
+
+// CompactionConfig contains configuration for the compaction controller.
+type CompactionConfig struct {
+	// EventsThreshold is Total number of events that can be allowed before a compaction job is triggered
+	EventsThreshold int64
+	// ActiveDeadlineDuration is the duration after which a running compaction job will be killed (Ex: "300ms", "20s", "-1.5h" or "2h45m")
+	ActiveDeadlineDuration string
```
timuthy, line 22: Why not `time.Duration`?

main.go
```diff
@@ -126,11 +128,31 @@ func main() {
...
+	}
+
 	if err := custodian.SetupWithManager(ctx, mgr, custodianWorkers); err != nil {
 		setupLog.Error(err, "Unable to create controller", "Controller", "Etcd Custodian")
 		os.Exit(1)
 	}
 
+	lc, err := controllers.NewLeaseControllerWithImageVector(mgr, controllersconfig.CompactionConfig{
+		EventsThreshold:        1000000,
+		ActiveDeadlineDuration: "2m",
```
timuthy, line 143: You introduced command line flags to configure these values, but here they are set them statically. Either overwrite them here with the configured ones or remove the CMD flags which is probably not intended.

controllers/lease_controller.go
```diff
@@ -0,0 +1,506 @@
...
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+	controllersconfig "github.com/gardener/etcd-druid/controllers/config"
+	"github.com/gardener/etcd-druid/pkg/common"
+	"github.com/gardener/etcd-druid/pkg/utils"
+	"github.com/gardener/gardener/pkg/utils/imagevector"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+)
+
+// LeaseController reconciles compaction job
+type LeaseController struct {
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/lease_controller.go
```diff
@@ -0,0 +1,506 @@
...
+		Spec: batchv1.JobSpec{
+			ActiveDeadlineSeconds: pointer.Int64Ptr(int64(activeDeadlineSeconds.Seconds())),
+			BackoffLimit:          pointer.Int32Ptr(0),
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: etcd.Spec.Annotations,
+					Labels:      getLabels(etcd),
+				},
+				Spec: v1.PodSpec{
+					ServiceAccountName: fmt.Sprintf("%s-serviceaccount", etcd.Name),
```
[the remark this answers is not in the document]
abdasgupta: Yes. Done

controllers/lease_controller.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	backupValues := etcd.Spec.Backup
+	if backupValues.EtcdSnapshotTimeout != nil {
+		command = append(command, "--etcd-snapshot-timeout="+backupValues.EtcdSnapshotTimeout.Duration.String())
+	}
+	storeValues := etcd.Spec.Backup.Store
+	if storeValues != nil {
+		if storeValues.Provider != nil {
+			command = append(command, "--storage-provider="+string(*storeValues.Provider))
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/lease_controller.go
```diff
@@ -0,0 +1,506 @@
...
+	// Delete job and reque if the job doesn't have label with this etcd instance
+	if val, ok := job.Labels["instance"]; ok {
+		if val != etcd.Name {
+			err = lc.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground))
+			if err != nil {
+				return nil, err
+			}
+			return nil, fmt.Errorf("job doesn't have label for this etcd instance")
+		}
+	}
```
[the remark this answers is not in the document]
abdasgupta: Removed it. It's not necessary

controllers/lease_controller.go
```diff
@@ -0,0 +1,506 @@
...
+func (lc *LeaseController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	lc.logger.Info("Lease controller reconciliation started")
+	etcd := &druidv1alpha1.Etcd{}
+	if err := lc.Get(ctx, req.NamespacedName, etcd); err != nil {
+		if errors.IsNotFound(err) {
+			// Object not found, return. Created objects are automatically garbage collected.
+			// For additional cleanup logic use finalizers.
+			return ctrl.Result{
+				RequeueAfter: 10 * time.Second,
+			}, err
```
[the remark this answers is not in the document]
abdasgupta: removed it

controllers/lease_controller.go
```diff
@@ -0,0 +1,506 @@
...
+		}
+		// Error reading the object - requeue the request.
+		return ctrl.Result{
+			RequeueAfter: 10 * time.Second,
+		}, err
+	}
+
```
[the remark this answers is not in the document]
abdasgupta: Right

abdasgupta:
Rebase done @stoyanr @kris94 would you please take a look at the PR to ensure that I have rebased it properly?

controllers/lease_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	cronJob, err := lc.cleanCronJobs(ctx, logger, etcd)
+	if err != nil {
+		return ctrl.Result{
+			RequeueAfter: 10 * time.Second,
+		}, err
+	}
+
+	if cronJob != nil {
+		return ctrl.Result{Requeue: false}, nil
+	}
```
timuthy, line 161: See previous comment - I propose to delete `CronJobs` in `etcd_controller.go`

timuthy (changes requested):

controllers/lease_controller.go
```diff
@@ -0,0 +1,580 @@
...
+
+	if etcd.Spec.Backup.Store == nil {
+		return vs
+	}
+
+	storeValues := etcd.Spec.Backup.Store
+	provider, err := utils.StorageProviderFromInfraProvider(storeValues.Provider)
+	if err != nil {
+		return vs
```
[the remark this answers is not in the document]
abdasgupta: I want to treat it same as when Backup.Store is nil. But should I print the error in a log?

controllers/lease_controller.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+
+	if cronJob.Name == getCronJobName(etcd) && cronJob.DeletionTimestamp == nil {
+		logger.Info("Deleting cron job", "cronjob", kutil.ObjectName(cronJob))
+		if err := lc.Delete(ctx, cronJob, client.PropagationPolicy(metav1.DeletePropagationForeground)); client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{
+				RequeueAfter: 10 * time.Second,
+			}, fmt.Errorf("error while deleting compaction cron job: %v", err)
+		}
+	}
```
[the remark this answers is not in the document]
abdasgupta: I think, we should take care of the cronjob only during job reconciliation. We have completely decoupled cronjob/job from` etcd_controller.go`.

controllers/lease_controller.go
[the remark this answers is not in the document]
timuthy: I think it would be beneficial to have an indicator in the logs that tell about a compaction job failure because of an unknown provider.

main.go
```diff
@@ -141,6 +150,20 @@ func main() {
 
 	if err := etcdCopyBackupsTask.SetupWithManager(mgr, etcdCopyBackupsTaskWorkers); err != nil {
 		setupLog.Error(err, "Unable to create controller", "Controller", "EtcdCopyBackupsTask")
+	}
+
+	lc, err := controllers.NewLeaseControllerWithImageVector(mgr, controllersconfig.CompactionConfig{
+		EventsThreshold:        eventsThreshold,
+		ActiveDeadlineDuration: activeDeadlineDuration,
```
aaronfern, line 157: Shouldn't we also have to pass `enableCompactionJob` here?

aaronfern (changes requested):
Just a couple of things

As part of the backup-restore options, the following commands need to be added in `etcd-statefulset.yaml` if `enableCompactionJob` is set to true to allow br to update snapshot leases
```
- --enable-snapshot-lease-renewal=true
- --full-snapshot-lease-name={{ .Values.fullSnapLeaseName }}
- --delta-snapshot-lease-name={{ .Values.deltaSnapLeaseName }}
```

abdasgupta:
> enable-snapshot-lease-renewal

Please do it in a separate PR just after this PR is merged. Adding the said line to yaml won't just do, we have to modify the test case as well. I don't want to make any additional changes in this PR other than refactoring

controllers/lease_controller.go
abdasgupta: Done

controllers/lease_controller.go
```diff
@@ -0,0 +1,580 @@
...
+		logger.Info("Couldn't fetch delta snap lease because: " + err.Error())
+
+		return ctrl.Result{
+			RequeueAfter: 10 * time.Second,
+		}, err
+	}
+
+	// Run compaction job
+	if etcd.Spec.Backup.Store != nil && lc.config.EnableCompactionJob {
+		hi, err := strconv.ParseInt(*deltaLease.Spec.HolderIdentity, 10, 64)
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/config/compaction.go
```diff
@@ -0,0 +1,27 @@
...
+package config
+
+import "time"
+
+// CompactionConfig contains configuration for the compaction controller.
+type CompactionConfig struct {
+	// ActiveDeadlineDuration is the duration after which a running compaction job will be killed (Ex: "300ms", "20s", "-1.5h" or "2h45m")
+	ActiveDeadlineDuration time.Duration
+	// EnableCompactionJob enables compaction job to run in parallel
+	EnableCompactionJob bool
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/etcd_controller_test.go
```diff
@@ -29,26 +29,20 @@ import (
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 	"github.com/gardener/etcd-druid/pkg/common"
 	"github.com/gardener/etcd-druid/pkg/utils"
-	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
```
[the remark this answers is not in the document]
abdasgupta: Right. Whole big mess. I corrected it. Please check.

controllers/etcd_controller.go
```diff
@@ -801,138 +820,151 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
 
-	if err := decodeObject(renderedChart, cronJobPath, &decoded); err != nil {
-		return nil, err
-	}
+	return deltaLease, nil
+}
 
-	return decoded, nil
+func getDeltaLease(etcd *druidv1alpha1.Etcd) string {
+	return fmt.Sprintf("%s-delta-snap", string(etcd.UID[:6]))
```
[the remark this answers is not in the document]

controllers/etcd_controller.go
```diff
@@ -801,138 +820,151 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
-		logger.Info("cronjob precondition doesn't hold, skip updating it", "cronjob", kutil.Key(cjCopy.Namespace, cjCopy.Name).String())
-		err = nil
-	}
-	if err != nil {
-		return nil, err
-	}
-
-	return cjCopy, err
+func getFullLease(etcd *druidv1alpha1.Etcd) string {
+	return fmt.Sprintf("%s-full-snap", string(etcd.UID[:6]))
```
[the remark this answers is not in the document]

controllers/lease_controller.go
```diff
@@ -0,0 +1,580 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
+
+	// Calculate time elapsed since the cron job is scheduled
+	timeElapsed := time.Since(cronJob.Status.LastScheduleTime.Time).Seconds()
+	if timeElapsed > lc.config.ActiveDeadlineDuration.Seconds() {
+		err = lc.Delete(ctx, cronJob, client.PropagationPolicy(metav1.DeletePropagationForeground))
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/lease_controller.go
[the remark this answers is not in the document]
abdasgupta: Done

controllers/lease_controller.go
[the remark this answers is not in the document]
abdasgupta: Yes

main.go
```diff
@@ -73,6 +75,9 @@ func main() {
 	flag.IntVar(&custodianWorkers, "custodian-workers", 3, "Number of worker threads of the custodian controller.")
 	flag.IntVar(&etcdCopyBackupsTaskWorkers, "etcd-copy-backups-task-workers", 3, "Number of worker threads of the EtcdCopyBackupsTask controller.")
 	flag.DurationVar(&custodianSyncPeriod, "custodian-sync-period", 30*time.Second, "Sync period of the custodian controller.")
+	flag.BoolVar(&enableCompactionJob, "enable-compaction-job", true, "Enables compaction job to run in parallel. Default is true.")
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/etcd_controller.go
```diff
@@ -1287,6 +1091,40 @@ func (r *EtcdReconciler) removeFinalizersToDependantSecrets(ctx context.Context,
 	return nil
 }
 
+func (r *EtcdReconciler) removeLeases(ctx context.Context, etcd *druidv1alpha1.Etcd) error {
```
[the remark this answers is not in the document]
abdasgupta: Right

timuthy:
> Just a couple of things
> 
> As part of the backup-restore options, the following commands need to be added in `etcd-statefulset.yaml` if `enableCompactionJob` is set to true to allow br to update snapshot leases
> 
> ```
> - --enable-snapshot-lease-renewal=true
> - --full-snapshot-lease-name={{ .Values.fullSnapLeaseName }}
> - --delta-snapshot-lease-name={{ .Values.deltaSnapLeaseName }}
> ```

PR #254 does that already. Once this PR is merged, I can rebase #254 if that's ok for you.

controllers/lease_controller.go
```diff
@@ -0,0 +1,544 @@
...
+}
+
+// SetupWithManager sets up manager with a new controller and ec as the reconcile.Reconciler
+func (lc *LeaseController) SetupWithManager(mgr ctrl.Manager, workers int) error {
+	builder := ctrl.NewControllerManagedBy(mgr).WithOptions(controller.Options{
+		MaxConcurrentReconciles: workers,
+	})
+
+	return builder.
+		For(&druidv1alpha1.Etcd{}).
```
timuthy, line 541: This will result in reconciliations when there are updates (spec or status) to the `etcd` resource as well. IIUC, we don't need to reconcile at all if there are updates to `etcd` but only react to events from `lease`s, right? Maybe we can add a predicate here to skip enqueuing.

timuthy (changes requested):
Some final comments from my side. Otherwise it looks good. I'm also going to test it locally one more time.

controllers/lease_controller.go
```diff
@@ -0,0 +1,544 @@
...
+	job := &batchv1.Job{}
+	err := lc.Get(ctx, types.NamespacedName{Name: getJobName(etcd), Namespace: etcd.Namespace}, job)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return ctrl.Result{RequeueAfter: 10 * time.Second}, fmt.Errorf("error while fetching compaction job: %v", err)
+		}
+		return ctrl.Result{Requeue: false}, nil
+	}
+
+	if job.Name == getJobName(etcd) && job.DeletionTimestamp == nil {
```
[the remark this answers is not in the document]
abdasgupta: Right.  Problem is, `DeletionTimestamp` can be nil both for job which was found and job which was not found (i.e empty jobs. See the snippet before this call). So, `job.Name == getJobName(etcd)` seemed to be a good trick to find if the job was found. Otherwise, even if a job was not found, then also it was trying to enter into the `if` block and throwing error

controllers/lease_controller.go
abdasgupta: Should the predicate process creation, deletion events for leases? WDYT? I think no. Because the leases are created with values "0" in holderIdentity always

controllers/lease_controller.go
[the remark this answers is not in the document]
timuthy: Hm, I don't understand. When the job was not found, then the function is already left before in line [225](https://github.com/gardener/etcd-druid/pull/235/files/8f3d21beb570ef785d44340e6d55e5aceeb985aa#diff-a51e70a68174a40a7e3aa8b6bc4445c4327005a6734c13d07e0f615ed40765c6R225), right?

controllers/lease_controller.go
timuthy: `Create` events happen when the controller is started for all existing leases. Thus, I think it's a good idea that we react to `Create` events.

controllers/lease_controller.go
```diff
@@ -0,0 +1,544 @@
...
+		if err != nil {
+			logger.Error(err, "Can't convert holder identity of delta snap lease to integer")
+			return ctrl.Result{
+				RequeueAfter: 10 * time.Second,
+			}, err
+		}
+
+		diff := delta - full
+
+		// Reconcile job only when current revision is more than 1 million (this vale is configurable through `events-threshold` option in druid) than
```
shreyas-s-rao, line 160: Change the comment to something more readable, like `Reconcile job only when number of accumulated revisions over the last full snapshot is more than the configured threshold value via 'events-threshold' flag`

controllers/lease_controller.go
```diff
@@ -0,0 +1,544 @@
...
+}
+func getCmpctJobVolumeMounts(etcd *druidv1alpha1.Etcd, logger logr.Logger) []v1.VolumeMount {
+	vms := []v1.VolumeMount{
+		{
+			Name:      "etcd-workspace-dir",
+			MountPath: "/var/etcd/data",
+		},
+	}
+
+	if etcd.Spec.Backup.Store == nil {
```
shreyas-s-rao, line 341: When would this condition be true? Compaction job is reconciled only when `etcd.Spec.Backup.Store != nil` right? And we know that this field is technically immutable now, so it can't change mid-reconciliation.

controllers/lease_controller.go
```diff
@@ -0,0 +1,544 @@
...
+func getCompactJobCommands(etcd *druidv1alpha1.Etcd) []string {
+	command := []string{"" + "etcdbrctl"}
+	command = append(command, "compact")
+	command = append(command, "--data-dir=/var/etcd/data")
+	command = append(command, "--snapstore-temp-directory=/var/etcd/data/tmp")
+	command = append(command, "--enable-snapshot-lease-renewal=true")
+	command = append(command, "--full-snapshot-lease-name="+getFullLease(etcd))
+	command = append(command, "--delta-snapshot-lease-name="+getDeltaLease(etcd))
+
+	var quota int64 = 8 * 1024 * 1024 * 1024 // 8Gi
```
shreyas-s-rao, line 501: Can we fetch this value from a constant instead of configuring it here within the reconciliation logic?

shreyas-s-rao (dismissed):
@abdasgupta Thanks for the well-written PR! I had a few comments. Can you please address them? Also, please update the `compactor.md` in the documentation with the new events-threshold-based design. Thanks.

controllers/lease_controller.go
abdasgupta: It is to avoid any unintended nil pointer error due to any unknown circumstances(a particular memory address might be wiped out for unknown reason). It would even help to write future unit test cases where developer unknowingly changed the value to nil at some place else.

controllers/etcd_controller.go
```diff
@@ -287,10 +314,58 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
+	return cronJob, nil
+}
+
 func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (ctrl.Result, error) {
 	logger := r.logger.WithValues("etcd", kutil.Key(etcd.Namespace, etcd.Name).String(), "operation", "delete")
 	logger.Info("Starting operation")
 
+	// TODO(abdasgupta) : This is for backward compatibility towards ETCD-Druid 0.6.0. Remove it.
+	cronJob := &batchv1beta1.CronJob{}
+	if err := client.IgnoreNotFound(r.Get(ctx, types.NamespacedName{Name: getCronJobName(etcd), Namespace: etcd.Namespace}, cronJob)); err != nil {
```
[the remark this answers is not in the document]
abdasgupta: Actually, two serves different purpose and that's why written likewise. CleanCronJobs is for soft deletion of existing cron jobs. It may let some cron jobs run. But delete function is to preempt deletion for all cron jobs. I think it's okay to keep both of them for now. We will anyway remove all of them after two more releases. So maintenance cost should not be much higher even for duplicated codes here.

controllers/etcd_controller.go
```diff
@@ -287,10 +314,58 @@ func (r *EtcdReconciler) reconcile(ctx context.Context, etcd *druidv1alpha1.Etcd
...
+		if err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
+
+	// Calculate time elapsed since the cron job is scheduled
+	timeElapsed := time.Since(cronJob.Status.LastScheduleTime.Time).Seconds()
+	// Delete the cron job if it's running for more than 3 hours
+	if timeElapsed > time.Duration(3*time.Hour).Seconds() {
```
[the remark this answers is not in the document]
abdasgupta: Bringing `activeDeadlineDuration` from other controller needs more work. And as end user need not worry about older cron jobs anymore, I don't think it rational to give them the option of configuring when their older cron jobs would be deleted. I will use a constant in the field though.

controllers/controllers_suite_test.go
```diff
@@ -125,6 +125,18 @@ var _ = BeforeSuite(func(done Done) {
 	err = etcdCopyBackupsTaskReconciler.SetupWithManager(mgr, 1)
 	Expect(err).NotTo(HaveOccurred())
 
+	activeDeadlineDuration, err = time.ParseDuration("2m")
+	Expect(err).NotTo(HaveOccurred())
+
+	lc, err := NewLeaseControllerWithImageVector(mgr, controllersconfig.CompactionConfig{
+		EventsThreshold:        1000000,
```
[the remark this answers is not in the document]
abdasgupta: We are not creating delta events for unit test cases so far. We are just updating the lease holderIdentity value. So, any value is fine here for current usage.

controllers/controller_ref_manager.go
```diff
@@ -551,3 +523,247 @@ func CheckStatefulSet(etcd *druidv1alpha1.Etcd, statefulSet *appsv1.StatefulSet)
 
 	return nil
 }
+
+func getMapFromEtcd(im imagevector.ImageVector, etcd *druidv1alpha1.Etcd) (map[string]interface{}, error) {
```
stoyanr, line 527: Why has this function been moved to this file? As far as I can see, it's still used only in `etcd_controller.go` and so it belongs to that file (or to its own file, or even its own type, something like e.g. `ValuesProvider`), but definitely not here?

main.go
```diff
@@ -128,6 +132,11 @@ func main() {
 		SyncPeriod: custodianSyncPeriod,
 	})
 
+	if err != nil {
+		setupLog.Error(err, "Unable to initialize etcd custodian controller with image vector")
+		os.Exit(1)
+	}
```
stoyanr, line 138: This error checking here seems to be misplaced - `err` is not being assigned anything since the last check.

main.go
```diff
@@ -141,6 +150,20 @@ func main() {
 
 	if err := etcdCopyBackupsTask.SetupWithManager(mgr, etcdCopyBackupsTaskWorkers); err != nil {
 		setupLog.Error(err, "Unable to create controller", "Controller", "EtcdCopyBackupsTask")
+	}
+
+	lc, err := controllers.NewLeaseControllerWithImageVector(mgr, controllersconfig.CompactionConfig{
```
stoyanr, line 155: * We have "lease controller" but "compaction config". I would propose to harmonize the naming, that is use either "lease" or "compaction" (or perhaps `CompactionLease`), but not both.
* Shouldn't we be able to configure the number of workers for this controller, similarly to other controllers?

controllers/controllers_suite_test.go
```diff
@@ -116,6 +114,8 @@ var _ = BeforeSuite(func(done Done) {
 		},
 	})
 
+	Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
stoyanr: This error checking here seems to be misplaced - `err` is not being assigned anything since the last check. The same thing happens also in `main.go`. Perhaps it's a leftover from a previous state.

controllers/etcdcopybackupstask_controller.go
```diff
@@ -369,7 +369,7 @@ func getConditionType(jobConditionType batchv1.JobConditionType) druidv1alpha1.C
 	return ""
 }
 
-func getJobName(task *druidv1alpha1.EtcdCopyBackupsTask) string {
+func getCopyJobName(task *druidv1alpha1.EtcdCopyBackupsTask) string {
```
stoyanr, line 372: Let's use "copy backups" to be more concrete and stay aligned with other similar names.
```suggestion
func getCopyBackupsJobName(task *druidv1alpha1.EtcdCopyBackupsTask) string {
```

controllers/lease_controller.go
abdasgupta: Done

controllers/lease_controller.go
abdasgupta: Done

controllers/etcd_controller.go
```diff
@@ -801,138 +876,151 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
+				},
+				Spec: coordinationv1.LeaseSpec{
+					HolderIdentity: pointer.StringPtr("0"),
+					RenewTime:      &renewTime,
+				},
+			}
+			err2 := r.Create(ctx, deltaLease)
+			if err2 != nil {
+				logger.Error(err2, "Delta snap lease "+deltaSnapshotRevisions+" couldn't be created")
+				return nil, err2
```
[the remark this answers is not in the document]

controllers/etcd_controller.go
```diff
@@ -801,138 +876,151 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
-	if err == errorsutil.ErrPreconditionViolated {
-		logger.Info("cronjob precondition doesn't hold, skip updating it", "cronjob", kutil.Key(cjCopy.Namespace, cjCopy.Name).String())
-		err = nil
-	}
-	if err != nil {
-		return nil, err
-	}
-
-	return cjCopy, err
+func getFullLease(etcd *druidv1alpha1.Etcd) string {
```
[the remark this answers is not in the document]

controllers/etcd_controller.go
```diff
@@ -801,138 +876,151 @@ func (r *EtcdReconciler) getStatefulSetFromEtcd(etcd *druidv1alpha1.Etcd, values
...
+				return nil, err2
 			}
-
-			if foundEtcd.UID != etcd.UID {
-				return nil, fmt.Errorf("original %v/%v etcd gone: got uid %v, wanted %v", etcd.Namespace, etcd.Name, foundEtcd.UID, etcd.UID)
+		} else {
+			return nil, err1
+		}
+	}
+	if !checkEtcdOwnerReference(fullLease.GetOwnerReferences(), etcd) {
```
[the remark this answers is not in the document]
abdasgupta: Not needed. Removed

controllers/controllers_suite_test.go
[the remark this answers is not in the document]
abdasgupta: Removed

main.go
```diff
@@ -73,6 +75,9 @@ func main() {
 	flag.IntVar(&custodianWorkers, "custodian-workers", 3, "Number of worker threads of the custodian controller.")
 	flag.IntVar(&etcdCopyBackupsTaskWorkers, "etcd-copy-backups-task-workers", 3, "Number of worker threads of the EtcdCopyBackupsTask controller.")
 	flag.DurationVar(&custodianSyncPeriod, "custodian-sync-period", 30*time.Second, "Sync period of the custodian controller.")
+	flag.IntVar(&compactionWorkers, "enable-compaction-job", 1, "Number of worker threads of the CompactionJob controller.")
+	flag.Int64Var(&eventsThreshold, "events-threshold", 1000000, "Total number of events that can be allowed before a compaction job is triggered")
+	flag.DurationVar(&activeDeadlineDuration, "active-deadline-duration", 3*time.Hour, "Duration after which a running compaction job will be killed (Ex: \"300ms\", \"20s\", \"-1.5h\" or \"2h45m\")")
```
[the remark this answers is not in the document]
abdasgupta: Done

main.go
```diff
@@ -73,6 +75,9 @@ func main() {
 	flag.IntVar(&custodianWorkers, "custodian-workers", 3, "Number of worker threads of the custodian controller.")
 	flag.IntVar(&etcdCopyBackupsTaskWorkers, "etcd-copy-backups-task-workers", 3, "Number of worker threads of the EtcdCopyBackupsTask controller.")
 	flag.DurationVar(&custodianSyncPeriod, "custodian-sync-period", 30*time.Second, "Sync period of the custodian controller.")
+	flag.IntVar(&compactionWorkers, "enable-compaction-job", 1, "Number of worker threads of the CompactionJob controller.")
+	flag.Int64Var(&eventsThreshold, "events-threshold", 1000000, "Total number of events that can be allowed before a compaction job is triggered")
```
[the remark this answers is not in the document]
abdasgupta: Done

main.go
```diff
@@ -73,6 +75,9 @@ func main() {
 	flag.IntVar(&custodianWorkers, "custodian-workers", 3, "Number of worker threads of the custodian controller.")
 	flag.IntVar(&etcdCopyBackupsTaskWorkers, "etcd-copy-backups-task-workers", 3, "Number of worker threads of the EtcdCopyBackupsTask controller.")
 	flag.DurationVar(&custodianSyncPeriod, "custodian-sync-period", 30*time.Second, "Sync period of the custodian controller.")
+	flag.IntVar(&compactionWorkers, "enable-compaction-job", 1, "Number of worker threads of the CompactionJob controller.")
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/lease_controller.go
abdasgupta: Done

controllers/lease_controller.go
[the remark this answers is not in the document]
abdasgupta: Sorry! My bad. I misunderstood something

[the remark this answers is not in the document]
abdasgupta: Removed

controllers/lease_controller.go
```diff
@@ -0,0 +1,544 @@
...
+	if job.Status.Failed > 0 {
+		err = lc.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground))
+		if err != nil {
+			return ctrl.Result{
+				RequeueAfter: 10 * time.Second,
+			}, fmt.Errorf("error while deleting failed compaction job: %v", err)
+		}
+		return ctrl.Result{
+			RequeueAfter: 10 * time.Second,
+		}, fmt.Errorf("existing job status failed")
```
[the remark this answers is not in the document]
abdasgupta: Yes

controllers/lease_controller.go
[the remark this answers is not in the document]
abdasgupta: Done

controllers/lease_controller_test.go
```diff
@@ -0,0 +1,1004 @@
...
+			// The active job should exist
+			Eventually(func() error { return jobIsCorrectlyReconciled(c, instance, j) }, timeout, pollingInterval).Should(BeNil())
+		})
+
+		AfterEach(func() {
+			Expect(c.Delete(context.TODO(), instance)).To(Succeed())
+			Eventually(func() error { return etcdRemoved(c, instance) }, timeout, pollingInterval).Should(BeNil())
+		})
+	})
+	Context("when an existing cron job from older version is already present", func() {
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/lease_controller_test.go
```diff
@@ -0,0 +1,1004 @@
...
+			err = kutil.TryUpdate(context.TODO(), retry.DefaultBackoff, c, deltaLease, func() error {
+				deltaLease.Spec.HolderIdentity = pointer.StringPtr("1000000")
+				renewedTime := time.Now()
+				deltaLease.Spec.RenewTime = &metav1.MicroTime{Time: renewedTime}
+				return nil
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			//Instead of failed one a new job should be created
+			Eventually(func() error { return jobIsCorrectlyReconciled(c, instance, j) }, timeout, pollingInterval).Should(BeNil())
```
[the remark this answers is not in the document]
abdasgupta: Changed

controllers/lease_controller.go
```diff
@@ -0,0 +1,544 @@
...
+	lc.ImageVector = imageVector
+	return lc, nil
+}
+
+// +kubebuilder:rbac:groups=druid.gardener.cloud,resources=etcds,verbs=get;list;watch
+// +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;create;list;watch;update;patch;delete
+
+// Reconcile reconciles the compaction job.
+func (lc *LeaseController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
stoyanr, line 87: I find it a bit difficult to read and understand this controller, e.g. whether cases such as reconciling a creation or update and reconciling a deletion of an object are handled appropriately, whether the status is always correctly maintained, etc. With its current "monolithic" implementation, it would also be rather difficult to cover this implementation with unit tests, rather than with envtest, and I understand this is something we might want to do in the future.

I would therefore like to make a general suggestion. I think it's very useful to structure controllers in terms of "reconciler" and "actuator". This is an approach that is already used in g/g and extensions because of its "separation of concerns" and testability advantages.
* The "reconciler" is responsible for reading the object that is being reconciled, managing finalizers, annotations, and status updates, and making the distinction between reconciling a creation / update (what we call "reconcile") and reconciling a deletion (what we call "delete"). Reconciliation logic is usually similar / generic and so often the same reconciler can be used with different actuators - but even if this is not the case, the distinction is still useful.
* The "actuator" is responsible for the actual reconciliation / deletion logic and returning an appropriately updated status (but not actually updating it).

I used this approach in the `EtcdCopyBackupsTaskController` though I stopped short of actually extracting an actuator, so the actuator is "hidden" in the `doReconcile` and `doDelete` methods. It would be however easy to extract these methods to a separate type implementing an `Actuator` interface, and then test both the (much simpler now) `Reconciler` and `Actuator` in isolation from each other using unit tests with generated mocks.

In addition, whenever an actuator needs to produce a set of values for a chart, it could make use of a `ValuesProvider` interface and implementation that can again be further tested in isolation, since the logic for producing the correct values is often one of the most important pieces. Other helper types can be introduced as well when it makes sense for similar reasons.

As further examples of this approach, please see our provider extensions as well as the `ManagedSeed` and `ManagedSeedSet` controllers in gardener:
* https://github.com/gardener/gardener/tree/master/pkg/gardenlet/controller/managedseed
* https://github.com/gardener/gardener/tree/master/pkg/controllermanager/controller/managedseedset

stoyanr (changes requested):
I have a few small comments and one general suggestion that I believe should be at least partially addressed if we are seriously thinking about making our controllers more modular and testable.

main.go
abdasgupta: I didn't understand the second point. `if err := lc.SetupWithManager(mgr, compactionWorkers); err != nil {` is there at line no ~160. The `compactionWorkers` is configurable. Am I missing something ?

abdasgupta: Hmm.. For the first point, we introduced lease controller because it will handle all the leases (like member leases in future for multi node ETCD). So we should name our controller as lease controller. But the configs are currently only for compaction job. But in future, it may hold config for other things that are dependent on lease. We can change the naming of this config. Probably, `leaseConfig` ?

timuthy: > For the first point, we introduced lease controller because it will handle all the leases (like member leases in future for multi node ETCD)

I'm a little bit surprised, TBH. Why would we do we want to do that?

main.go
```diff
@@ -75,9 +75,9 @@ func main() {
 	flag.IntVar(&custodianWorkers, "custodian-workers", 3, "Number of worker threads of the custodian controller.")
 	flag.IntVar(&etcdCopyBackupsTaskWorkers, "etcd-copy-backups-task-workers", 3, "Number of worker threads of the EtcdCopyBackupsTask controller.")
 	flag.DurationVar(&custodianSyncPeriod, "custodian-sync-period", 30*time.Second, "Sync period of the custodian controller.")
-	flag.IntVar(&compactionWorkers, "enable-compaction-job", 1, "Number of worker threads of the CompactionJob controller.")
-	flag.Int64Var(&eventsThreshold, "events-threshold", 1000000, "Total number of events that can be allowed before a compaction job is triggered")
-	flag.DurationVar(&activeDeadlineDuration, "active-deadline-duration", 3*time.Hour, "Duration after which a running compaction job will be killed (Ex: \"300ms\", \"20s\", \"-1.5h\" or \"2h45m\")")
+	flag.IntVar(&compactionWorkers, "compaction-Workers", 3, "Number of worker threads of the CompactionJob controller. CompactionJob controller runs backup compaction job in parallel. CompactionJob controller can be disabled if this flag is set to 0.")
```
timuthy, line 78:
```suggestion
      flag.IntVar(&compactionWorkers, "compaction-workers", 3, "Number of worker threads of the CompactionJob controller. The controller creates a backup compaction job if a certain etcd event threshold is reached. Setting this flag to 0 disabled the controller.")
```

pkg/predicate/predicate_test.go
```diff
@@ -108,6 +111,66 @@ var _ = Describe("Druid Predicate", func() {
...
+	Describe("#Lease", func() {
+		var pred predicate.Predicate
+
+		JustBeforeEach(func() {
+			pred = LeaseHolderIdentityChange()
+		})
+
+		Context("when holder identity matches", func() {
+			BeforeEach(func() {
+				renewTime := metav1.NewMicroTime(time.Now())
```
timuthy, line 123: No need to set `RenewTime` in these tests.

timuthy (changes requested):
Just some last comments. Thanks for your effort 👍

main.go
stoyanr: > I didn't understand the second point. if err := lc.SetupWithManager(mgr, compactionWorkers); err != nil { is there at line no ~160. The compactionWorkers is configurable. Am I missing something ?

You are right, everything is ok actually. I was again misled by the naming mismatch between "lease" and "compaction", so it wasn't immediately obvious that `compactionWorkers` is actually used with the "lease controller". Hopefully this could be improved.

abdasgupta: Done

controllers/etcdcopybackupstask_controller.go
abdasgupta: Done

timuthy (approved):
/lgtm
I tested and played around with it w/o facing any issues 👍

controllers/compaction_lease_controller.go
```diff
@@ -0,0 +1,546 @@
...
+const DefaultETCDQuota = 8 * 1024 * 1024 * 1024 // 8Gi
+// CompactionLeaseController reconciles compaction job
+type CompactionLeaseController struct {
+	client.Client
+	logger      logr.Logger
+	ImageVector imagevector.ImageVector
+	config      controllersconfig.CompactionLeaseConfig
+}
+
+// NewCompactionLeaseController creates a new LeaseController object
```
stoyanr, line 56:
```suggestion
// NewCompactionLeaseController creates a new CompactionLeaseController object
```

controllers/compaction_lease_controller.go
```diff
@@ -0,0 +1,546 @@
...
+func NewCompactionLeaseController(mgr manager.Manager, config controllersconfig.CompactionLeaseConfig) *CompactionLeaseController {
+	return &CompactionLeaseController{
+		Client: mgr.GetClient(),
+		logger: log.Log.WithName("compaction-lease-controller"),
+		config: config,
+	}
+}
+
+// NewCompactionLeaseControllerWithImageVector creates a new LeaseController object
```
stoyanr, line 65:
```suggestion
// NewCompactionLeaseControllerWithImageVector creates a new CompactionLeaseController object
```

controllers/compaction_lease_controller.go
```diff
@@ -0,0 +1,546 @@
...
+func NewCompactionLeaseControllerWithImageVector(mgr manager.Manager, config controllersconfig.CompactionLeaseConfig) (*CompactionLeaseController, error) {
+	lc := NewCompactionLeaseController(mgr, config)
+	return lc.InitializeControllerWithImageVector()
+}
+
+// InitializeControllerWithImageVector will use LeaseController client to initialize image vector for etcd
```
stoyanr, line 71:
```suggestion
// InitializeControllerWithImageVector will use CompactionLeaseController client to initialize image vector for etcd
```

stoyanr (approved):
/lgtm 

There are a few comments that still use the old name `LeaseController` that you may want to fix.

aaronfern (approved):
/lgtm

timuthy (approved):
/lgtm

controllers/lease_controller.go
unmarshall: Hello @stoyanr,
Thanks for providing details on how g/g is structuring their code.

IMHO there is no clear distinction between actuator and reconciler in the example that you have provided (e.g ManagedSeed). There are recent PRs which also move away from actuator for `BackupEntry`, `BackupBucket` etc. (https://github.com/gardener/gardener/pull/6926) 

Actuator is useful when you have common code in the reconciler that one wishes to re-use and then delegate the responsibility to reconcile different resources to resource specific actuators. IMO it should not be generally applied. The point of testability is important and for that writing smaller functions with one clear purpose and single responsibility are usually sufficient unless there is a violation of DRY where additional interfaces like `Actuator` can be thought of. Having said that we will keep your suggestions in mind.

PR #246, opened 2021-10-29
Updated the MultiNode proposal: To take out-of-sync snapshots delegate the incoming http request to backup-leader.

ishan16696:
This PR updated the MultiNode proposal:
To take [out-of-sync  full/delta snapshots](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/httpAPI.go#L103-L105) delegate the incoming http request to `backup-restore leader` and Let the `backup-restore leader` take/upload the out-of-sync snapshot(full as well as incremental)
I had already created a issue regarding this: https://github.com/gardener/etcd-backup-restore/issues/354

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/docs/proposals/multi-node/README.md
+++ b/docs/proposals/multi-node/README.md
@@ -1242,7 +1242,7 @@ The life-cycle of these work-flows is shown below.
 
 ### Work-flows independent of leader election in all members
 
-- Serve the HTTP API that all members are expected to support currently
+- Serve the [HTTP API](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/httpAPI.go#L101-L107) that all members are expected to support currently but some HTTP API call which are used to take [out-of-sync delta or full snapshot](https://github.com/gardener/etcd-backup-restore/blob/5dfcc1f848a9f325d41a24eae4defb70d997c215/pkg/server/httpAPI.go#L103-L105) should delegate the incoming HTTP requests to the `leading-sidecar` and one of the possible approach to achieve this is via an [HTTP reverse proxy](https://pkg.go.dev/net/http/httputil#ReverseProxy.ServeHTTP).
 - Check the health of the respective etcd member and renew the corresponding [member `lease`](#member-leases).
 
 ### Work-flows only on the leading member


timuthy (commented):
~I'd suggest to first discuss details in gardener/etcd-backup-restore#354 before we move on and merge this PR.~

docs/proposals/multi-node/README.md
```diff
@@ -1242,7 +1242,7 @@ The life-cycle of these work-flows is shown below.
 
 ### Work-flows independent of leader election in all members
 
-- Serve the HTTP API that all members are expected to support currently
+- Serve the [HTTP API](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/httpAPI.go#L101-L107) that all members are expected to support currently but HTTP API call which are used to take [out-of-sync delta or full snapshot](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/httpAPI.go#L103-L105) should delegate the incoming HTTP requests to the `leading-sidecar`.
```
timuthy, line 1245:
```suggestion
- Serve the [HTTP API](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/server/httpAPI.go#L101-L107) that all members are expected to support currently but HTTP API call which are used to take [out-of-sync delta or full snapshot](https://github.com/gardener/etcd-backup-restore/blob/5dfcc1f848a9f325d41a24eae4defb70d997c215/pkg/server/httpAPI.go#L103-L105) should delegate the incoming HTTP requests to the `leading-sidecar` via an HTTP reverse proxy.
```

ishan16696: > via an HTTP reverse proxy

This is an implementation detail and I don’t think we should mention this in doc. The design document shouldn’t recommend the implementation method unless it is the only method available to achieve the functionality. WDYT?

timuthy: Fair enough, but can you add it as an example -> (e.g. via an HTTP reverse proxy)? The problem today is that we don't have a dedicated documentation about multi-node etcd, only the proposal as a reference document. I think it's beneficial for every interested party to know more details w/o looking at the code.

ishan16696: make sense,I will mention this as one of the possible approach.

ishan16696: done

PR #248, opened 2021-10-29
Use g/g chart applier

timuthy:

/area dev-productivity
/kind cleanup

This PR removes the `chartrenderer` and `applier` implementation from the Druid code as these were solely copied from `gardener/gardener`. Instead, we can directly use the renderer and applier from the `gardener/gardener` dependency.

**Release note**:

```other operator
NONE
```


timuthy:
@aaronfern, @abdasgupta or @ishan16696 kindly asking for a review. Thanks 🙂

aaronfern (approved):
Thanks for the PR @timuthy 
/lgtm

PR #250, opened 2021-11-02
Deploy/maintain the correct PodDisruptionBudget configuration according to the Etcd resource status conditions

### LINKED ISSUE #165: [Feature] Deploy/maintain the correct `PodDisruptionBudget` configuration according to the Etcd resource status conditions

Deploy/maintain the correct [`PodDisruptionBudget`](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#poddisruptionbudget) configuration according to the `Etcd` resource [status](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#status) [conditions](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#conditions).

Pick individually executable pieces of the [multi-node proposal](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md).

The deployment of `PodDisruptionBudget` resource is probably best done in the main controller and the dynamic modification of the resource based on the `Etcd` resource `status` is best done in the [custodian controller](https://github.com/gardener/etcd-druid/pull/133).

It is probably better to deploy the `PodDisruptionBudget` only for the multi-node case (`spec.replicas > 1`) because deploying it for the single-node case might block node drain.

breuerfelix:

/area high-availability
/kind enhancement

It is needed for Multi Node etcd deployment as described [here](https://github.com/breuerfelix/etcd-druid/tree/master/docs/proposals/multi-node#poddisruptionbudget)

Fixes #165 

I haven't written any tests so far since it isn't allowed to specify replicas > 1 anyways currently. I guess it is best to write tests for etcd HA once all aspects of the GEP are implemented.
/edit Questions below already answered.
This PR is currently work in progress / blocked since i would like to get some opinions from maintainers in order to beautify the PR and write some tests.  
1. `etcd.Status.Members` never gets set on current `master` because the `Checker` does not return `true` even though all replicas are `ready`. This leads to `etcd.Status.clusterSize` being stuck at 1. Is this a fault on my side or is this feature just not implemented yet?
2. Instead of creating / deleting the `PodDisruptionBudget` when switching between single and multi-node, i just set `minAvailable` to `0` for single node etcd. What do you think about this approach?
3. The Issue recommends to implement the logic for `minAvailable` in the custodian controller but this controller only ever updates `status` fields of objects whereas `minAvailable` is part of the pdb `Spec` so i implemented this in the `etcd_controller`. It just seemed to be the better fit in my opinion. Thoughts?

**Release note**:
```other operator
feature operator
Deploying and maintaining the correct PodDisruptionBudget configuration according to Etcd resource status
```


gardener-robot-ci-3:
Thank you @breuerfelix for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

timuthy:
Thanks a lot for you PR @breuerfelix. Already some comments about you concerns:

> etcd.Status.Members never gets set on current master because the Checker does not return true even though all replicas are ready. This leads to etcd.Status.clusterSize being stuck at 1. Is this a fault on my side or is this feature just not implemented yet?

This is the case because there still no member `Lease` objects created in the cluster. https://github.com/gardener/etcd-druid/issues/158 tackles this requirement and since the latest release of etcd-backup-restore, I can continue working on that item.

> Instead of creating / deleting the PodDisruptionBudget when switching between single and multi-node, i just set minAvailable to 0 for single node etcd. What do you think about this approach?

Generally fine for me even though it makes Druid to only work with Kubernetes >= 1.15 ([ref](https://github.com/kubernetes/kubernetes/pull/69867)). We should document this restriction somewhere.

> The Issue recommends to implement the logic for minAvailable in the custodian controller but this controller only ever updates status fields of objects whereas minAvailable is part of the pdb Spec so i implemented this in the etcd_controller. It just seemed to be the better fit in my opinion. Thoughts?

👍 makes sense

breuerfelix:
@timuthy Thanks for your input! That resolved all of my questions :)  
I refactored the logic so the PDB gets only deployed in multi node deployments (as earlier suggested) so single node instance work for < 1.15 as before aswell.  
Merging this PR won't change anything to current deployments since replicas are fixed to `1` which won't trigger a PDB deployment.

Let me know if something needs to be changed in order for this PR to get merged :)

controllers/etcd_controller.go
```diff
@@ -588,6 +594,151 @@ const (
...
+			condition.Status == druidv1alpha1.ConditionStatus(druidv1alpha1.EtcdMemberStatusReady) {
+			allMembersReady = true
+			continue
+		}
+
+		if condition.Type == druidv1alpha1.ConditionTypeBackupReady &&
+			condition.Status == druidv1alpha1.ConditionTrue {
+			backupReady = true
+			continue
+		}
```
timuthy, line 616: This is another use-case why the custodian_controller ought to patch the PDB as well. As soon as the condition change, the proper minAvailable value must be updated.

charts/etcd/values.yaml
```diff
@@ -3,6 +3,10 @@ uid: uuid-of-etcd-resource
 serviceName: test
 configMapName: test
 
+pdb:
+  name: test-pod-disruption-budget
```
timuthy, line 7: Any reason why we need a different name? Otherwise, I propose to to remove this and use `name` instead.

controllers/etcd_controller.go
```diff
@@ -588,6 +594,151 @@ const (
 	noOp        operationResult = "none"
 )
 
+func calculatePDBminAvailable(etcd *druidv1alpha1.Etcd) int32 {
```
timuthy, line 597: Can you add tests for this function?

timuthy (changes requested):
Thanks for you PR! Overall, it looks very good and I really appreciate that you closely adhered to the multi-node etcd proposal. However, there are two use cases that probably need an implementation in the `custodian_controller` as mentioned in the inline comments.

timuthy:
Very valid questions @breuerfelix, thanks for bringing them up.

> Isn't it enough to sync minAvailable in the etcd_controller? If the Status of the etcd changes (ie switch from inBootstrap to ready), the etcd_controller also gets triggered which leads to updating the pdb aswell or am i wrong?

`etcd_controller` only reconciles `etcd` resources in specific cases. You can take the configured predicates [here](https://github.com/gardener/etcd-druid/blob/265052b6ece30459de180e43a78dccd9d290434d/controllers/etcd_controller.go#L1661) as a reference. 

> So overall you suggest to deploy the PDB with minAvailable: 0 in the etcd_controller (in single and multi node deployments) and remove the whole pdb syncing (since the only thing that can change is the minAvailable value) from etcd_controller and move the updating of minAvailable to etcd_custodian_controller?
That would also remove the need to use optimistic locking since only the etcd_custiodian_controller can update the spec.

Please note that operations like this need to be idempotent, i.e. `etcd_controller` must not overwrite the `minAvailable` configuration set by `etcd_custodian_controller`. I think the `etcd_controller` should cover the following cases:
1. PDB does not exist
- Create a new PDB with `0` in case `status.clusterSize` is not set yet.
- Create a new PDB with `status.clusterSize/2 + 1`.
- Create a new PDB with `status.clusterSize` if backup condition is false.

2. PDB exists
- Patch PDB with existing `minAvailable` value.

> Since this is my first touch with the etcd-druid codebase, i am still unsure about the difference of etcd_controller and etcd_custodian_controller and what "task" belongs to which controller. Is there a resource i can consume to understand the controllers a little more?

The `etcd_controller` is mainly responsible for applying all necessary resources (`StatefulSet`, `Service`, `ConfigMap`, etc.) to match the desired spec given in the `etcd` resource. Cleaning-up is also done by this controller if the `etcd` resource is deleted.

The `etcd_custodian_controller` in contrast reacts on **events of the etcd cluster** itself like maintaining `etcd.status.members` or `etcd.status.conditions`. These cluster events usually happen outside the common reconciliation window  the `etcd_controller`. A dynamic `minAvailable` that is based on these cluster events, ought to be set in the Custodian as well.

breuerfelix:
@timuthy Thanks for answering the questions again!  
I updated my PR so the `minAvailable` value gets updated by etcd controller and custodian controller in a way, that conflicts are resolved if they both edit at the same time.

I moved the calculation based on conditions to the `custodian` controller and the calculations based on `clusterSize` to the `etcd-controller`.

Looking forward to your Feedback and Review!

timuthy (changes requested):
> I updated my PR so the minAvailable value gets updated by etcd controller and custodian controller in a way, that conflicts are resolved if they both edit at the same time.

Thanks for the update. Looking at the logic, I'd say that we don't need to resolve conflicts by maintaining annotations when we do the following:

- etcd_controller only creates the `PDB` with:
```
minAvailable: 0                   # if etcd.spec.replicas <= 1
minAvailable: etcd.spec.replicas  #otherwise
```
- etcd_controller claims `PDB` which is the only update it does for this object.
- etcd_druid_controller updates `minAvailable` in all remaining use-cases described in the mutli-node proposal. Not found errors are handled gracefully.

Does that make sense?

breuerfelix:
@timuthy makes sense :)

I updated the PR to match your review! Let me know what you think :)

controllers/etcd_controller.go
```diff
@@ -1121,6 +1183,13 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
 		return nil, nil, err
 	}
 
+	{
```
timuthy, line 1186:
```suggestion

```

controllers/etcd_controller.go
```diff
@@ -1121,6 +1183,13 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
...
+		_, err := r.reconcilePodDisruptionBudget(ctx, logger, etcd, values)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
```
timuthy, line 1191:
```suggestion

```

controllers/etcd_controller.go
```diff
@@ -42,6 +42,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	eventsv1 "k8s.io/api/events/v1"
 	eventsv1beta1 "k8s.io/api/events/v1beta1"
+	policyv1 "k8s.io/api/policy/v1beta1"
```
timuthy, line 45:
```suggestion
	policyv1beta1 "k8s.io/api/policy/v1beta1"
```

controllers/controller_ref_manager.go
```diff
@@ -33,6 +33,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1beta1"
```
timuthy, line 36:
```suggestion
	policyv1beta1 "k8s.io/api/policy/v1beta1"
```

charts/etcd/templates/etcd-poddisruptionbudget.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: policy/v1
```
timuthy, line 1:
```suggestion
apiVersion: policy/v1beta1
```
PDBs were only graduated in Kubernetes 1.21 ([ref](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125)).

controllers/etcd_custodian_controller.go
```diff
@@ -168,6 +176,92 @@ func (ec *EtcdCustodian) updateEtcdStatus(ctx context.Context, logger logr.Logge
...
+
+	if err != nil {
+		return err
+	}
+
+	logger.Info("Claiming pdb object")
+	claimedPdb, err := refMgr.ClaimPodDisruptionBudget(ctx, pdb)
+	if err != nil {
+		return err
+	}
```
timuthy, line 250:
```suggestion
```
You can drop this, as this is already done by the `etcd_controller`.

controllers/etcd_custodian_controller.go
```diff
@@ -168,6 +176,92 @@ func (ec *EtcdCustodian) updateEtcdStatus(ctx context.Context, logger logr.Logge
...
+	minAvailable := calculatePDBminAvailable(etcd)
+	if claimedPdb.Spec.MinAvailable.IntValue() == minAvailable {
+		// do not update, nothing changed
+		return nil
+	}
+
+	// update fields
+	converted := intstr.FromInt(minAvailable)
+	claimedPdb.Spec.MinAvailable = &converted
+	return ec.Update(ctx, claimedPdb)
```
timuthy, line 262: Can you use `patch` instead? Please find an example [here](https://github.com/gardener/etcd-druid/blob/f1cd8a82acde5d4c26d52448424f453d5b689019/controllers/etcd_controller.go#L499-L514 (you can leave out the `RetryOnConflict` part).

controllers/etcd_custodian_controller.go
```diff
@@ -168,6 +176,92 @@ func (ec *EtcdCustodian) updateEtcdStatus(ctx context.Context, logger logr.Logge
 	})
 }
 
+func calculatePDBminAvailable(etcd *druidv1alpha1.Etcd) int {
```
timuthy, line 179: Since we have many cases here it'd be awesome to have proper unit tests for this function.

controllers/etcd_custodian_controller.go
```diff
@@ -168,6 +176,92 @@ func (ec *EtcdCustodian) updateEtcdStatus(ctx context.Context, logger logr.Logge
...
+
+		if condition.Type == druidv1alpha1.ConditionTypeBackupReady &&
+			condition.Status == druidv1alpha1.ConditionTrue {
+			backupReady = true
+			continue
+		}
+	}
+
+	clusterSize := int(*etcd.Status.ClusterSize)
+	values := make([]int, int(math.Floor(float64(clusterSize)/float64(2))+1))
```
timuthy, line 207: Hm, you might want to check this line 👀 I rather expected `int(math.Floor(float64(clusterSize)/float64(2))+1)` to be added to the values slice but instead it's used as `len`?

timuthy (changes requested):
Thanks for the changes which look good in general. I found some remaining points which I've commented.

controllers/etcd_custodian_controller.go
breuerfelix: since the implementation is in the correct spot now, i will write some tests for this :)

breuerfelix:
@timuthy updated :)  
Let me know what you think or if i should write some more test cases.

controllers/etcd_custodian_controller.go
```diff
@@ -168,6 +176,89 @@ func (ec *EtcdCustodian) updateEtcdStatus(ctx context.Context, logger logr.Logge
...
 
+func calculatePDBminAvailable(etcd *druidv1alpha1.Etcd) int {
+	// do not enable in single mode
+	if etcd.Spec.Replicas < 2 {
+		return 0
+	}
+
+	// cluster is in bootstrap
+	if etcd.Status.ClusterSize == nil {
+		return 0
```
timuthy, line 187: We should not return `0` here because Custodian can accidentally reset `minAvailable: 0` when bootstrapping is happening and the value should actually be `int(etcd.Spec.Replicas)`. I suggest to do nothing in case of bootstrapping and leave it up to the `etcd_controller` to set the correct value.

timuthy (changes requested):
Looks very good except for one case I left comment about 👍

breuerfelix:
@timuthy i thought about also returning an error in the calculatePDB function but in case of the bootstrapping error we would simple do nothing and not escalate that error to the reconciler.  
Other (future) errors might be escalated which would lead to checking which could end up in a switch statement for error checking.  
i hope the implementation with negative values which mean "do nothing" is fine since `pdbminavailable` must be positive anyways.

timuthy (approved):
/lgtm
Thanks a lot 🙂

Issue #147, opened 2021-03-10
[Feature] Failed backups to not block incoming traffic and trigger high prio alert instead


amshuman-kr:
**Feature (What you would like to be added):**
Currently, the health check of the etcd pods is linked to the backup health (last backup upload succeeded) in addition to just etcd health. But as long as etcd data is backed by persistent volumes (it is now), we can afford for etcd to continue serve income requests even when backup upload fails as long as high priority alerts are triggered when backup upload fails and follow up is done to resolve the issue.

**Motivation (Why is this needed?):**
Avoid bringing down the whole shoot cluster control-plane when backup upload fails as that basically brings the cluster to a grinding halt. This might be affordable if etcd data is backed by persistent volumes because for data loss to occur a further data corruption in the persistent volumes is required (while backup upload is failing) to cause a data loss. 

See also https://github.tools.sap/kubernetes-canary/issues-canary/issues/599

**Approach/Hint to the implement solution (optional):**
The following tasks might have to checked/evaluated.

- [x] Trigger high priority alert when backup upload fails (for full or incremental backups) https://github.com/gardener/gardener/pull/4094
- [x] Change the health check criteria for the readinessProbe in the etcd pods to not take backup health (last backup upload) into consideration https://github.com/gardener/etcd-backup-restore/pull/411
- [ ] ~Enhance the [multi-node etcd proposal](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md) to address this new requirement.~

vlerenc:
As we have seen lately with the off-by-one (32/33 chunks) on GCP, doesn't it make sense to give this one a higher prio @amshuman-kr?

ishan16696:
> But as long as etcd data is backed by persistent volumes (it is now), we can afford for etcd to continue serve income requests even when backup upload fails

Just wanted to mention one point: In [multi-node etcd](https://github.com/gardener/etcd-druid/tree/master/docs/proposals/multi-node) we have a plan to use [Ephemeral volume](https://github.com/gardener/etcd-druid/tree/master/docs/proposals/multi-node#ephemeral), if we choose to go with Ephemeral volume then we might lose the data in worst case scenario as data wasn’t backed up.

timuthy:
> > But as long as etcd data is backed by persistent volumes (it is now), we can afford for etcd to continue serve income requests even when backup upload fails
> 
[...]

PR #256, opened 2021-11-08
Move ephemeral volumes to follow-up

timuthy:

/area documentation
/kind epic

As discussed in https://github.com/gardener/etcd-druid/issues/147#issuecomment-961918646 we consider the ephemeral volume use-case being a follow-up item.

**Release note**:

```doc operator
The multi-node etcd proposal has been updated and now considers ephemeral volumes not being an integral part of the multi-node feature set.
```


ishan16696 (approved):
LGTM!!

PR #260, opened 2021-11-17
Fix `etcd.status.clusterSize` update

### LINKED ISSUE #255: [BUG] Cluster size calculation in bootstrap case is buggy
**Describe the bug:**
The `etcd.status.cluster.clusterSize` remains empty if the `etcd` creation failed or is interrupted after the `StatefulSet` was created. As a result, the `Ready` condition cannot be calculated:

```
    - lastTransitionTime: "2021-11-05T07:39:42Z"
      lastUpdateTime: "2021-11-05T11:14:17Z"
      message: Cannot determine readiness of cluster since no cluster size has been
        calculated
      reason: ClusterSizeUnknown
      status: Unknown
      type: Ready
```

**Expected behavior:**
`etcd.status.cluster.clusterSize` should be set.

 - Etcd-druid version/commit ID :  222a1cc0030846cc137c2aabfd4250cc5172f4ad

timuthy:

/area robustness
/kind bug

This PR fixes a referenced bug by making the `etcd.status.clusterSize` update idempotent.

Fixes #255

**Release note**:

```bugfix operator
A bug has been fixed which caused the `etcd.status.clusterSize` only being set for new `etcd` resources
```

### DIFF
--- a/controllers/etcd_controller.go
+++ b/controllers/etcd_controller.go
@@ -1530,6 +1530,7 @@ func (r *EtcdReconciler) updateEtcdErrorStatus(ctx context.Context, op operation
 		if sts != nil {
 			ready := CheckStatefulSet(etcd, sts) == nil
 			etcd.Status.Ready = &ready
+			etcd.Status.Replicas = pointer.Int32PtrDerefOr(sts.Spec.Replicas, 0)
 
 			if op == bootstrapOp {
 				// Reset members in bootstrap phase to ensure dependent conditions can be calculated correctly.
@@ -1548,6 +1549,7 @@ func (r *EtcdReconciler) updateEtcdStatus(ctx context.Context, op operationResul
 		etcd.Status.ServiceName = &svcName
 		etcd.Status.LastError = nil
 		etcd.Status.ObservedGeneration = &etcd.Generation
+		etcd.Status.Replicas = pointer.Int32PtrDerefOr(sts.Spec.Replicas, 0)
 
 		if op == bootstrapOp {
 			// Reset members in bootstrap phase to ensure dependent conditions can be calculated correctly.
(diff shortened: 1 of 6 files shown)


abdasgupta (approved):
Looks good to me.

controllers/etcd_controller.go
```diff
@@ -1519,40 +1510,47 @@ func canDeleteStatefulset(sts *appsv1.StatefulSet, etcd *druidv1alpha1.Etcd) boo
...
 func bootstrapReset(etcd *druidv1alpha1.Etcd) {
 	etcd.Status.Members = nil
-	etcd.Status.ClusterSize = pointer.Int32Ptr(int32(etcd.Spec.Replicas))
+	etcd.Status.ClusterSize = pointer.Int32Ptr(etcd.Spec.Replicas)
 }
 
-func (r *EtcdReconciler) updateEtcdErrorStatus(ctx context.Context, op operationResult, etcd *druidv1alpha1.Etcd, sts *appsv1.StatefulSet, lastError error) error {
+func clusterInBootstrap(etcd *druidv1alpha1.Etcd) bool {
+	return etcd.Status.Replicas == 0 ||
+		(etcd.Spec.Replicas > 1 && etcd.Status.Replicas == 1)
```
aaronfern, line 1518: Why do we need this check? is it for an upgrade case from single node to multi node?

aaronfern (commented):
Looks really good!
Just one query from me, thanks

controllers/etcd_controller.go
timuthy: Yes, exactly. According to [this comment](https://github.com/gardener/etcd-druid/pull/188#discussion_r644553365) scaling from 1 -> 3 (or similar) is the same as bootstrapping, i.e. we don't plan to add members via the [members API](https://etcd.io/docs/v3.3/op-guide/runtime-configuration/#add-a-new-member).

aaronfern: Thanks for clarifying

aaronfern (approved):
/lgtm

PR #262, opened 2021-11-25
Create member, delta/full-snapshot `Leases`

timuthy:

/area control-plane
/kind enhancement

This PR adds member, full snapshot and delta snapshot `Leases` which are maintained by `etcd-backup-restore`.

Fixes part of #158
Fixes part of #278

A "component" concept was introduced in this PR which in the long run can replace the chart that we use and instead build the K8s resources within the code to achieve a better test-ability and more control how the client is used to reconcile the objects (e.g. used verb `UPDATE` vs `PATCH`) or how merging objects will be implemented. The same approach was taken in `gardener/gardener`.

Please see [this comment](https://github.com/gardener/gardener/issues/2754#issuecomment-678301676) for more information.

**Release note**:

```feature operator
Etcd-Druid now creates member `Lease` objects which enables the heartbeat functionality for etcd members. Along the way a new flag `--etcd-member-unknown-threshold` was introduced. It determines the duration after which a etcd member's state is considered `unknown` when the member `Lease` is not renewed.
```


abdasgupta (commented):
Thanks Tim for a well-written PR. I just have few minor suggestions.

pkg/component/etcd/lease/values.go
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lease
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type Values struct {
```
abdasgupta, line 21: Instead of struct, should we make Values as `map[string]interface{}`? If we make it as struct, we need to pass separate structs to the `getMapFromEtcd` function for separate components in future. WDYT?

abdasgupta (changes requested):

pkg/component/etcd/lease/values.go
timuthy: Hm yes, got your point. However, `getMapFromEtcd` will only be there as long as we have not entirely transitioned all manifests to components. For the transitioning phase it should be fine to do so, wdyt?

The awkward point about `map[string]interface{}` is the missing type safety for values and that you need to create constants for the keys, so that other components can use them.

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -210,6 +210,10 @@ spec:
 {{- end }}
         - --snapstore-temp-directory={{ .Values.backup.snapstoreTempDir }}
         - --etcd-process-name=etcd
+{{- if .Values.etcd.heartbeatDuration }}
+        - --enable-member-lease-renewal=true
+        - --k8s-heartbeat-duration={{ .Values.etcd.heartbeatDuration }}
```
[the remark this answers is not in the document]
timuthy: Hm, the flag is implemented in `gardener/etcd-backup-restore` and describes the duration after which the member lease is updated. I don't think we should change this name and wait for another release of `etcd-backup-restore` before we continue here.

api/v1alpha1/etcd_types.go
```diff
@@ -201,6 +201,9 @@ type EtcdConfig struct {
 	// EtcdDefragTimeout defines the timeout duration for etcd defrag call
 	// +optional
 	EtcdDefragTimeout *metav1.Duration `json:"etcdDefragTimeout,omitempty"`
+	// HeartbeatDuration defines the time members send a heartbeat. The default value is 10s.
```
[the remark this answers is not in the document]
timuthy: Can you explain why timeout? It's actually the duration after which a member renews the member lease.

timuthy:
Thanks a lot for your review @abdasgupta. Please see my comments.

pkg/component/etcd/lease/values_helper.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// GetDeltaSnapshotLeaseName returns the name of the delta snapshot lease based on the given `etcd` object.
+func GetDeltaSnapshotLeaseName(etcd *druidv1alpha1.Etcd) string {
+	return fmt.Sprintf("delta-snapshot-%s", etcd.Name)
+}
+
+// GetFullSnapshotLeaseName returns the name of the full snapshot lease based on the given `etcd` object.
+func GetFullSnapshotLeaseName(etcd *druidv1alpha1.Etcd) string {
+	return fmt.Sprintf("full-snapshot-%s", etcd.Name)
```
aaronfern, line 42: Since the lease names are being changed from `%s-full-snap` to `full-snapshot-%s` and the previous named leases are already being used, we should add a step that deletes any leases with the previous name if present on the cluster.
Could you please add this?

pkg/component/etcd/lease/lease_test.go
```diff
@@ -0,0 +1,268 @@
...
+				}))).To(Succeed())
+
+				Expect(leases.Items).To(BeEmpty())
+			})
+		})
+	})
+})
+
+func getSnapshotLease(ctx context.Context, c client.Client, namespace, name string) (*coordinationv1.Lease, error) {
+	fullLease := coordinationv1.Lease{}
```
aaronfern, line 206: Nit
Change var used in this func to `snapshotLease` as it's general purpose
```suggestion
	snapshotLease := coordinationv1.Lease{}
```

aaronfern (changes requested):
Thanks @timuthy for the PR!
I have one suggestion and one minor nitpick

pkg/component/etcd/lease/values_helper.go
timuthy: Nice catch! It was actually never my intention to change this name 😦 I reverted it to use the previous naming scheme.

pkg/component/etcd/lease/lease_test.go
timuthy: Makes sense, I changed it.

timuthy:
Thanks a lot for your review @aaronfern. I addressed your feedback, PTAL.

aaronfern (approved):
Thanks for making the changes @timuthy 
/lgtm

pkg/component/etcd/lease/values.go
abdasgupta: Hmm.. Sadly, until getMapFromEtcd is removed completely, parameter list of `getMapFromEtcd` for each component type

api/v1alpha1/etcd_types.go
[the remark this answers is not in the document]
abdasgupta: Okay. Then, instead of `time` we should use `duration`

pkg/component/etcd/lease/values.go
timuthy: I added another `componentetcd.Values` struct which aggregates the different sub values. I hope that's fine.

timuthy:
Thanks a lot for your review @abdasgupta. I addressed your comments in 735d69c, PTAL.

abdasgupta (approved):

PR #263, opened 2021-11-26
Fix etcd update conflicts

timuthy:

/area quality
/kind bug
/priority 1

This PR fixes several update conflicts that occurs because of competing updates between `etcd_controller` and `etcd_custodian_controller`.

With this PR:
- The operation annotation (`gardener.cloud/operation: reconcile`) is patched out instead of removed via update
- The Custodian controller does not react on updates as long as the ETCD controller reconciles the resource.

**Release note**:

``` bugfix operator
A bug has been fixed that led to multiple update conflicts when the `etcd` resource was reconciled.
```

### DIFF
--- a/pkg/predicate/predicate.go
+++ b/pkg/predicate/predicate.go
@@ -20,7 +20,9 @@ import (
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	appsv1 "k8s.io/api/apps/v1"
 	coordinationv1 "k8s.io/api/coordination/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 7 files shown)


pkg/predicate/predicate.go
```diff
@@ -20,33 +20,31 @@ import (
...
+	return obj.GetAnnotations()[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationReconcile
+}
+
 // HasOperationAnnotation is a predicate for the operation annotation.
 func HasOperationAnnotation() predicate.Predicate {
-	f := func(obj runtime.Object) bool {
-		etcd, ok := obj.(*druidv1alpha1.Etcd)
-		if !ok {
-			return false
-		}
```
shreyas-s-rao, line 37: Why aren't we retaining the nil check on the etcd object here? If the object is `nil`, wouldn't `hasOperationAnnotation` call fail?

shreyas-s-rao (changes requested):
@timuthy Thanks a lot for fixing this issue and raising the PR! I just had one small nitpick before we can merge the PR. PTAL.

pkg/predicate/predicate.go
timuthy: Actually, this is not a nil check. If you pass the following, the code will panic also with the previous code in place:
```
var etcd *druidv1alpha1.Etcd = nil

createEvent := event.CreateEvent{Object: etcd}
```
So, I wondered about the benefit of having this check at all?

shreyas-s-rao: Oh ok. I was just intrigued since the predicates returned in `EtcdReconciliationFinished` had these checks in place. I still feel it's better to keep this check, to be able to return `false` early rather than panic. WDYT?

timuthy: > I was just intrigued since the predicates returned in EtcdReconciliationFinished had these checks in place.

There it's needed because we access a particular field of the `etcd` resource which is `etcd.status.ready`.

Here, we only need to get the annotations which is possible via the `client.Object` interface that is guaranteed through the `{Create,Update,Delete,Generic}Event` object.

A panic is not prevented by the earlier check because you actually need a [typed nil check](https://stackoverflow.com/questions/13476349/check-for-nil-and-nil-interface-in-go) for this which has never been used.

shreyas-s-rao (approved):

PR #271, opened 2021-12-06
Added Backupready condition to the etcd status

### LINKED ISSUE #251: [Feature] Introduce BackupReady condition
The multi-node etcd proposal includes the introduction of a `BackupReady` condition [ref](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/multi-node/README.md#cutting-off-client-requests).

The `BackupReady` condition ~will influence the decision if client traffic to the etcd cluster is cut-off.~ will serve as an indicator whether the backup functionality is working as expected.

This condition is supposed to deduce the healthiness of backups from the snapshot `Lease` object maintained by the `etcd-backup-restore` side-car. Hence, the backup is considered unhealthy if the renewal time exceeds a configured threshold.

aaronfern:

/area backup
/area disaster-recovery
/kind enhancement

This PR adds the `BackupReady` condition to the etcd status. 
This condition indicates the health of etcd-backup-restore and whether it is able to regularly take and save data snapshots

Fixes #251 

**Release note**:

```other operator
Added a new condition `BackupReady` to the etcd status
```


pkg/health/condition/check_all_members.go
```diff
@@ -14,12 +14,17 @@
...
 
-import druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+import (
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
 
-type allMembersReady struct{}
+type allMembersReady struct {
+	cl client.Client
```
timuthy, line 23: What do we need the client for?

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+const (
+	FullBackupSucceeded string = "FullBackupSucceeded"
+	IncrBackupSucceeded string = "IncrementalBackupSucceeded"
+	BackupFailed        string = "BackupFailed"
+	Unknown             string = "Unknown"
+)
+
+func (a *backupReadyCheck) Check(etcd druidv1alpha1.Etcd) Result {
```
timuthy, line 40:
```suggestion
func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
```
Can you make the `Check` function accepting a `context.Context` that you can use for further calls instead of `context.TODO()`?

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,102 @@
...
+		reason:  Unknown,
+		message: "Cannot determine etcd backup status since no snapshot leases present",
+	}
+
+	//Fetch snpashot leases
+	fullSnapLease := &coordinationv1.Lease{}
+	deltaSnapLease := &coordinationv1.Lease{}
+	var fullSnapErr, incrSnapErr error
+	fullSnapErr = a.cl.Get(context.TODO(), types.NamespacedName{Name: getFullSnapLeaseName(&etcd), Namespace: "default"}, fullSnapLease)
+	incrSnapErr = a.cl.Get(context.TODO(), types.NamespacedName{Name: getDeltaSnapLeaseName(&etcd), Namespace: "default"}, deltaSnapLease)
```
timuthy, line 54: - See previous comment about context.
- Please use the namespace of the `etcd` resources instead of `"default"`.

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,102 @@
...
+	IncrBackupSucceeded string = "IncrementalBackupSucceeded"
+	BackupFailed        string = "BackupFailed"
+	Unknown             string = "Unknown"
+)
+
+func (a *backupReadyCheck) Check(etcd druidv1alpha1.Etcd) Result {
+	//Default case
+	result := &result{
+		conType: druidv1alpha1.ConditionTypeBackupReady,
+		status:  druidv1alpha1.ConditionFalse,
```
timuthy, line 44:
```suggestion
		status:  druidv1alpha1.ConditionUnknown,
```

We should only set it to `druidv1alpha1.ConditionFalse` after a configurable amount of time has been elapsed.

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	//Fetch snpashot leases
+	fullSnapLease := &coordinationv1.Lease{}
+	deltaSnapLease := &coordinationv1.Lease{}
+	var fullSnapErr, incrSnapErr error
+	fullSnapErr = a.cl.Get(context.TODO(), types.NamespacedName{Name: getFullSnapLeaseName(&etcd), Namespace: "default"}, fullSnapLease)
+	incrSnapErr = a.cl.Get(context.TODO(), types.NamespacedName{Name: getDeltaSnapLeaseName(&etcd), Namespace: "default"}, deltaSnapLease)
+
+	if fullSnapErr != nil && incrSnapErr != nil {
```
timuthy, line 56:
```suggestion
	if fullSnapErr != nil || incrSnapErr != nil {
```
Both leases are created at almost the same time, so you can drop any `incrSnapErr == nil` or `fullSnapErr == nil` checks below.

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,102 @@
...
+			result.reason = IncrBackupSucceeded
+			result.message = "Delta snapshot taken"
+			result.status = druidv1alpha1.ConditionTrue
+			return result
+
+		}
+	}
+	if fullSnapErr == nil && fullSnapLease.Spec.RenewTime != nil {
+		fullLeaseRenewTime := fullSnapLease.Spec.RenewTime
+		if time.Since(fullLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
```
timuthy, line 73: Hm, full snapshot leases are only renewed after full snapshot or compaction. So the check here will very likely fail, right?

timuthy (changes requested):

pkg/health/condition/check_all_members.go
aaronfern: Right, not needed. Removed

pkg/health/condition/check_backup_ready.go
aaronfern: Done

pkg/health/condition/check_backup_ready.go
aaronfern: Sorry, this completely escaped my attention
Thanks. Corrected this

pkg/health/condition/check_backup_ready.go
aaronfern: Done

pkg/health/condition/check_backup_ready.go
aaronfern: Done.
Also, expanded this condition to return if both snapshots have empty holderIDs (during the startup scenario when we have empty leases)

pkg/health/condition/check_backup_ready.go
aaronfern: Yes, this will be false almost always except when a full snapshot is taken. 
In that case, the delta snapshot case just before this would decide the condition.

Do you think it makes sense to combine this with the block above and just consider to be true if either snapshot is renewed with a common message instead?

timuthy: Good question. Does it make sense to check if the last full backup at least happened in the last 24h (this is the only thing we can presume)?

aaronfern: On it's own, I don't think that would be a good idea. 24h would be too large of a window to consider while setting the condition to true.
If you mean a check to see if a full snap has happened in the past 24h together with the delta snap check, that would make sense

timuthy: > If you mean a check to see if a full snap has happened in the past 24h together with the delta snap check, that would make sense

Yep, that's what I meant 👍

aaronfern: Yup, this makes sense, I'll add it and push a new commit

pkg/health/condition/check_all_members.go
```diff
@@ -14,12 +14,17 @@
...
+
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
 
 type allMembersReady struct{}
 
-func (a *allMembersReady) Check(status druidv1alpha1.EtcdStatus) Result {
-	if len(status.Members) == 0 {
+func (a *allMembersReady) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
```
timuthy, line 26: Nit
```suggestion
func (a *allMembersReady) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
```

pkg/health/condition/check_all_members.go
```diff
@@ -49,6 +54,6 @@ func (a *allMembersReady) Check(status druidv1alpha1.EtcdStatus) Result {
 }
 
 // AllMembersCheck returns a check for the "AllMembersReady" condition.
-func AllMembersCheck() Checker {
+func AllMembersCheck(cl client.Client) Checker {
```
timuthy, line 57: Nit
```suggestion
func AllMembersCheck(_ client.Client) Checker {
```

timuthy (changes requested):
Just found two more nits and waiting for the change mentioned in https://github.com/gardener/etcd-druid/pull/271/files#r766817769.

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	//Fetch snpashot leases
+	fullSnapLease := &coordinationv1.Lease{}
+	deltaSnapLease := &coordinationv1.Lease{}
+	var fullSnapErr, incrSnapErr error
+	fullSnapErr = a.cl.Get(ctx, types.NamespacedName{Name: getFullSnapLeaseName(&etcd), Namespace: etcd.ObjectMeta.Namespace}, fullSnapLease)
+	incrSnapErr = a.cl.Get(ctx, types.NamespacedName{Name: getDeltaSnapLeaseName(&etcd), Namespace: etcd.ObjectMeta.Namespace}, deltaSnapLease)
+
+	if fullSnapErr != nil || incrSnapErr != nil || (fullSnapLease.Spec.HolderIdentity == nil && deltaSnapLease.Spec.HolderIdentity == nil) {
```
timuthy, line 56: I wonder about how this behaves for new clusters as I expect there will be a period where only one of the leases (usually `fullSnapLease.Spec.HolderIdentity`) is maintained. Then we'd add a failing backup condition until the next snapshot happens, right? Should we be more tolerant in this case?

aaronfern: Hm, yes, you're right there
Will rework the conditions that follow to be a bit more tolerant

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,111 @@
...
+func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
+	//Default case
+	result := &result{
+		conType: druidv1alpha1.ConditionTypeBackupReady,
+		status:  druidv1alpha1.ConditionUnknown,
+		reason:  Unknown,
+		message: "Cannot determine etcd backup status since no snapshot leases present",
+	}
+
+	//Fetch snpashot leases
```
timuthy, line 48:
```suggestion
	//Fetch snapshot leases
```

timuthy (changes requested):
Thanks for the update @aaronfern. Just two more things 🙂

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,111 @@
...
+		if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration && time.Since(fullLeaseCreateTime.Time) < 24*time.Hour {
+			result.reason = BackupSucceeded
+			result.message = "Delta snapshot backup succeeded"
+			result.status = druidv1alpha1.ConditionTrue
+			return result
+		}
+	} else if deltaSnapLease.Spec.HolderIdentity == nil && fullLeaseRenewTime != nil {
+		//Most probable during a startup scenario for new clusters
+		//Special case. Return Unknown condition for some time to allow delta backups to start up
+		if time.Since(fullLeaseRenewTime.Time) < 5*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
```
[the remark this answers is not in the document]
aaronfern: You're right, I see your point here

I still need we need this check, but I'll propose something slightly different on the whole
1. instead of having any `nil` checks for `HolderIdentity`, I'll instead use `RenewTime` as `HolderIdentity` won't be set without `RenewTime`. It should amount to the same thing, but will be more significant with point 2

2. I've already proposed a change in lease update logic in [etcd-backup-restore#410](https://github.com/gardener/etcd-backup-restore/pull/410) where even if a delta snapshot is not created, the snapshot lease is still renewed (needed as the lease renewals are heartbeats, and we shouldn't skip a beat if there's no change in the cluster)

With these two changes combined, I think the condition should make more sense as a whole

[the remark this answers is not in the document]
timuthy: > With these two changes combined, I think the condition should make more sense as a whole

Makes sense. However, we now have a dependency and can (strictly) only merge this PR when [etcd-backup-restore#410](https://github.com/gardener/etcd-backup-restore/pull/410) is merged and released. So either we set this PR on hold for the time being or we remove this check and add it again when etcd-br is released. I personally tend to do the latter so you don't have to worry about rebasing.

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	deltaLeaseRenewTime := deltaSnapLease.Spec.RenewTime
+	fullLeaseRenewTime := fullSnapLease.Spec.RenewTime
+	fullLeaseCreateTime := &fullSnapLease.ObjectMeta.CreationTimestamp
+
+	if fullLeaseRenewTime == nil && deltaLeaseRenewTime != nil && fullLeaseCreateTime != nil {
+		//Most probable during reconcile of existing clusters if fresh leases are created
+		//Treat backup as succeeded if delta snap lease renewal happens in the required time window and full snap lease not more than 24h old
+		if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration && time.Since(fullLeaseCreateTime.Time) < 24*time.Hour {
```
ishan16696, line 66:
```suggestion
		if time.Since(deltaLeaseRenewTime.Time) < 1.1*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration && time.Since(fullLeaseCreateTime.Time) < 24*time.Hour {
```
why `2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration` ? What about `1.1` if we want more to be more precise

ishan16696 (changes requested):
I also have a similar doubts as @timuthy ,
Let's discuss after this [PR](https://github.com/gardener/etcd-backup-restore/pull/410)

ishan16696:
One more thing, the multi-node design document suggested that `.etcd.Status.BackupReady` condition should be maintained by leading sidecar(check [here](https://github.com/gardener/etcd-druid/tree/master/docs/proposals/multi-node#status)).
I think here it is maintained by custodian-controller using snapshot leases as heartbeat. 
Am I missing something?

aaronfern:
@ishan16696 
yes, you're right, the custodian controller maintains that now.
We took a call some time back that this approach would be preferable over letting the leading sidecar maintain the status as that will avoid multiple actors trying to modify the etcd object 
The doc was not modified to capture this point

pkg/health/condition/check_backup_ready.go
aaronfern: That's a little bit arbitrary 
2 in the sense that we can tolerate one missed update, but 2 missed updated in a row would be reason to suspect some component is failing

wdyt? would you recommend something else?

pkg/health/condition/check_backup_ready.go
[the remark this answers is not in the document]
aaronfern: makes sense
Removed this condition

pkg/health/condition/check_backup_ready.go
ishan16696: > 2 in the sense that we can tolerate one missed update, but 2 missed updated in a row would be reason to suspect some component is failing

if that’s the case then it this condition get fails when `time.Since(deltaLeaseRenewTime.Time)=10min` and `2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration=10min`

aaronfern: Right
in this case since since lease had not been updated in 10 mins (2 missed updates) this is expected to fail

ishan16696:
Isn't we are giving too much work to custodian controller that now custodian controller have to maintain `.etcd.Status. BackupReady` by keep checking the snapshot leases(delta snapshot lease will get update for every 5mins).
And custodian controller have to check the snapshot leases of all shoots backup-restore.
WDYT?

cc @abdasgupta @shreyas-s-rao

ishan16696:
> Isn't we are giving too much work to custodian controller that now custodian controller have to maintain .etcd.Status. BackupReady by keep checking the snapshot leases(delta snapshot lease will get update for every 5mins).
And custodian controller have to check the snapshot leases of all shoots backup-restore.

As discussed to avoid race condition we can't let both backup-restore(leading sidecar) as well as controller to update the `.etcd.Status`, hence we decided to go with custodian controller maintaining `.etcd.Status.BackupReady` using snapshot leases as heartbeat but as I described above this can leads to performance issue which we need to verify.

pkg/health/condition/check_backup_ready.go
ishan16696: but what I'm saying is lets suppose `time.Since(deltaLeaseRenewTime.Time)=10min` and deltaSnashot is taken but Snapshotter is about to update the delta lease but before that custodian controller checks lease then isn't this condition can occur.

timuthy: Hm, should we first transition to `Unknown` and then to `False` eventually? This can help to only fire alerts when the condition is `False` and to handle intermittent states as mentioned by @ishan16696 gracefully.

aaronfern: Hm, you're right, that will help
I've added the intermediate state in [`f61fb00`](https://github.com/gardener/etcd-druid/pull/271/commits/f61fb002123b5b69fd231ea7057ce2446c709cb2)

timuthy: The changes look good. However, I noticed that we have the same dependency as discussed [here](https://github.com/gardener/etcd-druid/pull/271#discussion_r774607346), right? We actually need [etcd-backup-restore#410](https://github.com/gardener/etcd-backup-restore/pull/410) to be merged and released first. Should we make it a TODO as well or do you have any other suggestion?

aaronfern:
Dropped  [`b07222b`](https://github.com/gardener/etcd-druid/commit/b07222b0a312a03fe1e5ac396a7a5ad0f63c5113) as we agreed to complete this PR before putting it `on-hold` until a new release of etcd-backup-restore is available with [etcd-backup-restore/410](https://github.com/gardener/etcd-backup-restore/pull/410) included

pkg/health/condition/check_backup_ready.go
timuthy: https://github.com/gardener/etcd-druid/pull/271#issuecomment-1012259584

ishan16696 (approved):
LGTM!!

ishan16696:
@timuthy do you also wanted to take a look?

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+type backupReadyCheck struct {
+	cl client.Client
+}
+
+const (
+	BackupSucceeded string = "BackupSucceeded"
+	BackupFailed    string = "BackupFailed"
+	Unknown         string = "Unknown"
```
timuthy, line 36: Doc strings are missing.

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,129 @@
...
+	//Set status to Unknown if errors in fetching snapshot leases or lease never renewed
+	if fullSnapErr != nil || incrSnapErr != nil || (fullSnapLease.Spec.RenewTime == nil && deltaSnapLease.Spec.RenewTime == nil) {
+		return result
+	}
+
+	deltaLeaseRenewTime := deltaSnapLease.Spec.RenewTime
+	fullLeaseRenewTime := fullSnapLease.Spec.RenewTime
+	fullLeaseCreateTime := &fullSnapLease.ObjectMeta.CreationTimestamp
+
+	if fullLeaseRenewTime == nil && deltaLeaseRenewTime != nil && fullLeaseCreateTime != nil {
```
timuthy, line 64: Above there is:
```
fullLeaseCreateTime := &fullSnapLease.ObjectMeta.CreationTimestamp
```
So `fullLeaseCreateTime` will never be `nil`. Can you double check?

pkg/health/condition/check_backup_ready.go
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
timuthy, line 1:
```suggestion
// Copyright (c) 2022 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

timuthy (changes requested):
Just found some minor points. Thanks for the PR.

pkg/health/condition/check_backup_ready.go
aaronfern: Snapshot leases contain `CreationTimestamp` when they're created and any updates to the snapshot leases happen through patches
That should be sufficient to guarantee that `CreationTimestamp` always exists, right?

pkg/health/condition/check_backup_ready.go
aaronfern: Added

pkg/health/condition/check_backup_ready.go
timuthy: I think you don't need to check this because in line 57 you retrieve the full snapshot lease, then in line 61 you check for errors and if there isn't one then you certainly got the object from the Apiserver (no not found error) and thus it'll always have a `CreationTimestamp`.

aaronfern: Ah, okay
So `fullLeaseCreateTime != nil` is kind of redundant in this condition
did I understand correctly what you're trying to imply?

aaronfern: Updated condition

aaronfern:
Thanks for your review @timuthy 
I've addressed your feedback. PTAL

timuthy (approved):
/lgtm
Thank you 🙂

PR #208, opened 2021-07-28
Simplify reconciliation

amshuman-kr:
**How to categorize this PR?**

/area control-plane
/kind technical-debt

**What this PR does / why we need it**:

Simplified claim logic in reconciliation by replacing selecting existing objects based on a selector with `CreateOrUpdate` or recreating them if necessary.

**Which issue(s) this PR fixes**:
Fixes #186 

**Special notes for your reviewer**:
This PR is based on #203 and ensures that all the test cases as of #203 are still working after the simplification. So, only the commits https://github.com/gardener/etcd-druid/commit/67df3c125fc0518c0f9274cb44960c217c599617 and https://github.com/gardener/etcd-druid/commit/36a8ed60695e286b850149aa9cc9bbfa6bda78a5 are relevant in this PR.

This PR should be merged only after rebasing after #203 is merged.

**Release note**:

```other operator
Simplified claim logic in reconciliation by replacing selecting existing objects based on a selector with `CreateOrUpdate` or recreating them if necessary.
```

PR #273, opened 2021-12-09
Add service component and peer service

timuthy:

/area control-plane
/kind enhancement

This PR does two things:
- Adds a `service` component (similar to #262). As part of this transition the earlier used claim logic was removed as proposed in #208.
- Adds a new `service` object for peer-communication. 

Fixes parts of #158

~This PR is based on #262 and therefore place ignore b3a437c415004fe401e1459ff99c3b8208149a68 during your review. Because of this dependency, the draft state will remain until #262 is merged.~

**Release note**:

```feature operator
A new service (`<etcd-name>-peer`) for etcd peer communication (default port `2380`) is now created by Etcd-Druid.
```
```breaking operator
Using the etcd `client` service for server communication (default port `2380`) has been deprecated. The port will be removed from the service in the near future. If necessary, switch to the new `peer` service instead.
```
```breaking operator
The claiming logic for services has been removed from Etcd-Druid. This means that existing `service` objects cannot be adopted anymore but a new and dedicated object is created. Please check any usages for already adopted `services` and switch to the dedicated `<etcd-name>-client` service.
```

### DIFF
--- a/controllers/etcd_controller.go
+++ b/controllers/etcd_controller.go
@@ -411,36 +411,6 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
 	return ctrl.Result{}, nil
 }
 
-func (r *EtcdReconciler) syncServiceSpec(ctx context.Context, logger logr.Logger, svc *corev1.Service, etcd *druidv1alpha1.Etcd, renderedChart *chartrenderer.RenderedChart) (*corev1.Service, error) {
-	decoded, err := r.getServiceFromEtcd(etcd, renderedChart)
-	if err != nil {
-		return nil, err
-	}
-
-	if reflect.DeepEqual(svc.Spec, decoded.Spec) {
-		return svc, nil
-	}
-	svcCopy := svc.DeepCopy()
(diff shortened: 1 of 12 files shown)


timuthy:
PR is ready for a review since #262 is merged.

controllers/etcd_controller.go
```diff
@@ -410,87 +411,6 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
-	}
-
-	if err := controllerutil.SetControllerReference(etcd, svc, r.Scheme); err != nil {
-		return nil, err
-	}
-
-	return svc.DeepCopy(), err
-}
-
 func (r *EtcdReconciler) syncServiceSpec(ctx context.Context, logger logr.Logger, svc *corev1.Service, etcd *druidv1alpha1.Etcd, renderedChart *chartrenderer.RenderedChart) (*corev1.Service, error) {
```
aaronfern, line 414: We don't need this function anymore, right?

timuthy: Nice catch 👍  Removed in a895b0c6d0eec7a9e7d4fee2bf230c1fd8e20cba.

aaronfern (approved):
/lgtm

PR #276, opened 2021-12-17
Enhancement.reconciliation

timuthy:

/area cost
/area robustness
/area performance
/kind enhancement

This PR enhances several reconcile logics in Druid controllers, mainly to reduce the amount of unnecessary reconciliation. For instance, earlier an `etcd` resource was at least reconciled twice in a row when it was created or updated. Please have a look at the individual commits for more information.

**Release note**:

```other operator
The amount of active reconciliations has been limited to improve performance, CPU, memory and network consumption.
```


aaronfern (approved):
/lgtm

timuthy:
/hold
I want to double check https://github.com/gardener/etcd-druid/pull/276/commits/e1cbc6e4a7e2230f1cc774ffeb5f175f6b3942b8

timuthy:
I dropped e1cbc6e because we cannot simply prematurely remove the annotation w/o setting `etcd` to not ready. Otherwise controllers expect that the resource is successfully reconciled. I will work on another approach and will file it in a separate PR.
/unhold

Issue #299, opened 2022-03-02
[Feature] Enhance the Etcd StatefulSet chart to support usage of the `Local` provider in a container environment.


kris94:
**Feature (What you would like to be added):**
Enhance the Etcd StatefulSet chart to support usage of the `Local` provider in a container environment.
**Motivation (Why is this needed?):**
The scenario in which this is needed is described [here](https://github.com/gardener/gardener/issues/5500)
**Approach/Hint to the implement solution (optional):**
In the stateful set there must be added a volume that points to `/etc/gardener/etcd-local` + {{ .Values.store.storageContainer }}. This is the directory that will be mounted in the container that hosts the control plane in a `local provider` scenario. Also in the `backup-restore` container a volume mount must be added with `mounthPath` = {{ .Values.store.storageContainer }}.

PR #300, opened 2022-03-02
Add support for ETCD backups in the local provider extension

kris94:

/area backup
/kind enhancement

This PR adds support for ETCD backups in the local provider extension by enhancing the Etcd StatefulSet and the Etcd-copy-backup-job yaml files to support the `Local` provider.
Fixes #
#299 

**Release note**:

```other operator
The ETCD backup restore is now configured to support Local provider in container environment.
```

### DIFF
--- a/charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
+++ b/charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
@@ -108,6 +108,14 @@ spec:
           value: "/root/source-etcd-backup"
 {{- end }}
         volumeMounts:
+{{- if eq .Values.targetStore.storageProvider "Local" }}
+        - name: target-host-storage
+          mountPath: {{ .Values.targetStore.storageContainer }}
+{{- end }}
+{{- if eq .Values.sourceStore.storageProvider "Local" }}
+        - name: source-host-storage
+          mountPath: {{ .Values.sourceStore.storageContainer }}
+{{- end }}
 {{- if eq .Values.targetStore.storageProvider "GCS" }}
         - name: etcd-backup
           mountPath: "/root/.gcp/"
@@ -149,6 +157,18 @@ spec:
           mountPath: "/root/source-etcd-backup"
 {{- end }}
       volumes:
+{{- if eq .Values.targetStore.storageProvider "Local" }}
+      - name: target-host-storage
+        hostPath:
+          path: {{ .Values.targetStore.storageMountPath }}/{{ .Values.targetStore.storageContainer }}
+          type: Directory
+{{- end }}
+{{- if eq .Values.sourceStore.storageProvider "Local" }}
+      - name: source-host-storage
+        hostPath:
+          path: {{ .Values.sourceStore.storageMountPath }}/{{ .Values.sourceStore.storageContainer }}
+          type: Directory
+{{- end }}
 {{- if eq .Values.targetStore.storageProvider "GCS" }}
       - name: etcd-backup
         secret:
--- a/charts/etcd/templates/etcd-statefulset.yaml
+++ b/charts/etcd/templates/etcd-statefulset.yaml
@@ -307,6 +307,10 @@ spec:
               optional: true
 {{- end }}
         volumeMounts:
+{{- if eq .Values.store.storageProvider "Local" }}
+        - name: host-storage
+          mountPath: {{ .Values.store.storageContainer }}
+{{- end }}
         - name: {{ .Values.volumeClaimTemplateName }}
           mountPath: /var/etcd/data
         - name: etcd-config-file
@@ -345,6 +349,12 @@ spec:
             - SYS_PTRACE
       shareProcessNamespace: true
       volumes:
+{{- if eq .Values.store.storageProvider "Local" }}
+      - name: host-storage
+        hostPath:
+          path: {{ .Values.store.storageMountPath }}/{{ .Values.store.storageContainer }}
+          type: Directory
+{{- end }}
       - name: etcd-config-file
         configMap:
           name: {{ .Values.configMapName }}
--- a/pkg/utils/miscellaneous.go
+++ b/pkg/utils/miscellaneous.go
@@ -15,13 +15,24 @@
 package utils
 
 import (
+	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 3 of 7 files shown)


charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
```diff
@@ -232,6 +240,18 @@ spec:
           mountPath: "/root/.source-gcp/"
 {{- end }}
       volumes:
+{{- if eq .Values.targetStore.storageProvider "Local" }}
+      - name: target-host-storage
+        hostPath:
+          path: /etc/gardener/local-backupbuckets/{{ .Values.targetStore.storageContainer }}
```
rfranzke, line 246: This path is hard-coded, so when it changes on g/g side it will break. Maybe it should be configurable?

kris94: Any ideas how we can deliver this property from the `local-extension-provider` or the `gardnlet` to the druid in most elegant way without having to extend the `BackupBucket` or the `ETCD` objects just for the specific local stuff? Probably dumb idea, but we can add `bucketPath` to `etcd-backup` secret just for provider local. It would be something like:
```yaml
apiVersion: v1
data:
  bucketName: <bucket name>
  bucketPath: /etc/gardener/local or whatever the path would be
  kubeconfig: <kubeconfig>
  kind: Secret
....
```

rfranzke: Yes, sounds like a good pragmatic idea!

pkg/utils/miscellaneous.go
```diff
@@ -163,6 +169,13 @@ func GetStoreValues(store *druidv1alpha1.StoreSpec) (map[string]interface{}, err
 		"storePrefix":     store.Prefix,
 		"storageProvider": storageProvider,
 	}
+	if strings.EqualFold(string(*store.Provider), Local) {
```
plkokanov, line 172: I'm wondering whether you should just always try to call getMountPathFromSecretRef(), and if it cannot find the proper data field in the secret, just to not add the storeValues["storageMountPath"] value.

charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
plkokanov, line 246: Getting the path from the etcd-secret should be added here as well.

charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
```diff
@@ -232,6 +240,18 @@ spec:
...
+          type: Directory
+{{- end }}
+{{- if eq .Values.sourceStore.storageProvider "Local" }}
+      - name: source-host-storage
+        hostPath:
+          path: /etc/gardener/local-backupbuckets/{{ .Values.sourceStore.storageContainer }}
```
plkokanov, line 252:
```suggestion
          path: {{ .Values.store.storageMountPath }}/{{ .Values.store.storageContainer }}
```

pkg/utils/miscellaneous.go
```diff
@@ -172,6 +185,20 @@ func GetStoreValues(store *druidv1alpha1.StoreSpec) (map[string]interface{}, err
 	return storeValues, nil
 }
 
+func getMountPathFromSecretRef(ctx context.Context, client client.Client, store *druidv1alpha1.StoreSpec, namespace string) (string, error) {
```
plkokanov, line 188:
```suggestion
func getHostMountPathFromSecretRef(ctx context.Context, client client.Client, store *druidv1alpha1.StoreSpec, namespace string) (string, error) {
```

plkokanov (changes requested):

pkg/utils/miscellaneous.go
kris94: I was thinking the same, but I have decided to save one unnecessary call to the Api Server that would be valid only for the local provider and will be unnecessary in 99% of the use cases anyway.

plkokanov: Yep thought, you did it because of that :)

kris94:
/ hold 

I need to add [this](https://github.com/gardener/gardener/pull/5514#discussion_r823420777) first

charts/etcd-copy-backups/templates/etcd-copy-backups-job.yaml
```diff
@@ -221,8 +221,16 @@ spec:
               name: {{ .Values.sourceStore.storeSecret }}
               key: "accessKeyID"
 {{- end }}
-{{- if or (eq .Values.targetStore.storageProvider "GCS") (eq .Values.sourceStore.storageProvider "GCS") }}
+{{- if or (eq .Values.targetStore.storageProvider "GCS") (eq .Values.sourceStore.storageProvider "GCS") (eq .Values.sourceStore.storageProvider "Local") (eq .Values.sourceStore.storageProvider "Local")}}
```
plkokanov, line 224:
```suggestion
{{- if or (eq .Values.targetStore.storageProvider "GCS") (eq .Values.sourceStore.storageProvider "GCS") (eq .Values.targetStore.storageProvider "Local") (eq .Values.sourceStore.storageProvider "Local")}}
```

pkg/utils/miscellaneous.go
```diff
@@ -15,15 +15,27 @@
...
 	"fmt"
+	"strings"
 	"time"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
plkokanov, line 28:
```suggestion
```

plkokanov (changes requested):

plkokanov (approved):
/lgtm

ishan16696:
Hi @kris94 ,
kindly asking you rebase this PR on latest master

ishan16696 (approved):
LGTM!!

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -307,6 +307,10 @@ spec:
               optional: true
 {{- end }}
         volumeMounts:
+{{- if eq .Values.store.storageProvider "Local" }}
+        - name: host-storage
+          mountPath: {{ .Values.store.storageContainer }}
+{{- end }}
```
ishan16696, line 313: so, you are mounting a volume for `Local` provider for data-storage ... right ?

kris94: Yes, exactly. That directory will be used to write the backups there

PR #309, opened 2022-03-22
[GEP-18] Make CA certificate data key configurable

rfranzke:
**How to categorize this PR?**

/area security
/kind enhancement
/merge squash

**What this PR does / why we need it**:
As part of https://github.com/gardener/gardener/issues/3292#issuecomment-1036058890, the CA bundle secrets should be used which may contain both old and new CA certificate during rotation. However, such bundle secrets expose the data under `.data["bundle.crt"]` despite of `.data["ca.crt"]`, so the data key should be configurable for users.

**Which issue(s) this PR fixes**:
Part of https://github.com/gardener/gardener/issues/3292#issuecomment-1036058890

**Special notes for your reviewer**:
/cc @ishan16696 

**Release note**:

```noteworthy developer
The CA data key is now configurable under `.spec.etcd.tls.tlsCASecretRef.dataKey`. It still defaults to `ca.crt` if not provided.
```

PR #310, opened 2022-03-22
Add secret controller to ensure finalizers for referenced secrets

rfranzke:

/area usability
/kind bug
/merge squash

While testing https://github.com/gardener/etcd-druid/pull/309, I noticed that `etcd-druid` does not remove finalizer from old secrets in case the references changes. This breaks the CA rotation scenario where the secrets are immutable and their names unique (i.e., when the certificates are regenerated then secrets with new names will be created).

This PR introduces a secret controller which is now responsible for adding and removing finalizers from referenced (and no longer referenced) secrets in `Etcd` resources.
It watches `corev1.Secret`s and maps references in `druidv1alpha1.Etcd` resources.

Required for https://github.com/gardener/gardener/issues/3292#issuecomment-1036058890

**Release note**:

```bugfix operator
A bug has been fixed which caused the `etcd-druid` not removing its finalizers from referenced secrets in `Etcd` resources when those references changed.
```


rfranzke:
I don't see how this is easily possible given that the `Etcd` controller operators on the current `Etcd` object and performs a lot of operations other than just maintaining finalizers for secret references. How would it ensure that it removes finalizers from old secrets when the references changes? I am open for suggestions, what do you have in mind?

ishan16696:
I also don’t have any hard opinion on this, just wanted to check as we already have 4 controllers (etcd-controller, custodian, compaction, copyback controllers) but as you mention to ensure the removal of finalizers from old secrets you introduced new secret controller, that's fine for me. Just one more doubt:
More controller will decrease the performance as they have their separate workers ... Am I right ?

ishan16696:
ahh ... ok, thanks for explaining.

ishan16696 (approved):

PR #312, opened 2022-03-24
Remove default CPU and memory limits

stoyanr:

/kind fix

Removes the default CPU and memory limits from the `etcd` and `backup-restore` containers. This is done in order to enable removal of limits via the `Etcd` resource. Currently, the default limits are used if no limits are specified in the `Etcd` resource.

Fixes #

**Release note**:
```bugfix operator
The default CPU and memory limits for `etcd` and `backup-restore` containers have been removed to enable removal of limits via the `Etcd` resource.
```

### DIFF
--- a/charts/etcd/values.yaml
+++ b/charts/etcd/values.yaml
@@ -23,9 +23,6 @@ etcd:
   metrics: basic
   etcdDefragTimeout: 8m
   resources:
-    limits:
-      cpu: 100m
-      memory: 512Gi
     requests:
       cpu: 50m
       memory: 128Mi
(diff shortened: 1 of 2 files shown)


ishan16696:
> This is done in order to enable removal of limits via the `Etcd` resource.

I didn’t understand the purpose to removing the limits for etcd and backup-restore. Is this because we want `etcd` and `backup-restore` should not get failed due to these limits ?

stoyanr:
> I didn’t understand the purpose to removing the limits for etcd and backup-restore. Is this because we want etcd and backup-restore should not get failed due to these limits ?

Yes, please see the change in g/g that removes the CPU limits from many components (https://github.com/gardener/gardener/pull/5627), a similar change is being prepared for memory limits.  The default limits used here are far too low and should never be used anyway. Please let's remove them as this is blocking other changes.

ishan16696 (approved):

PR #314, opened 2022-03-30
Updated configmap mountpoints based on ETCD replicas.

abdasgupta:

/area TODO
/kind TODO

This PR is rework of #193. 
Fixes #

**Release note**:

**breaking**
```other operator
- If Spec.Replicas in ETCD CR is greater than 0 and a even number, then no statefulset for ETCD nodes will be created and so the ETCD cluster won't be setup by Druid
```

```other operator
- Deployed configmap programmatically as component instead of chart
- configmap configures ETCD config based on the number of nodes in cluster. Number of nodes in cluster is derived from spec.Replicas of ETCD CR
- Introduced separate TLS config for client and peer communication with ETCD cluster
```


config/rbac/leader_election_role.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: leader-election-role
 rules:
 - apiGroups:
-  - ""
+  - "coordination.k8s.io
```
aaronfern, line 8:
```suggestion
  - "coordination.k8s.io"
```

aaronfern (changes requested):
Can you please remove `charts/etcd/templates/etcd-configmap.yaml` as it's not required anymore

config/rbac/leader_election_role.yaml
ishan16696: are you sure `coordination.k8s.io` about this as when I tested druid before release I didn’t face any error.

controllers/etcd_controller.go
```diff
@@ -940,21 +823,15 @@ func (r *EtcdReconciler) reconcileRoleBinding(ctx context.Context, logger logr.L
...
-		Lease:   componentlease.GenerateValues(etcd),
-		Service: componentservice.GenerateValues(etcd),
-	}
-
-	values, err := getMapFromEtcd(ctx, r.Client, r.ImageVector, etcd, val, r.disableEtcdServiceAccountAutomount)
-
-	if err != nil {
-		return nil, nil, err
+	// Check if Spec.Replicas is odd or even.
+	if etcd.Status.Replicas > 1 && etcd.Spec.Replicas&1 == 0 {
```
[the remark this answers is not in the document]
abdasgupta: Yes. I already realised it. Fixed it.

config/rbac/leader_election_role.yaml
abdasgupta: Fixed

aaronfern (approved):
Thanks for the PR!
/lgtm

Issue #107, opened 2020-10-21
Multi-Node/Clustered ETCD


amshuman-kr:
# Stories
Please provide stories that we plan to tackle:
- As operator ...
- As provider ...

# Motivation
We should support provisioning and management of multi-node etcd clusters via etcd-druid to serve the following goals:
- Generally, better robustness, resilience, and high availability (HA)
- Zero downtime maintenance (ZDM)
- Non-disruptive scaling
- Single-zone outage/fault tolerance
- Ephemeral (or in-memory) persistence for better performance

# Acceptance Criteria
- [x] First increment (please break-down into different increments rather than by component, if possible)
  <details open>
  <summary>Tasks</summary>

  - [x] Documentation/Proposal
    - [x] gardener/etcd-backup-restore#115
    - [x] gardener/etcd-backup-restore#152
    - [x] gardener/etcd-backup-restore#496

  - [x] Etcd-Druid
    - [x] gardener/etcd-backup-restore#168
    - [x] gardener/etcd-backup-restore#151
    - [x] gardener/etcd-backup-restore#165
    - [x] gardener/etcd-backup-restore#235
    - [x] gardener/etcd-backup-restore#349 
    - [x] gardener/etcd-backup-restore#158
    - [x]  gardener/etcd-backup-restore#354
    - [x] gardener/etcd-backup-restore#340
    - [x] gardener/etcd-backup-restore#358  
    - [x] gardener/etcd-backup-restore#221
    - [x] gardener/etcd-backup-restore#242
    - [x] gardener/etcd-backup-restore#436
    - [x]  gardener/etcd-backup-restore#437
    - [x] gardener/etcd-backup-restore#368
    - [x] gardener/etcd-backup-restore#371
    - [x] gardener/etcd-backup-restore#251
    - [x] gardener/etcd-backup-restore#147
    - [x] gardener/etcd-backup-restore#286

  - [x] Etcd-Backup-Restore
    - [x] https://github.com/gardener/etcd-backup-restore/issues/333
    - [x]  https://github.com/gardener/etcd-backup-restore/issues/321
    - [x] https://github.com/gardener/etcd-backup-restore/issues/354
    - [x] https://github.com/gardener/etcd-backup-restore/issues/322
    - [x] https://github.com/gardener/etcd-backup-restore/issues/489
    - [x] https://github.com/gardener/etcd-backup-restore/issues/323
    - [x] https://github.com/gardener/etcd-wrapper/issues/7

[...]

PR #329, opened 2022-04-08
Add SchedulingConstraints to Etcd spec

shreyas-s-rao:

/area high-availability
/kind enhancement

This PR adds support for adding scheduling constraints to the etcd pod(s), via the `SchedulingConstraints` field in the Etcd template. The currently supported constraints are [`Affinity`](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) and [`TopologySpreadConstraints`](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/). This is required by https://github.com/gardener/gardener/pull/5741 .

Fixes partially #107 

cc @abdasgupta @unmarshall @timuthy @ashwani2k 

**Release note**:

```feature user
Add `SchedulingConstraints` field to Etcd spec. The currently supported constraints are `Affinity` and `TopologySpreadConstraints`.
```


charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -46,6 +46,16 @@ spec:
...
 {{- end }}
       serviceAccountName: {{ .Values.serviceAccountName }}
+{{- if .Values.schedulingConstraints }}
+{{- if .Values.schedulingConstraints.affinity }}
+      affinity:
+{{ toYaml .Values.schedulingConstraints.affinity | indent 8 }}
+{{- end }}
+{{- if .Values.schedulingConstraints.topologySpreadConstraints }}
+      topologySpreadConstraints:
+{{ toYaml .Values.schedulingConstraints.topologySpreadConstraints | indent 6 }}
```
abdasgupta, line 56: Should it not be `indent 8` ?

abdasgupta (changes requested):
Overall looks good. I posted just one doubt as review. Also, unit test is failing

charts/etcd/templates/etcd-statefulset.yaml
shreyas-s-rao: No, it's 6. Because the contents of this field is an array, so the hyphen starts at indentation 6. You can try it by running `helm template`

shreyas-s-rao:
@abdasgupta the test has passed now. It was from an unrelated error.

config/samples/druid_v1alpha1_etcd.yaml
```diff
@@ -62,6 +62,9 @@ spec:
   sharedConfig:
     autoCompactionMode: periodic
     autoCompactionRetention: "30m"
+  schedulingConstraints:
+    affinity: {}
+    topologySpreadConstraints: {}
```
aaronfern, line 67:
```suggestion
    topologySpreadConstraints: []
```
Since `topologySpreadConstraints` is an array

aaronfern (changes requested):
Thanks for the PR!
Looks good, just one small nit

shreyas-s-rao:
@aaronfern thank you so much for the review. I've fixed it now. PTAL

api/v1alpha1/etcd_types.go
```diff
@@ -242,6 +242,20 @@ type SharedConfig struct {
...
+// Currently supported constraints are Affinity and TopologySpreadConstraints.
+type SchedulingConstraints struct {
+	// Affinity defines the various affinity and anti-affinity rules for a pod
+	// that are honoured by the kube-scheduler.
+	// +optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
+	// TopologySpreadConstraints describes how a group of pods ought to spread across topology domains,
+	// that are honoured by the kube-scheduler.
+	// +optional
+	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
```
unmarshall, line 256: Minor point: do you still want to keep TSC? We decided that we will not be using TSC. When these are stable and later if we feel that these should be used, then we can add them as well.

shreyas-s-rao: We are still unsure whether we want to use TSC or not. Maybe it still might make sense for our usecase since e won't need to scale beyond 5 replicas. With this PR, I have just provided the option of using these scheduling constraints, while the gardenlet can actually take the decision of whether to use TSC or affinity or maybe a combination of both. This PR allows us to keep our options open.

unmarshall: Ok, this is fine.

aaronfern (approved):

charts/etcd/templates/etcd-statefulset.yaml
abdasgupta: Okay

PR #335, opened 2022-04-13
Added restoration indicator lease.

abdasgupta:

/area TODO
/kind TODO

The druid controller now creates a lease that will indicate whether a multinode ETCD cluster is restored from backup bucket or not. This lease is not used for single node ETCD cluster.
Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/pkg/component/etcd/lease/lease.go
+++ b/pkg/component/etcd/lease/lease.go
@@ -35,8 +35,9 @@ type component struct {
 
 func (c *component) Deploy(ctx context.Context) error {
 	var (
-		deltaSnapshotLease = c.emptyLease(c.values.DeltaSnapshotLeaseName)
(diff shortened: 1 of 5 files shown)


pkg/component/etcd/lease/lease.go
```diff
@@ -68,6 +74,10 @@ func (c *component) Destroy(ctx context.Context) error {
 		return err
 	}
 
+	if err := c.deleteSnapshotLease(ctx, clusterRestoreLease); err != nil {
```
aaronfern, line 77:
```suggestion
	if err := c.deleteClusterRestoreLease(ctx, clusterRestoreLease); err != nil {
```

aaronfern (changes requested):

timuthy:
@abdasgupta can we close this PR in favor of a planned single-node cluster restoration + expansion?

PR #336, opened 2022-04-23
fix deployment issue of etcd druid

jia-jerry:

/area ops-productivity
/kind bug

Fix issue that the etcd-druid-controller can't be deployed from the kustomize files which are located in /config folder.
Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,69 +7,151 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
-  - batch
+  - ""
   resources:
-  - jobs
+  - pods
   verbs:
-  - create
+  - list
+  - watch
   - delete
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  - endpoints
+  verbs:
   - get
   - list
   - patch
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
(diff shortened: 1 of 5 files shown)


config/samples/druid_v1alpha1_etcd.yaml
```diff
@@ -64,7 +64,13 @@ spec:
...
   schedulingConstraints:
     affinity: {}
-    topologySpreadConstraints: []
+    topologySpreadConstraints:
+    - maxSkew: 1
+      topologyKey: topology.kubernetes.io/zone
+      whenUnsatisfiable: DoNotSchedule
+      labelSelector:
+        matchLabels:
+          app: etcd-statefulset
```
jia-jerry, line 73:
```suggestion
```

config/rbac/role.yaml
```diff
@@ -7,69 +7,151 @@ metadata:
   name: manager-role
```
aaronfern, line 7: In this file, there are a few rules that have the same `apiGroups` and  `verbs`.
Could you please consolidate them into a single rule rather than have multiple different ones?

aaronfern (changes requested):
Thanks for the PR @jia-jerry!

One observation from me

config/rbac/role.yaml
jia-jerry: I just split the rules by different `apiGroups` and `verbs`. I think it might be more clear.

aaronfern: Alright, fair enough

aaronfern (approved):

PR #343, opened 2022-05-06
Add etcd object labels to configmap

aaronfern:

/area robustness
/kind bug

Fixes a bug where etcd object labels are not added to the configmap
This is required in the event of a rollback to a previous `etcd-druid` version. In the current implementation, `conigmap` is created with 2 standard labels using the component concept. However in the event of a rollback to a previous `etcd-druid` version, the `claimConfigmap` func returns no matching cm found as it tries to list based on the etcd object labels. `etcd-druid` then assumes that the `configmap` does not exist and tries to create a new `configmap` which then fails (with the error shown below) as the `configmap` actually exists.
This can lead to a scenario where etcd reconciliation gets stuck

```
2022-05-06T19:33:36.023+0530	ERROR	controller.etcd	Reconciler error	{"reconciler group": "druid.gardener.cloud", "reconciler kind": "Etcd", "name": "etcd-main", "namespace": "default", "error": "configmaps \"etcd-bootstrap-cd7e2b\" already exists"}
```

Fixes #

cc @ishan16696 

**Release note**:

```other operator

```

### DIFF
--- a/pkg/component/etcd/configmap/configmap.go
+++ b/pkg/component/etcd/configmap/configmap.go
@@ -229,6 +229,9 @@ func (c *component) emptyConfigmap(name string) *corev1.ConfigMap {
 
 func getObjectMeta(val *Values) metav1.ObjectMeta {
 	labels := map[string]string{"name": "etcd", "instance": val.EtcdName}
+	for key, value := range val.Labels {
+		labels[key] = value
+	}
 	return metav1.ObjectMeta{
 		Name:      val.ConfigMapName,
 		Namespace: val.EtcdNameSpace,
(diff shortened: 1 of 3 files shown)


ishan16696:
We will be having a hotfix release of backup-restore and druid for this PR https://github.com/gardener/etcd-backup-restore/pull/476.
Should we inlcude this also @aaronfern ?

aaronfern:
gardener/etcd-backup-restore#476 would go in `v0.8.4`, right?
This PR should be going in `v0.9.x` since thats where we use the component concept for configmaps

ishan16696:
yes, my bad

timuthy (changes requested):
Can you please enhance the unit tests to also cover labels and add a proper release note? The rest looks fine to me.

aaronfern:
Thanks for the review @timuthy!
I updated the tests and added a release note. PTAL

timuthy (approved):
/lgtm
Thanks for the fix.

Issue #280, opened 2021-12-22
[Feature] Liveness Probe on multi-node etcd


ishan16696:
**Feature (What you would like to be added):**
Configure the correct livenessProbe in `multi-node etcd`.

**Motivation (Why is this needed?):**
Right now I found 2 ways to configure the livenessProbe:
1.  `/health`  endpoint
`/health` endpoint returns `false` if one of the following conditions is met ([source](https://github.com/etcd-io/etcd/blob/v3.4.14/etcdserver/api/etcdhttp/metrics.go#L106-L119)):
    - there is no etcd leader or leader-election is currently going on.
    - the latency of a QGET request exceeds 1sec

 **Disadvantage** of Method 1 (`/health` endpoint).
   -  If there is no Quorum, etcd-follower will get restart by kubelet but etcd's restart isn't a right behavior.
   -  If etcd cluster is overloaded which is bad, but restart will generate even more load

2.  `ETCDCTL_API=3 etcdctl endpoint health --endpoints=${ENDPOINTS} --command-timeout=Xs` 
`etcdctl endpoint health` command performs a GET on the "health" key([source](https://github.com/etcd-io/etcd/blob/main/etcdctl/ctlv3/command/ep_command.go#L126-L133))
    - fails if there is no etcd leader or leader-election is going on as I think GET request will fail if there is no etcd leader.

  **Disadvantage** of Method 2 (`etcdctl endpoint health`).
  -  If there is no Quorum, etcd-follower will get restart by kubelet but etcd's restart isn't a right behavior.

This issue of liveness is already reported on etcd [here](https://github.com/etcd-io/etcd/issues/13340) and it has been fixed by this [PR](https://github.com/etcd-io/etcd/pull/13399) but this feature will come with etcd v3.6 and we don't know the timeline when it will be released.

**What options we have:**
1. Go with the method 2(`etcdctl endpoint health`) and whenever etcd release v3.6, switch to that version and configure the correct liveness probe.
2. Other way is fork the etcd v3.4.13 and make the necessary changes and use the custom image.

ishan16696:
cc @abdasgupta @aaronfern @timuthy

timuthy:
Reading this I wonder about the option keeping the current liveness probe but setting `--consistency="s"`? This mimics the implementation of the linked PR https://github.com/etcd-io/etcd/pull/13399. WDYT @ishan16696?

ishan16696:
[...]

PR #357, opened 2022-06-08
Update the readiness/liveness Probes configuration.

ishan16696:
1. Adds `failureThreshold` to **5** for both readiness and liveness Probe.
2. Update the livenessProbe for etcd making the consistency `serializable` using `--consistency="s"`

Fixes https://github.com/gardener/etcd-druid/issues/280 partially as there is still a room for improvement([more info](https://github.com/gardener/etcd-druid/issues/280#issuecomment-1143639829))

For values of initialDelaySeconds and failureThreshold can be discussed. I'm open for change.
cc @timuthy 

**Release note**:
```other operator
livenessProbe of etcd container has been updated to `ETCDCTL_API=3 etcdctl get foo --consistency=s` making the consistency `serializable`. 
```
```other operator
failureThreshold has been updated to `5` for both livenessProbe and readinessProbe of etcd.
```


ishan16696:
Should we include this PR as well in our upcoming release of druid ?

cc @abdasgupta @timuthy @aaronfern

charts/etcd/templates/etcd-statefulset.yaml
```diff
@@ -99,8 +100,10 @@ spec:
 # {{- end }}
             - get
             - foo
-          initialDelaySeconds: 15
+            - --consistency="s"
+          initialDelaySeconds: 30
```
timuthy, line 104: I suggest to first stick to the `15s` as long as we don't see a need for it. Otherwise this setting will stretch the start up period for all etcds.

timuthy (commented):
Thanks for the PR @ishan16696. Please also add a release note so that users are aware of the change.

ishan16696:
Added the release note. PTAL @timuthy

timuthy (approved):
/lgtm

PR #360, opened 2022-06-23
Change base image from alpine to distroless

dimityrmirchev:

/area open-source security
/kind enhancement

This PR changes the base image for `etcd-druid` from `alpine` to [distroless](https://github.com/GoogleContainerTools/distroless). The processes will now use a non root user for their execution. This will reduce the attack surface of the image.

Fixes #

As far as I saw there is no need to have `bash` installed in the `etcd-druid` container.

**Release note**:

```other operator
The `etcd-druid` now uses `distroless` instead of `alpine` as a base image.
```


timuthy (approved):
/lgtm

Issue #278, opened 2021-12-21
Implement  "component" concept for all the K8s resources


abdasgupta:
**Feature (What you would like to be added):**
Implement  "component" concept for all the K8s resources to replace the [chart](https://github.com/gardener/etcd-druid/tree/v0.7.3/charts/etcd/templates) that we use in Druid and instead build the K8s resources within the code

**Motivation (Why is this needed?):**
To achieve a better test-ability and more control how the client is used to reconcile the objects (e.g. used verb `UPDATE` vs `PATCH`) or how merging objects will be implemented. The same approach was taken in `gardener/gardener`.

Please see [this comment](https://github.com/gardener/gardener/issues/2754#issuecomment-678301676) for more information.

**Approach/Hint to the implement solution (optional):**
Please follow [this PR](https://github.com/gardener/etcd-druid/pull/262) for the idea to implement.

- [x] Lease: #262
- [x] Service: #273
- [x] ConfigMap: #193
- [x] PodDisruptionBudget #430 
- [x] StatefulSet #365
- [x] RBAC (Role + RoleBinding) #538 #539 
- [x] ServiceAccount #535 
- [x] https://github.com/gardener/etcd-druid/pull/559

abdasgupta:
cc @ashwani2k @aaronfern @ishan16696 @timuthy

abdasgupta:
We have two of the k8s resources to be left to be converted with golang code. We will pick them up when we are less overloaded.

shreyas-s-rao:
Added another task to the original list in issue description: `Job (for EtcdCopyBackupsTask Controller)`

seshachalam-yv:
Raised PRs for followings components,

- ServiceAccount - #535
- Role - #538
- RoleBinding - #539

PR #365, opened 2022-07-04
Deploy statefulset as component.

abdasgupta:

/area robustness
/kind enhancement

This PR converts the statefulset yaml to client-go code to create ETCD statefulset programmatically. Earlier an example yaml was decoded and merged with correct values to create statefulset yaml. But this PR removes the use of yaml. Instead, it fetches values from ETCD spec and directly creates statefulset using client-go for kubernetes API.

This PR also removes the logic for claiming statefulset based on label selector.

Fixes part of #278 

**Release note**:

```feature developer
Deploying the etcd StatefulSet through a Helm chart has been abandoned. A codified version (component concept) is now used for this purpose.
```
```breaking operator
`etcd` Statefulsets are not claimed anymore based on labels. Instead, the statefulsets are fetched using Name and Namespace combination. Thus, `etcd.spec.selector` does not have an effect on statefulsets anymore.
```


aaronfern (changes requested):
Please remove [this](https://github.com/gardener/etcd-druid/blob/009a00f84103d61214e07d37db6a7030be15047e/controllers/controller_ref_manager.go#L229) function as it is not used anymore

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,683 @@
...
+func getEtcdCommand() []string {
+	command := []string{"" + "/var/etcd/bin/bootstrap.sh"}
+
+	return command
+}
+
+func getReadinessProbeCommand(val Values) []string {
+	command := []string{"" + "/usr/bin/curl"}
+
+	if val.ClientUrlTLS != nil {
```
shreyas-s-rao, line 133: Since the URL points to either etcd server or backup-restore server based on number of replicas, the first decision point should be the replicas, and next decision point should be for whether TLS is enabled on the corresponding server, ie, if replicas==1, then check for whether `backup.tls` is present, else check for whether `etcd.clientUrlTls` is present. Please re-structure this function accordingly.

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,683 @@
...
+		quota = val.Quota.Value()
+	}
+
+	command = append(command, "--embedded-etcd-quota-bytes="+fmt.Sprint(quota))
+
+	if pointer.BoolDeref(val.EnableProfiling, false) {
+		command = append(command, "--enable-profiling=true")
+	}
+
+	if val.BackupTLS != nil {
```
shreyas-s-rao, line 232: The check here should be for `etcd.clientUrlTls`, since the endpoint being called is an etcd server endpoint.

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -0,0 +1,239 @@
...
+								InitialDelaySeconds: 15,
+								PeriodSeconds:       5,
+								FailureThreshold:    5,
+							},
+							Ports:     getEtcdPorts(c.values),
+							Resources: getEtcdResources(c.values),
+							Env: []v1.EnvVar{
+								{
+									Name:  "ENABLE_TLS",
+									Value: c.values.EnableClientTLS,
```
shreyas-s-rao, line 142: `ENABLE_TLS` is actually used to check whether backup-restore server has TLS enabled. So the check here should be for `backupTLS`.

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -0,0 +1,239 @@
...
+								},
+								{
+									Name:  "BACKUP_ENDPOINT",
+									Value: strconv.FormatBool(c.values.BackupTLS != nil),
```
shreyas-s-rao, line 146: `BACKUP_ENDPOINT` is the backup-restore server endpoint, which would look like `http(s)://......:8080`. Please also set the correct `http`/`https` value based on whether `backupTLS` is enabled or not.

shreyas-s-rao (changes requested):
@abdasgupta thanks for the PR. I have done a partial review of this PR, mainly keeping https://github.com/gardener/etcd-druid/issues/374 in mind. This PR seems to be the best place to fix the particular bug. Please check my comments and make the necessary changes.
You can also include the above issue in your PR note, mentioning that this PR will fix it.

controllers/etcd_controller.go
```diff
@@ -869,6 +642,8 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
 		return nil, nil, err
 	}
 
+	val.StatefulSet = statefulset.GenerateValues(etcd)
```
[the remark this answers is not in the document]
aaronfern: Resolved in an ad-hoc sync

controllers/etcd_controller.go
```diff
@@ -889,11 +664,24 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
...
+
 	if err != nil {
 		return nil, nil, err
 	}
 
+	sts := &appsv1.StatefulSet{}
+	err = r.Get(ctx, types.NamespacedName{Name: etcd.Name, Namespace: etcd.Namespace}, sts)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Cound not fetch statefulset after deploying a statefulset")
+	}
```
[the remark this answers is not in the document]
aaronfern: Resolved in an ad-hoc sync

pkg/component/etcd/statefulset/values_helper.go
abdasgupta: Made the necessary changes

pkg/component/etcd/statefulset/statefulset.go
abdasgupta: Done

pkg/component/etcd/statefulset/statefulset.go
abdasgupta: Done

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,695 @@
...
+	command := []string{"" + "/var/etcd/bin/bootstrap.sh"}
+
+	return command
+}
+
+func getReadinessProbeCommand(val Values) []string {
+	command := []string{"" + "/usr/bin/curl"}
+
+	protocol := "http"
+	if val.BackupTLS != nil || val.ClientUrlTLS != nil {
```
shreyas-s-rao, line 134:
```suggestion
	if (val.Replicas == 1 && val.BackupTLS != nil) || (val.Replicas != 1 && val.ClientUrlTLS != nil) {
```
To ensure that backupTLS and ClientUrlTLS are decoupled.

shreyas-s-rao (dismissed):
@abdasgupta Thanks for making the requested changes. Just one small suggested change from my side before the PR is ready from the perspective of https://github.com/gardener/etcd-druid/issues/374. PTAL

controllers/etcd_controller.go
```diff
@@ -468,228 +463,6 @@ func (r *EtcdReconciler) getPodDisruptionBudgetFromEtcd(etcd *druidv1alpha1.Etcd
...
-
-	if reflect.DeepEqual(ss.Spec, decoded.Spec) {
-		return ss, nil
-	}
-
-	ssCopy := ss.DeepCopy()
-	ssCopy.Spec.Replicas = decoded.Spec.Replicas
-	ssCopy.Spec.UpdateStrategy = decoded.Spec.UpdateStrategy
-
-	recreateSTS := false
```
timuthy, line 612: Earlier we supported the recreation of the StatefulSet which was for instance required if the cluster was scaled up and the `spec.ServiceName` was changed in that scope (see following lines). I see that this logic has been removed but not added to the component. Am I missing something?

controllers/etcd_controller.go
```diff
@@ -939,6 +729,14 @@ func (r *EtcdReconciler) getMapFromEtcd(im imagevector.ImageVector, etcd *druidv
...
 	}
 
+	if etcd.Spec.Etcd.ClientPort != nil {
+		etcd.Spec.Etcd.ClientPort = pointer.Int32Ptr(val.Service.ClientPort)
+	}
+
+	if etcd.Spec.Etcd.ServerPort != nil {
+		etcd.Spec.Etcd.ServerPort = pointer.Int32Ptr(val.Service.ServerPort)
+	}
+
```
timuthy, line 739: Why is the `etcd` object changed here?

controllers/controller_ref_manager.go
```diff
@@ -210,56 +210,6 @@ func (m *EtcdDruidRefManager) FetchStatefulSet(ctx context.Context, etcd *druidv
...
-// Optional: If one or more filters are specified, a Statefulset will only be claimed if
-// all filters return true.
-//
-// A non-nil error is returned if some form of reconciliation was attempted and
-// failed. Usually, controllers should try again later in case reconciliation
-// is still needed.
-//
-// If the error is nil, either the reconciliation succeeded, or no
-// reconciliation was necessary. The list of statefulsets that you now own is returned.
-func (m *EtcdDruidRefManager) ClaimStatefulsets(ctx context.Context, statefulSetList *appsv1.StatefulSetList, filters ...func(*appsv1.StatefulSet) bool) ([]*appsv1.StatefulSet, error) {
```
timuthy, line 229: This is an incompatible change that really must be mentioned in the release notes.

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,695 @@
...
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: defaultLocalPrefix + "/" + *storeValues.Container,
+					Type: &hpt,
+				},
+			},
+		})
+	}
+
+	if provider == "GCS" || provider == "S3" || provider == "OSS" || provider == "ABS" || provider == "Swift" || provider == "OCS" {
```
timuthy, line 571: Can you use public constants for these providers? Maybe make those (https://github.com/gardener/etcd-druid/blob/1cc064288605568386a89b1a8944b57fdd0fb31a/pkg/utils/miscellaneous.go#L52) public and use them here.

controllers/etcd_controller.go
```diff
@@ -889,12 +664,27 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
 		return nil, nil, err
 	}
 
-	sts, err := r.reconcileStatefulSet(ctx, logger, etcd, values)
+	stsDeployer := componentsts.New(r.Client, etcd.Namespace, val.StatefulSet)
+	err = stsDeployer.Deploy(ctx)
+	// Ignore the precondition violated error, this machine is already updated
+	// with the desired label.
+	if err == errorsutil.ErrPreconditionViolated {
```
timuthy, line 671: I think you can remove this. You don't find it in the most of our components either and TBH I've never experienced it myself (it's not even used in the K8s org https://github.com/search?q=org%3Akubernetes+ErrPreconditionViolated&type=code).

timuthy (changes requested):
Happy to see also the StatefulSet being transitioned to the component concept 👏 
Few points from my side and PLEASE update the PR description and add RELEASE NOTES.

controllers/etcd_controller.go
timuthy: Some fields are immutable, so you need to re-create the entire `StatefulSet` if the fields are changed for the supported use-cases.

abdasgupta: What I see that the recreation was called if a claimed statefulset was needed to be updated with immutable fields. But we are not claiming statefulset anymore. So, why do we need to update immutable fields of statefulset? Is there any need of claiming an existing statefulset? I may be missing something here

abdasgupta: Okay. I think, I got it. Will address this in next update of the PR shortly. Thanks.

timuthy: I don't see a pure relation to the claim logic. It is required to support:
- Changing `statefulset.spec.selector` (see [here](https://github.com/gardener/etcd-druid/blob/009a00f84103d61214e07d37db6a7030be15047e/controllers/etcd_controller.go#L613-L615)). Maybe this case is not very important anymore, esp. when we drop the claim logic.
- Changing `statefulset.spec.serviceName` (see [here](https://github.com/gardener/etcd-druid/blob/009a00f84103d61214e07d37db6a7030be15047e/controllers/etcd_controller.go#L617-L623)). Please see the comment in the code why this is required and important to support.

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -0,0 +1,229 @@
...
+	client    client.Client
+	namespace string
+
+	values Values
+}
+
+func (c *component) Deploy(ctx context.Context) error {
+	var (
+		etcdMainSts = c.emptyStatefulset(c.values.EtcdMainStsName)
+		//etcdEventSts = c.emptyStatefulset(c.values.EtcdEventStsName)
```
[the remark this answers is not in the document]
abdasgupta: Removed

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -0,0 +1,229 @@
...
+	)
+
+	if err := c.syncEtcdMainSts(ctx, etcdMainSts); err != nil {
+		return err
+	}
+
+	/*if err := c.syncEtcdEventSts(ctx, etcdEventSts); err != nil {
+		return err
+	}*/
```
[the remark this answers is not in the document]
abdasgupta: Removed

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -0,0 +1,229 @@
...
+type component struct {
+	client    client.Client
+	namespace string
+
+	values Values
+}
+
+func (c *component) Deploy(ctx context.Context) error {
+	var (
+		etcdMainSts = c.emptyStatefulset(c.values.EtcdMainStsName)
```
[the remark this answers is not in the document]
abdasgupta: Took care

pkg/component/etcd/statefulset/statefulset_test.go
```diff
@@ -0,0 +1,716 @@
...
+	serverPort       int32 = 2380
+	backupPort       int32 = 8080
+	uid                    = "a9b8c7d6e5f4"
+	imageEtcd              = "eu.gcr.io/gardener-project/gardener/etcd:v3.4.13-bootstrap"
+	imageBR                = "eu.gcr.io/gardener-project/gardener/etcdbrctl:v0.12.0"
+	snapshotSchedule       = "0 */24 * * *"
+	defragSchedule         = "0 */24 * * *"
+	container              = "default.bkp"
+	storageCapacity        = resource.MustParse("5Gi")
+	//defaultStorageCapacity        = resource.MustParse("16Gi")
```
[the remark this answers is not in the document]
abdasgupta: Removed

pkg/component/etcd/statefulset/statefulset_test.go
```diff
@@ -0,0 +1,716 @@
...
+	storageClass            = "gardener.fast"
+	priorityClassName       = "class_priority"
+	deltaSnapShotMemLimit   = resource.MustParse("100Mi")
+	autoCompactionMode      = druidv1alpha1.Periodic
+	autoCompactionRetention = "2m"
+	quota                   = resource.MustParse("8Gi")
+	//provider                      = druidv1alpha1.StorageProvider("Local")
```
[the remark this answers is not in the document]
abdasgupta: Removed

pkg/component/etcd/statefulset/statefulset_test.go
```diff
@@ -0,0 +1,716 @@
...
+	prefix                  = "/tmp"
+	volumeClaimTemplateName = "etcd-main"
+	garbageCollectionPolicy = druidv1alpha1.GarbageCollectionPolicy(druidv1alpha1.GarbageCollectionPolicyExponential)
+	metricsBasic            = druidv1alpha1.Basic
+	//maxBackups                    = 7
```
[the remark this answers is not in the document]
abdasgupta: Removed

pkg/component/etcd/statefulset/values.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type Values struct {
```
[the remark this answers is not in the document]
abdasgupta: Added

controllers/etcd_controller.go
abdasgupta: Addressed

pkg/component/etcd/statefulset/values_helper.go
abdasgupta: Done

controllers/etcd_controller.go
abdasgupta: Removed

pkg/component/etcd/statefulset/values_helper.go
abdasgupta: Done

controllers/controller_ref_manager.go
abdasgupta: Done

controllers/etcd_controller.go
abdasgupta: It was needed because statefulset needs the value of ports from ETCD spec. I needed to pass the port values to syncStatefulset() function. For that, either I needed to pass the port values as arguments to GenerateValues() function or pass it through ETCD spec. I took the later approach as the earlier one would need to write a function with long parameter list.

timuthy: I'd nevertheless propose to do it via the GenerateValues() function because changing the `etcd` object in place can cause unexpected results. So if you want to actually write back the port that we should also consider updating the etcd object at the end. If this is not intended or necessary then we should also not manipulate the object.

abdasgupta: Okay. Addressed

pkg/component/etcd/statefulset/statefulset_test.go
```diff
@@ -0,0 +1,713 @@
...
+
+				sts := &appsv1.StatefulSet{}
+
+				Expect(cl.Get(ctx, kutil.Key(namespace, values.StsName), sts)).To(Succeed())
+				checkStatefulset(sts, values)
+
+			})
+		})
+
+		Context("when statefulset exists", func() {
```
timuthy, line 147: Can you add another case for which `spec.serviceName` is changed?

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,696 @@
...
+	storeValues := val.BackupStore
+
+	env = append(env, getEnvVarFromValues("STORAGE_CONTAINER", *storeValues.Container))
+
+	provider, err := utils.StorageProviderFromInfraProvider(val.BackupStore.Provider)
+	if err != nil {
+		return env
+	}
+
+	if provider == "S3" {
```
timuthy, line 371: It's not necessary to go through all cases as there can only be one provider. I suggest to have a `switch case` here.

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,696 @@
...
+	endpoint := fmt.Sprintf("%s://%s-local:%d", protocol, val.EtcdName, pointer.Int32Deref(val.BackupPort, defaultBackupPort))
+
+	var env []corev1.EnvVar
+	env = append(env, getEnvVarFromValues("ENABLE_TLS", strconv.FormatBool(val.BackupTLS != nil)))
+	env = append(env, getEnvVarFromValues("BACKUP_ENDPOINT", endpoint))
+
+	return env
+}
+
+func getStsEnvVar(val Values) []corev1.EnvVar {
```
timuthy, line 352:
```suggestion
func getBackupRestoreEnvVar(val Values) []corev1.EnvVar {
```

timuthy (changes requested):
Please see my comments after another review round. In addition:
- Can you delete `charts/etcd/templates/etcd-statefulset.yaml` and the unused values in `charts/etcd/values.yaml`?
- To demonstrate some points I changed the code myself and added it to my fork https://github.com/timuthy/etcd-druid/commits/pr.365. I referred to specific commits in the comments. Please let me know if you want to discuss further.

ashwani2k:
Somehow this doesn't read correct to me in the release notes.
```
- yaml is not used anymore to create ETCD statefulset. Instead, ETCD statefulsets are created through client-go.
```

I checked the previous PR's for component model for configmap #314  and that reads differently and more apt.  
Why is there a different release note for this change viz-a-viz past changes of similar nature.

timuthy:
In addition I see that the at the release note about a breaking change is not assigned the right category. It's a **breaking** change for **operators** and thus it should have:


\`\`\`breaking operator
ETCD Statefulsets are not claimed anymore based on labels. Instead, the statefulsets are fetched using Name and Namespace combination. `ClaimStatefulsets` function has been removed.
\`\`\`

timuthy:
FYI: The PR is still WIP. During tests I found some differences in the yaml based etcd <-> component etcd which are not supposed to happen.

timuthy:
@gardener/etcd-druid-maintainers the PR is ready for reviewing.

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,352 @@
...
+import (
+	"fmt"
+
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+	"github.com/gardener/etcd-druid/pkg/utils"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/utils/pointer"
+)
+
+var (
```
unmarshall, line 26: All of these except `defaultStorageCapacity` should ideally be declared as constants.

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,352 @@
...
+
+	return values
+}
+
+func getEtcdCommand() []string {
+	return []string{"/var/etcd/bin/bootstrap.sh"}
+}
+
+func getReadinessProbeCommand(val Values) []string {
+	command := []string{"" + "/usr/bin/curl"}
```
unmarshall, line 132: why add an empty string?

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,352 @@
...
+
+	command = append(command, "--data-dir=/var/etcd/data/new.etcd")
+
+	if val.BackupStore != nil {
+		store, _ := utils.StorageProviderFromInfraProvider(val.BackupStore.Provider)
+		command = append(command, "--storage-provider="+store)
+		command = append(command, "--store-prefix="+string(val.BackupStore.Prefix))
+	}
+
+	var quota int64 = 8 * 1024 * 1024 * 1024 // 8Gi
```
unmarshall, line 237: Minor: Can be defined as a `defaultQuota` constant

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -0,0 +1,352 @@
...
+		command = append(command, "--server-key=/var/etcd/ssl/client/server/tls.key")
+	}
+
+	command = append(command, "--etcd-connection-timeout="+defaultEtcdConnectionTimeout)
+
+	if val.DeltaSnapshotPeriod != nil {
+		command = append(command, "--delta-snapshot-period="+val.DeltaSnapshotPeriod.Duration.String())
+	}
+
+	var deltaSnapshotMemoryLimit int64 = 100 * 1024 * 1024 // 100Mi
```
unmarshall, line 272: Minor: can be defined above as constant

pkg/utils/miscellaneous.go
```diff
@@ -211,20 +219,20 @@ func StorageProviderFromInfraProvider(infra *druidv1alpha1.StorageProvider) (str
 	}
 
 	switch *infra {
-	case aws, s3:
```
unmarshall, line 214: Could not leave the comment at an appropriate place, so leaving it here. Function `ValueExists` is not used anywhere. Do you need it?

controllers/controller_ref_manager.go
unmarshall, line 229: Why is this no longer needed? Can you explain?

controllers/etcd_controller.go
```diff
@@ -379,10 +377,6 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
 		return ctrl.Result{
 			Requeue: true,
 		}, err
-	} else if waitForStatefulSetCleanup {
```
unmarshall, line 382: stsDeployer.Destroy currently does not wait. Should this also call `WaitCleanup`? Maybe use `gardenercomponent.OpWaiter`?

controllers/etcd_controller.go
```diff
@@ -840,31 +612,48 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
...
-		Service:   componentservice.GenerateValues(etcd),
+	etcdImage, etcdBackupImage, err := getEtcdImages(r.ImageVector, etcd)
+	if err != nil {
+		return nil, nil, err
 	}
 
-	leaseDeployer := componentlease.New(r.Client, etcd.Namespace, val.Lease)
+	if etcd.Spec.Etcd.Image == nil {
+		if etcdImage == "" {
+			return nil, nil, fmt.Errorf("either etcd resource or image vector should have %s image while deploying statefulset", common.Etcd)
```
unmarshall, line 622: This will result in re-queueing of request and since the image is now nil this will keep getting re-queued right?

controllers/etcd_controller.go
```diff
@@ -840,31 +612,48 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
...
+		}
+	} else {
+		etcdImage = *etcd.Spec.Etcd.Image
+	}
+
+	if etcd.Spec.Backup.Image == nil {
+		if etcdBackupImage == "" {
+			return nil, nil, fmt.Errorf("either etcd resource or image vector should have %s image while deploying statefulset", common.BackupRestore)
```
unmarshall, line 630: This will result in re-queueing of request and since the image is now nil this will keep getting re-queued right?

pkg/component/etcd/statefulset/values.go
```diff
@@ -0,0 +1,128 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Values contains the values necessary for creating ETCD statefulset.
+type Values struct {
+	// Name is the name of the etcd resource.
+	Name string
```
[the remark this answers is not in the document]
timuthy: Hm, Name is the name to be used for the `StatefulSet` and since it's in the StatefulSet component package, this is evident I think.
Unlike UID which is really the UID of the `etcd` and not the one from the StatefulSet or any other object.

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -0,0 +1,726 @@
...
+		default:
+			return retry.SevereError(err)
+		}
+	})
+}
+
+func (c *component) syncStatefulset(ctx context.Context, sts *appsv1.StatefulSet, create bool) error {
+	patch := client.StrategicMergeFrom(sts.DeepCopy())
+
+	sts.ObjectMeta = getObjectMeta(&c.values)
```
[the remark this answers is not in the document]
timuthy: How are the `createStatefulset` and `patchStatefulset` functions supposed to look like? I assume you then have to create another function like `prepareStatefulSetObject` that contains the common logic how the object is being constructed. I can change that structure if you insist but I don't see a big difference.

pkg/utils/statefulset.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+// CheckStatefulSet checks whether the given StatefulSet is healthy.
+// A StatefulSet is considered healthy if its controller observed its current revision,
+// it is not in an update (i.e. UpdateRevision is empty) and if its current replicas are equal to
+// desired replicas specified in ETCD specs.
+func CheckStatefulSet(etcdReplicas int32, statefulSet *appsv1.StatefulSet) error {
```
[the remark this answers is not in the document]
timuthy: The function actually does more than just returning an error. It also tells the reason for the error which is logged by the controller (currently via the `Wait` function: https://github.com/gardener/etcd-druid/blob/954f6ef747b7ffa4cc08d71f77e90aa4660b4677/pkg/component/etcd/statefulset/statefulset.go#L125)

```
last error: not enough ready replicas (2/3)	{"etcd": "default/etcd-test2"}
```

controllers/etcd_controller.go
timuthy: Actually it'll only be re-queued if the images are nil and `etcd` or `etcd-backup-restore` images cannot be found in `charts/images.yaml`, see https://github.com/gardener/etcd-druid/blob/954f6ef747b7ffa4cc08d71f77e90aa4660b4677/controllers/etcd_controller.go#L992

controllers/etcd_controller.go
timuthy: Nice catch 👍

controllers/controller_ref_manager.go
timuthy: Yes, we agreed a longer time ago that we're going to drop the claim functionality to simplify the entire reconciliation logic and handling.

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -0,0 +1,726 @@
...
+}
+
+const (
+	// defaultInterval is the default interval for retry operations.
+	defaultInterval = 5 * time.Second
+	// defaultTimeout is the default timeout for retry operations.
+	defaultTimeout = 90 * time.Second
+)
+
+func (c *component) Wait(ctx context.Context) error {
```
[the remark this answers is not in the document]
timuthy: Hm what should I say 😄 Now that I became used to the interface, I didn't notice it but essential you're right. However, it's not changeable in this PR as you already mentioned. I hope that it's fine to resolve this comment.

timuthy:
Thanks for your valuable feedback @unmarshall. I addressed most of the points in 9d31a9810f4db96e9b087bf3cb1829e6037d2ee3 and replied to the rest of the comments.

controllers/etcd_controller.go
unmarshall: So if the images cannot be found, re-queue will also never help, so continuously re-queueing will block on worker all the time with no real advantage. Is there any alert raised?

pkg/component/etcd/statefulset/values.go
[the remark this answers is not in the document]
unmarshall: Ok then the comment needs to be corrected.

unmarshall (approved):
/lgtm

controllers/etcd_controller.go
timuthy: Yes, you are right. I added a TODO that this is supposed to be part of a validating webhook. No, there is no alert but the error message is put to `etcd.status.lastError`.

timuthy (approved):

PR #372, opened 2022-07-11
Added pod permissions in etcd-role

aaronfern:

/area disaster-recovery

This PR adds a new rule in `etcd-role` to allow `etcd-backup-restore` to be able to access pods as needed by gardener/etcd-backup-restore#403

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/charts/etcd/templates/etcd-role.yaml
+++ b/charts/etcd/templates/etcd-role.yaml
@@ -37,3 +37,13 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch


charts/etcd/templates/etcd-role.yaml
```diff
@@ -36,4 +36,14 @@ rules:
...
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - patch
+  - update
```
timuthy, line 48: Why is `patch` and `update` required?

aaronfern: You're right, `patch` and `update` aren't required
Thanks for pointing out :)

timuthy (approved):
/lgtm

PR #373, opened 2022-07-12
command is /etcd-druid

vasu1124:
/area ops-productivity
/kind bug

kustomize uers want to directly reference the `/config/default` folder.
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

namespace: etcd-druid

resources:
  - https://github.com/gardener/etcd-druid/config/default?ref=vTAG
```

Fixes the wrong command `/bin/etcd-druid`. Alternatively, the deployment could remove the command.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -18,7 +18,7 @@ spec:
     spec:
       containers:
         - command:
-            - /bin/etcd-druid
+            - /etcd-druid
           args:
             - --enable-leader-election
           name: druid


aaronfern:
Thanks for the PR @vasu1124!

Can you also make a change [here](https://github.com/gardener/etcd-druid/blob/009a00f84103d61214e07d37db6a7030be15047e/config/default/manager_image_patch.yaml#L11) to update the `etcd-druid` image used to `v0.11.0`? This is the version that actually uses the `/etcd-druid` entry point

timuthy (approved):
/lgtm
I'll follow up with @aaronfern's comment https://github.com/gardener/etcd-druid/pull/373#issuecomment-1181737376.

PR #375, opened 2022-07-19
Update to Golang 1.18.4

timuthy:

/area security
/kind enhancement

This PR updates the used Golang version to `1.18.4`.

**Release note**:

```other operator
Etcd-Druid's Golang version has been update to `1.18.4.`.
```

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -24,11 +24,11 @@ etcd-druid:
                 build: ~
     steps:
       check:
-        image: 'golang:1.17.6'
+        image: 'golang:1.18.4'
       test:
-        image: 'golang:1.17.6'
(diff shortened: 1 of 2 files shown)


aaronfern (commented):
Don't we also need to update the go version specified in `go.mod`?

timuthy:
> Don't we also need to update the go version specified in `go.mod`?

That's a good point. It's not a must and I thought that I don't want to enforce developers to have Golang 1.18 on their machines to compile this project locally, as long as 1.17 is still supported.

aaronfern:
Fair enough. Although can you please open an issue about the go version in `go.mod` so that we don't forget about it?

aaronfern (approved):
/lgtm

PR #376, opened 2022-07-20
Added a design doc for single-member restoration in multi-node cluster.

### LINKED ISSUE #371: [Feature] Design doc on restoration of single member of a ETCD cluster when PVC is corrupted
We need to have a design discussion on restoration of single member with corrupted PVC in a multinode ETCD cluster

ishan16696:
This PR adds a design document for single-member restoration in multi-node cluster.

Fixes #371 

**Release note**:
```other operator
None
```

### DIFF
--- a/docs/single_member_restoration.md
+++ b/docs/single_member_restoration.md
@@ -1,4 +1,4 @@
-## Restoration of a single member in multi-node etcd deployed by etcd-druid.
+# Restoration of a single member in multi-node etcd deployed by etcd-druid.
 
 **Note**:
 - For a cluster with n members, we are proposing the solution to only single member restoration within a etcd cluster not the quorum loss scenario (when majority of members within a cluster fail).
@@ -8,9 +8,9 @@
 If a single etcd member within a multi-node etcd cluster goes down due to DB corruption/PVC corruption/Invalid data-dir then it needs to be brought back. Unlike in the single-node case, a minority member of a multi-node cluster can't be restored from the snapshots present in storage container as you can't restore from the old snapshots as it contains the metadata information of cluster which leads to **memberID mismatch** that prevents the new member from coming up as new member is getting it's metadata information from db which got restore from old snapshots.
 
 ## Solution
-- Corresponding backup-restore sidecar if detects that its corresponding etcd is down due to [data-dir corruption](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L177) or [Invalid data-dir](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L204)
-- Then, if it's a cluster with cluster size `>1` then backup-restore will first remove the etcd failed member using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir of failed etcd member.
+- If a corresponding backup-restore sidecar detects that its corresponding etcd is down due to [data-dir corruption](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L177) or [Invalid data-dir](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L204)
+- Then backup-restore will first remove the failing etcd member from the cluster using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir of failed etcd member.
 - It won't affect the etcd cluster as quorum is still maintained.
-- After successfully removing failed member from the cluster, backup-restore sidecar will try to add a new member in a cluster to get the same cluster size as before.
-- Backup-restore firstly adds new member as a [Learner](https://etcd.io/docs/v3.3/learning/learner/) using the [MemberAddAsLearner API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L42-L43) call, once learner was added to the cluster and it's get in sync with leader and becomes up-to-date then promote the learner to a voting member using [MemberPromote API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L51-L52) call.
+- After successfully removing failed etcd member from the cluster, backup-restore sidecar will try to add a new etcd member to a cluster to get the same cluster size as before.
+- Backup-restore firstly adds new member as a [Learner](https://etcd.io/docs/v3.3/learning/learner/) using the [MemberAddAsLearner API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L42-L43) call, once learner is added to the cluster and it's get in sync with leader and becomes up-to-date then promote the learner(non-voting member) to a voting member using [MemberPromote API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L51-L52) call.
 - So, the failed member first needs to be removed from the cluster and then added as a new member.
--- a/docs/single_member_restoration.md
+++ b/docs/single_member_restoration.md
@@ -1,0 +1,15 @@
+## Restoration of a single member in multi-node etcd-backup-restore cluster.
+
+**Note**:
+Please note here we are proposing the solution to only single member restoration within a cluster not the quorum loss scenario(when majority of members within a cluster fail).
+
+### Motivation:
+If a single etcd member within a multi-node etcd cluster goes down due to PVC corruption(invalid data-dir) then it needs to be brought back. But it can't be restore from the snapshots present in storage container as you can't restore a part of whole cluster.
+
+### Solution
+- Corresponding backup-restore sidecar if detects that its corresponding etcd is down due to [data-dir corruption](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L177) or [Invalid data-dir](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L204)
+- Then, if its is a cluster with cluster size>1 then backup-restore will first remove the etcd failed member using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir of failed etcd member.
+- It won't affect the etcd cluster as quorum is still maintained.
+- After successfully removing failed member from the cluster, backup-restore sidecar will try to add a new member in a cluster to get the same cluster size as before.
+- Backup-restore firstly add new member as a [Learner](https://etcd.io/docs/v3.3/learning/learner/) using the [MemberAddAsLearner API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L42-L43) call, once learner added to the cluster and its get in sync with leader and becomes up-to-date then promote the learner to a voting member using [MemberPromote API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L51-L52) call.
+- So, the failed member first needs to be removed removed from the cluster and then add a new member.
--- a/docs/single_member_restoration.md
+++ b/docs/single_member_restoration.md
@@ -1,15 +1,16 @@
 ## Restoration of a single member in multi-node etcd-backup-restore cluster.
 
 **Note**:
-Please note here we are proposing the solution to only single member restoration within a cluster not the quorum loss scenario(when majority of members within a cluster fail).
+- Here we are proposing the solution to only single member restoration within a cluster not the quorum loss scenario (when majority of members within a cluster fail).
+- In this proposal we are not targetting the recovery of single member which got separated from cluster due to [network partition](https://etcd.io/docs/v3.3/op-guide/failures/#network-partition).
 
-### Motivation:
-If a single etcd member within a multi-node etcd cluster goes down due to PVC corruption(invalid data-dir) then it needs to be brought back. But it can't be restore from the snapshots present in storage container as you can't restore a part of whole cluster.
+## Motivation
+If a single etcd member within a multi-node etcd cluster goes down due to PVC corruption(invalid data-dir) then it needs to be brought back. Unlike in the single-node case, a minority member of a multi-node cluster can't be restored from the snapshots present in storage container as you can't restore a part of whole cluster.
 
-### Solution
+## Solution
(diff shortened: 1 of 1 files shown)


docs/single_member_restoration.md
```diff
@@ -0,0 +1,15 @@
+## Restoration of a single member in multi-node etcd-backup-restore cluster.
+
+**Note**:
+Please note here we are proposing the solution to only single member restoration within a cluster not the quorum loss scenario(when majority of members within a cluster fail).
```
timuthy, line 4:
```suggestion
Please note here we are proposing the solution to only single member restoration within a cluster not the quorum loss scenario (when majority of members within a cluster fail).
```

docs/single_member_restoration.md
```diff
@@ -0,0 +1,15 @@
...
+
+### Motivation:
+If a single etcd member within a multi-node etcd cluster goes down due to PVC corruption(invalid data-dir) then it needs to be brought back. But it can't be restore from the snapshots present in storage container as you can't restore a part of whole cluster.
```
timuthy, line 7:
```suggestion
If a single etcd member within a multi-node etcd cluster goes down due to PVC corruption (invalid data-dir) then it needs to be brought back. Unlike in the single-node case, a member of a multi-node cluster can't restored from snapshots because of a metadata mismatch that prevents to run as the same member again.
```

IIUC 🙂

docs/single_member_restoration.md
```diff
@@ -0,0 +1,15 @@
...
+
+### Solution
+- Corresponding backup-restore sidecar if detects that its corresponding etcd is down due to [data-dir corruption](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L177) or [Invalid data-dir](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L204)
+- Then, if its is a cluster with cluster size>1 then backup-restore will first remove the etcd failed member using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir of failed etcd member.
```
timuthy, line 11:
```suggestion
- Then, if it's a cluster with cluster size `>1` then backup-restore will first remove the etcd failed member using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir.
```

docs/single_member_restoration.md
```diff
@@ -0,0 +1,15 @@
...
+- It won't affect the etcd cluster as quorum is still maintained.
+- After successfully removing failed member from the cluster, backup-restore sidecar will try to add a new member in a cluster to get the same cluster size as before.
+- Backup-restore firstly add new member as a [Learner](https://etcd.io/docs/v3.3/learning/learner/) using the [MemberAddAsLearner API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L42-L43) call, once learner added to the cluster and its get in sync with leader and becomes up-to-date then promote the learner to a voting member using [MemberPromote API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L51-L52) call.
```
timuthy, line 14:
```suggestion
- Backup-restore firstly adds new member as a [Learner](https://etcd.io/docs/v3.3/learning/learner/) using the [MemberAddAsLearner API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L42-L43) call, once learner was added to the cluster, it's in sync with leader and becomes up-to-date then promote the learner to a voting member using [MemberPromote API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L51-L52) call.
```

docs/single_member_restoration.md
```diff
@@ -0,0 +1,15 @@
...
+- So, the failed member first needs to be removed removed from the cluster and then add a new member.
```
timuthy, line 15:
```suggestion
- So, the failed member first needs to be removed removed from the cluster and then added as a new member.
```

docs/single_member_restoration.md
```diff
@@ -0,0 +1,15 @@
+## Restoration of a single member in multi-node etcd-backup-restore cluster.
```
timuthy, line 1:
```suggestion
# Restoration of a single member in multi-node etcd cluster.
```

timuthy (changes requested):
Thanks for the document. I'm wondering why the PR is in WIP?

ishan16696:
>  I'm wondering why the PR is in WIP?

Right now this is the design/approach which I'm following but I kept it as WIP as we might need to revise our design may be unknown edge case which may/may not arise. 
Moreover, I also wants to add a diagram/picture to explain design in a more better way.

docs/single_member_restoration.md
ishan16696: > because of a metadata mismatch that prevents to run as the same member again.

yes, you are correct but I don't want to go details as it might be ambiguous for a person who is not very familiar with etcd .... doubts like "Why snapshots(snapshot of whole cluster) not holding the metadata, why metadata mismatch? Can't we store metadata somewhere else and restore from snapshots and then update the member with correct metadata ?" 😅 .... these are some possibilities of a doubts.
May be I will try to add some points of `metadata mismatch` but I don't want to remove `as you can't restore a part of whole cluster`

docs/single_member_restoration.md
ishan16696: hmm .... Isn't mentioning only `multi-node etcd` might be ambiguous as we want to restore `multi-node etcd` but it's possible only when backup-restore should be running as sidecar. That's why I mentioned `multi-node etcd-backup-restore`. Although I don’t have any hard opinion on this. I'm open for change.

timuthy: Technically, it's an `etcd` cluster and there is no such thing as a `etcd-backup-restore` cluster. I think that's pretty clear if one uses Druid that a backup-restore side-car is involved.
However, I get your point and if you'd like to emphasize this then I propose to just mentioned it in an additional sentence.

docs/single_member_restoration.md
timuthy: Both sounds quite ambiguous as one who knows etcd a little understands that every member holds the data of the entire cluster, so by saying "restore a part of whole cluster" I got confused. I don't have a strong opinion but maybe you can find a better phrasing.

ishan16696: Updated .PTAL

docs/single_member_restoration.md
ishan16696: Updated. PTAL

docs/single_member_restoration.md
```diff
@@ -0,0 +1,16 @@
...
+- Then, if it's a cluster with cluster size `>1` then backup-restore will first remove the etcd failed member using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir of failed etcd member.
+- It won't affect the etcd cluster as quorum is still maintained.
+- After successfully removing failed member from the cluster, backup-restore sidecar will try to add a new member in a cluster to get the same cluster size as before.
+- Backup-restore firstly adds new member as a [Learner](https://etcd.io/docs/v3.3/learning/learner/) using the [MemberAddAsLearner API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L42-L43) call, once learner was added to the cluster and it's get in sync with leader and becomes up-to-date then promote the learner to a voting member using [MemberPromote API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L51-L52) call.
```
unmarshall, line 15: Is it possible that backup-restore container fails to promote a member even after several retries? At what point a corrective action be taken, what will it be and which actor will take that action? You cannot try and promote the member indefinitely as then you increase the risk of running the entire etcd cluster with zero tolerance to yet another failure.

docs/single_member_restoration.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Solution
+- Corresponding backup-restore sidecar if detects that its corresponding etcd is down due to [data-dir corruption](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L177) or [Invalid data-dir](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L204)
+- Then, if it's a cluster with cluster size `>1` then backup-restore will first remove the etcd failed member using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir of failed etcd member.
+- It won't affect the etcd cluster as quorum is still maintained.
```
unmarshall, line 13: So lets take the following case:
There are 3 etcd members - `etcd-1`, `etcd-2` and `etcd-3`.
There was data corruption in `etcd-2` - back-restore triggers the process of cleaning the data dir, removing the member and promoting the member. While this is being done, there is another member for which data corruption happens, back-up restore for that pod also tries the same thing. It removes the member but when it attempts to add the member as a learner it fails as you can have atmost one learner at a time. How do you propose to handle this situation? Also will Abhishek's proposal of quorum loss kick in here which yet another actor (etcd-druid) handles?

docs/single_member_restoration.md
ishan16696: > Is it possible that backup-restore container fails to promote a member 

`MemberPromote` call can be failed ....it is possible. 

> even after several retries?

I'm not sure even after several retries `MemberPromote` can be failed. 

> t what point a corrective action be taken, what will it be and which actor will take that action? You cannot try and promote the member indefinitely as then you increase the risk of running the entire etcd cluster with zero tolerance to yet another failure.

Do you have anything on your mind .... what action can be taken or when ..... but I would suggest that’s also depends on error we are getting while promoting the member.

docs/single_member_restoration.md
ishan16696: > add the member as a learner it fails as you can have atmost one learner at a time.

no, this will fail as there is no quorum present not because of `one learner at a time`.
And for quorum loss:
yes, I'm assuming Abhishek's proposal of quorum loss should take care of this as I mentioned in doc as well that .... `this proposal doesn’t take care of quorum loss`.

docs/single_member_restoration.md
```diff
@@ -0,0 +1,16 @@
+## Restoration of a single member in multi-node etcd deployed by etcd-druid.
```
timuthy, line 1:
```suggestion
# Restoration of a single member in multi-node etcd deployed by etcd-druid.
```

docs/single_member_restoration.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Motivation
+If a single etcd member within a multi-node etcd cluster goes down due to DB corruption/PVC corruption/Invalid data-dir then it needs to be brought back. Unlike in the single-node case, a minority member of a multi-node cluster can't be restored from the snapshots present in storage container as you can't restore from the old snapshots as it contains the metadata information of cluster which leads to **memberID mismatch** that prevents the new member from coming up as new member is getting it's metadata information from db which got restore from old snapshots.
+
+## Solution
+- Corresponding backup-restore sidecar if detects that its corresponding etcd is down due to [data-dir corruption](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L177) or [Invalid data-dir](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L204)
```
timuthy, line 11:
```suggestion
- If a corresponding backup-restore sidecar detects that its corresponding etcd is down due to [data-dir corruption](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L177) or [Invalid data-dir](https://github.com/gardener/etcd-backup-restore/blob/7d27a47f5793b0949492d225ada5fd8344b6b6a2/pkg/initializer/validator/datavalidator.go#L204)
```

docs/single_member_restoration.md
```diff
@@ -0,0 +1,16 @@
...
+- Then, if it's a cluster with cluster size `>1` then backup-restore will first remove the etcd failed member using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir of failed etcd member.
```
timuthy, line 12:
```suggestion
- Then backup-restore will first remove the failing etcd member using the [MemberRemove API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L45-L46) call and clean the data-dir.
```

It's already mentioned in the headline that this is only for multi-node clusters.

docs/single_member_restoration.md
```diff
@@ -0,0 +1,16 @@
...
+- It won't affect the etcd cluster as quorum is still maintained.
+- After successfully removing failed member from the cluster, backup-restore sidecar will try to add a new member in a cluster to get the same cluster size as before.
```
timuthy, line 14:
```suggestion
- After successfully removing failed member from the cluster, backup-restore sidecar will try to add a new member to a cluster to get the same cluster size as before.
```

docs/single_member_restoration.md
timuthy: > You cannot try and promote the member indefinitely as then you increase the risk of running the entire etcd cluster with zero tolerance to yet another failure.

Same is true for the scale-up case. Any suggestion?

timuthy (commented):
Only some nits from my side.

docs/single_member_restoration.md
```diff
@@ -0,0 +1,23 @@
+# Restoration of a single member in multi-node etcd deployed by etcd-druid.
+
+**Note**:
+- For a cluster with n members, we are proposing the solution to only single member restoration within a etcd cluster not the quorum loss scenario (when majority of members within a cluster fail).
+- In this proposal we are not targetting the recovery of single member which got separated from cluster due to [network partition](https://etcd.io/docs/v3.3/op-guide/failures/#network-partition).
+
+## Motivation
+If a single etcd member within a multi-node etcd cluster goes down due to DB corruption/PVC corruption/Invalid data-dir then it needs to be brought back. Unlike in the single-node case, a minority member of a multi-node cluster can't be restored from the snapshots present in storage container as you can't restore from the old snapshots as it contains the metadata information of cluster which leads to **memberID mismatch** that prevents the new member from coming up as new member is getting it's metadata information from db which got restore from old snapshots.
```
timuthy, line 8:
```suggestion
If a single etcd member within a multi-node etcd cluster goes down due to DB corruption/PVC corruption/Invalid data-dir then it needs to be brought back. Unlike in the single-node case, a minority member of a multi-node cluster can't be restored from the snapshots present in storage container as you can't restore from the old snapshots as it contains the metadata information of cluster which leads to **memberID mismatch** that prevents the new member from coming up as new member is getting its metadata information from db which got restore from old snapshots.
```

docs/single_member_restoration.md
```diff
@@ -0,0 +1,23 @@
...
+- It won't affect the etcd cluster as quorum is still maintained.
+- After successfully removing failed etcd member from the cluster, backup-restore sidecar will try to add a new etcd member to a cluster to get the same cluster size as before.
+- Backup-restore firstly adds new member as a [Learner](https://etcd.io/docs/v3.3/learning/learner/) using the [MemberAddAsLearner API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L42-L43) call, once learner is added to the cluster and it's get in sync with leader and becomes up-to-date then promote the learner(non-voting member) to a voting member using [MemberPromote API](https://github.com/etcd-io/etcd/blob/ae9734ed278b7a1a7dfc82e800471ebbf9fce56f/clientv3/cluster.go#L51-L52) call.
+- So, the failed member first needs to be removed from the cluster and then added as a new member.
+
+### Example: 
+1. If a `3` member etcd cluster has 1 downed member(due to invalid data-dir), the cluster can still make forward progress because the quorum is `2`.
+2. Backup-restore detects this downed member and then removed the downed member from cluster.
```
timuthy, line 20: Can you explain that this only happens if the `etcd` container restarts and the detection is triggered through the initialization step?

ishan16696: PTAL. I mention some necessary informariton, don't want to go much in details.

timuthy (approved):
/lgtm

PR #381, opened 2022-07-22
Fix golangci for 1.18

timuthy:

/area delivery
/kind bug

This PR fixes the current build which is blocked by an `Golangci` failure after upgrading to Go 1.18.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ IMAGE_REPOSITORY    := $(REGISTRY)/etcd-druid
 IMAGE_TAG           := $(VERSION)
 BUILD_DIR           := build
 
+# TODO(timuthy): Remove this as soon as vendored to new gardener/gardener version.
+GOLANGCI_LINT_VERSION := v1.45.2
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -20,6 +20,9 @@ IMAGE_REPOSITORY    := $(REGISTRY)/etcd-druid
 IMAGE_TAG           := $(VERSION)
 BUILD_DIR           := build
 
+# TODO: Remove this as soon as vendored to new gardener/gardener version.
```
abdasgupta, line 23: Don't you want to put your name with TODO? Like TODO(timuthy) ? So that, in future if anybody wants to work on this TODO, they can connect you for context.

abdasgupta (approved):
Thanks for the PR. Just a minor comment.

Makefile
timuthy: Done. PTAL 🙂

shreyas-s-rao (approved):
/lgtm

PR #382, opened 2022-07-25
Added the logic for quorum loss scenario

### LINKED ISSUE #362: [Feature] Handle quorum loss scenario by Druid in ETCD multinode
We need to reconcile ETCD cluster in case of quorum loss

When n/2+1 number of ETCD members are down in a cluster of size n, the cluster becomes unavailable. This situation is called Quorum Loss. To recover from quorum loss scenario, ETCD Druid needs to delete ETCD statefulset and redeploy again.
ETCD druid will delete the statefulset and redeploy the statefulset with replicas  as many as replicas mentioned in ETCD CR

Flow:
1. There is already provision for Health check of cluster in ETCD Druid code base. The health check is based on Member Lease renewal time. It also deduct the quorum loss case. This health check will be used by Druid
2. Druid will monitor the cluster health at a regular interval (Or should it monitor any other resource?)
If quorum loss is detected, Druid will put an annotation in ETCD CR to indicate that cluster is being restarted.
3. Druid will delete the ETCD statefulset and delete the PVC
4. Druid will deploy the statefulset with replicas = 1
5. Druid will wait for the one ETCD pod to restore properly (What’s the best way to detect if the first pod is restored properly?)
6. Once the restoration of first pod is complete, Druid will scale up the statefulset equal to the replicas mentioned in ETCD CR. Rest of the things will be handled by ETCD scale up mechanism

abdasgupta:

/area control-plane
/kind enhancement

This PR restores a multinode ETCD cluster after the quorum is lost. The steps are like following:
1. There is already provision for Health check of cluster in ETCD Druid code base. The health check is based on Member Lease renewal time. It also deduct the quorum loss case. This health check is used by Druid
2. Druid custodion controller monitors the cluster health at a regular interval
3. If quorum loss is detected, custodion controller will put an annotation on ETCD CR to indicate that ETCD controller needs to fix quorum loss.
4. ETCD controller scales down the ETCD statefulset to 0 and delete the PVCs
5. ETCD controller deploys the statefulset with replicas = 1
6. ETCD controller scales up the statefulset equal to the replicas mentioned in ETCD CR. ETCD scale up mechanism in ETCD BR makes sure that the first instance of the StatefulSet is up and then the rests are added.
Fixes #362 

**Release note**:

```feature developer 
1. A new annotation `gardener.cloud/quorum-loss=true` is used on ETCD CR to indicate the quorum loss happened in the ETCD multi node cluster. If there is no quorum loss, either `gardener.cloud/quorum-loss=false` is set or the annotation is not set at all.
```
```feature operator
1. ETCD multinode cluster can recover now during quorum loss case. If quorum is lost for multinode cluster and the lost nodes can't be scheduled for some period, then during that period, a temporary one node ETCD cluster will serve the requests and when the new nodes can be scheduled again, rest of the nodes will join the cluster without any manual intervention.
```

### DIFF
--- a/controllers/etcd_controller.go
+++ b/controllers/etcd_controller.go
@@ -28,10 +28,10 @@ import (
 	componentconfigmap "github.com/gardener/etcd-druid/pkg/component/etcd/configmap"
 	componentlease "github.com/gardener/etcd-druid/pkg/component/etcd/lease"
 	componentservice "github.com/gardener/etcd-druid/pkg/component/etcd/service"
-	"github.com/gardener/etcd-druid/pkg/component/etcd/statefulset"
 	componentsts "github.com/gardener/etcd-druid/pkg/component/etcd/statefulset"
 	druidpredicates "github.com/gardener/etcd-druid/pkg/predicate"
 	"github.com/gardener/etcd-druid/pkg/utils"
+	coordinationv1 "k8s.io/api/coordination/v1"
 
 	extensionspredicate "github.com/gardener/gardener/extensions/pkg/predicate"
 	v1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
@@ -41,6 +41,7 @@ import (
 	gardenercomponent "github.com/gardener/gardener/pkg/operation/botanist/component"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	gardenerretry "github.com/gardener/gardener/pkg/utils/retry"
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
@@ -105,31 +106,25 @@ type EtcdReconciler struct {
 	ImageVector                        imagevector.ImageVector
 	logger                             logr.Logger
 	disableEtcdServiceAccountAutomount bool
-}
-
-// NewReconcilerWithImageVector creates a new EtcdReconciler object with an image vector
-func NewReconcilerWithImageVector(mgr manager.Manager, disableEtcdServiceAccountAutomount bool) (*EtcdReconciler, error) {
-	etcdReconciler, err := NewEtcdReconciler(mgr, disableEtcdServiceAccountAutomount)
-	if err != nil {
-		return nil, err
-	}
-	return etcdReconciler.InitializeControllerWithImageVector()
+	//waitTimeForTests allow some waiting time between certain operations. This variable help some unit test cases
+	waitTimeForTests time.Duration
 }
 
 // NewEtcdReconciler creates a new EtcdReconciler object
-func NewEtcdReconciler(mgr manager.Manager, disableEtcdServiceAccountAutomount bool) (*EtcdReconciler, error) {
+func NewEtcdReconciler(mgr manager.Manager, disableEtcdServiceAccountAutomount bool, waitTimeForTests time.Duration) (*EtcdReconciler, error) {
 	return (&EtcdReconciler{
 		Client:                             mgr.GetClient(),
 		Config:                             mgr.GetConfig(),
 		Scheme:                             mgr.GetScheme(),
 		logger:                             log.Log.WithName("etcd-controller"),
 		disableEtcdServiceAccountAutomount: disableEtcdServiceAccountAutomount,
+		waitTimeForTests:                   waitTimeForTests,
 	}).InitializeControllerWithChartApplier()
 }
 
 // NewEtcdReconcilerWithImageVector creates a new EtcdReconciler object
-func NewEtcdReconcilerWithImageVector(mgr manager.Manager, disableEtcdServiceAccountAutomount bool) (*EtcdReconciler, error) {
-	ec, err := NewEtcdReconciler(mgr, disableEtcdServiceAccountAutomount)
+func NewEtcdReconcilerWithImageVector(mgr manager.Manager, disableEtcdServiceAccountAutomount bool, waitTimeForTests time.Duration) (*EtcdReconciler, error) {
+	ec, err := NewEtcdReconciler(mgr, disableEtcdServiceAccountAutomount, waitTimeForTests)
 	if err != nil {
 		return nil, err
 	}
@@ -206,11 +201,15 @@ func (r *EtcdReconciler) SetupWithManager(mgr ctrl.Manager, workers int, ignoreO
 
 func buildPredicate(ignoreOperationAnnotation bool) predicate.Predicate {
 	if ignoreOperationAnnotation {
-		return predicate.GenerationChangedPredicate{}
+		return predicate.Or(
+			predicate.GenerationChangedPredicate{},
+			druidpredicates.HasQuorumLossAnnotation(),
+		)
 	}
 
 	return predicate.Or(
 		druidpredicates.HasOperationAnnotation(),
+		druidpredicates.HasQuorumLossAnnotation(),
 		druidpredicates.LastOperationNotSuccessful(),
 		extensionspredicate.IsDeleting(),
 	)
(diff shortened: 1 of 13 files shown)


controllers/cluster_mgmt_controller.go
```diff
@@ -0,0 +1,192 @@
...
+				EtcdMemberUnknownThreshold:  unknownThreshold,
+			},
+		})
+
+		results := checker.Check(context.Background(), *etcd)
+
+		totalReadyMembers := 0
+
+		for _, result := range results {
+			if result.Status() == "LeaseSucceeded" {
```
unmarshall, line 125: status is compared against result.Reason. The correct status value to compare against should be `druidv1alpha1.EtcdMemberStatusReady`

controllers/cluster_mgmt_controller.go
```diff
@@ -0,0 +1,192 @@
...
+				totalReadyMembers = totalReadyMembers + 1
+			}
+		}
+
+		quorum := int(etcd.Spec.Replicas)/2 + 1
```
unmarshall, line 130: quorum can be computed just once per etcd object since it will not change and can be done before starting the for loop.

controllers/cluster_mgmt_controller.go
```diff
@@ -0,0 +1,192 @@
...
+
+		if totalReadyMembers < quorum {
+			// scale down the statefulset to 0
+			sts := &appsv1.StatefulSet{}
```
unmarshall, line 134: **Case 1**
when `readyCheck.Check` is invoked then then members which are currently in bootstrapping phase are not considered and therefore not added to the []Result. In this case the `totalReadyMembers < quorum` will be true and it will try and scale down sts when there are already members that are in bootstrapping phase. So yes there is a quorum loss but should we not wait for the bootstrapping to finish and the member to join the cluster instead of scaling down the entire cluster to 0?

**Case 2**
There was an error fetching leases. In this case in `readyCheck.Check` an empty []Result will be returned. In this case also is it prudent to scale down the entire sts to 0?

**Case 3**
In `readyCheck.Check` members whose containers are not yet ready are added with status `druidv1alpha1.EtcdMemberStatusNotReady`. In this case also will it be prudent to scale down the sts cluster to 0?

Similarly please consider all cases where an action to scale down sts would make sense.

controllers/cluster_mgmt_controller.go
```diff
@@ -0,0 +1,192 @@
...
+			if _, err := controllerutils.GetAndCreateOrStrategicMergePatch(ctx, cmc.Client, sts, func() error {
+				sts.Spec.Replicas = &etcd.Spec.Replicas
+				return nil
+			}); err != nil {
+				return ctrl.Result{
+					RequeueAfter: 10 * time.Second,
+				}, fmt.Errorf("cound not scale up statefulset to replica number : %v", err)
+			}
+
+			continue
```
unmarshall, line 180: what is the need of `continue` here?

controllers/cluster_mgmt_controller.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		// Error reading the object - requeue the request.
+		return ctrl.Result{
+			RequeueAfter: 10 * time.Second,
+		}, err
+	}
+
+	logger := cmc.logger.WithValues("etcd", kutil.Key(etcd.Namespace, etcd.Name).String())
+
+	// run a loop every 5 minutes that will monitor the cluster health and take action if members in the etcd cluster are down
```
unmarshall, line 104: It is written that the loop runs every 5 mins i could not see the delay between 2 loop runs. If there is no error then this loop can run forever and then there is no chance of processing of the next reconcile request. Maybe i miss something. Please check.

controllers/cluster_mgmt_controller.go
```diff
@@ -0,0 +1,192 @@
...
+			// scale down the statefulset to 0
+			sts := &appsv1.StatefulSet{}
+			err := cmc.Get(ctx, types.NamespacedName{Name: etcd.Name, Namespace: etcd.Namespace}, sts)
+			if err != nil {
+				return ctrl.Result{
+					RequeueAfter: 10 * time.Second,
+				}, fmt.Errorf("cound not fetch statefulset: %v", err)
+			}
+
+			if _, err := controllerutils.GetAndCreateOrStrategicMergePatch(ctx, cmc.Client, sts, func() error {
```
unmarshall, line 142: Ideally these steps of scaling down, deletion of PVC and scaling up to 1 should in separate functions and one can use flow which is also used in gardener for this.

controllers/cluster_mgmt_controller.go
```diff
@@ -0,0 +1,192 @@
...
+				}, fmt.Errorf("cound not scale up statefulset to replica number : %v", err)
+			}
+
+			continue
+		}
+	}
+}
+
+func getMatchingLabels(sts *appsv1.StatefulSet) map[string]string {
+	labels := make(map[string]string)
```
unmarshall, line 186: use make with initial size. Here you have a fixed size of 2.

controllers/cluster_mgmt_controller.go
```diff
@@ -0,0 +1,192 @@
...
+
+	labels["name"] = sts.Labels["name"]
```
unmarshall, line 188: will label `name` and `instance` be always present in `sts`? If not then you should check for existence of key before adding to the new map.

unmarshall (commented):
I could only review one file. Will push further review comments tomorrow.

abdasgupta:
Added a flag in ETCD druid that will not allow Druid to apply quorum-loss annotation in case of quorum-loss. The flag needs to be set true if somebody wants to allow Druid to handle quorum loss automatically

pkg/health/etcdmember/check_ready.go
```diff
@@ -73,7 +73,7 @@ func (r *readyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) []Resul
 		renew := lease.Spec.RenewTime
 		if renew == nil {
 			r.logger.Info("Member hasn't acquired lease yet, still in bootstrapping phase", "name", lease.Name)
-			continue
+			return []Result{}
```
aaronfern, line 76: Why is this needed?
Won't returning an empty result set if any lease has an empty `RenewTime` result in an empty member list?

pkg/health/status/check.go
```diff
@@ -120,6 +126,13 @@ func (c *checker) executeConditionChecks(ctx context.Context, etcd *druidv1alpha
 // executeEtcdMemberChecks runs all registered etcd member checks **sequentially**.
 // The result of a check is passed via the `status` sub-resources to the next check.
 func (c *checker) executeEtcdMemberChecks(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd) error {
+	// Don't execute member checks until bootstrapping is going on
```
aaronfern, line 129:
```suggestion
	// Don't execute member checks while bootstrapping is in progress
```

aaronfern (changes requested):

aaronfern (changes requested):
Please add permissions for `etcds` in `charts/etcd/templates/etcd-role.yaml` as etcd-backup-restore is now intended to read the etcd resource (ref: [here](https://github.com/gardener/etcd-backup-restore/pull/528/files#diff-dc0aa715ea97877db344484988dcdffda1fc0eaeda89257be6c70e9d724721d7R80))

controllers/etcd_custodian_controller.go
```diff
@@ -108,6 +122,37 @@ func (ec *EtcdCustodian) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{}, err
 	}
 
+	conLength := len(etcd.Status.Conditions)
+	if conLength > 0 && etcd.Status.Conditions[conLength-1].Reason == "QuorumLost" && etcd.Spec.Replicas > 1 {
```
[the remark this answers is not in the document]
abdasgupta: I don't think we should consider any earlier conditions from the list. It might be very disastrous

controllers/etcd_custodian_controller.go
```diff
@@ -108,6 +122,37 @@ func (ec *EtcdCustodian) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+		logger.Info("Quorum loss detected. Taking measures to fix it.")
+		if !ec.config.EnableAutomaticQuorumLossHandling {
```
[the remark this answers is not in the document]
abdasgupta: Why would we want to update ETCD status if the quorum is lost? If quorum is lost and ETCD status is updated, ETCD controller may start working to fix the ETCD . This will give totally unintended result

controllers/etcd_custodian_controller.go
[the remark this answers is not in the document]
aaronfern: I don't mean considering any other condition. What I meant is that the `Ready` condition that gives `Reason == "QuorumLost"` might not always be at `conLength-1` and it might be good if possible to parse through all the conditions and make sure

pkg/health/status/check.go
abdasgupta: Done

controllers/etcd_controller.go
```diff
@@ -238,6 +237,102 @@ func (r *EtcdReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+			r.logger.Info("Scaling up the statefulset to the number of replicas mentioned in ETCD spec")
+			if _, err := controllerutils.GetAndCreateOrStrategicMergePatch(ctx, r.Client, sts, func() error {
+				sts.Spec.Replicas = &etcd.Spec.Replicas
+				return nil
+			}); err != nil {
+				return ctrl.Result{
+					RequeueAfter: 10 * time.Second,
+				}, fmt.Errorf("cound not scale up statefulset to replica number while tackling quorum loss scenario in ETCD multi node cluster : %v", err)
+			}
+			if err = r.removeQuorumLossAnnotation(ctx, r.logger, etcd); err != nil {
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/etcd_controller.go
```diff
@@ -692,10 +787,35 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
 		deployWaiter = gardenercomponent.OpWaiter(stsDeployer)
 	)
 
+	if _, err := stsDeployer.Get(ctx); apierrors.IsNotFound(err) {
+
+		logger.Info("As statefulsets are not yet created, it's bootstrap case. Adding annotation to ETCD CR for it")
```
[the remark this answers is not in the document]

controllers/etcd_controller.go
```diff
@@ -692,10 +787,35 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
...
+		withBootstrapAnnotation := etcd.DeepCopy()
```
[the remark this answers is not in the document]

controllers/etcd_custodian_controller.go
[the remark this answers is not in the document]
abdasgupta: Suppose, there is a quorum loss. Action is taken on it and then there is some other condition appeared. If we are parsing the whole list every time then we may pick old quorum loss case  which is already taken care of. Is it not?

pkg/health/etcdmember/check_ready.go
abdasgupta: Couldn't get you what you asked

abdasgupta:
> Please add permissions for `etcds` in `charts/etcd/templates/etcd-role.yaml` as etcd-backup-restore is now intended to read the etcd resource (ref: [here](https://github.com/gardener/etcd-backup-restore/pull/528/files#diff-dc0aa715ea97877db344484988dcdffda1fc0eaeda89257be6c70e9d724721d7R80))

Like this:
```
- apiGroups:
  - druid.gardener.cloud/v1alpha1
  resources:
  - etcds
  verbs:
  - get
  - list
  - patch
  - update
  - watch
  ```

**Comment by aaronfern** [comment]
yes, but afaik we don't need the `v1aplha1`
```
- apiGroups:
  - druid.gardener.cloud
  resources:
  - etcds
  verbs:
  - get
  - list
  - patch
  - update
  - watch
```

controllers/etcd_custodian_controller.go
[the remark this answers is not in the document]
aaronfern: The list would have the same conditions. The reasons would change depending on the state of the cluster. 
When we recover from quorum loss for instance, the condition would be updated and hence picking up an old quorum loss case will not happen imo

controllers/etcd_custodian_controller.go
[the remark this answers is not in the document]
aaronfern: Ah, okay
Fair enough

ishan16696:
@abdasgupta I guess we can close this PR ?

PR #388, opened 2022-07-29
Add service-endpoints to etcd-br

timuthy:

/area control-plane
/kind bug

This PR adds the `--service-endpoints` flag to the `etcd-backup-restore` side-car. This let's the side-car connect to the intended etcd client service whereas earlier the peer service was used which doesn't expose `2379`.

/cc @aaronfern

**Release note**:

```bugfix operator
A bug has been fixed that caused the `etcd-backup-restore` side-car to connect to the etcd cluster via the `peer-service` URL. The side-car is supposed to use the `client-service` instead since it a) exposes client port `2379` and b)  redirects traffic only to members which are ready to service traffic.
```


aaronfern (approved):
/lgtm

PR #389, opened 2022-08-01
Clarify `etcd.status.ready` field

timuthy:

/area usability
/kind enhancement

This PR defines the meaning of `etcd.status.ready` more precisely. Earlier, when Etcd-Druid could only handle single-node etcds the meaning and implementation clearer because with only one replica the cluster was either ready or not ready but evaluating this count.

This semantic changed with multi-node (cluster is ready as long as quorum is maintained) which is at the same time rather covered by a dedicated condition (`type: Ready`). Thus, the `etcd.status.ready` field  became quite ambiguous. 

**Release note**:

```other operator
The definition of the `etcd.status.ready` field was defined more precisely due to changed semantics of multi-node etcd clusters. `etcd.status.ready` is `true` whenever all underlying etcd replicas are ready. Please note, that the implementation for this check was not changed.
```

### DIFF
--- a/api/v1alpha1/etcd_types.go
+++ b/api/v1alpha1/etcd_types.go
@@ -417,7 +417,7 @@ type EtcdStatus struct {
 	// ReadyReplicas is the count of replicas being ready in the etcd cluster.
 	// +optional
 	ReadyReplicas int32 `json:"readyReplicas,omitempty"`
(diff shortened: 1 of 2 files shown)


shreyas-s-rao (approved):
@timuthy thanks a lot for the clarification of the field.
/lgtm

PR #393, opened 2022-08-03
StatefulSet component follow ups

timuthy:

/area quality
/kind technical-debt

This PR is a follow up for #365:

- Addresses https://github.com/gardener/etcd-druid/pull/365#discussion_r932905215
- Addresses https://github.com/gardener/etcd-druid/pull/365#discussion_r933740382
- Removes unused `etcd-statefulset.yaml`
- Removes unused Values from etcd Helm chart
- Resolves some TODOs as we expect an etcd restart due to the update of `etcd-backup-restore` (https://github.com/gardener/etcd-druid/pull/387).

**Release note**:

```other operator
NONE
```


pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -652,6 +640,8 @@ func getBackupRestoreEnvVars(val Values) []corev1.EnvVar {
 		return env
 	}
 
+	env = append(env, getEnvVarFromValue("STORAGE_CONTAINER", pointer.StringDeref(val.BackupStore.Container, "")))
```
shreyas-s-rao, line 643: Don't we need the check here for `val.BackupStore != nil` to avoid pointer dereference error?

shreyas-s-rao (commented):
@timuthy thanks for making these changes and removing deprecated code. Just one small comment from my side. Other than that, looks good.

pkg/component/etcd/statefulset/statefulset.go
unmarshall: @shreyas-s-rao There is a nil check
```go
if storeValues == nil {
		return env
}
```
If the storeValues which is `val.BackupStore` is nil then you there is an early return from this function. So its fine not to check for nil again. Also `pointer.StringDeref` anyways take care of the case where `Container` could also be nil.

unmarshall (approved):
/lgtm

pkg/component/etcd/statefulset/statefulset.go
shreyas-s-rao: Oops, I missed that, thanks.

shreyas-s-rao (approved):

PR #396, opened 2022-08-05
Fix liveness and readiness probes

timuthy:

/area quality
/kind bug

This PR fixes several issues for the currently used `liveness` and `readiness` checks and also adds a `startup` probe for single- and multi-node etcds.

Fixes #

Issue with current liveness probe:

* When passing multiple arguments to `/bin/sh -ec` then only the first argument is considered, i.e. `/bin/sh -ec ETCDCTL_API=3` is always successful.

```
            - /bin/sh
            - -ec
            - ETCDCTL_API=3
            - etcdctl
            - --cert=/var/etcd/ssl/client/client/tls.crt
            - --key=/var/etcd/ssl/client/client/tls.key
            - --cacert=/var/etcd/ssl/client/ca/ca.crt
            - --endpoints=https://etcd-aws-local:2379/
            - get
            - foo
            - --consistency=s
```

Issue with current readiness probe:

* The `exec` command did not evaluate the return code of the HTTP response and thus the container was considered `ready` even though the `/health(z)` endpoint returned `!= 200`.

```
            - /usr/bin/curl
            - --cert
            - /var/etcd/ssl/client/client/tls.crt
            - --key
            - /var/etcd/ssl/client/client/tls.key
            - --cacert
            - /var/etcd/ssl/client/ca/ca.crt
            - https://etcd-aws-local:8080/healthz
```

For single-node it's possible to switch to an [http probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes) to solve the explained issue.

For multi-node it's necessary to use an [exec probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes) because the `/health` endpoint of `etcd` is protected by mutual TLS (also see https://github.com/etcd-io/etcd/pull/12370) and providing a client cert is not supported via Kubernetes http probes.

The new liveness probe is now accurate, but still fails after few seconds due to the backup sidecar requiring a long time to promote its etcd member from learner. This leads to the etcd container being restarted, and the backup sidecar's initialization fails, and begins another initialization when the etcd container comes back up. This cycle continues, and the etcd pods never become ready. This problem is solved by using a startup probe of 2 minutes to allow the initialization to complete without interruptions due to etcd container restarts.

**Release note**:

```bugfix operator
An issue has been fixed that caused the `liveness` and `readiness` probes of `etcd` to always succeed even though an error was reported. This prevented defective etcd pods from being restarted automatically and caused unready candidates being considered as ready to serve traffic via the `etcd service`.
```
```bugfix operator
A `startup` probe has been added to `etcd` to allow 2 minutes of initialization time before checking for etcd liveness.
```
```feature developer
Add support for running envtest on M1 Macbooks.
```

### DIFF
--- a/pkg/component/etcd/statefulset/values_helper.go
+++ b/pkg/component/etcd/statefulset/values_helper.go
@@ -16,6 +16,7 @@ package statefulset
 
 import (
 	"fmt"
+	"strings"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 	"github.com/gardener/etcd-druid/pkg/utils"
@@ -119,8 +120,16 @@ func GenerateValues(
 	}
 
 	values.EtcdCommand = getEtcdCommand()
-	values.ReadinessProbeCommand = getReadinessProbeCommand(values)
-	values.LivenessProbCommand = getLivenessProbeCommand(values)
+
+	// Use linearizability for readiness probe so that pod is only considered ready
+	// when it has an active connection to the cluster and the cluster maintains a quorum.
+	values.ReadinessProbeCommand = getProbeCommand(values, linearizable)
+
+	// Use serializability for liveness probe because we are only interested in the health of this particular member.
+	// Those read requests will especially succeed if quorum is lost. Usually it doesn't make sense to also restart
+	// members which are healthy in general, independent of the quorum situation.
+	values.LivenessProbCommand = getProbeCommand(values, serializable)
+
 	values.EtcdBackupCommand = getBackupRestoreCommand(values)
 
 	return values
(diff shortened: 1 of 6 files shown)


pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -119,8 +120,16 @@ func GenerateValues(
...
-	values.LivenessProbCommand = getLivenessProbeCommand(values)
+
+	// Use linearizability for readiness probe so that pod is only considered ready
+	// when it has an active connection to the cluster and the cluster maintains a quorum.
+	values.ReadinessProbeCommand = getProbeCommand(values, linearizable)
+
+	// Use serializability for liveness probe because we are only interested in the health of this particular member.
+	// Those read requests will especially succeed if quorum is lost. Usually it doesn't make sense to also restart
+	// members which are healthy in general, independent of the quorum situation.
+	values.LivenessProbCommand = getProbeCommand(values, serializable)
```
shreyas-s-rao, line 131: Typo: `LivenessProbCommand`

shreyas-s-rao (approved):
@timuthy Thanks for the quick fix! Overall LGTM except for a small nit.
/lgtm

pkg/component/etcd/statefulset/values_helper.go
aaronfern, line 131: This here lead to a case of both livenessProbe and readinessProbe having similar values
```
livenessProbe:
          exec:
            command:
            - /bin/sh
            - -ec
            - ETCDCTL_API=3 etcdctl --endpoints=http://etcd-main-local:2379 get foo
              --consistency=s

readinessProbe:
          exec:
            command:
            - /bin/sh
            - -ec
            - ETCDCTL_API=3 etcdctl --endpoints=http://etcd-main-local:2379 get foo
              --consistency=l
```

This isn't the intention, right?

aaronfern (changes requested):
Thanks for the PR @timuthy!

One comment from me

pkg/component/etcd/statefulset/values_helper.go
shreyas-s-rao: This is the intended behavior. The goal is to get a response from the etcd - in this case via an exec call due to the client cert issue mentioned earlier. `consistency` decides whether the response requires quorum or not - liveness probe doesn't require the member to have quorum, whereas readiness probe requires a quorate cluster. The readiness probe is essentially, in function, similar to a http call to the etcd member's `/health` endpoint, since it too returns true for a quorate cluster.

timuthy:
Thanks for the reviews @shreyas-s-rao and @aaronfern 🚀 I fixed the type, PTAL.

shreyas-s-rao:
@timuthy the unit tests for sts component are failing because the probes in `validateEtcd` and `validateEtcdWithDefaults` functions in the test need to be adapted. Can you please make that change? Thanks!

aaronfern (approved):
/lgtm

abdasgupta (approved):

PR #400, opened 2022-08-09
Fix storage class in statefulset

shreyas-s-rao:

/area backup
/kind bug

This PR fixes an issue with populating `storageClassName` for the `volumeClaimTemplate` in the etcd statefulset. It adds an extra layer of validation for storageClass value to ensure that if it is unset or empty, it should remain unset in the statefulset spec, rather than being set to empty string `""`.

Fixes #

Co-author: @aaronfern 

**Release note**:

```bugfix user
Fix statefulset volumeClaimTemplate `StorageClassName` value population if etcd storageClass is an empty string.
```


ishan16696 (approved):
thanks @shreyas-s-rao and @aaronfern for taking care of this fix.
LGTM.

PR #402, opened 2022-08-10
Fix PodManagementPolicy updation for etcd statefulset

shreyas-s-rao:

/area backup
/kind bug

This PR temporarily fixes the issue where `PodManagementPolicy` was trying to be updated from `OrderedReady` to `Parallel` for older shoots (created using etcd-druid:v0.8.5 and before), but the statefulset forbids updates to this field.

Fixes #

The statefulset claim/recreation logic needs to be re-looked at, since the only way to update the `PodManagementPolicy` from `OrderedReady` to `Parallel` would be to delete and recreate the statefulset, which the current component model for statefulset does not do. PodManagementPolicy `Parallel` is important during shoot unhibernation, for the etcd cluster to be bootstrapped.  

**Release note**:

```bugfix user
Temporarily fix issue where `PodManagementPolicy` was trying to be updated from `OrderedReady` to `Parallel` for older shoots (created using etcd-druid:v0.8.5 and before), but the statefulset forbids updates to this field.
```

### DIFF
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -165,7 +165,6 @@ func (c *component) syncStatefulset(ctx context.Context, sts *appsv1.StatefulSet
 
 	sts.ObjectMeta = getObjectMeta(&c.values)
 	sts.Spec = appsv1.StatefulSetSpec{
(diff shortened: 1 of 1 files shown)


aaronfern (approved):
/lgtm

PR #403, opened 2022-08-10
Fix ServiceName updation for etcd statefulset

shreyas-s-rao:

/area backup
/kind bug

This PR temporarily fixes an issue where druid tries to set `spec.ServiceName` to `PeerServiceName` by default, although older single-node etcds would have this field set to `ClientServiceName`, and updation of statefulset `spec.ServiceName` field is forbidden.

Fixes #

The permanent solution to this is to introduce sts recreation logic for updating any fields for which updates are forbidden.

**Release note**:

```bugfix user
Temporarily fixes an issue where druid tries to set `spec.ServiceName` to `PeerServiceName` by default, although older single-node etcds would have this field set to `ClientServiceName`, and updation of statefulset `spec.ServiceName` field is forbidden.
```


aaronfern (approved):
/lgtm

Issue #404, opened 2022-08-10
[BUG] Statefulset updation fails as druid attempts to update forbidden fields


shreyas-s-rao:
**Describe the bug:**
With the new component model for statefulset, and the removal of claim/recreation logic for statefulsets, druid uses certain default values in the statefulset spec. This works well for newly created etcds, but older ones that were created with the old default values, fail to be updated by druid since updates to certain fields in the statefulset spec is forbidden, with the error message like
```
Error while waiting for Etcd xyz to become ready: StatefulSet.apps "etcd-main" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

Examples of such "forbidden" fields are `PodManagementPolicy` and `ServiceName`.

**Expected behavior:**
Druid should have respected the old values of the statefulset, or recreated the statefulset with the new values for these forbidden fields.

**How To Reproduce (as minimally and precisely as possible):**
Create an etcd resource using the [sample etcd resource](https://github.com/gardener/etcd-druid/blob/master/config/samples/druid_v1alpha1_etcd.yaml), using etcd-druid `v0.8.5` or below. Once the etcd becomes ready, stop running druid, checkout `v0.12.0` and re-run druid. Reconciliation of the etcd will fail.

**Logs:**

**Screenshots (if applicable):**

**Environment (please complete the following information):**
 - Etcd version/commit ID :
 - Etcd-druid version/commit ID : `v0.12.0`
 - Cloud Provider [All/AWS/GCS/ABS/Swift/OSS]:

**Anything else we need to know?**:
Temporarily fixed with https://github.com/gardener/etcd-druid/pull/401, https://github.com/gardener/etcd-druid/pull/402 and https://github.com/gardener/etcd-druid/pull/403.
The permanent solution is for druid to check if the statefulset patch includes changes to forbidden/immutable fields, and if required recreates the statefulset with the new values.

timuthy:
FMPOV this issue has two sides:

[...]

PR #406, opened 2022-08-15
[rel-v0.12] Fix podManagementPolicy for scale-up scenario

timuthy:

/area high-availability
/kind bug

This PR fixes a bug which caused a scaled-up etcd (`etcd.spec.replicas: 1 -> 3`) cluster having the wrong `sts.spec.podManamgentPolicy`.
This field must be set to `Parallel`, esp. if the cluster is afterwards scaled down (`etcd.spec.replicas: 0`) and scaled up (`etcd.spec.replicas: 3`) again. This is a common case when Gardener hibernates and wakes up shoot clusters.

/cc @shreyas-s-rao @aaronfern

**Release note**:

```bugfix operator
This PR fixes an issue which caused the `sts.spec.podManagementPolicy` not to be updated to `Parallel` if an existing etcd cluster is scaled-up from `1 -> x`. This can cause an issue if the cluster is afterwards completely scaled-down (aka hibernation) and scaled-up again.
```

### DIFF
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -72,19 +72,20 @@ func (c *component) Deploy(ctx context.Context) error {
 		sts = c.emptyStatefulset()
 	}
 
-	if sts.Generation > 1 && sts.Spec.ServiceName != c.values.PeerServiceName {
-		// Earlier clusters referred to the client service in `sts.Spec.ServiceName` which must be changed
-		// when a multi-node cluster is used, see https://github.com/gardener/etcd-druid/pull/293.
-		if clusterScaledUpToMultiNode(c.values) {
-			deleteAndWait := gardenercomponent.OpDestroyAndWait(c)
-			if err := deleteAndWait.Destroy(ctx); err != nil {
-				return err
-			}
-			sts = c.emptyStatefulset()
+	if sts.Generation > 1 && clusterScaledUpToMultiNode(c.values) && immutableFieldUpdate(sts, c.values) {
+		// Several immutable fields must be reset for the multi-node use-case.
+		deleteAndWait := gardenercomponent.OpDestroyAndWait(c)
+		if err := deleteAndWait.Destroy(ctx); err != nil {
+			return err
 		}
+		sts = c.emptyStatefulset()
 	}
 
 	return c.syncStatefulset(ctx, sts)
+}
+
+func immutableFieldUpdate(sts *appsv1.StatefulSet, val Values) bool {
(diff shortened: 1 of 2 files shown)


pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -72,21 +72,22 @@ func (c *component) Deploy(ctx context.Context) error {
...
 		}
+		sts = c.emptyStatefulset()
 	}
 
 	return c.syncStatefulset(ctx, sts)
 }
 
+func immutableFieldUpdate(sts *appsv1.StatefulSet, val Values) bool {
+	return sts.Spec.ServiceName != val.PeerServiceName || sts.Spec.PodManagementPolicy != appsv1.ParallelPodManagement
+}
```
aaronfern, line 89: Should we generalise this and let it return true if any immutable field in the sts is being updated? (any field apart from `replicas`, `template`, `updateStrategy`, or `minReadySeconds`)
wdyt?

aaronfern (commented):
Thanks for the PR @timuthy!

Just one general question

pkg/component/etcd/statefulset/statefulset.go
timuthy: Good question. FMPOV we should not do this in general because such a deletion should remain an exception, i.e. it should be a conscious decision of Druid under which circumstances the deletion and thus an interruption is acceptable, see https://github.com/gardener/etcd-druid/issues/404#issuecomment-1215129643. WDYT?

unmarshall: AFAIK immutability is currently only enforced via webhooks. There is no way currently to know by just looking at the spec which fields are immutable. The [proposal](https://github.com/kubernetes/enhancements/pull/1265) for that is merged but i am unaware if the implementation has been. Anything generic will require this support, right?

timuthy: > AFAIK immutability is currently only enforced via webhooks.

I don't see any issues with that. We do this for provider extensions in Gardener as well. I think it's more dangerous to just allow a field to be changed, delete the sts by default and thus cause a unexpected disruption.

unmarshall: Yes there is nothing wrong in using a webhook as that is the only way to enforce it. My comment was for:
>Should we generalise this and let it return true if any immutable field in the sts is being updated?

To  enforce of all immutable fields in a spec generically you need to be aware of all immutable fields. I was merely pointing to the fact that this is not possible for any spec today as immutable field identification is not yet part of a spec but is only coded in the webhook.

aaronfern: Fair enough, thank you for pointing this out 
Since I see we're all in agreement, I will resolve this conversation :)

aaronfern (approved):
/lgtm

PR #408, opened 2022-08-18
Fix serviceName and podManagementPolicy field updates

### LINKED ISSUE #404: [BUG] Statefulset updation fails as druid attempts to update forbidden fields

timuthy:

/area control-plane
/kind bug

This PR fixes an issue that made Etcd-Druid to update `sts.spec.serviceName` and `sts.spec.podManagementPolicy` for older `etcd` resources. Those are immutable fields that should only be updated when a scale-up is performed and thus a disruption is excepted (by first deleting the statefulset and re-creating it afterwards).

Fixes #404

**Release note**:

```bugfix operator
An issue has been fixed that caused Etcd-Druid to update immutable fields `sts.spec.serviceName` and `sts.spec.podManagementPolicy` for older `etcd` resources that had different values configured. These updates must only happen when a etcd cluster is scaled up for the first time (`1 -> x`) because (a) then these values are mandatory and (b) a disruption is accepted.
```

### DIFF
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -72,19 +72,20 @@ func (c *component) Deploy(ctx context.Context) error {
 		sts = c.emptyStatefulset()
 	}
 
-	if sts.Generation > 1 && sts.Spec.ServiceName != c.values.PeerServiceName {
-		// Earlier clusters referred to the client service in `sts.Spec.ServiceName` which must be changed
-		// when a multi-node cluster is used, see https://github.com/gardener/etcd-druid/pull/293.
-		if clusterScaledUpToMultiNode(c.values) {
-			deleteAndWait := gardenercomponent.OpDestroyAndWait(c)
-			if err := deleteAndWait.Destroy(ctx); err != nil {
-				return err
-			}
-			sts = c.emptyStatefulset()
+	if sts.Generation > 1 && clusterScaledUpToMultiNode(c.values) && immutableFieldUpdate(sts, c.values) {
+		// Several immutable fields must be reset for the multi-node use-case.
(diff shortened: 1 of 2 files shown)


aaronfern (approved):
/lgtm

PR #414, opened 2022-08-22
Added the metrics needed to capture for multi-node etcd.

### LINKED ISSUE #340: [Feature] List down all the metrics necessary for maintaining multinode
Create a collective list of all the metrics that is needed to maintain multi-node ETCD. Update the [file](https://github.com/gardener/etcd-druid/blob/master/docs/multinode-metrics.md) with the list. This is a running document that will capture all the metrics that will be exposed through prometheus for mutinode ETCD.

Once etcd-druid starts managing multi-node etcd clusters, it would perform operations such as cluster scale-up, scale-down, recovery quorum losses, forced restorations, etc. Currently druid does not expose any metrics about its operations, and these metrics will become imperative for the [multi-node story](https://github.com/gardener/etcd-druid/issues/107), especially for understanding and debugging druid behaviour during etcd cluster failures. 

ishan16696:
To add metrics for multi-node etcd and backup-restore this PR adds a document which mentioned the important metrics which needs to be captured or expose in order to debug/fine-tune multi-node etcd.

Fixes #340 

For backup-restore metrics: some of metrics will remain same like full snapshots, delta snapshots.

**Release note**:
```other operator
Adds a document mentioning the metrics for multi-node etcd.
```

### DIFF
--- a/docs/multinode-metrics.md
+++ b/docs/multinode-metrics.md
@@ -29,17 +29,13 @@
 
 #### Backup-restore
 
-|  No. | Metrics Name | Description  | Comments|
-| -- | -- | -- | -- |
+|  No. | Metrics Name | Description |
+| -- | -- | -- | 
 | 1. | etcdbr_cluster_size | to capture the scale-up/scale-down scenarios. |
 | 2. | etcdbr_is_learner | whether or not this member is a learner. 1 if it is, 0 otherwise. |
-| 3. | etcdbr_has_learner | whether or not this cluster has a learner. 1 if it is, 0 otherwise. |
+| 3. | etcdbr_is_learner_count_total | total number times member added as the learner. |
 | 4. | etcdbr_restoration_duration_seconds | total latency distribution required to restore the etcd member. |
 | 5. | etcdbr_add_learner_duration_seconds | total latency distribution of adding the etcd member as a learner to the cluster. |
 | 6. | etcdbr_member_remove_duration_seconds | total latency distribution removing the etcd member from the cluster. |
 | 7. | etcdbr_member_promote_duration_seconds | total latency distribution of promoting the learner to the voting member. |
 | 8. | etcdbr_defragmentation_duration_seconds | total latency distribution of defragmentation of each etcd cluster member. |
-
-
-
-
--- a/docs/multinode-metrics.md
+++ b/docs/multinode-metrics.md
@@ -1,1 +1,45 @@
+## Metrics: Multi-node etcd
+
+#### Etcd
+
+| No. | Metrics Name | Description  | Comments|
+| -- | -- | -- | -- |
+| 1 |  etcd_disk_wal_fsync_duration_seconds | latency distributions of fsync called by WAL.  | High disk operation latencies indicate disk issues. |
+| 2 |  etcd_disk_backend_commit_duration_seconds | latency distributions of commit called by backend. | High disk operation latencies indicate disk issues. |
+| 3 |  etcd_server_has_leader | whether or not a leader exists. 1: leader exists, 0: leader not exists. | To capture quorum loss or to check the availability of etcd cluster. |
+| 4 | etcd_server_is_leader |  whether or not this member is a leader. 1 if it is, 0 otherwise. | |
+| 5 |  etcd_server_leader_changes_seen_total | number of leader changes seen. | Helpful in fine tuning the zonal cluster like etcd-heartbeat time etc, it can also indicates the etcd load and network issues.  |
+| 6 | etcd_server_is_learner | whether or not this member is a learner. 1 if it is, 0 otherwise. | |
+| 7 | etcd_server_learner_promote_successes | total number of successful learner promotions while this member is leader. | Might be helpful in checking the success of API calls called by backup-restore. |
+| 8 |  etcd_network_client_grpc_received_bytes_total | total number of bytes received from grpc clients. | Client Traffic In. |
+| 9 | etcd_network_client_grpc_sent_bytes_total | total number of bytes sent to grpc clients. | Client Traffic Out. |
+| 10 | etcd_network_peer_sent_bytes_total | total number of bytes sent to peers. | Useful for network usage. |
+| 11 | etcd_network_peer_received_bytes_total | total number of bytes received from peers. | Useful for network usage. |
+| 12 | etcd_network_active_peers | current number of active peer connections. | Might be useful in detecting issues like network partition. |
+| 13 | etcd_server_proposals_committed_total | total number of consensus proposals committed. | A consistently large lag between a single member and its leader indicates that member is slow or unhealthy. |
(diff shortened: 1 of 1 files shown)


ashwani2k:
Thanks for the proposal @ishan16696. 
I was wondering if it makes sense to also check on the usage of the following -- 

**ETCD**
For each member below 2 can help during operations to see if some event is peculiar to a leader or learner to map it to an observed anomaly if any. 

Per MEMBER
```
# name: "etcd_server_is_leader"
# description: "Whether or not this member is a leader. 1 if is, 0 otherwise."
# type: "gauge"
etcd_server_is_leader

# name: "etcd_server_is_learner"
# description: "Whether or not this member is a learner. 1 if is, 0 otherwise."
# type: "gauge"
etcd_server_is_learner
```

Overall -- 
```
# name: "etcd_server_heartbeat_send_failures_total"
# description: "The total number of leader heartbeat send failures (likely overloaded from slow disk)."
# type: "counter"
etcd_server_heartbeat_send_failures_total
```

**BACKUP-RESTORE**
If there is a way to achieve these as well might be of help to gauge the time it takes 

```
# name: "last_restoration_timestamp"
# description: "Timestamp of when the last restoration was done"
# type: "gauge"

# name: "last_restoration_duration"
# description: "Time take for restoring an etcd member, time it takes for etcd member to join"
# type: "gauge"

# name: "avg_restoration_duration"
# description: "Average time taken for restoring an etcd member, time it takes for etcd member to join"
# type: "gauge"
```

**Nice to have but not required immediately** 
This is just for completeness sake and might come handy even for single node etcds.
```
# name: "last_compaction_run_timestamp"
# description: "Timestamp when the last compaction was done"
# type: "gauge"

# name: "last_compaction_duration"
# description: "Time taken during compaction"
# type: "gauge"

# name: "avg_compaction_duration"
# description: "Avg time taken during compaction"
# type: "gauge"

# name: "avg_compaction_size"
# description: "Timestamp when the last compaction was done"
# type: "gauge"

# name: "failed_compaction_jobs_total"
# description: "No of compaction jobs failed in last 24hrs or since last full snapshot"
# type: "counter"
```

ishan16696:
Thanks @ashwani2k for suggestions, I will also take them into consideration while adding/exposing metrics for etcd and backup-restore.

ishan16696:
To capture the `Compaction` metrics: I will open a separate issue to track that. Right now we can focus on multi-node etcd metrics as compaction is not enable on all landscapes. Hope that’s make sense.

docs/multinode-metrics.md
```diff
@@ -1 +1,34 @@
+## Metrics to enhance the grafana for multi-node etcd
+
+
+#### Etcd
+
+| Sr No | Metrics Name | Description  | Comments|
```
timuthy, line 6: What does `Sr` stand for? 👀

docs/multinode-metrics.md
```diff
@@ -1 +1,34 @@
+## Metrics to enhance the grafana for multi-node etcd
+
+
+#### Etcd
```
timuthy, line 4: What about the following metrics?

- `etcd_network_peer_round_trip_time_seconds`
- `etcd_server_heartbeat_send_failures`
- `etcd_server_slow_apply_total`
- `etcd_server_slow_read_indexes_total`

docs/multinode-metrics.md
```diff
@@ -1 +1,34 @@
+## Metrics to enhance the grafana for multi-node etcd
```
timuthy, line 1: Can you please refine this and describe what the purpose of this document is? It's not quite accurate to use "Garfana" here because it's only one way to visualize the captured metrics. This is rather a document of metrics that we consider important to observe the runtime behavior of multi-node etcd clusters. How it's scraped, stored and visualized is up to the operator.

docs/multinode-metrics.md
```diff
@@ -1 +1,34 @@
...
+
+
+#### Etcd
+
+| Sr No | Metrics Name | Description  | Comments|
+| -- | -- | -- | -- |
+| 1 |  etcd_disk_wal_fsync_duration_seconds | latency distributions of fsync called by wal  | Should be captured for all Etcd members |
```
timuthy, line 8: Fussy but can you please use punctuation consistently?

timuthy (changes requested):

docs/multinode-metrics.md
ishan16696: This is just a PR, motive is to get start working on metrics with basic metrics, We will keep adding whatever we feel needs to added or needed to expose.

`etcd_network_peer_round_trip_time_seconds` 
As Shesha and Madhav worked with etcd network traffic and I guess if they feel it will be helpful in fine tuning the zonal cluster then we can add. I don't have a strong opinion on this, I will add this if they feel it should be added.
cc @unmarshall  

`etcd_server_heartbeat_send_failures`
What information can we get from this ? 
there will be a hearbeat failures which will result into leader-election may be but that's a normal scenario isn't ? especially in etcd zonal cluster.

timuthy: It'll be useful if we see an unexpected amount of leader election and double checking `etcd_server_heartbeat_send_failures` can help to find the root cause for it.

ishan16696: added

ishan16696: `etcd_server_slow_apply_total`, `etcd_server_slow_read_indexes_total` not sure about these 2 metrics but I will add them as well ... no harm in it.

ishan16696: added.

ishan16696:
I have updated the document of metrics. PTAL.

docs/multinode-metrics.md
```diff
@@ -1 +1,45 @@
...
+| 17 | etcd_mvcc_db_total_size_in_bytes | total size of the underlying database physically allocated in bytes. |  |
+| 18 | etcd_server_heartbeat_send_failures_total | total number of leader heartbeat send failures. | Might be helpful in fine-tuning the cluster or detecting slow disk or any network issues. |
+| 19 | etcd_network_peer_round_trip_time_seconds | round-trip-time histogram between peers. | Might be helpful in fine-tuning network usage specially for zonal etcd cluster. |
+| 20 | etcd_server_slow_apply_total | total number of slow apply requests. | Might indicate overloaded from slow disk. |
+| 21 | etcd_server_slow_read_indexes_total | total number of pending read indexes not in sync with leader's or timed out read index requests. | |
+
+
+#### Backup-restore
+
+|  No. | Metrics Name | Description  | Comments|
```
timuthy, line 32: Can you remove the comments column if there are no comments?

timuthy (commented):
Just one nit. LGTM otherwise.

docs/multinode-metrics.md
ishan16696: done. PTAL

ishan16696:
Seems pipeline is having some issue. I will re-push to run the pipeline checks again.

timuthy (approved):
/lgtm

PR #415, opened 2022-08-25
Enhance BackupReady condition handling

### LINKED ISSUE #413: [Enhancement] Exceptions for `BackupReady` condition
Today, `Etcd-Druid` maintains the `BackupReady` condition even for `etcd` resources that don't define a backup. Even though the `status` of this conditions remains `Unknown`, we should rather not maintain the condition at all to not cause any confusion.

```
status:
  conditions:
  - lastTransitionTime: "2022-07-28T05:48:24Z"
    lastUpdateTime: "2022-08-22T15:21:11Z"
    message: Cannot determine etcd backup status
    reason: Unknown
    status: Unknown
    type: BackupReady
```

Furthermore, we should correctly maintain the `BackupReady` condition when `etcd.spec.replicas: 0` is set (aka hibernation). Today, the conditions becomes `false`:

```
  - lastTransitionTime: "2022-07-28T05:48:24Z"
    lastUpdateTime: "2022-08-22T15:23:08Z"
    message: Stale snapshot leases. Not renewed in a long time
    reason: BackupFailed
    status: "False"
    type: BackupReady
```

aaronfern:

/area backup
/kind enhancement

This PR enhances the `BackupReady` condition handling to take into account cases where the backup section is not defined and in cases where the etcd statefulset is scaled down

Fixes #413 

**Release note**:

```other operator
Enhance `BackupReady` condition to take into account statefulset being scaled down and the backup section not being defined
```

### DIFF
--- a/pkg/health/condition/builder.go
+++ b/pkg/health/condition/builder.go
@@ -24,10 +24,10 @@ import (
 )
 
 //
-var skipMergeConditions = map[druidv1alpha1.ConditionType]bool{
-	druidv1alpha1.ConditionTypeReady:           true,
-	druidv1alpha1.ConditionTypeAllMembersReady: true,
-	druidv1alpha1.ConditionTypeBackupReady:     true,
+var skipMergeConditions = map[druidv1alpha1.ConditionType]struct{}{
+	druidv1alpha1.ConditionTypeReady:           struct{}{},
+	druidv1alpha1.ConditionTypeAllMembersReady: struct{}{},
+	druidv1alpha1.ConditionTypeBackupReady:     struct{}{},
 }
 
 // Builder is an interface for building conditions.
@@ -118,7 +118,7 @@ func (b *defaultBuilder) Build() []druidv1alpha1.Condition {
 
 	for _, condition := range b.old {
 		// Do not add conditions that are part of the skipMergeConditions list
-		ok := skipMergeConditions[condition.Type]
+		_, ok := skipMergeConditions[condition.Type]
 		if ok {
 			continue
 		}
--- a/pkg/health/condition/builder.go
+++ b/pkg/health/condition/builder.go
@@ -23,6 +23,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+//
+var skipMergeConditions = map[druidv1alpha1.ConditionType]bool{
+	druidv1alpha1.ConditionTypeReady:           true,
+	druidv1alpha1.ConditionTypeAllMembersReady: true,
+	druidv1alpha1.ConditionTypeBackupReady:     true,
+}
+
 // Builder is an interface for building conditions.
 type Builder interface {
 	WithOldConditions(conditions []druidv1alpha1.Condition) Builder
@@ -105,6 +112,17 @@ func (b *defaultBuilder) Build() []druidv1alpha1.Condition {
 		condition.Message = res.Message()
 		condition.Reason = res.Reason()
 
+		conditions = append(conditions, condition)
+		delete(b.old, condType)
+	}
+
+	for _, condition := range b.old {
+		// Do not add conditions that are part of the skipMergeConditions list
+		ok := skipMergeConditions[condition.Type]
+		if ok {
+			continue
+		}
+		// Add existing conditions as they were. This needs to be changed when SSA is used.
 		conditions = append(conditions, condition)
 	}
 
--- a/pkg/health/condition/builder.go
+++ b/pkg/health/condition/builder.go
@@ -23,7 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-//
+// skipMergeConditions contain the list of conditions we dont want to add to the list if not recalculated
 var skipMergeConditions = map[druidv1alpha1.ConditionType]struct{}{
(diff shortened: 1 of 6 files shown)


pkg/health/condition/check_backup_ready.go
```diff
@@ -48,6 +49,34 @@ func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) R
 		message: "Cannot determine etcd backup status",
 	}
 
+	// Special case of etcd not being configured to take snapshots
+	// Do not add the BackupReady condition if backup is not configured
+	if etcd.Spec.Backup.Store == nil {
+		return nil
```
timuthy, line 55: Can you double check if this also works when backups are disabled afterwards, i.e. it was enabled before? I assume that the Backup condition will still be present and we probably want to condition to be removed. WDYT?

pkg/health/condition/check_backup_ready.go
```diff
@@ -48,6 +49,34 @@ func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) R
...
+			}
+		}
+
+		if bckpCond.Status == "" {
+			result.status = druidv1alpha1.ConditionUnknown
+		} else {
+			result.status = bckpCond.Status
+		}
+		result.reason = ConditionNotChecked
+		result.message = "Statefulset has been scaled down"
```
timuthy, line 76:
```suggestion
		result.message = "etcd cluster has been scaled down"
```

timuthy (changes requested):
Looks good in genera, thanks for the contribution 🙂

pkg/health/condition/check_backup_ready.go
aaronfern: Hmm, I checked this and it does not remove the condition in this case
Let me try and see if we can support this case

aaronfern: Updated the builder function to not add any old conditions to the condition set
PTAL whenever possible

timuthy: To be on the safe side, can you only not add old conditions for known types `ConditionTypeReady`, `ConditionTypeAllMembersReady` and `ConditionTypeBackupReady`?

aaronfern: Updated to include this

pkg/health/condition/builder.go
```diff
@@ -23,6 +23,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+//
+var skipMergeConditions = map[druidv1alpha1.ConditionType]bool{
```
timuthy, line 27:
```suggestion
var skipMergeConditions = map[druidv1alpha1.ConditionType]struct{}{
```
Since the value doesn't matter.

timuthy (changes requested):

pkg/health/condition/builder.go
aaronfern: Thanks for pointing this out :) 
Updated

pkg/health/condition/builder.go
```diff
@@ -23,6 +23,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+//
```
ishan16696, line 26: do you want to add something like in comment for what `skipMergeConditions` is used

pkg/health/condition/check_backup_ready.go
```diff
@@ -48,6 +50,34 @@ func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) R
 		message: "Cannot determine etcd backup status",
 	}
 
+	// Special case of etcd not being configured to take snapshots
+	// Do not add the BackupReady condition if backup is not configured
+	if etcd.Spec.Backup.Store == nil {
```
ishan16696, line 55: Can you also add a case when `etcd.Spec.Backup.Store.Provider` is passed as empty string.

ishan16696 (changes requested):

pkg/health/condition/check_backup_ready.go
aaronfern: Added

timuthy (approved):
/lgtm
Thanks for taking care 🙂

pkg/health/condition/check_backup_ready.go
```diff
@@ -48,6 +50,12 @@ func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) R
 		message: "Cannot determine etcd backup status",
 	}
 
+	// Special case of etcd not being configured to take snapshots
+	// Do not add the BackupReady condition if backup is not configured
+	if etcd.Spec.Backup.Store == nil || etcd.Spec.Backup.Store.Provider == nil {
```
ishan16696, line 55: Please add this condition as well.
```suggestion
	if etcd.Spec.Backup.Store == nil || etcd.Spec.Backup.Store.Provider == nil  || len(*etcd.Spec.Backup.Store.Provider)==0{
```

ishan16696 (changes requested):

pkg/health/condition/check_backup_ready.go
aaronfern: Added

PR #418, opened 2022-08-28
Add e2e tests for multi-node etcd 

seshachalam-yv:

/area testing
/kind enhancement

This PR adds  e2e tests for multi-node etcd for below cases,

 **Etcd cluster operations:**
- [x] Creation of 3-member etcd cluster
- [x] Hibernation
    -  [x] Scale down from 3->0
    - [x] Scale up from 0->3
- [x] Zero downtime rolling updates 
- [x] Zero downtime maintenance operations - Defragmentation
- [x] Deletion of 3-member etcd cluster

 **Scaling**
- [x]  Scale up of a healthy cluster (from 1->3)
- [x]  Scale down of a healthy cluster (from 3->1) - (but commented, because this case not yet supported.)

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/e2e/etcd_backup_test.go
+++ b/test/e2e/etcd_backup_test.go
@@ -22,7 +22,6 @@ import (
 	"github.com/gardener/etcd-druid/api/v1alpha1"
 
 	brtypes "github.com/gardener/etcd-backup-restore/pkg/types"
-	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/test/matchers"
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo"
@@ -194,7 +193,7 @@ func createAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logge
 
 func checkEtcdReady(ctx context.Context, cl client.Client, logger logr.Logger, etcd *v1alpha1.Etcd) {
 	logger.Info("Waiting for etcd to become ready")
-	Eventually(func() error {
+	EventuallyWithOffset(2, func() error {
 		ctx, cancelFunc := context.WithTimeout(context.TODO(), timeout)
 		defer cancelFunc()
 
@@ -212,7 +211,7 @@ func checkEtcdReady(ctx context.Context, cl client.Client, logger logr.Logger, e
 		}
 
 		if *etcd.Status.ClusterSize != etcd.Spec.Replicas {
-			return fmt.Errorf("etcd %s cluster size is %v, but its not expected size as %v",
+			return fmt.Errorf("etcd %s cluster size is %v, but it's not expected size as %v",
 				etcd.Name, etcd.Status.ClusterSize, etcd.Spec.Replicas)
 		}
 
@@ -232,37 +231,37 @@ func checkEtcdReady(ctx context.Context, cl client.Client, logger logr.Logger, e
 
 	logger.Info("Checking statefulset")
 	sts := &appsv1.StatefulSet{}
-	ExpectWithOffset(1, cl.Get(ctx, client.ObjectKeyFromObject(etcd), sts)).To(Succeed())
-	ExpectWithOffset(1, sts.Status.ReadyReplicas).To(Equal(etcd.Spec.Replicas))
+	ExpectWithOffset(2, cl.Get(ctx, client.ObjectKeyFromObject(etcd), sts)).To(Succeed())
+	ExpectWithOffset(2, sts.Status.ReadyReplicas).To(Equal(etcd.Spec.Replicas))
 
 	logger.Info("Checking configmap")
 	cm := &corev1.ConfigMap{}
-	ExpectWithOffset(1, cl.Get(ctx, client.ObjectKey{Name: "etcd-bootstrap-" + string(etcd.UID[:6]), Namespace: etcd.Namespace}, cm)).To(Succeed())
+	ExpectWithOffset(2, cl.Get(ctx, client.ObjectKey{Name: "etcd-bootstrap-" + string(etcd.UID[:6]), Namespace: etcd.Namespace}, cm)).To(Succeed())
 
 	logger.Info("Checking client service")
 	svc := &corev1.Service{}
-	ExpectWithOffset(1, cl.Get(ctx, client.ObjectKey{Name: etcd.Name + "-client", Namespace: etcd.Namespace}, svc)).To(Succeed())
+	ExpectWithOffset(2, cl.Get(ctx, client.ObjectKey{Name: etcd.Name + "-client", Namespace: etcd.Namespace}, svc)).To(Succeed())
 }
 
 func deleteAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logger, etcd *v1alpha1.Etcd) {
 	ExpectWithOffset(1, cl.Delete(ctx, etcd, client.PropagationPolicy(metav1.DeletePropagationForeground))).To(Succeed())
 
 	logger.Info("Checking if etcd is gone")
-	Eventually(func() error {
+	EventuallyWithOffset(1, func() error {
 		ctx, cancelFunc := context.WithTimeout(ctx, timeout)
 		defer cancelFunc()
 		return cl.Get(ctx, client.ObjectKeyFromObject(etcd), etcd)
 	}, timeout*3, pollingInterval).Should(matchers.BeNotFoundError())
 
 	logger.Info("Checking if statefulset is gone")
-	Eventually(func() error {
+	EventuallyWithOffset(1, func() error {
 		ctx, cancelFunc := context.WithTimeout(ctx, timeout)
 		defer cancelFunc()
 		return cl.Get(ctx, client.ObjectKeyFromObject(etcd), &appsv1.StatefulSet{})
 	}, timeout, pollingInterval).Should(matchers.BeNotFoundError())
 
 	logger.Info("Checking if configmap is gone")
-	Eventually(func() error {
+	EventuallyWithOffset(1, func() error {
 		ctx, cancelFunc := context.WithTimeout(ctx, timeout)
 		defer cancelFunc()
 
@@ -270,7 +269,7 @@ func deleteAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logge
 	}, timeout, pollingInterval).Should(matchers.BeNotFoundError())
 
 	logger.Info("Checking client service is gone")
-	Eventually(func() error {
+	EventuallyWithOffset(1, func() error {
 		ctx, cancelFunc := context.WithTimeout(ctx, timeout)
 		defer cancelFunc()
 
@@ -284,10 +283,14 @@ func deleteAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logge
 
 func purgeEtcdPVCs(ctx context.Context, cl client.Client, etcdName string) {
 	r, _ := k8s_labels.NewRequirement("instance", selection.Equals, []string{etcdName})
-	opts := &client.ListOptions{LabelSelector: k8s_labels.NewSelector().Add(*r)}
-	pvcList := &corev1.PersistentVolumeClaimList{}
-	cl.List(ctx, pvcList, client.InNamespace(namespace), opts)
-	for _, pvc := range pvcList.Items {
-		ExpectWithOffset(1, kutil.DeleteObject(ctx, cl, &pvc)).To(Succeed())
-	}
+	pvc := &corev1.PersistentVolumeClaim{}
+	delOptions := client.DeleteOptions{}
+	delOptions.ApplyOptions([]client.DeleteOption{client.PropagationPolicy(metav1.DeletePropagationForeground)})
+	ExpectWithOffset(1, client.IgnoreNotFound(cl.DeleteAllOf(ctx, pvc, &client.DeleteAllOfOptions{
+		ListOptions: client.ListOptions{
+			Namespace:     namespace,
+			LabelSelector: k8s_labels.NewSelector().Add(*r),
+		},
+		DeleteOptions: delOptions,
+	}))).ShouldNot(HaveOccurred())
 }
--- a/test/e2e/etcd_backup_test.go
+++ b/test/e2e/etcd_backup_test.go
@@ -137,7 +137,7 @@ var _ = Describe("Etcd Backup", func() {
 
 					logger.Info("waiting for sts to become unready", "statefulSetName", etcdName)
 					Eventually(func() error {
-						ctx, cancelFunc := context.WithTimeout(context.TODO(), singleNodeEtcdTimeout)
+						ctx, cancelFunc := context.WithTimeout(context.Background(), singleNodeEtcdTimeout)
 						defer cancelFunc()
 
 						sts := &appsv1.StatefulSet{}
@@ -153,7 +153,7 @@ var _ = Describe("Etcd Backup", func() {
 
 					logger.Info("waiting for sts to become ready again", "statefulSetName", etcdName)
 					Eventually(func() error {
-						ctx, cancelFunc := context.WithTimeout(context.TODO(), singleNodeEtcdTimeout)
+						ctx, cancelFunc := context.WithTimeout(context.Background(), singleNodeEtcdTimeout)
 						defer cancelFunc()
 
 						sts := &appsv1.StatefulSet{}
--- a/test/e2e/etcd_backup_test.go
+++ b/test/e2e/etcd_backup_test.go
@@ -28,7 +28,6 @@ import (
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8s_labels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
@@ -84,7 +83,7 @@ var _ = Describe("Etcd Backup", func() {
 					objLogger := logger.WithValues("etcd", client.ObjectKeyFromObject(etcd))
 
 					By("Create etcd")
-					createAndCheckEtcd(ctx, cl, objLogger, etcd)
+					createAndCheckEtcd(ctx, cl, objLogger, etcd, singleNodeEtcdTimeout)
 
 					By("Check initial snapshot is available")
 					var podName = fmt.Sprintf("%s-0", etcdName)
@@ -138,7 +137,7 @@ var _ = Describe("Etcd Backup", func() {
 
 					logger.Info("waiting for sts to become unready", "statefulSetName", etcdName)
 					Eventually(func() error {
-						ctx, cancelFunc := context.WithTimeout(context.TODO(), timeout)
+						ctx, cancelFunc := context.WithTimeout(context.TODO(), singleNodeEtcdTimeout)
 						defer cancelFunc()
 
 						sts := &appsv1.StatefulSet{}
@@ -149,12 +148,12 @@ var _ = Describe("Etcd Backup", func() {
 							return fmt.Errorf("sts %s is still in ready state", etcdName)
 						}
 						return nil
-					}, timeout*3, pollingInterval).Should(BeNil())
+					}, singleNodeEtcdTimeout, pollingInterval).Should(BeNil())
 					logger.Info("sts is unready", "statefulSetName", etcdName)
 
 					logger.Info("waiting for sts to become ready again", "statefulSetName", etcdName)
 					Eventually(func() error {
-						ctx, cancelFunc := context.WithTimeout(context.TODO(), timeout)
+						ctx, cancelFunc := context.WithTimeout(context.TODO(), singleNodeEtcdTimeout)
 						defer cancelFunc()
 
 						sts := &appsv1.StatefulSet{}
@@ -165,7 +164,7 @@ var _ = Describe("Etcd Backup", func() {
 							return fmt.Errorf("sts %s unready", etcdName)
 						}
 						return nil
-					}, timeout*3, pollingInterval).Should(BeNil())
+					}, singleNodeEtcdTimeout, pollingInterval).Should(BeNil())
 					logger.Info("sts is ready", "statefulSetName", etcdName)
 
 					// verify existence and correctness of keys 1 to 30
@@ -178,7 +177,7 @@ var _ = Describe("Etcd Backup", func() {
 					}
 
 					By("Delete etcd")
-					deleteAndCheckEtcd(ctx, cl, objLogger, etcd)
+					deleteAndCheckEtcd(ctx, cl, objLogger, etcd, singleNodeEtcdTimeout)
 
 				})
 			})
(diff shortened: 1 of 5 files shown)


test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,441 @@
...
+
+			By("Wakeup etcd (Scale up from 0->3)")
+			wakeupEtcd(ctx, cl, objLogger, etcd)
+
+			By("Zero downtime rolling updates")
+			job := startEtcdZeroDownTimeValidatorJob(ctx, cl, etcd, "rolling-update")
+			// this defer ensures to remove the job, if test case breaks before deleting the job.
+			defer cleanUpTestHelperJob(ctx, cl, client.ObjectKeyFromObject(job))
+
+			ExpectWithOffset(1, cl.Get(ctx, client.ObjectKeyFromObject(etcd), etcd)).To(Succeed())
```
timuthy, line 91:
```suggestion
			Expect(cl.Get(ctx, client.ObjectKeyFromObject(etcd), etcd)).To(Succeed())
```

`ExpectWithOffset` is only needed when called in a helper function.

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,441 @@
...
+	logger.Info("Checking statefulset")
+	Eventually(func() error {
+		sts := &appsv1.StatefulSet{}
+		ExpectWithOffset(1, cl.Get(ctx, client.ObjectKeyFromObject(etcd), sts)).To(Succeed())
+
+		if sts.Status.ReadyReplicas != 0 {
+			return fmt.Errorf("sts %s not ready", etcd.Name)
+		}
+		return nil
+	}, timeout*3, pollingInterval).Should(BeNil())
```
timuthy, line 230: Can you check the `statefulset` after checking the `etcd` resource? This can then also be done without `Eventually` because the `etcd` resource should only meet the requirements after the `statefulset` is ready.

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,441 @@
...
+		}
+		// TODO: uncomment me once scale down is supported,
+		// currently ClusterSize is not updated while scaling down.
+		// if *etcd.Status.ClusterSize != 0 {
+		// 	return fmt.Errorf("etcd %q cluster size is %d, but expected to be 0",
+		// 		etcdName, *etcd.Status.ClusterSize)
+		// }
+
+		for _, c := range etcd.Status.Conditions {
+			if c.Status != v1alpha1.ConditionUnknown {
```
timuthy, line 257: FYI: This will change with https://github.com/gardener/etcd-druid/pull/415, i.e. `status` can also remain `true`.

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,441 @@
...
+func wakeupEtcd(ctx context.Context, cl client.Client, logger logr.Logger, etcd *v1alpha1.Etcd) {
+	ExpectWithOffset(1, cl.Get(ctx, client.ObjectKeyFromObject(etcd), etcd)).To(Succeed())
+	etcd.Spec.Replicas = multiNodeEtcdReplicas
+	updateAndCheckEtcd(ctx, cl, logger, etcd)
+}
+
+// updateAndCheckEtcd updates the given etcd obj in the Kubernetes cluster.
+func updateAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logger, etcd *v1alpha1.Etcd) {
+	sts := &appsv1.StatefulSet{}
+	ExpectWithOffset(1, cl.Get(ctx, client.ObjectKeyFromObject(etcd), sts)).To(Succeed())
```
timuthy, line 278: This helper func is called in tests directly as well as by other helper funcs. Thus, there isn't always an offset of `1`.

timuthy (changes requested):
I'm not completely done with the review and plan to do some test runs. Nevertheless, please see already some feedback. In general it looks good, thanks a lot for the tests 🙂

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,441 @@
...
+			Expect(deployBackupSecret(parentCtx, cl, logger, provider, etcdNamespace, storageContainer))
+
+		})
+
+		AfterEach(func() {
+			// remove etcd objects if any old etcd objects exists.
+			purgeEtcd(parentCtx, cl, providers)
+		})
+
+		It("should perform etcd operations", func() {
```
unmarshall, line 70: Would it not make sense to use `BeforeAll`, `AfterAll` and `Ordered` containers?
One test here is doing way too much. I can see that to circumvent it you use `By` clause.
I generally prefer creating smaller tests where i can see all assertions and then individual test will only assert one behavior. Since it is ordered, you can use the state from the previous test as well.

If you see this one single BIG test then it tests for hibernation and wake up from hibernation, then it also tests  rolling update, defragmentation etc. These are not really correlated.

timuthy: > Since it is ordered, you can use the state from the previous test as well.

Can you explain what you mean by ordered? There is an order decorator in Ginkgo 2.x which we currently don't use [ref](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#ordered-decorator). Otherwise there is no order of `It`s as far as I can see 👀

unmarshall: Yes [this](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#ordered-decorator) is correct. I see that etcd-druid currently only uses v1 - :(

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,441 @@
...
+			storageContainer string
+			parentCtx        context.Context
+			provider         TestProvider
+			providers        []TestProvider
+			err              error
+		)
+
+		BeforeEach(func() {
+			parentCtx = context.Background()
+			providers, err = getProviders()
```
unmarshall, line 44: This is already done in `suite_test.go` in `BeforeSuite`. Do we gain something by doing this again here?

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,441 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			// take first provider
+			provider = providers[0]
+			cl, err = getKubernetesClient(kubeconfigPath)
```
unmarshall, line 48: This is already done in suite_test.go in BeforeSuite. Do we gain something by doing this again here?

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,441 @@
...
+			providers        []TestProvider
+			err              error
+		)
+
+		BeforeEach(func() {
+			parentCtx = context.Background()
+			providers, err = getProviders()
+			Expect(err).ToNot(HaveOccurred())
+			// take first provider
+			provider = providers[0]
```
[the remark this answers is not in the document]
seshachalam-yv: Hi @unmarshall, 
Yes we wanted to test first backup provider.  
In e2e etcd multi-node test, wanted to focus on etcd multi-node cases with single backup provider.
Backup tests for all providers are covered [here](https://github.com/gardener/etcd-druid/blob/master/test/e2e/etcd_backup_test.go)  

> How does one test it against a specific provider. Ideally that should be input to ginkgo maybe via environment variable?

Yes, we can specify specific provider by setting env variable 
```
 make \
  KUBECONFIG="$HOME/.kube/config" \
  PROVIDERS="aws" \
  TEST_ID="some-test-id" \
  STEPS="test" \
test-e2e
```

test/e2e/etcd_multi_node_test.go
seshachalam-yv: Fixed, `updateAndCheckEtcd` func is called in tests only.

test/e2e/etcd_backup_test.go
```diff
@@ -185,7 +189,10 @@ var _ = Describe("Etcd Backup", func() {
 
 func createAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logger, etcd *v1alpha1.Etcd) {
 	ExpectWithOffset(1, cl.Create(ctx, etcd)).ShouldNot(HaveOccurred())
+	checkEtcdReady(ctx, cl, logger, etcd)
+}
 
+func checkEtcdReady(ctx context.Context, cl client.Client, logger logr.Logger, etcd *v1alpha1.Etcd) {
 	logger.Info("Waiting for etcd to become ready")
 	Eventually(func() error {
```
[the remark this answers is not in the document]
seshachalam-yv: Fixed ❤️

test/e2e/etcd_multi_node_test.go
seshachalam-yv: Fixed, `providers`  used from `suite_test.go`.

test/e2e/utils.go
```diff
@@ -728,3 +741,120 @@ func getEnvAndExpectNoError(key string) string {
...
+							"if [ $? -gt 0 ] ; then let failed++; echo \"etcd is unhealthy and retrying\"; continue;  fi ; " +
+							"echo \"etcd is healthy\";  touch /tmp/healthy; let failed=0; " +
+							"sleep 1; done;  echo \"etcd is unhealthy\"; exit 1;" +
+							"' > test.sh && sh test.sh",
+					},
+					ReadinessProbe: &v1.Probe{
+						InitialDelaySeconds: int32(5),
+						FailureThreshold:    int32(1),
+						PeriodSeconds:       int32(1),
+						SuccessThreshold:    int32(3),
```
ishan16696, line 818: why `successThreshold` is set to `3` ? Is successThreshold:`1` not good enough ?

test/e2e/utils.go
```diff
@@ -728,3 +741,120 @@ func getEnvAndExpectNoError(key string) string {
...
+			Containers: []corev1.Container{
+				{
+					Name:    "etcd-zero-down-time-validator-" + testName,
+					Image:   "alpine/curl",
+					Command: []string{"/bin/sh"},
+					//To avoid flakiness, consider downtime when curl fails consecutively back-to-back.
+					Args: []string{"-ec",
+						"echo '" +
+							"failed=0 ; threshold=2 ; " +
+							"while [ $failed -lt $threshold ] ; do  " +
```
ishan16696, line 807: > //To avoid flakiness, consider downtime when curl fails consecutively back-to-back.

Is consecutively failure of `curl` determine by `threshold=2` ?
And how it helps in removing the flakiness ?

test/e2e/utils.go
```diff
@@ -728,3 +741,120 @@ func getEnvAndExpectNoError(key string) string {
...
+			Template: corev1.PodTemplateSpec{
+				Spec: *podSpec,
+			},
+			BackoffLimit: pointer.Int32(0),
+		},
+	}
+}
+
+// etcdZeroDownTimeValidatorJob returns k8s job which ensures
+// Etcd cluster zero down time by continuously checking etcd cluster health.
```
ishan16696, line 763:
```suggestion
// Etcd cluster(size>1) zero down time by continuously checking etcd cluster health.
```

ishan16696 (changes requested):
Hi @seshachalam-yv ,
Thanks for PR. Can you also please rebase this PR on latest master ?

test/e2e/utils.go
seshachalam-yv: Zero downtime validator Job which continuously `curls` **Etcd cluster health** before triggering our test case.
let say when we trigger rolling update `etcd-03` pod will be restarted  at the same moment curl may point to IP of  `etcd-3 pod` because we are using `etcd-client` svc (This happens in rare case).

test/e2e/utils.go
seshachalam-yv: After two Curl fails continuously, we exit with code 1.
successThreshold is `3` pod to be ready, at least two curls to be succeeded out of 3 curls, this will reduce  fake signal of cluster health before starting next test case

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,435 @@
...
+		})
+	})
+
+	Context("when single-node is configured", func() {
+		var (
+			etcdName         string
+			storageContainer string
+			parentCtx        context.Context
+			provider         TestProvider
+		)
```
ishan16696, line 124: I guess there is no need to re-declare them if you declare them in inside `Describe(){...}`

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,435 @@
...
+)
+
+var _ = Describe("Etcd", func() {
+	Context("when multi-node is configured", func() {
+		var (
+			etcdName         string
+			storageContainer string
+			parentCtx        context.Context
+			provider         TestProvider
+		)
```
ishan16696, line 38: you can move this inside the describe, so you can re-use them in each `Context`

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,435 @@
...
+			// purge any existing backups in bucket
+			Expect(purgeSnapstore(store)).To(Succeed())
+			Expect(deployBackupSecret(parentCtx, cl, logger, provider, etcdNamespace, storageContainer))
+
+		})
+
+		AfterEach(func() {
+			// remove etcd objects if any old etcd objects exists.
+			purgeEtcd(parentCtx, cl, providers)
+		})
```
ishan16696, line 62: In this 1 `Context` there is only 1 `It(){...}` and similar `BeforeEach` and `AfterEach` is used again for another `Context` , then why not move this `BeforeEach(){...}` and `AfterEach(){...}` inside the `Describe` ?

ishan16696 (changes requested):

test/e2e/etcd_multi_node_test.go
seshachalam-yv: Fixed

test/e2e/etcd_multi_node_test.go
seshachalam-yv: Fixed

ishan16696 (approved):
LGTM!!

timuthy (commented):
No further feedback from my side. Due to other HA tasks in `gardener/gardener` I'll only be able to take another look after next week. Please feel free to merge since @ishan16696 already gave his agreement.

ishan16696:
@unmarshall do you want to take another look ?

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,414 @@
...
+func getEtcdLeaderPodName(ctx context.Context, cl client.Client, namespace string) (*types.NamespacedName, error) {
+	leaseList := &v1.LeaseList{}
+	opts := &client.ListOptions{Namespace: namespace}
+	ExpectWithOffset(1, cl.List(ctx, leaseList, opts)).ShouldNot(HaveOccurred())
+
+	for _, lease := range leaseList.Items {
+		if lease.Spec.HolderIdentity == nil {
+			return nil, fmt.Errorf("error occurred while finding leader, etcd lease %q spec.holderIdentity is nil", lease.Name)
+		}
+		if strings.Contains(*lease.Spec.HolderIdentity, "Leader") {
```
unmarshall, line 328: If you see the backup-restore code the convention of memberID is "memberName:<Leader | Member>
So you accurate test should be: `strings.HasSuffix(holderIdentity, ":Leader")`

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+// getPodLogs returns logs for the given pod
+func getPodLogs(ctx context.Context, PodKey *types.NamespacedName, opts *corev1.PodLogOptions) (string, error) {
+	typedClient, err := getKubernetesTypedClient(kubeconfigPath)
+	if err != nil {
+		return "", err
+	}
+
+	req := typedClient.CoreV1().Pods(namespace).GetLogs(PodKey.Name, opts)
```
unmarshall, line 342: depending upon pod logs is very error prone, instead it should be reflected in the status as a condition. So for now checking the logs is fine but this should be corrected.

ishan16696: I agree with @unmarshall , so what I would suggest is get the db size of each member (before and after defragmentation call), now compare the db size, if new db size of each member is less then old db size then defragmentation was successful. check [here](https://github.com/gardener/etcd-backup-restore/blob/master/pkg/defragmentor/defrag_test.go#L297-L340)
My only fear is it might cause flaky tests as we are doing the similar thing in unit tests of defragmentation in backup-restore and it becomes flaky.

seshachalam-yv: @ishan16696 
How do we i know defragmention is completed or not ?
Is it based on db size ?

ishan16696: yes, basically defragmentation freeup the database space but before calling the defragmentation you have to put some dummy data into etcd cluster, so that database get increase somewhat  and defragmentation will be effective.

test/e2e/etcd_backup_test.go
```diff
@@ -197,83 +203,94 @@ func createAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logge
...
+	// because sometimes k8s garbage collection is delayed to remove PVCs before starting next tests.
+	purgeEtcdPVCs(ctx, cl, etcd.Name)
+}
+
+func purgeEtcdPVCs(ctx context.Context, cl client.Client, etcdName string) {
+	r, _ := k8s_labels.NewRequirement("instance", selection.Equals, []string{etcdName})
+	pvc := &corev1.PersistentVolumeClaim{}
+	delOptions := client.DeleteOptions{}
+	delOptions.ApplyOptions([]client.DeleteOption{client.PropagationPolicy(metav1.DeletePropagationForeground)})
+	ExpectWithOffset(1, client.IgnoreNotFound(cl.DeleteAllOf(ctx, pvc, &client.DeleteAllOfOptions{
```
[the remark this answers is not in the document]
seshachalam-yv: Sorry @unmarshall, there was confusion.
[Here](https://github.com/gardener/etcd-druid/pull/418/files#diff-13ee08488a82e7ced9773ac78e459f75ef48784b404b7c2f7e065175f21f530aR254) we are deleting the k8s objects, So we need  `matchers.BeNotFoundError()` which checks if error is NotFound.
But in this case deleting pvc and want to Ignore NotFound error.

test/e2e/etcd_multi_node_test.go
seshachalam-yv: @ishan16696 
 I hope @unmarshall point is that we should update the status of defgramentation as one of the status conditions.
 Right now we rely on pod logs for degfragmention status and  I dont think so  checking the database size is a proper way of confirming the degferamention is completed or not.

test/e2e/etcd_backup_test.go
```diff
@@ -185,9 +188,12 @@ var _ = Describe("Etcd Backup", func() {
 
 func createAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logger, etcd *v1alpha1.Etcd) {
 	ExpectWithOffset(1, cl.Create(ctx, etcd)).ShouldNot(HaveOccurred())
+	checkEtcdReady(ctx, cl, logger, etcd)
+}
 
+func checkEtcdReady(ctx context.Context, cl client.Client, logger logr.Logger, etcd *v1alpha1.Etcd) {
 	logger.Info("Waiting for etcd to become ready")
-	Eventually(func() error {
+	EventuallyWithOffset(2, func() error {
```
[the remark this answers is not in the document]
seshachalam-yv: Fixed as part of this [commit](https://github.com/gardener/etcd-druid/pull/418/commits/c7342d18b9438f1ba659b68c093aefbae9d0a10c#diff-13ee08488a82e7ced9773ac78e459f75ef48784b404b7c2f7e065175f21f530aR196)

test/e2e/etcd_backup_test.go
```diff
@@ -197,83 +203,94 @@ func createAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logge
 		}
 
 		if &etcd.Status == nil || etcd.Status.Ready == nil || *etcd.Status.Ready != true {
-			return fmt.Errorf("etcd %s not ready", etcd.Name)
```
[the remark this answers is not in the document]
seshachalam-yv: Fixed as part of this [commit](https://github.com/gardener/etcd-druid/pull/418/commits/c7342d18b9438f1ba659b68c093aefbae9d0a10c#diff-13ee08488a82e7ced9773ac78e459f75ef48784b404b7c2f7e065175f21f530aR200).

test/e2e/etcd_backup_test.go
```diff
@@ -197,83 +203,94 @@ func createAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logge
...
+		if etcd.Status.ClusterSize == nil {
+			return fmt.Errorf("etcd %s cluster size is empty", etcd.Name)
 		}
 
+		if *etcd.Status.ClusterSize != etcd.Spec.Replicas {
+			return fmt.Errorf("etcd %s cluster size is %v, but it's not expected size as %v",
+				etcd.Name, etcd.Status.ClusterSize, etcd.Spec.Replicas)
+		}
+
+		for _, c := range etcd.Status.Conditions {
```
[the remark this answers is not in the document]
seshachalam-yv: Fixed as part of this [commit](https://github.com/gardener/etcd-druid/pull/418/commits/c7342d18b9438f1ba659b68c093aefbae9d0a10c#diff-13ee08488a82e7ced9773ac78e459f75ef48784b404b7c2f7e065175f21f530aR217-R220) (somehow its not reflecting here)

test/e2e/etcd_multi_node_test.go
```diff
@@ -0,0 +1,414 @@
...
+	ExpectWithOffset(1,
+		client.IgnoreNotFound(
+			cl.Delete(ctx,
+				&batchv1.Job{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      jobName,
+						Namespace: namespace,
+					},
+				},
+				client.PropagationPolicy(metav1.DeletePropagationForeground),
```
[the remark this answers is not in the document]
seshachalam-yv: Sorry @unmarshall, there was confusion.
[Here](https://github.com/gardener/etcd-druid/pull/418/files#diff-13ee08488a82e7ced9773ac78e459f75ef48784b404b7c2f7e065175f21f530aR254) we are deleting the k8s objects, So we need matchers.BeNotFoundError() which checks if error is NotFound.
But in this case deleting obj and want to Ignore NotFound error(if that object already deleted).

test/e2e/etcd_multi_node_test.go
ishan16696: @unmarshall may be we can go ahead with what @seshachalam-yv is doing and may be I will try to pick and improve this. WDYT?

ishan16696:
Hi @seshachalam-yv ,
Can you re-push a commit to re-run the pipeline checks as I'm unable to retrigger the pipeline tests as one of the checks is pending.

test/e2e/etcd_backup_test.go
```diff
@@ -146,12 +148,12 @@ var _ = Describe("Etcd Backup", func() {
...
 						}
 						return nil
-					}, timeout*3, pollingInterval).Should(BeNil())
+					}, singleNodeEtcdTimeout, pollingInterval).Should(BeNil())
 					logger.Info("sts is unready", "statefulSetName", etcdName)
 
 					logger.Info("waiting for sts to become ready again", "statefulSetName", etcdName)
 					Eventually(func() error {
-						ctx, cancelFunc := context.WithTimeout(context.TODO(), timeout)
+						ctx, cancelFunc := context.WithTimeout(context.TODO(), singleNodeEtcdTimeout)
```
unmarshall, line 156: **Nit pick:**
I have seen that there is a general confusion on when to use `context.TODO` and when to use `context.Background`. Both of these return an empty context and the difference is more semantic.
If you look at the documentation for these 2 methods then it will become clear https://github.com/golang/go/blob/master/src/context/context.go#L204-L218
Given the description i would suppose that context.Background would be semantically more appropriate.

test/e2e/etcd_multi_node_test.go
unmarshall: ok as i indicated in the first comment, checking logs is fine for now, but this needs to be changed. let us discuss this separately.

unmarshall (approved):
/lgtm with minor comments

test/e2e/etcd_backup_test.go
seshachalam-yv: Fixed

timuthy (approved):

Issue #419, opened 2022-08-29
[BUG] Cluster scale up - one pod not joining the cluster


aaronfern:
**Describe the bug:**
When an etcd cluster is scaled up, there is a rare case where the third pod cannot join the cluster. Hence, we have only 2 members working is expected ways. What this results in is a cluster is size 2 with a learner that cannot join.
Although this means that the quorum is maintained, it also means that we lose a bit of resilience as one we are now just pod away from quorum loss.

The third pod complains of a `cluster ID mismatch` error and hence refuses to join the cluster. A possible cause of this is incorrect configuration passed to the etcd container before it is ready to join the cluster. The root cause however, is not very clear.

**Expected behavior:**
Cluster scale up to always succeed 

**How To Reproduce (as minimally and precisely as possible):**
It is not yet clear how to consistently reproduce this. It is seen quite sporadically.

abdasgupta:
Closing this issue as the problem is resolved

PR #421, opened 2022-08-30
Peer TLS not enabled after rolling update

### LINKED ISSUE #420: [BUG] Peer TLS not enabled after rolling update
**Describe the bug:**
Configuring `etcd.spec.etcd.peerUrlTLS` to activate TLS for peer communication **for an existing** etcd cluster will trigger a rolling update, i.e. all pods receive the required configuration and certificates. However, after the rolling update is finished, the peer communication does not properly work via TLS.

If TLS is enabled later, the `peer` address(es) in the cluster have to be changed which is done by `etcd-backup-restore`. After this change another restart is required which is currently not triggered. 

**Expected behavior:**
TLS communication works properly after changing the `etcd` resource.

1. Create a multi-node `etcd` cluster w/o peer tls
2. Configure `etcd.spec.etcd.peerUrlTLS`
3. Observe that member pods don't start because of broken TLS communication.

unmarshall:

/area control-plane
/kind bug

For a single node etcd the peer URL is not TLS enabled. This causes an issue when the same etcd cluster has to be scaled up from 1 to 3. All the new etcd pods will try and reach the existing etcd member over TLS enabled peer URL and will get a response in HTTP which will get rejected. As a result all the new pods will go into `CrashLoopBackOff` and will never recover.

Fixes #420 and #419

@timuthy  e2e tests are currently pending.
I have marked it currently as WIP as i have some unit tests to write for the sts component changes.
**Release note**:

```other operator
Enables etcd-druid to trigger restarts of the etcd pods when peer URL is TLS enabled.
```


unmarshall:
unit tests added

pkg/component/etcd/statefulset/values.go
```diff
@@ -124,4 +124,7 @@ type Values struct {
 	AutoCompactionRetention *string
 	// ConfigMapName is the name of the configmap that holds the ETCD config.
 	ConfigMapName string
+	// PeerUrlTLSAlreadyEnabled captures the existing enablement state for PeerUrl TLS. This will be used to compare the updated PeerUrlTLS
+	// to determine if action to update of PeerUrlTLS should be taken
+	PeerUrlTLSAlreadyEnabled *bool
```
timuthy, line 129:
```suggestion
	StatusPeerUrlTLSEnabled *bool
```

Similar to `StatusReplicas`.

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -71,19 +72,49 @@ func (c *component) Deploy(ctx context.Context) error {
...
+}
 
+func (c *component) getNumTimesToRecreateSts(sts *appsv1.StatefulSet) uint8 {
 	if sts.Generation > 1 && clusterScaledUpToMultiNode(c.values) && immutableFieldUpdate(sts, c.values) {
-		// Several immutable fields must be reset for the multi-node use-case.
-		deleteAndWait := gardenercomponent.OpDestroyAndWait(c)
-		if err := deleteAndWait.Destroy(ctx); err != nil {
+		return 1
+	} else if c.IsPeerUrlTLSChangedToEnabled() {
+		c.logger.Info("PeerUrl TLS has been enabled for etcd. To reflect this change etcd StatefulSet has to be re-created 2 times", "namespace", c.values.Namespace, "name", c.values.Name, "etcdUID", c.values.EtcdUID)
```
timuthy, line 87: Can you add a comment explaining why we need two restarts of all etcd pods?

timuthy (changes requested):
Thanks for the changes. I added a few in-line comments.

controllers/etcd_controller.go
```diff
@@ -686,21 +703,109 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
...
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      configMapNamespacedName.Name,
+			Namespace: configMapNamespacedName.Namespace,
+		},
+	}
+	err := r.Client.Get(ctx, configMapNamespacedName, cm)
+	switch {
+	case apierrors.IsNotFound(err):
+		return nil, nil
```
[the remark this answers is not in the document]
unmarshall: It was an open point and my initial assumption was that if the config map is not there then i cannot assume any value for peerUrlTLS and therefore just return nil with the hope that it will get created. But I think this assumption is true in the gardener context and if we see usage of etcd-druid sans gardener then maybe it is better to just return the error and not ignore it.

controllers/etcd_controller.go
```diff
@@ -686,21 +703,109 @@ func (r *EtcdReconciler) reconcileEtcd(ctx context.Context, logger logr.Logger,
...
+		peerUrlTLSEnabled: peerUrlTLSEnabled,
+		svcName:           &serviceValues.ClientServiceName,
+		sts:               sts,
+	}, err
+}
+
+// -------------------------- BEGIN: To be removed later --------------------------------------
+
+// capturePeerUrlTLSEnabledStatus captures the state of peer URL TLS enablement in etcd status.
+// It will fetch any existing ConfigMap and will derive the current state of PeerTLS enablement.
```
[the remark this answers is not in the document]
unmarshall: Yes this was said in context to etcd so it was implicit. But may be it is better to make it explicit, removes mis-interpretations

controllers/etcd_controller.go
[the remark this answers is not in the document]
timuthy: Got it, thank. Why don't we call `capturePeerUrlTLSEnabledStatus` later in the flow, i.e. right before the `StatefulSet` is deployed? We don't even need to consult the `ConfigMap`, but we can enhance the ConfigMap-Component as it already has the data in memory.

pkg/component/etcd/statefulset/statefulset.go
unmarshall: Ok i will add it.

controllers/etcd_controller.go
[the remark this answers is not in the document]
unmarshall: The reason to always depend upon ConfigMap was that in the event of a crash, the in-memory state is not lost. Can you elaborate of what you have in mind?

ishan16696:
Is this PR still in WIP state ?

unmarshall:
> Is this PR still in WIP state @unmarshall ?

No. I have removed WIP from the subject line. Kindly review it critically.

pkg/component/etcd/lease/lease.go
```diff
@@ -75,10 +82,23 @@ func (c *component) Destroy(ctx context.Context) error {
...
 
+func (c *component) GetPeerURLTLSEnabledStatus(ctx context.Context) (bool, error) {
+	tlsEnabledValues, err := c.getTLSEnabledAnnotationValues(ctx, c.values.EtcdName)
+	if err != nil {
+		return false, err
+	}
+	tlsEnabled := true
+	for _, v := range tlsEnabledValues {
+		tlsEnabled = tlsEnabled && v
+	}
```
ishan16696, line 93: AFAIU, you are checking the annotation in lease object of all etcd clusters  member , right but why ? 
Is there a reason behind it ?

ishan16696:
`e2e-tests` have passed successfully  for this PR for storageProvider: Azure and AWS.

pkg/component/etcd/lease/lease.go
unmarshall: The idea is to check if there is any member who still has a value of false. If the annotation is non-existent then it is ignored because its possible that the etcdbr has still not been able to set this annotation.
In the case the annotation is false then it requires 2 restarts in total.

Having said that, ideally this will return false when there is only a single-member etcd in case of gardener. Because today a single member etcd cluster is launched with a peer URL which does not have TLS enabled. When there is a multi-node cluster then it will be ensured from gardener that all new members will always have peer URL TLS enabled.

However, the above assumption is only relevant in case of gardener.

ishan16696: ok, got it

aaronfern (approved):
Looks good from my point of view
I've tested this out and all flows are working as expected
/lgtm

ishan16696 (approved):

timuthy (commented):
Unfortunately, I can only have another look beginning next week. @aaronfern or @ishan16696 please feel free to merge if everything was tested successfully and the PR is in good shape, since I don't want to block the release.

PR #422, opened 2022-08-30
Sort members in etcd status

timuthy:

/area control-plane
/kind enhancement

This PR makes the members check to sort the array before the `etcd` status is updated.

**Release note**:

```other operator
NONE
```


aaronfern (approved):
/lgtm

PR #424, opened 2022-09-01
Remove startup and liveness probes

aaronfern:
The enablement of startup/liveness probes through https://github.com/gardener/etcd-druid/pull/396 showed that they cause more harm than good:
- The startup time of etcds can vary depending on the state and amount of data
- If startup does not happen in the expected time, the failing probes kill the container which does not help to solve the issue at all but will end in a endless loop of restarts
- Liveness probes had been disabled for a long time before which never caused issues in our experience.
- Other communities have come to a similar conclusion, see https://github.com/improbable-eng/etcd-cluster-operator/blob/master/docs/operations.md#why-arent-there-liveness-probes-for-the-etcd-pods

/area control-plane
/kind enhancement

Cherry pick of [`a8cb33`](https://github.com/gardener/etcd-druid/pull/423/commits/a8cb336d3d4dca9100f673b2baa96ad3e01b8bba)

Fixes #

**Release note**:

```other operator @timuthy 
Liveness and startup probes for etcd were removed. After activating them, we noticed that they cause more harm than good since the startup time for etcd clusters varies and isn't predicable. Killing the `etcd` container in such a case doesn't solve the situation and will rather end in an endless loop of restarts. This change will cause a restart of etcd clusters.
```

### DIFF
--- a/controllers/etcd_controller_test.go
+++ b/controllers/etcd_controller_test.go
@@ -995,33 +995,6 @@ func validateEtcdWithDefaults(instance *druidv1alpha1.Etcd, s *appsv1.StatefulSe
 								"PeriodSeconds":       Equal(int32(5)),
 								"FailureThreshold":    Equal(int32(5)),
 							})),
-							"LivenessProbe": PointTo(MatchFields(IgnoreExtras, Fields{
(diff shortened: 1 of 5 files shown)


timuthy (approved):
Thank you!!
/lgtm

PR #430, opened 2022-09-08
Handle PDB API version upgrade

### LINKED ISSUE #429: [Feature] Update API versions for Cronjobs and PDBs
From kubernetes v1.25 onwards, `Cronjob`s and `PDB`s are [depricated](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25) from the `v1beta1` API and will instead be in the `v1` API

This issue recommends updating `etcd-druid` to handle both API versions based on the detected kubernetes version

Druid needs to run of various kubernetes clusters and should not have a issues with specific versions

Use the discovery API to find out kubetnetes version and take appropriate action

aaronfern:

/area control-plane
/kind api-change
/kind cleanup

1. This PR handles the API version for `PodDisruptionBudget` (ref: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25)
2. The claim logic for PDBs has been removed as it's no longer needed
3. Drops support for `CronJob` deletion that was present because of etcd-druid v0.6.0

Fixes #429 

**Release note**:

```other operator
Etcd druid will now use `policy/v1` for `PodDisruptionBudget`s for kubernetes >= 1.21. However, for kubernetes < 1.21, `PodDisruptionBudget`s will still default to `policy/v1beta1`
```
```breaking operator
Claim logic for `PodDisruptionBudget`s have been removed. Already existing `PodDisruptionBudget` objects cannot be adopted anymore
```
```breaking operator
Logic for deleting any leftover `CronJob`s created by `etcd-druid:v0.6.0` has been removed. Please deploy a lower version of `etcd-druid` before upgrading if you still have any leftover `CronJob`s or manually delete them
```


controllers/etcd_controller.go
```diff
@@ -406,59 +348,81 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
 
 func (r *EtcdReconciler) reconcilePodDisruptionBudget(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd, values map[string]interface{}) error {
 	logger.Info("Reconcile PodDisruptionBudget")
-	pdb := &policyv1beta1.PodDisruptionBudget{}
-	err := r.Get(ctx, types.NamespacedName{Name: etcd.Name, Namespace: etcd.Namespace}, pdb)
 
-	if err == nil {
-		// pdb already exists, claim it
+	// Find out k8s version
+	targetCoreClient, _ := k8s.NewForConfig(r.Config)
```
shafeeqes, line 353: You could return the error.

controllers/etcd_controller.go
```diff
@@ -406,59 +348,81 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
+	targetServerVersion, _ := targetCoreClient.Discovery().ServerVersion()
```
shafeeqes, line 354: same here.

controllers/etcd_controller.go
```diff
@@ -406,59 +348,81 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
+	targetKubernetesVersion, _ := semver.NewVersion(targetServerVersion.GitVersion)
```
shafeeqes, line 355: same here.

pkg/utils/k8sutils/version.go
```diff
@@ -0,0 +1,47 @@
...
+func init() {
+	var err error
+
+	ConstraintK8sGreaterEqual121, err = semver.NewConstraint(">= 1.21")
+	utilruntime.Must(err)
+}
+
+// CheckVersionMeetsConstraint returns true if the <version> meets the <constraint>.
+func CheckVersionMeetsConstraint(version, constraint string) (bool, error) {
+	c, err := semver.NewConstraint(constraint)
```
unmarshall, line 36: is there a reason to add a dependency on another lib just to check version ?

timuthy (changes requested):
Thanks for the PR. I added some comments at places that need a rework as it's currently not working for `1.25` clusters as expected. There is also a PDB handling in the `custodian-controller` which hasn't been adjusted. Can you please check this?

pkg/utils/k8sutils/version.go
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright (c) 2022 SAP SE or an SAP affiliate company. All rights reserved.
```
unmarshall, line 2: any reason to create k8sutil under util?
We could have just called it version.go or k8s_version.go and directly put it under utils package.

controllers/etcd_controller.go
```diff
@@ -414,59 +357,90 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
-		selector, err := metav1.LabelSelectorAsSelector(etcd.Spec.Selector)
-		if err != nil {
-			logger.Error(err, "Error converting etcd selector to selector")
-			return err
-		}
 
-		logger.Info("Claiming pdb object")
-		_, err = r.claimPodDisruptionBudget(ctx, etcd, selector, pdb)
+	// Find out k8s version
+	targetCoreClient, err := k8s.NewForConfig(r.Config)
```
unmarshall, line 362: can you move this to a utility function in utils/k8s_version.go? So essentially create the following function:
```go
func GetTargetK8sVersion(config *rest.Config) (*semver.Version, error) {
}
```

controllers/etcd_controller.go
```diff
@@ -414,59 +357,90 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
-	renderedChart, err := r.chartApplier.Render(chartPath, etcd.Name, etcd.Namespace, values)
-	if err != nil {
-		return nil, err
-	}
-	if content, ok := renderedChart.Files()[pdbPath]; ok {
-		decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader([]byte(content)), 1024)
-		if err = decoder.Decode(&decoded); err != nil {
-			return nil, err
+		// Required podDisruptionBudget doesn't exist. Create new
+		pdbPath := getChartPathForPodDisruptionBudget()
```
unmarshall, line 388: the chart contains `v1beta1` for PDB, will this create a PDB version version = v1 when k8s version > 1.21 and there is no existing PDB? Can you double check this?

controllers/etcd_controller.go
```diff
@@ -414,59 +357,90 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
+		if err != nil {
+			return err
+		}
+
+		logger.Info("Creating PodDisruptionBudget", "poddisruptionbudget", kutil.Key(pdb.Namespace, pdb.Name).String())
+		if err := r.Create(ctx, pdb); err != nil {
+			return err
+		}
+	} else {
+		pdb := &policyv1beta1.PodDisruptionBudget{}
```
unmarshall, line 409: There is a lot of code repetition. One proposal is to use generics.  Something similar to the following code. I have not tested it obviously but this just hints at how to use it.
Unfortunately go generics is quite half baked (that is what happens when it is not included in the original language) so you cannot use generics for methods - https://github.com/golang/go/issues/49085
```go
func createPdbIfNotExists[T client.Object](ctx context.Context, cl client.Client, chartApplier kubernetes.ChartApplier, etcdNsName types.NamespacedName, values map[string]interface{}, pdbObj T) error {
	err := cl.Get(ctx, etcdNsName, pdbObj)
	if err == nil {
		return nil
	}
	if !apierrors.IsNotFound(err) {
		return err
	}

	pdbPath := getChartPathForPodDisruptionBudget()
	chartPath := getChartPath()
	renderedChart, err := chartApplier.Render(chartPath, etcdNsName.Name, etcdNsName.Namespace, values)
	if err != nil {
		return err
	}
	if content, ok := renderedChart.Files()[pdbPath]; ok {
		decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader([]byte(content)), 1024)
		if err = decoder.Decode(&pdbObj); err != nil {
			return err
		}
	}
	if err != nil {
		return err
	}

	if err := cl.Create(ctx, pdbObj); err != nil {
		return err
	}
	return nil
}
```

controllers/etcd_controller.go
```diff
@@ -414,59 +357,90 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
-
-	if !apierrors.IsNotFound(err) {
+	targetServerVersion, err := targetCoreClient.Discovery().ServerVersion()
+	if err != nil {
 		return err
 	}
-
-	// Required podDisruptionBudget doesn't exist. Create new
-	pdb, err = r.getPodDisruptionBudgetFromEtcd(etcd, values, logger)
+	targetKubernetesVersion, err := semver.NewVersion(targetServerVersion.GitVersion)
```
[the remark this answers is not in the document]
shafeeqes: But won't the target version always be a semver compliant one? IIUC, the above issue arises becauses it's a seed component.

pkg/component/etcd/poddisruptionbudget/values.go
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
unmarshall, line 1: should be 2022 and not 2021 as this is a new file

pkg/component/etcd/poddisruptionbudget/values_helper.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
unmarshall, line 1: should be 2022 and not 2021 as this is a new file

pkg/utils/k8s_version.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2019 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
unmarshall, line 1: should be 2022 and not 2019 as this is a new file

pkg/utils/poddisruptionbudget.go
```diff
@@ -0,0 +1,67 @@
...
+package utils
+
+import (
+	"math"
+
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+)
+
+// CalculatePDBminAvailable calculates the minimum available value for the PDB
+func CalculatePDBminAvailable(etcd *druidv1alpha1.Etcd) int {
```
unmarshall, line 24: Name should be `CalculatePDBMinAvailable` and this is only used in pdb package. No reason to move this to the utils package. It can be moved to values_helper.go

pkg/utils/poddisruptionbudget.go
```diff
@@ -0,0 +1,67 @@
...
+	// do not enable in single mode
+	if etcd.Spec.Replicas < 2 {
+		return 0
+	}
+
+	// cluster is in bootstrap
+	if etcd.Status.ClusterSize == nil {
```
unmarshall, line 31: Is this check sufficient to check if the cluster is in bootstrap stage?

pkg/utils/poddisruptionbudget.go
```diff
@@ -0,0 +1,67 @@
...
+			allMembersReady = true
+			break
+		}
+	}
+
+	clusterSize := int(*etcd.Status.ClusterSize)
+	values := make([]int, 0)
+	values = append(values, int(math.Floor(float64(clusterSize)/float64(2))+1))
+
+	if !allMembersReady {
```
unmarshall, line 48: Is this check required at all? In which case would you use the check for all members ready. 
In all cases would it not be sufficient to have PDB always set to quorum size?

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,175 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (c *component) syncPodDisruptionBudget(ctx context.Context, pdb interface{}) error {
+	switch pdb := pdb.(type) {
+	case *policyv1.PodDisruptionBudget:
+		_, err := controllerutils.GetAndCreateOrMergePatch(ctx, c.client, pdb, func() error {
```
unmarshall, line 130: Will `OperationResult` which is the ignored argument contain additional information when `GetAndCreateOrMergePatch` returns an error?

controllers/etcd_custodian_controller.go
```diff
@@ -130,7 +131,9 @@ func (ec *EtcdCustodian) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{}, err
 	}
 
-	if err := ec.updatePodDisruptionBudget(ctx, logger, etcd, refMgr); err != nil {
+	pdbValues := componentpdb.GenerateValues(etcd)
+	pdbDeployer := componentpdb.New(ec.Client, etcd.Namespace, &pdbValues, ec.restConfig)
+	if err := pdbDeployer.Deploy(ctx); err != nil {
```
unmarshall, line 136: why is PDB being now created from 2 controllers?

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if err := c.syncPodDisruptionBudget(ctx, pdb); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *component) syncPodDisruptionBudget(ctx context.Context, pdb interface{}) error {
+	switch pdb := pdb.(type) {
```
shafeeqes, line 128: You could do the type switch inside the function in `GetAndCreateOrMergePatch`, like
```
_, err := controllerutils.GetAndCreateOrMergePatch(ctx, k.client.Client(), obj, func() error {
		switch pdb := obj.(type) {
		case *policyv1.PodDisruptionBudget:
...
```

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,175 @@
...
+	k8sversion, _ := utils.GetTargetK8sVersion(c.controllerConfig)
+	pdb := c.emptyPodDisruptionBudget(c.values.EtcdName, k8sversion)
+	switch pdb.(type) {
+	case *policyv1.PodDisruptionBudgetList:
+		pdbv1 := pdb.(*policyv1.PodDisruptionBudget)
+		return client.IgnoreNotFound(c.client.Delete(ctx, pdbv1))
+	case *policyv1beta1.PodDisruptionBudgetList:
+		pdbv1beta1 := pdb.(*policyv1.PodDisruptionBudget)
+		return client.IgnoreNotFound(c.client.Delete(ctx, pdbv1beta1))
+	}
```
shafeeqes, line 173:
```suggestion
pdb := c.emptyPodDisruptionBudget(c.values.EtcdName, k8sversion)
return client.IgnoreNotFound(c.client.Delete(ctx, pdb))
```
`client.Delete` takes `client.Object` as argument, you can directly pass the pdb.

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,175 @@
...
+			}
+			return nil
+		})
+		return err
+	}
+	return nil
+}
+
+func (c *component) Destroy(ctx context.Context) error {
+	k8sversion, _ := utils.GetTargetK8sVersion(c.controllerConfig)
```
shafeeqes, line 164: maybe, return the error?

shafeeqes (changes requested):

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
aaronfern: `emptyPodDisruptionBudget` returns an `interface{}` and I can't pass that to `Delete(..)`

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
aaronfern: Does this really help?
I really don't think it brings some benefit nor reduces the code size by much
wdyt?

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+// getLabels returns the labels for the PDB object
+func getLabels(val *Values) map[string]string {
+	labels := map[string]string{
+		"instance": val.EtcdName,
+	}
+
+	for k, v := range val.Labels {
```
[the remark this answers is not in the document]
aaronfern: Thanks for pointing this out
Updated

pkg/utils/poddisruptionbudget.go
aaronfern: I think so, yes
Custodian controller will take over after that and update the `minAvailable`

controllers/etcd_custodian_controller.go
aaronfern: `Deploy` does two things, creates a PDB and updates it. 
Here, it is intended to keep the PDB up to date
Custodian controller won't start it's job until the sts is up. Etcd controller creates the PDB before creating the sts

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
aaronfern: `OperationResult` is just a string that indicates the executed operation (ref: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#OperationResult)

It's just "unchanged" in case of an error, so no, no additional in case of an error

controllers/etcd_controller.go
```diff
@@ -414,59 +357,90 @@ func (r *EtcdReconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (
...
+			if err = decoder.Decode(&pdb); err != nil {
+				return err
+			}
+		}
+		if err != nil {
+			return err
+		}
+
+		logger.Info("Creating PodDisruptionBudget", "poddisruptionbudget", kutil.Key(pdb.Namespace, pdb.Name).String())
+		if err := r.Create(ctx, pdb); err != nil {
```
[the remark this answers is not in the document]
aaronfern: Moved PDBs to now be deployed using a component and not with charts

pkg/utils/k8sutils/version.go
```diff
@@ -0,0 +1,47 @@
...
+package k8sutils
+
+import (
+	"github.com/Masterminds/semver"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+var (
+	// ConstraintK8sGreaterEqual121 is a version constraint for versions >= 1.21.
+	ConstraintK8sGreaterEqual121 *semver.Constraints
```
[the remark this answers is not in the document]
aaronfern: Other packages should be able to use the constraint. Hence exposed

pkg/utils/k8sutils/version.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func init() {
+	var err error
+
+	ConstraintK8sGreaterEqual121, err = semver.NewConstraint(">= 1.21")
```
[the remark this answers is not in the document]
aaronfern: No hard reason for this. Constraint seems constant and something that will never change at runtime, hence init seemed apt
wdyt?

controllers/etcd_controller.go
aaronfern: Done

controllers/etcd_controller.go
aaronfern: PDBs are now created using components, so this should be resolved now

pkg/utils/k8sutils/version.go
```diff
@@ -0,0 +1,47 @@
...
+	utilruntime.Must(err)
+}
+
+// CheckVersionMeetsConstraint returns true if the <version> meets the <constraint>.
+func CheckVersionMeetsConstraint(version, constraint string) (bool, error) {
```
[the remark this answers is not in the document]
aaronfern: I added it for completeness. I have no strong preference though
wdyt? should it be removed?

pkg/utils/k8sutils/version.go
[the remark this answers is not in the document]
unmarshall: Is there any package using it currently?

pkg/utils/k8sutils/version.go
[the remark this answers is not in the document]
unmarshall: dead code should be removed.

pkg/utils/k8sutils/version.go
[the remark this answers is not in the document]
unmarshall: In general if you make function calls that can return error then it makes it harder for you to test.

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
unmarshall: ok

controllers/etcd_custodian_controller.go
unmarshall: Custodian controller should only update the status and not the object, right? Here there is a potential to actually create/update the PDB causing 2 actors calling `Deploy`. Is there any other resource for which similar thing is done in the custodian controller?

pkg/utils/k8sutils/version.go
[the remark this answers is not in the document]
aaronfern: Currently just the `poddisruptionbudget` package

controllers/etcd_custodian_controller.go
aaronfern: No, there is no other resource for which a similar thing is done in custodian controller

However, I've updated the PR to address this concern. The PDB component now has a method to only sync the `minAvailable` field that's used by the custodian controller

pkg/utils/k8sutils/version.go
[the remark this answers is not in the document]
aaronfern: Removed

controllers/etcd_controller.go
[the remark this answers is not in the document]
aaronfern: Thanks for pointing this out
Adapted this PR to consider k8s versions with suffixes

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,206 @@
...
+type component struct {
+	client               client.Client
+	namespace            string
+	values               *Values
+	controllerConfig     *rest.Config
+	k8sVersionGetterFunc K8sVersionGetter
+}
+
+// K8sVersionGetter defines a func type that accepts a rest configuration and returns the semantic k8s version of the k8s cluster
+type K8sVersionGetter func(config *rest.Config) (*semver.Version, error)
```
shafeeqes, line 45: Instead of using this function and increasing the complexity, Can we just pass it to `New` and store it in the `Values` of the component? This is what we do for every component in gardener as well.

pkg/utils/k8s_version.go
```diff
@@ -0,0 +1,85 @@
...
+	v, err := semver.NewVersion(normalizeVersion(version))
+	if err != nil {
+		return false, err
+	}
+
+	return c.Check(v), nil
+}
+
+// GetTargetK8sVersion returns the semver version of the cluster k8s version
+func GetTargetK8sVersion(config *rest.Config) (*semver.Version, error) {
```
shafeeqes, line 69: Usually, the target means the Shoot, but here the cluster is the Seed. So can we not use it, to avoid confusion?

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
shafeeqes: It's more compact, more readable IMO.
You can even reuse ObjectMeta to reduce more duplication.

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
aaronfern: Updated

pkg/utils/k8s_version.go
aaronfern: Updated this function to drop the `target` prefix for variables and renamed the func to `GetClusterK8sVersion`

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, err
+	}
+	if k8sVersionGreaterEqual121 {
+		return &policyv1.PodDisruptionBudget{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: c.namespace,
+			},
+		}, nil
+	} else {
```
abdasgupta, line 125: Add a TODO with your username to delete this else part when we won't be supporting k8s version lesser than 1.21. WDYT?

pkg/component/etcd/poddisruptionbudget/values.go
```diff
@@ -0,0 +1,30 @@
...
+package poddisruptionbudget
+
+import "k8s.io/apimachinery/pkg/types"
+
+type Values struct {
+	// EtcdName is the name of the etcd resource.
+	EtcdName string
+	// EtcdNameSpace is the namespace of etcd resource
+	EtcdNameSpace string
+	// EtcdID is the UID of the etcd resource.
```
abdasgupta, line 24: //EtcdUID is the UID of the etcd resource

abdasgupta (changes requested):
Overall very good PR. Just few nitpicks

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
shafeeqes: Why not return `client.Object` from `emptyPodDisruptionBudget `?

aaronfern: Thanks for pointing this out
Updated to `emptyPodDisruptionBudget` to return `client.Object`

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
aaronfern: Makes sense
Added comment

pkg/component/etcd/poddisruptionbudget/values.go
aaronfern: Done

shafeeqes:
No more comments from my side.

rfranzke:
Can we continue with this PR?

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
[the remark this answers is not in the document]
unmarshall: Also the labels added in this map are mostly static. Is it then not possible to fully populate the labels in the values_helper.go? You are copying the labels from `val.Labels` and adding 3 additional labels all of which could have ben created when creating the values for PDB.

pkg/component/etcd/poddisruptionbudget/values_helper.go
```diff
@@ -0,0 +1,44 @@
...
+		EtcdName:      etcd.Name,
+		EtcdNameSpace: etcd.Namespace,
+		EtcdUID:       etcd.UID,
+		Labels:        etcd.Spec.Labels,
+		MinAvailable:  int32(CalculatePDBMinAvailable(etcd)),
+	}
+}
+
+// CalculatePDBMinAvailable calculates the minimum available value for the PDB
+func CalculatePDBMinAvailable(etcd *druidv1alpha1.Etcd) int {
```
unmarshall, line 35: There is no unit test.

pkg/utils/k8s_version.go
```diff
@@ -0,0 +1,85 @@
...
+	v := strings.Replace(version, "v", "", -1)
+	idx := strings.IndexAny(v, "-+")
+	if idx != -1 {
+		v = v[:idx]
+	}
+	return v
+}
+
+// CheckVersionMeetsConstraint returns true if the <version> meets the <constraint>.
+func CheckVersionMeetsConstraint(version, constraint string) (bool, error) {
```
unmarshall, line 54: No unit tests

pkg/utils/k8s_version.go
```diff
@@ -0,0 +1,85 @@
...
+	v, err := semver.NewVersion(normalizeVersion(version))
+	if err != nil {
+		return false, err
+	}
+
+	return c.Check(v), nil
+}
+
+// GetClusterK8sVersion returns the semver version of the cluster k8s version
+func GetClusterK8sVersion(config *rest.Config) (*semver.Version, error) {
```
unmarshall, line 69: No unit tests

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,190 @@
...
+
+	labels[instanceKey] = val.EtcdName
+	labels[nameKey] = "etcd"
+	labels[appKey] = "etcd-statefulset"
+
+	return labels
+}
+
+// getSelectorLabels returns the labels used in the selector of the PDB spec
+func getSelectorLabels(val *Values) map[string]string {
```
unmarshall, line 90: This also could be moved to values_helper.go

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,190 @@
...
+	labels := map[string]string{
+		instanceKey: val.EtcdName,
+		nameKey:     "etcd",
+	}
+
+	return labels
+}
+
+// getAnnotations returns the annotations for the PDB object
+func getAnnotations(val *Values) map[string]string {
```
unmarshall, line 100: This also could be moved to values_helper.go

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget_test.go
```diff
@@ -0,0 +1,362 @@
...
+					values = GenerateValues(etcd)
+					pdbDeployer = New(cl, namespace, &values, *k8sVersion_1_20_0)
+					Expect(pdbDeployer.Deploy(ctx)).To(Succeed())
+
+					pdb := &policyv1beta1.PodDisruptionBudget{}
+					Expect(cl.Get(ctx, kutil.Key(namespace, values.EtcdName), pdb)).To(Succeed())
+					Expect(pdb.Spec.MinAvailable.IntVal).To(BeNumerically("==", 0))
+					checkV1beta1PDB(pdb, &values)
+
+					Expect(cl.Delete(ctx, pdb)).To(Succeed())
```
unmarshall, line 151: You should ideally define an `AfterEach` and do it there.

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,60 @@
...
+		Context("Comparing between versions 1.20.0-dev and 1.25.0-dev", func() {
+			It("Should return true if the constraint is 1.20.0-dev < 1.25.0-dev", func() {
+				result, err := CompareVersions("1.20.0-dev", "<", "1.25.0-dev")
+				Expect(err).To(BeNil())
+				Expect(result).To(BeTrue())
+			})
+		})
+	})
+	Describe("#CheckVersionMeetsConstraint", func() {
+		Context("COmparing between versions 1.21.0 and 1.21.1", func() {
```
unmarshall, line 42:
```suggestion
		Context("Comparing between versions 1.21.0 and 1.21.1", func() {
```

unmarshall (commented):
As discuss there are some minor comments. Please make the required changes. Other than that it is /lgtm from my end.

unmarshall (approved):
/lgtm

pkg/component/etcd/poddisruptionbudget/valueshelper.go
```diff
@@ -0,0 +1,62 @@
...
+		Labels:         labels,
+		SelectorLabels: selectorLabels,
+		Annotations:    annotations,
+		MinAvailable:   int32(CalculatePDBMinAvailable(etcd)),
+	}
+}
+
+// CalculatePDBMinAvailable calculates the minimum available value for the PDB
+func CalculatePDBMinAvailable(etcd *druidv1alpha1.Etcd) int {
+	// do not enable in single mode
```
ishan16696, line 54:
```suggestion
	// do not enable for single node cluster
```

pkg/component/etcd/poddisruptionbudget/valueshelper.go
```diff
@@ -0,0 +1,62 @@
...
+	if etcd.Spec.Replicas < 2 {
+		return 0
+	}
+
+	clusterSize := int(etcd.Spec.Replicas)
+	return int(math.Floor(float64(clusterSize)/float64(2)) + 1)
```
ishan16696, line 60: Why you are converting it to float, taking a floor and converting back to int.
you can directly use this: `clusterSize/2 + 1` as it is already int and int will truncate the result which is equivalent to taking a floor.

pkg/component/etcd/poddisruptionbudget/valueshelper_test.go
```diff
@@ -0,0 +1,59 @@
...
+	})
+	Describe("#CalculatePDBMinAvailable", func() {
+		Context("With etcd replicas less than 2", func() {
+			It("Should return MinAvailable 0 if replicas is set to 1", func() {
+				etcd.Spec.Replicas = 1
+				Expect(CalculatePDBMinAvailable(etcd)).To(Equal(0))
+			})
+		})
+		Context("With etcd replicas more than 2", func() {
+			It("Should return MinAvailable 3 if replicas is set to 5", func() {
```
ishan16696, line 49:
```suggestion
			It("Should return MinAvailable as quorum size of cluster", func() {
```

pkg/component/etcd/poddisruptionbudget/valueshelper_test.go
```diff
@@ -0,0 +1,59 @@
...
+				etcd.Spec.Replicas = 5
+				Expect(CalculatePDBMinAvailable(etcd)).To(Equal(3))
+			})
+			It("Should return MinAvailable 4 if replicas is set to 6", func() {
```
ishan16696, line 53:
```suggestion
			It("Should return MinAvailable as quorum size of cluster", func() {
```

ishan16696 (changes requested):
few nits from my side.

pkg/component/etcd/poddisruptionbudget/valueshelper.go
aaronfern: Ah, thanks for pointing this out
This was a function that was already present in custodian_controller and I just moved it here, hence I didn't scrutinise it much
Updated

ishan16696 (approved):
LGTM!!

abdasgupta (approved):

PR #438, opened 2022-09-22
Support annotations for etcd client service

### LINKED ISSUE #435: [Feature] Support annotations for `etcd` service
Etcd-Druid creates a `client` Kubernetes Service for etcd clusters. We should allow users to specify annotations in the `etcd` resource which are added to the before mentioned `service`.

Some Kubernetes features, e.g. [Topology Aware Hints](https://kubernetes.io/docs/concepts/services-networking/topology-aware-hints/), need `service`s to have special annotations. These features can currently not be used as there are no means to manipulate the client `service`.

I propose to enhance the `etcd` API:

```
spec:
  etcd:
    clientService: # optional field
      annotations: # optional field
```

The [Service component](https://github.com/gardener/etcd-druid/tree/master/pkg/component/etcd/service) is supposed to consider this configuration then creating or patching the client `service`.

aaronfern:

/area control-plane
/kind api-change

Adds a field to the `etcd` CRD to allow users to specify annotations which are intended to be added to the client `service`.
These annotations are then intended to be consumed by certain kubernetes features (i.e. topology aware hints)

Fixes #435 

**Release note**:

```other operator
Added new field to the etcd CRD `etcd.Spec.Etcd.clientService`
Users can now add annotations to the client service by specifying them in the above field of the etcd resource
```


api/v1alpha1/etcd_types.go
```diff
@@ -230,6 +230,13 @@ type EtcdConfig struct {
 	// HeartbeatDuration defines the duration for members to send heartbeats. The default value is 10s.
 	// +optional
 	HeartbeatDuration *metav1.Duration `json:"heartbeatDuration,omitempty"`
+	// ClientService defines specified annotations that should be added to the client service
+	// +optional
+	ClientService ClientServiceSpec `json:"clientService"`
```
unmarshall, line 235: Shouldn't this be a pointer? ClientService should be made optional (as you have already indicated by `+ optional` comment . In additional you have already made annotations optional which is fine. If additional fields are added you can define optional/required for those fields individually.

config/crd/bases/10-crd-druid.gardener.cloud_etcds.yaml
```diff
@@ -284,6 +284,14 @@ spec:
               etcd:
                 description: EtcdConfig defines parameters associated etcd deployed
                 properties:
+                  clientService:
+                    description: ClientService defines specified annotations that should be added to the client service
```
unmarshall, line 288: The description specifies only annotations. If that is the case then why have another struct for that? If you feel that you can have additional client service config added then you will have to change the description.

pkg/component/etcd/service/values_helper.go
```diff
@@ -33,6 +33,7 @@ func GenerateValues(etcd *druidv1alpha1.Etcd) Values {
 		BackupPort:        pointer.Int32Deref(etcd.Spec.Backup.Port, defaultBackupPort),
 		ClientPort:        pointer.Int32Deref(etcd.Spec.Etcd.ClientPort, defaultClientPort),
 		ClientServiceName: utils.GetClientServiceName(etcd),
+		ClientAnnotations: etcd.Spec.Etcd.ClientService.Annotations,
```
unmarshall, line 36: service name is captured as `ClientServiceName`, shouldn't annotations be named as `ClientServiceAnnotations`?

aaronfern: Yes, that makes things more clear
Updated

config/crd/bases/10-crd-druid.gardener.cloud_etcds.yaml
aaronfern: Yes, you're right
Updated description

api/v1alpha1/etcd_types.go
```diff
@@ -230,6 +230,15 @@ type EtcdConfig struct {
 	// HeartbeatDuration defines the duration for members to send heartbeats. The default value is 10s.
 	// +optional
 	HeartbeatDuration *metav1.Duration `json:"heartbeatDuration,omitempty"`
+	// ClientService defines specified annotations that should be added to the client service
+	// +optional
+	ClientService *ClientServiceSpec `json:"clientService"`
```
timuthy, line 235:
```suggestion
	ClientService *ClientService `json:"clientService,omitempty"`
```

api/v1alpha1/etcd_types.go
```diff
@@ -230,6 +230,15 @@ type EtcdConfig struct {
...
+}
+
+type ClientServiceSpec struct {
```
timuthy, line 238:
```suggestion
type ClientService struct {
```

Please also add a doc string.

pkg/component/etcd/service/values_helper.go
```diff
@@ -29,14 +29,21 @@ const (
 
 // GenerateValues generates `service.Values` for the service component with the given `etcd` object.
 func GenerateValues(etcd *druidv1alpha1.Etcd) Values {
+	var clientServiceAnnotations map[string]string
+	if etcd.Spec.Etcd.ClientService != nil {
+		clientServiceAnnotations = etcd.Spec.Etcd.ClientService.Annotations
+	} else {
+		clientServiceAnnotations = nil
+	}
```
timuthy, line 37: No need for this `else` branch since the map is `nil` be default when you declared the var.

api/v1alpha1/etcd_types.go
```diff
@@ -230,6 +230,15 @@ type EtcdConfig struct {
 	// HeartbeatDuration defines the duration for members to send heartbeats. The default value is 10s.
 	// +optional
 	HeartbeatDuration *metav1.Duration `json:"heartbeatDuration,omitempty"`
+	// ClientService defines specified annotations that should be added to the client service
```
timuthy, line 233: Can you please check this doc string? I think it's not generic enough if we want to expand the `ClientService` struct in the future, i.e. don't mention annotations here.

timuthy (changes requested):
Thanks a lot for the PR. Some minor comments, but otherwise it looks good 👍

aaronfern:
Thanks for taking a look @timuthy 
I've made changes. PTAL

timuthy (approved):
/lgtm

PR #441, opened 2022-09-28
Remove BackupReady condition from PDB calculation

timuthy:

/area high-availability
/kind bug

Earlier failing backups resulted in PDBs not tolerating any disruptions.
This becomes esp. critical if such failures need to be fixed through rolling updates which are then blocked by the PDB configuration.

**Release note**:

```bugfix operator
A bug has been fixed that caused the wrong `minAvailable` configuration being calculated for multi-node etcd `PodDisruptionBudget`.
```
```other operator
The `BackupReady` condition is not considered anymore when the `PodDisruptionBudget` configuration is calculated. This earlier blocked rolling out fixes that potentially solved problems with backup procedures.
```

### DIFF
--- a/controllers/etcd_custodian_controller.go
+++ b/controllers/etcd_custodian_controller.go
@@ -193,18 +193,11 @@ func calculatePDBminAvailable(etcd *druidv1alpha1.Etcd) int {
 	}
 
 	allMembersReady := false
-	backupReady := false
 	for _, condition := range etcd.Status.Conditions {
 		if condition.Type == druidv1alpha1.ConditionTypeAllMembersReady &&
 			condition.Status == druidv1alpha1.ConditionTrue {
 			allMembersReady = true
-			continue
-		}
-
-		if condition.Type == druidv1alpha1.ConditionTypeBackupReady &&
-			condition.Status == druidv1alpha1.ConditionTrue {
-			backupReady = true
-			continue
+			break
 		}
 	}
 
@@ -220,10 +213,6 @@ func calculatePDBminAvailable(etcd *druidv1alpha1.Etcd) int {
 			}
 		}
 		values = append(values, readyMembers)
-	}
-
-	if backupReady {
-		values = append(values, clusterSize)
 	}
 
 	// calculate max value
--- a/controllers/etcd_custodian_controller_test.go
+++ b/controllers/etcd_custodian_controller_test.go
@@ -70,27 +70,12 @@ var _ = Describe("Custodian Controller", func() {
 
 			etcd.Status.Conditions = []druidv1alpha1.Condition{
 				{Type: druidv1alpha1.ConditionTypeAllMembersReady, Status: druidv1alpha1.ConditionFalse},
-				{Type: druidv1alpha1.ConditionTypeBackupReady, Status: druidv1alpha1.ConditionFalse},
 			}
 
 			It("should use all ready members if not all members are ready", func() {
 				Expect(calculatePDBminAvailable(etcd)).To(BeEquivalentTo(4))
 			})
 		})
-
-		When("having a multi node cluster", func() {
-			etcd := getEtcdWithStatus(5)
-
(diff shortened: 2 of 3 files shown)


aaronfern (approved):
/lgtm

ishan16696:
Hi @timuthy , thanks for the PR. One doubt:
For `etcd-events` when there is no storage-provider configured hence, no backup is taken then how PDB will behave ?
If PDB set the min-available to cluster-size in case of etcd-events then it is issue right ?

timuthy:
Nice catch @ishan16696 👍 I amended this in a354296b4105d86f62f4da1573c99abe8c777193.

controllers/etcd_custodian_controller.go
```diff
@@ -220,10 +220,10 @@ func calculatePDBminAvailable(etcd *druidv1alpha1.Etcd) int {
 			}
 		}
 		values = append(values, readyMembers)
-	}
-
-	if backupReady {
-		values = append(values, clusterSize)
+	} else {
```
unmarshall, line 229: The two conditions `!backReady` and `!allMembersReady` are disjoint and there is no reason to use it in if..else block.
Just have:
```go
if !backupReady {
	values = append(values, clusterSize)
}
if !allMembersReady {
	readyMembers := 0
	for _, member := range etcd.Status.Members {
		if member.Status == druidv1alpha1.EtcdMemberStatusReady {
			readyMembers += 1
		}
	}
	values = append(values, readyMembers)
}
```
In the code next you anyways take the max out of all the entries in the `value` slice, right?

unmarshall: If backReady = false and allMembersReady = false then in the proposed change you will only enter the `if` block and not the `else` block. If the backup is not ready then by setting `minAvailable` to `clusterSize` you essentially wish to prevent update of any member in the cluster. This would not happen then.

timuthy: I don't consider both conditions being disjoint in that case: It doesn't make sense to use the entire cluster size as `minAvailable` if both not all members and backups aren't ready, i.e. we should rather use the lower number in such a case to allow disruptions in case we want to ship a fix.

I even thought about disregarding the backup condition and remove it from the calculation because there is no benefit but rather disadvantages: If backups aren't ready because of a software issue, then the PDB would forbid rolling out a fix to the cluster. WDYT?

unmarshall: >we should rather use the lower number in such a case to allow disruptions in case we want to ship a fix.

If that is the case then there could be a bug due to which `backReady` will be false. Assume all members are ready and backReady is false in that case you will still set the minAvailable to cluster size thus preventing any fix to be applied which might correct the issue with taking backups. Isn't it?

unmarshall: I am fine with leaving out backupReady condition check altogether. Also we should force a re-election via the backup-restore sidecar container.

timuthy: > Today that is not done. Setting the minAvailable to clusterSize in that case will result in a situation where no backup is taken and you cannot even update the etcd cluster with a fix.

That's exactly the reason I think considering the backup condition isn't beneficial.

timuthy:
As discussed with @unmarshall, I changed the PR so that the `BackupReady` condition is not considered anymore when calculating the PDBs.

controllers/etcd_custodian_controller.go
unmarshall: I think it is just a safety net but without re-election this safety net is quite useless. The only case i can think of is if there is only a transient error in taking backups then retries from the leading backup-restore sidecar will happen. While this is happening no updates should be allowed to give the leading sidecar a slim chance to salvage data before updates roll in. Removing this would also remove this safety net :)

ishan16696 (approved):

unmarshall (approved):
/lgtm

PR #448, opened 2022-10-14
Harmonize kustomizations

afritzler:

/area dev-productivity
/kind enhancement

This PR add a missing `ServiceAccount` in the kustomization `/config` folder. Additionally the namespaces have been harmonised to use the same default namespace `etcd-druid-system`.

You can evaluate the output by running:

```shell
kustomize build config/default
```

**Release note**:

```other operator

```

### DIFF
--- a/config/rbac/service_account.yaml
+++ b/config/rbac/service_account.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: controller-manager
+  namespace: garden
--- a/config/rbac/service_account.yaml
+++ b/config/rbac/service_account.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: controller-manager
-  namespace: garden
+  namespace: system
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
-namespace: garden
+namespace: etcd-druid-system
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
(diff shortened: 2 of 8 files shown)


config/rbac/service_account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: controller-manager
```
aaronfern, line 4: Although the name here is `controller-manager`, when created I see a SA with the name `etcd-druid-controller-manager`
Is this expected?

aaronfern (commented):
Thanks for the PR @afritzler!

Just one question as I'm not too familiar with kustomize
Also, can you please add a release note?

config/rbac/service_account.yaml
afritzler: This is expected as all resources will be prefixed with `etcd-druid-` as defined in the `config/default` kustomization. In general it is a good practise to use generic names such as `controller-manager` in the sub-kustomizations and prefix is later. E.g. a user might have his own default and overwrite the `namePrefix: abcd-` and hence will end up with a `abcd-controller-manager SA.

afritzler:
One other thing which came to my mind is that we might want to change the controller `namespace` from `garden` to something more specific like `etcd-druid-system`.

config/rbac/service_account.yaml
aaronfern: Ah, that clarifies things
Thank you

aaronfern:
That's a good idea 
Can it be done as part of this PR, or do you suppose it's best to be taken up in some later PR?

afritzler:
Sorry for the delay. I just updated the PR branch with the correct namespace naming.

aaronfern:
Thanks for the changes! 
Looks good! I'll just test it out a bit before merging

aaronfern (approved):
/lgtm

unmarshall (approved):
/lgtm

PR #455, opened 2022-10-28
Update condition to add ScaleToMultiNode annotation

### LINKED ISSUE #452: [BUG] Permanent Quorum Loss Recovery Failure
**Describe the bug:**
While working through permanent quorum loss runbook, I noticed that the recovery often fails (at least for me). In most cases, one pod cannot join the cluster and "forms a new cluster", indicated by log messages such as `cluster ID mismatch: peer[dce2d98511a55225]=cdf818194e3a8c32, local=8879bbfa20c8a9c3` and the line `initial-cluster-state: new` in the ETCD config of the broken pod/etcd container (as noticed by @aaronfern in a call).

**How to reproduce:**
Basically, follow the playbook mentioned above. In my case, I ran the ETCD in a dedicated namespace (`shoot` namespace) in a dedicated shoot cluster in which I also run the druid (`druid` namespace): See cluster `staging/core/ha` for its configuration.

**Logs:**
[etcd-main-permanent-quorom-loss-recovery-issue.zip](https://github.com/gardener/etcd-druid/files/9869263/etcd-main-permanent-quorom-loss-recovery-issue.zip) (etcd-main-1 caused the trouble here)

**Environment:**
 - Etcd-druid version/commit ID: `eu.gcr.io/sap-se-gcr-k8s-public/eu_gcr_io/gardener-project/gardener/etcd-druid@sha256:8938ca5a4e7eed0b93ea8eae5844b0b681e9802a8e08293278c9674905a43c7d`
 - Cloud Provider: See shoot spec above

**Preliminary analysis/root cause:**
It seems that the annotation `gardener.cloud/scaled-to-multi-node="true"` (step 9 in the playbook mentioned above) gets removed by the druid too eagerly/early (observed by @aaronfern in a call). This seems to be a race condition, which explains why my etcd cluster is not always recovering or always not recovering. It may depend on how long the pod takes to start and should probably only be removed after all pods report `ready` instead (or maybe it's a mess-up because of step 8, i.e. steps 8 and 9 should probably be swapped in any case).

aaronfern:

/area robustness
/kind bug

This PR updates the condition that adds the `gardener.cloud/scaled-to-multi-node` annotation used to scale up the etcd cluster.
Previously the condition depended on `etcd.Status.Replicas` to make a decision of whether to add the annotation or not, which in incorrect in terms of quorum loss recovery as `etcd.Status.Replicas` is a mirror of `etcd.Spec.Replicas` which was `n` before the cluster crashed.
The change now is that we now use `sts.Spec.Replicas` to decide whether to add the annotation or not. This is more accurate as it now considers the actual number of pods deployed to decide whether it is a scale-up case or not

Fixes #452 

**Release note**:

```bugfix operator
Updated condition used to decide when the `gardener.cloud/scaled-to-multi-node` annotation is added
```

### DIFF
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -282,8 +282,11 @@ func immutableFieldUpdate(sts *appsv1.StatefulSet, val Values) bool {
 	return sts.Spec.ServiceName != val.PeerServiceName || sts.Spec.PodManagementPolicy != appsv1.ParallelPodManagement
 }
 
-func clusterScaledUpToMultiNode(val *Values) bool {
-	return val.Replicas > 1 && val.StatusReplicas == 1
+func clusterScaledUpToMultiNode(val *Values, sts *appsv1.StatefulSet) bool {
+	if sts.Spec.Replicas != nil {
+		return val.Replicas > 1 && *sts.Spec.Replicas == 1
+	}
+	return false
 }
 
 func (c *component) createOrPatch(ctx context.Context, sts *appsv1.StatefulSet, replicas int32) error {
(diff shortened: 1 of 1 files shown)


ishan16696 (commented):
Can we add a test case for this ?

ishan16696 (approved):
LGTM!! 
@aaronfern may add a test case in a separate PR.

PR #462, opened 2022-11-11
Fix make manifest target

aaronfern:

/area usability
/kind technical-debt

There is an issue where `config/rbac/role.yaml` gets incorrectly overwritten when `make manifests` is run. This is due to incorrect/incomplete kubebuilder RBACs 
This PR adds and corrects RBACs so that `make manifests` now generates the correct role file

Fixes #

**Release note**:

```other operator

```


ialidzhikov:
Please also backport to the release branch for v0.15 so people can easily find the version of the CRDs they need.

shreyas-s-rao (approved):
@aaronfern Thanks for fixing this.

/lgtm

PR #464, opened 2022-11-14
Minor cleanups

shreyas-s-rao:

/area testing
/kind cleanup

Minor cleanups

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/api/v1alpha1/etcd_types_test.go
+++ b/api/v1alpha1/etcd_types_test.go
@@ -91,12 +91,12 @@ func getEtcd(name, namespace string) *Etcd {
 	deltaSnapshotPeriod := metav1.Duration{
 		Duration: 300 * time.Second,
 	}
-	imageEtcd := "eu.gcr.io/gardener-project/gardener/etcd:v3.4.13-bootstrap"
-	imageBR := "eu.gcr.io/gardener-project/gardener/etcdbrctl:v0.12.0"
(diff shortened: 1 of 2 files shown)


aaronfern (approved):
/lgtm

api/v1alpha1/etcd_types_test.go
```diff
@@ -91,12 +91,12 @@ func getEtcd(name, namespace string) *Etcd {
 	deltaSnapshotPeriod := metav1.Duration{
 		Duration: 300 * time.Second,
 	}
-	imageEtcd := "eu.gcr.io/gardener-project/gardener/etcd:v3.4.13-bootstrap"
-	imageBR := "eu.gcr.io/gardener-project/gardener/etcdbrctl:v0.12.0"
+	imageEtcd := "eu.gcr.io/gardener-project/gardener/etcd:v3.4.13-bootstrap-8"
+	imageBR := "eu.gcr.io/gardener-project/gardener/etcdbrctl:v0.20.2"
```
ishan16696, line 95: Do we have to update this after every release backup-restore ?

ishan16696: can we make this automated like it can read the image tag from version may be ?

shreyas-s-rao: I think this will then be part of `make generate`. And since this is just an API test, the values don't even matter. We can use dummy values here if necessary, like the value for `container` field.

ishan16696 (approved):

PR #466, opened 2022-11-16
Updated condition that adds the scale-up annotation to the sts

aaronfern:

/area disaster-recovery
/area high-availability
/kind bug

`etcd-druid` contains a `scale-up` feature where a single member etcd can be scaled up to a multi member etcd cluster.

Ideally the `statefulset` has to be patched to update (increase) the number of replicas so that new members are added to the cluster. In addition, etcd cluster scale-up should also work when the `statefulset` is recreated rather than patched.

There was a bug where the etcd cluster would not scale up correctly if the `statefulset` was recreated and would lead to a broken cluster
This PR fixes a bug where the cluster would not scale up correctly for recreated `statefulsets`

Fixes #

**Release note**:

```other operator
The decision to add the scale-up annotation to the etcd sts now considers the etcd status if an existing sts is not present
```


aaronfern:
needs/cherry-pick

ishan16696 (approved):

PR #467, opened 2022-11-16
Make `GetConfigmapName` consistent

rfranzke:

/area usability
/kind enhancement

All other resources are prefixed with the `etcd.Name` but the `ConfigMap` is not. This PR fixes this.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/utils/names.go
+++ b/pkg/utils/names.go
@@ -37,7 +37,7 @@ func GetServiceAccountName(etcd *druidv1alpha1.Etcd) string {
 
 // GetConfigmapName returns the name of the configmap based on the given `etcd` object.
 func GetConfigmapName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("etcd-bootstrap-%s", string(etcd.UID[:6]))
+	return fmt.Sprintf("%s-bootstrap-%s", etcd.Name, string(etcd.UID[:6]))
 }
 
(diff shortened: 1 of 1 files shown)


shreyas-s-rao:
@rfranzke Thanks a lot for the PR. Can you please also adapt the tests to use "`etcdName`-bootstrap-xxxxxx" names for the configmaps? Also, does this work well for new druid reconciling older `Etcd` resources with the old configmap name?
And can you please also make related changes in the [quorum loss doc](https://github.com/gardener/etcd-druid/blob/master/docs/operation/Recover_From_Etcd_Permanent_Quorum_Loss.md)? Thanks!

pkg/utils/names.go
```diff
@@ -37,7 +37,7 @@ func GetServiceAccountName(etcd *druidv1alpha1.Etcd) string {
 
 // GetConfigmapName returns the name of the configmap based on the given `etcd` object.
 func GetConfigmapName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("etcd-bootstrap-%s", string(etcd.UID[:6]))
+	return fmt.Sprintf("%s-bootstrap-%s", etcd.Name, string(etcd.UID[:6]))
```
timuthy, line 40: I think we also need some cleanup logic.

unmarshall:
@rfranzke we will take this up later and make the name consistent. This also helps in looking at config-maps.

PR #469, opened 2022-11-18
Fix backup ready health check

timuthy:

/area ops-productivity
/kind bug

This PR fixes an issue for the backup ready health check, when the full snapshot is already taken but delta is missing.

**Release note**:

```bugfix operator
An issue has been fixed that caused the `BackupReady` condition to show `Unknown` when the cluster is newly created.
```


shreyas-s-rao (approved):
@timuthy thanks a lot for fixing this!
/lgtm

PR #474, opened 2022-11-27
Update golangci-lint version and bump golang to `v1.19`

shreyas-s-rao:

/area quality
/kind bug

Fix failing tests due to golang linting issue for go `v1.18`. Also bumps golang version to `v1.19`.

Fixes #

/cc @abdasgupta @timuthy 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -24,11 +24,11 @@ etcd-druid:
                 build: ~
     steps:
       check:
-        image: 'golang:1.19.2'
+        image: 'golang:1.19.3'
       test:
-        image: 'golang:1.19.2'
+        image: 'golang:1.19.3'
       build:
-        image: 'golang:1.19.2'
+        image: 'golang:1.19.3'
         output_dir: 'binary'
 
   jobs:
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/gardener/etcd-druid
 
-go 1.18
+go 1.19
 
 require (
 	github.com/Masterminds/semver v1.5.0
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,9 +1,7 @@
 run:
   concurrency: 4
   deadline: 10m
-  # some of the linters don't work correctly with 1.18, ref https://github.com/golangci/golangci-lint/issues/2649
-  # we are not using generics, so let's pin this to 1.17 until 1.18 is fully supported
-  go: "1.17"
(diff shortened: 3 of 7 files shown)


.ci/pipeline_definitions
```diff
@@ -24,11 +24,11 @@ etcd-druid:
                 build: ~
     steps:
       check:
-        image: 'golang:1.19.2'
+        image: 'golang:1.19.3'
```
ishan16696, line 27: can we stick to go `v1.19.2` as backup-restore `v0.21.0` is also using go `v1.19.2`.
IMO, can we have a same go version in both druid and backup-restore, so that it is easy to track

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/gardener/etcd-druid
 
-go 1.18
+go 1.19
```
ishan16696, line 3: Do we want to change go.mod as well ? I think this will force dev to move to go 1.19 as well right ?

ishan16696 (changes requested):

go.mod
unmarshall: I would recommend keeping the golang version same across CI build configuration and go.mod. There is no reason not to upgrade the dev environment to go 1.19

.ci/pipeline_definitions
shreyas-s-rao: It's not a necessity to keep them on the same version. Also due to vulnerabilities like https://nvd.nist.gov/vuln/detail/CVE-2022-41716, which are fixed in go `v1.19.3`, it makes sense to move to this version for etcd-backup-restore as well TBH.

go.mod
shreyas-s-rao: go `v1.19` brings many fixes, especially in this PR's case, for the golangci-linter. Again, as mentioned above, it also makes sense to move to go `v1.19` for security reasons.

.ci/pipeline_definitions
ishan16696: but we have already released etcd-backup-restore `v0.21` with go `v1.19.2`

ishan16696 (approved):
thanks for PR @shreyas-s-rao. LGTM!!

.ci/pipeline_definitions
shreyas-s-rao: Patch version shouldn't matter too much actually. Keeping the golang build version to the latest stable version is usually recommended.

PR #480, opened 2022-12-07
Run `make manifests`

ialidzhikov:

/area open-source
/kind bug

While looking for a CRD version that includes the clientService annotations from https://github.com/gardener/etcd-druid/pull/438, I noticed that `charts/druid/charts/crds/templates/10-crd-druid.gardener.cloud_etcds.yaml` does not include at all the corresponding fields.
Then I ran `make manifests` and it generated me the content from this PR.

**Release note**:

```other operator
The `etcds.druid.gardener.cloud` CRD does now include the `spec.etcd.clientService` field.
```

### DIFF
--- a/charts/druid/charts/crds/templates/10-crd-druid.gardener.cloud_etcds.yaml
+++ b/charts/druid/charts/crds/templates/10-crd-druid.gardener.cloud_etcds.yaml
@@ -300,6 +300,17 @@ spec:
                   clientPort:
                     format: int32
                     type: integer
+                  clientService:
+                    description: ClientService defines the parameters of the client
+                      service that a user can specify
+                    properties:
+                      annotations:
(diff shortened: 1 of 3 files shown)


ialidzhikov:
cc @aaronfern @shreyas-s-rao

shreyas-s-rao:
I'm unsure why `config/rbac/role.yaml` is getting updated. I see important permissions on resources like `job`s getting removed, and I also see this which looks strange to me:
```
- apiGroups:
  - druid.gardener.cloud
  resources:
  - secrets
  verbs:
  - ......
```

aaronfern:
I had noticed this inconsistency and have a PR already raised to solve this https://github.com/gardener/etcd-druid/pull/462
Can we close this PR in favour of that one?

PR #511, opened 2023-01-16
Enabled e2e tests to read KUBECONFIG from in-cluster variables along with environment variable

### LINKED ISSUE #465: [Feature] Kind-based druid e2e tests and pipeline integration
As a developer, I would like be able to easily run the e2e tests for etcd-druid without the need for an external cluster. Ideally, I would like to use a local [kind cluster](https://kind.sigs.k8s.io/docs/user/quick-start/) to do this, similar to [how gardener/gardener does it](https://github.com/gardener/gardener/blob/master/docs/development/testing.md#end-to-end-e2e-tests-using-provider-local).

I would also like to run these e2e tests as part of the CICD pipeline before every PR is merged, to ensure that we aren't required to manually run the e2e tests before every PR/release.  

Kind-based testing is quite convenient, as already demonstrated by gardener. It allows developers to easily test locally, and also to integrate with the CICD pipeline, which can spin up a kind cluster and run tests against it, rather than trying to procure a k8s cluster from elsewhere.

/area quality
/area testing

abdasgupta:

/area TODO
/kind TODO

Currently, `make test-e2e` needs KUBECONFIG to be provided through environment variable. But the command can not construct KUBECONFIG from the in-cluster pod environment variables. But to enable running Druid e2e tests through Prow jobs, we need to run the e2e tests inside a pod (as a job) and the tests need to figure out kubeconfig from in-cluster variables. So, this is PR is raised which enables to run e2e tests with kubeconfigs provided by either environment variables or through in-cluster variables. 
Fixes #465 

**Release note**:

```other developer
KUBECONFIG variable is set by environment flag instead of options during the execution of command `make test-e2e`

```

### DIFF
--- a/test/e2e/utils.go
+++ b/test/e2e/utils.go
@@ -40,11 +40,11 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/remotecommand"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


abdasgupta:
cc @seshachalam-yv

ishan16696:
Ok, I will test this PR but how this PR is fixing #429 as mentioned in the PR description ?

abdasgupta:
> Ok, I will test this PR but how this PR is fixing #429 as mentioned in the PR description ?

Sorry! My bad. I updated with the correct issue

ishan16696:
> I updated with the correct issue

 it is part of #465 but won't fixes the #465 right ?

test/e2e/utils.go
```diff
@@ -387,26 +387,27 @@ func getProviders() ([]TestProvider, error) {
...
 	if err != nil {
 		return nil, err
 	}
-
-	return client.New(config, client.Options{})
+	cl, err = client.New(restConfig, client.Options{})
+	if err != nil {
+		return nil, err
+	}
+	return cl, nil
```
shreyas-s-rao, line 410: I don't see any need for making changes to this code. The previous implementation of `return client.New(config, client.Options{})` was absolutely fine and neat to read as well. The new code introduced here simply does the exact same thing that the previous code did, but with more lines of unnecessary code. Please revert this change.

shreyas-s-rao (changes requested):
@abdasgupta thanks for your PR. Please add the `area` and `kind` labels to this PR, in line with best practices for this repo. Please also update the PR note to `Fixes partially #465`, because this PR does not fully fix issue #465. Let's close that issue only after the story is complete.

Now coming to the PR itself, you mention that `this is PR is raised which enables to run e2e tests with kubeconfigs provided by either environment variables or through in-cluster variables`, but this PR completely removes support for specifying KUBECONFIG env var for running tests locally against any arbitrary k8s cluster, which is still necessary for local testing. Please enhance the PR to allow both methods of getting restConfig for the k8s cluster. Hint: if KUBECONFIG env var is not present or is empty, then default to trying using in-cluster kubeconfig.

abdasgupta:
This PR does not fully resolve #465. Though it enabled reading KUBECONFIG from in-cluster as well as environment variables but it is not a must have requirement for now. I am closing this PR in favour of https://github.com/gardener/etcd-druid/pull/513 .. Will reopen the issue if reading kubeconfig from in-cluster variables are necessary in future
/close

PR #521, opened 2023-02-06
Removed unnecessary lebel selectors from the services created by Druid.

### LINKED ISSUE #534: [BUG] gardener upgrade tests are failing because of ETCD resources
**Describe the bug:**
Gardener upgrade was failing with ETCD pods being stuck as unhealthy in multinode cluster. Currently, we are defining selectors in ETCD statefulset and services same as ETCD labels. ETCD labels contain many labels which are related to other components of Gardener, like networking. If there are changes in those labels as part of gardener upgrade, those changes conflict with the already existing selectors in ETCD statefulset and service. So, it is decided that we keep the label selector of services and statefulsets created by druid as minimalistic as possible. Currently, we will use only `instance: <ETCD Name>` as selector in statefulsets and services. 

**Expected behavior:**
Gardener upgrade will be passed.

 - Etcd version/commit ID :

abdasgupta:

/area control-plane
/kind bug

Gardener upgrade was failing with ETCD pods being stuck as unhealthy in multinode cluster. The reason behind this is explained here https://sap-ti.slack.com/archives/C0177NLL8V9/p1675341746522519 . As per the discussion, it is decided that we keep the label selector of services created by druid as minimalistic as possible. I trimmed the unrelated label selector from the services in this PR

Fixes #534 

**Release note**:

```bugfix developer
ETCD Statefulset and Services only use `instance:<ETCD Name>` as selector field.
```


pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -309,12 +309,12 @@ func (c *component) createOrPatch(ctx context.Context, sts *appsv1.StatefulSet,
 		Replicas:    &replicas,
 		ServiceName: c.values.PeerServiceName,
 		Selector: &metav1.LabelSelector{
-			MatchLabels: getCommonLabels(&c.values),
+			MatchLabels: getSelectors(&c.values),
```
timuthy, line 312: I think this change would require a recreation of the entire `StatefulSet`

pkg/component/etcd/service/service.go
```diff
@@ -98,6 +98,18 @@ func getOwnerReferences(val Values) []metav1.OwnerReference {
...
 }
 
+func getSelectors(val Values) map[string]string {
+	selectors := map[string]string{
+		"instance": val.EtcdName,
+	}
+
+	for k, v := range val.Selectors {
+		selectors[k] = v
+	}
```
timuthy, line 108: I wonder if it's really beneficial to have additional selectors defined by the user while we already have a well known and unique selector with `"instance": val.EtcdName`.

timuthy (changes requested):
Thanks for PR!

While the purpose of this change is clear, I think it also entails some problems and confusion:
1.) `MatchLabels` are merged into `Labels` which is unexpected FMPOV.
2.) Using `etcd.Spec.Etcd.Selector` makes this change incompatible. Today the field is not respected at all and after merging the PR it can cause an outage of existing clusters since the selector in the `service` changes + the `statefulset` needs to be re-created (see inline comment).
3.) I wonder about the use-case and benefit of having `etcd.Spec.Etcd.Selector` at all. Earlier it was used to adopt existing `statefulset`s but now we want to semantically use it for a different purpose. Is there really a use case for users to define `etcd.Spec.Etcd.Selector`? (cc @shreyas-s-rao @unmarshall )

pkg/component/etcd/service/values_helper.go
```diff
@@ -47,6 +47,7 @@ func GenerateValues(etcd *druidv1alpha1.Etcd) Values {
 		EtcdName:                 etcd.Name,
 		EtcdUID:                  etcd.UID,
 		Labels:                   etcd.Spec.Labels,
```
unmarshall, line 49: do we need networking.gardener.cloud/* labels to be added to the service?

unmarshall:
> I wonder about the use-case and benefit of having etcd.Spec.Etcd.Selector at all. Earlier it was used to adopt existing statefulsets but now we want to semantically use it for a different purpose. Is there really a use case for users to define etcd.Spec.Etcd.Selector? (cc @shreyas-s-rao @unmarshall )

I agree that there is no need for users to define selectors for service. Following default set of labels should be set on every statefulset pod:
```yaml
labels:
  app: etcd-statefulset
  instance: <pod-name-prefix>
  gardener.cloud/role: controlplane
```
A webhook can be used to enforce that the default set of labels are not tampered with either be reverting any change done to the default set of labels or failing validation.
The same default set of labels can be used to define a default `selector` as well.

**There are some other observations as well:**

If you look at the etcd CRD that we have today. It additionally has the following:

- `etcd.clientService.annotations`, `etcd.clientService.labels`. We do not use these and instead in the component for the service we use `spec.etcd.labels` instead.  
- We do not have spec field for `peerService`. 
- `etcd.spec.selector` and `etcd.spec.labels` - do not tell to which resource are they applied. Etcd controller creates a lot of resources. There is also no documentation in the CRD for `etcd.spec.labels`. One could assume that this is only meant for the pods that are started, representing each etcd member OR specify it explicitly via `etcd.pod.labels`.

I still see a need to provide an ability to set labels. For instance all `networking.gardener.cloud` labels can be specified in the etcd spec.

pkg/component/etcd/configmap/values_helper.go
```diff
@@ -43,7 +43,7 @@ func GenerateValues(etcd *druidv1alpha1.Etcd) *Values {
 		AutoCompactionMode:      etcd.Spec.Common.AutoCompactionMode,
 		AutoCompactionRetention: etcd.Spec.Common.AutoCompactionRetention,
 		ConfigMapName:           utils.GetConfigmapName(etcd),
-		Labels:                  etcd.Spec.Labels,
+		Labels:                  etcd.Spec.Selector.MatchLabels,
```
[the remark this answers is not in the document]
unmarshall: `etcd.spec.labels` is meant for labels. Currently it is loosely defined and can be used for any resource that is created by the etcd controller. Ideally we should not be using `etcd.Spec.Selector.MatchLabels` as labels for configmap.

pkg/component/etcd/service/service.go
unmarshall: I agree it is not required.

pkg/component/etcd/service/values_helper.go
```diff
@@ -47,6 +47,7 @@ func GenerateValues(etcd *druidv1alpha1.Etcd) Values {
...
+		Selectors:                etcd.Spec.Selector.MatchLabels,
```
unmarshall, line 50: etcd CRD has `etcd.clientService.labels`? This section of the etcd CRD is not used today. Any reason why? Also is there a need for users to set additional labels for Services (client and peer)?

pkg/component/etcd/statefulset/statefulset.go
unmarshall: Yes, in the statefulset.go you should explicitly handle this case. If the selector has changed which will result in a new member not being able to be identified by the peer service then an STS recreation is required.

pkg/component/etcd/service/service_client.go
```diff
@@ -25,12 +25,12 @@ import (
 
 func (c *component) syncClientService(ctx context.Context, svc *corev1.Service) error {
 	_, err := controllerutils.GetAndCreateOrStrategicMergePatch(ctx, c.client, svc, func() error {
-		svc.Labels = utils.MergeStringMaps(getLabels(c.values), c.values.ClientServiceLabels)
+		svc.Labels = utils.MergeStringMaps(getLabels(c.values), c.values.ClientServiceLabels, getSelectors(c.values))
```
unmarshall, line 28: This seems unnecessary to me. We should just have a default set of labels for a client service and then add additional labels that are specified in the etcd CRD via `spec.clientService.Labels`

pkg/component/etcd/statefulset/statefulset.go
abdasgupta: Yes, recreation of statefulset is required. But without this future gardener upgrade (which handles label changes) won't be possible.

pkg/component/etcd/service/service.go
abdasgupta: In my opinion, we can keep it. If the users don't want to put any other unique selector then they may choose to not mention any in etcd.spec.selector, right?

pkg/component/etcd/service/service_client.go
abdasgupta: I did this for uniformity between service and statefulsets. I think, if it's not harming, we can keep it

unmarshall: network* specific labels are not really needed on a service. Having labels which were never meant for a resource is quite misleading. This was also a comment in the slack channel and i agree with it.

pkg/component/etcd/service/service.go
unmarshall: From a gardener perspective it is not required. From an open source perspective if you wish to open it up (and not assume any default labels and selectors) then you need to also handle the following:

1. A webook to validate if the labels and selectors are proper, else the pods will never be discovered by the respective services.
2. Ensure that if lable-selector has been changed in a manner that requires a recreation of STS then it needs to be handled in component/statefulset

pkg/component/etcd/statefulset/statefulset.go
unmarshall: Not necessarily, if there are default set of labels which are added to STS and the selector also has the same default set of labels then addition of labels only goes to STS and does not affect the discoverability of pods from the service.

pkg/component/etcd/service/service.go
timuthy: > If the users don't want to put any other unique selector then they may choose to not mention any in etcd.spec.selector, right?

Can you elaborate on this? I'd like to understand the use case of defining additional selectors via the `etcd` resource.

pkg/component/etcd/statefulset/values.go
```diff
@@ -42,6 +42,8 @@ type Values struct {
 	Annotations map[string]string
 	// Labels is the labels provided in ETCD spec.
 	Labels map[string]string
+	// Selectors are the service selectors
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/component/etcd/configmap/values_helper.go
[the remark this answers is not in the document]
abdasgupta: Rectified

pkg/component/etcd/service/values_helper.go
abdasgupta: I don't think that this PR is the place to address that. We need more context to rule out that label

pkg/component/etcd/service/service.go
abdasgupta: Currently, it is not required. I was talking mainly from open source point of view. I was thinking, let the code be in place, if somebody require to add something in future, they can do it at their choice.

timuthy: As written in chat:

There are however many downsides:
- Re-creation of STS is necessary
- Incompatible change
- Misusage of API field (earlier it was used to adopt STS … now it’s used for service selector)

In my eyes, we should only follow the approach in this PR if there is a valid and beneficial use-case for the user. So far I don't see a justification to accept the downsides listed above.

pkg/component/etcd/service/values_helper.go
unmarshall: Then let us define the scope of the issue that this PR addresses. Since there is no hurry to get this PR out we can use this as an opportunity to clean up technical debt and streamline the labels that are added to all components that are created by druid.

pkg/component/etcd/statefulset/statefulset.go
abdasgupta: In the current change in my PR, the selector is only `instance: ETCDName`. So, no change in ETCD selector field will recreate sts. However, currently the selector fields contains "networking" labels. So, when this PR is merged and the new druid release goes in, recreation of STS is required to reflect the changes in existing STS

pkg/component/etcd/service/service.go
abdasgupta: Made the changes

pkg/component/etcd/service/service_client.go
abdasgupta: Can it be part of the refactoring work that you are doing? Let it go through for now. Let the gardener upgrade tests pass and then we can discuss what can be the common labels for services

pkg/component/etcd/statefulset/statefulset.go
timuthy: Here again, why do we need to change this now? There has never been a problem reported by using the selector in the STS hat are used today. The PR description also says that it fixes the service selector, nothing else 🙁
Please revert this change.

pkg/component/etcd/service/service.go
timuthy: I don't see the changes being made. The service selector still considers `etcd.Spec.Selector.MatchLabels`.

abdasgupta:
I reverted back the changes for selectors introduced to statefulset @timuthy @unmarshall

timuthy (commented):
/lgtm

timuthy:
@unmarshall feel free to approve and merge if you are fine with the changes.

unmarshall (approved):

unmarshall:
@timuthy you will have to approve as well as its not allowing me to merge (requested change from you). I have re-requested a review from you. Maybe you can just approve it again.

pkg/component/etcd/service/service_peer.go
```diff
@@ -17,19 +17,20 @@ package service
...
+	"github.com/gardener/etcd-druid/pkg/utils"
 	"github.com/gardener/gardener/pkg/controllerutils"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 func (c *component) syncPeerService(ctx context.Context, svc *corev1.Service) error {
 	_, err := controllerutils.GetAndCreateOrStrategicMergePatch(ctx, c.client, svc, func() error {
-		svc.Labels = getLabels(c.values)
+		svc.Labels = utils.MergeStringMaps(getLabels(c.values), getSelectors(c.values))
```
aaronfern, line 28: Since `"instance": val.EtcdName` is now present as part of `getSelectors()`, it can be removed from `getLabels()` [here](https://github.com/gardener/etcd-druid/blob/master/pkg/component/etcd/service/service.go#L101-L111)

aaronfern (changes requested):
Looks good, just one change

aaronfern (approved):

PR #517, opened 2023-02-01
Cleanup CI scripts

shreyas-s-rao:
**How to categorize this PR?**

/area quality
/kind enhancement

**What this PR does / why we need it**:
This cleans up the CI scripts by pulling common functionality into a separate script instead of duplication. It also helps when we have new CI steps coming in, like `test_integration`. 

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
This PR is required to be merged to master so that https://github.com/gardener/etcd-druid/pull/506/ CI steps `build` and `test` can pass. You will also notice that the steps `build` and `test` in this PR are failing, due to the concourse CI issue/design that the base `.ci` directory comes from the master branch rather than the PR branch. You will see that once this PR is merged, the head-update job on the master branch will run successfully.
/cc @unmarshall @seshachalam-yv 

**Release note**:

```other developer
NONE
```

PR #523, opened 2023-02-09
Fix CI pipeline cleanup

shreyas-s-rao:

/area delivery
/kind bug

Fixes bug introduced in https://github.com/gardener/etcd-druid/pull/517 during cleanup of CI scripts. Pipeline job for etcd-druid head-update is failing for `build` and `test` steps due to incorrect path provided to the `source` command in the scripts (pwd is set to the parent of the repo root dir by default in each CICD step). This PR fixes this by correctly setting the `SOURCE_PATH` before sourcing the `common.sh` script.

Fixes #

/cc @unmarshall @seshachalam-yv 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.ci/build
+++ b/.ci/build
@@ -13,9 +13,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
+set -eo pipefail
 
-source .ci/common.sh
+# For all steps, concourse will set the following environment variables:
+# SOURCE_PATH - path to component repository root directory.
+if [[ -z "${SOURCE_PATH}" ]]; then
+  SOURCE_PATH="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
+else
+  SOURCE_PATH="$(readlink -f "${SOURCE_PATH}")"
+fi
+export SOURCE_PATH
 
 # For the build step concourse will additionally set the following environment variable:
 # BINARY_PATH - path to an existing (empty) directory to place build results into.
@@ -25,6 +32,8 @@ else
   BINARY_PATH="$(readlink -f "${BINARY_PATH}")/bin"
 fi
 export BINARY_PATH
+
+source "${SOURCE_PATH}/.ci/common.sh"
 
 ###############################################################################
 
--- a/.ci/common.sh
+++ b/.ci/common.sh
@@ -12,16 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
-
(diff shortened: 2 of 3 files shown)


seshachalam-yv (approved):
/lgtm

.ci/common.sh
```diff
@@ -31,6 +22,11 @@ REPOSITORY=${VCS}/${ORGANIZATION}/${PROJECT}
 # The `go <cmd>` commands requires to see the target repository to be part of a
 # Go workspace. Thus, if we are not yet in a Go workspace, let's create one
 # temporarily by using symbolic links.
+if [[ -z "$SOURCE_PATH" ]]; then
```
unmarshall, line 25: is this check required, considering that you source this script from `.ci/build` where you export SOURCE_PATH. Also you have set `set -eo pipefail` which will already exit if there was an issue setting this .

.ci/build
```diff
@@ -13,9 +13,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
+set -eo pipefail
 
-source .ci/common.sh
+# For all steps, concourse will set the following environment variables:
+# SOURCE_PATH - path to component repository root directory.
+if [[ -z "${SOURCE_PATH}" ]]; then
```
unmarshall, line 20: setting of `SOURCE_PATH` is done in both `.ci/build` and `.ci/test` - would it make sense to also move this to `.ci/common.sh`?

.ci/common.sh
shreyas-s-rao: `common.sh` is a bash script that can still be run independent of the other `.ci` scripts, so keeping this check made sense. Also, if an importer of common.sh does not set `SOURCE_PATH`, then this check allows us to fail early and with a meaningful error too.

.ci/build
shreyas-s-rao: That was the initial approach in https://github.com/gardener/etcd-druid/pull/517. But that did not work because the `.ci/build` and `.ci/test` scripts were unable to source `common.sh` unless they knew the repo root (absolute path) and sourced other scripts built using the absolute path. So in this PR I was forced to move the logic of determining `SOURCE_PATH` back into the individual scripts as the only way of fixing the [error we saw after merging that PR to master](https://concourse.ci.gardener.cloud/teams/gardener/pipelines/etcd-druid-master/jobs/master-head-update-job/builds/5).

unmarshall (approved):

PR #539, opened 2023-02-27
Refactor `RoleBinding` Helm charts into component

seshachalam-yv:

/area robustness
/kind enhancement

This PR refactors the `Etcd-druid` `RoleBinding` helm charts in Golang component and adds unit tests.

Fixes #

**Release note**:

```feature developer 
Eliminated `RoleBinding` helm charts and converted into Golang component with added unit tests.
```


pkg/component/etcd/rolebinding/values_helper.go
```diff
@@ -0,0 +1,32 @@
...
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
+	"github.com/gardener/etcd-druid/pkg/utils"
+)
+
+// GenerateValues generates `serviceaccount.Values` for the serviceaccount component with the given `etcd` object.
+func GenerateValues(etcd *druidv1alpha1.Etcd) *Values {
+	return &Values{
+		Name:               utils.GetRoleBindingName(etcd),
+		Namespace:          etcd.Namespace,
+		Labels:             etcd.Spec.Labels,
```
unmarshall, line 27: `etcd.Spec.Labels` have a lot of labels which includes `networking.gardener.cloud/*` labels as well. I am not sure why should all spec.Labels be applied on a `RoleBinding`. We have all the labels specified in `spec.etcd.Labels` today as well so maybe in this PR its ok to just do the port. However we should create another PR to clean up the labels that we are putting onto resources.

pkg/component/etcd/rolebinding/rolebinding_test.go
```diff
@@ -0,0 +1,145 @@
...
+		APIGroup: "rbac.authorization.k8s.io",
+		Kind:     "Role",
+		Name:     values.RoleName,
+	}))
+}
+func getRoleBindingKeyFromValue(value *rolebinding.Values) types.NamespacedName {
+	return client.ObjectKey{Name: value.Name, Namespace: value.Namespace}
+}
+
+func getTestValue() *rolebinding.Values {
```
unmarshall, line 137:
```suggestion
func getTestRoleBindingValues() *rolebinding.Values {
```

pkg/component/etcd/rolebinding/rolebinding_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var _ = Describe("RoleBinding Component", func() {
+	var (
+		ctx                  context.Context
+		c                    client.Client
+		value                *rolebinding.Values
```
unmarshall, line 38:
```suggestion
		values                *rolebinding.Values
```

pkg/component/etcd/rolebinding/rolebinding_test.go
```diff
@@ -0,0 +1,145 @@
...
+		value = getTestValue()
+		roleBindingComponent = rolebinding.New(c, value)
+	})
+
+	Describe("#Deploy", func() {
+		AfterEach(func() {
+			Expect(roleBindingComponent.Destroy(ctx)).NotTo(HaveOccurred())
+		})
+
+		It("should create and update the RoleBinding with the expected values", func() {
```
unmarshall, line 54: Ideally these are different tests and not just one test, right?
Would it make more sense to create separate `It("<test description>", func() {})` instead of clubbing all under just one test?

pkg/component/etcd/rolebinding/rolebinding_test.go
```diff
@@ -0,0 +1,145 @@
...
+			err := roleBindingComponent.Deploy(ctx)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("verifying that the RoleBinding is created on the K8s cluster as expected")
+			created := &rbacv1.RoleBinding{}
+			err = c.Get(ctx, getRoleBindingKeyFromValue(value), created)
+			Expect(err).NotTo(HaveOccurred())
+			verifyRoleBindingValues(created, value)
+
+			By("updating the RoleBinding")
```
unmarshall, line 65: Was any value updated? When you test patching a resource because it already exists then there needs to be some change made to it. Could not spot what change was made.

pkg/component/etcd/rolebinding/rolebinding_test.go
```diff
@@ -0,0 +1,145 @@
...
+		})
+	})
+
+	Describe("#Destroy", func() {
+		BeforeEach(func() {
+			Expect(roleBindingComponent.Deploy(ctx)).NotTo(HaveOccurred())
+		})
+
+		It("should delete the RoleBinding", func() {
+			roleBinding := &rbacv1.RoleBinding{}
```
unmarshall, line 97: Similarly these should be independent tests

pkg/component/etcd/rolebinding/value_helper_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/gardener/etcd-druid/api/v1alpha1"
+	"github.com/gardener/etcd-druid/pkg/component/etcd/rolebinding"
+	"github.com/gardener/etcd-druid/pkg/utils"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+var _ = Describe("RoleBindig", func() {
```
unmarshall, line 27:
```suggestion
var _ = Describe("RoleBinding", func() {
```

pkg/component/etcd/rolebinding/value_helper_test.go
```diff
@@ -0,0 +1,67 @@
...
+			Labels:             etcd.Spec.Labels,
+			RoleName:           utils.GetRoleName(etcd),
+			ServiceAccountName: utils.GetServiceAccountName(etcd),
+			OwnerReferences: []v1.OwnerReference{
+				{
+					APIVersion:         v1alpha1.GroupVersion.String(),
+					Kind:               etcd.Kind,
+					Name:               etcd.Name,
+					UID:                etcd.UID,
+					Controller:         pointer.BoolPtr(true),
```
unmarshall, line 54:
```suggestion
					Controller:         pointer.Bool(true),
```

pkg/component/etcd/rolebinding/value_helper_test.go
```diff
@@ -0,0 +1,67 @@
...
+					BlockOwnerDeletion: pointer.BoolPtr(true),
```
unmarshall, line 55:
```suggestion
					BlockOwnerDeletion: pointer.Bool(true),
```

pkg/component/etcd/rolebinding/rolebinding_test.go
seshachalam-yv: Thanks @unmarshall , fixed now.

unmarshall (commented):
Looks good to me. One more person can also have a look.

pkg/component/etcd/rolebinding/rolebinding_test.go
```diff
@@ -0,0 +1,144 @@
...
+		})
+		It("should not return an error when there is nothing to delete", func() {
+			err := roleBindingComponent.Destroy(ctx)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+})
+
+func verifyRoleBindingValues(expected *rbacv1.RoleBinding, values *rolebinding.Values) {
+	Expect(expected.Name).To(Equal(values.Name))
```
abdasgupta, line 102: Instead of `values.Name`, can we test it against actual string like `test-rolebinding` in this case? Can we do same for other validations?

abdasgupta (changes requested):
I just requested a minor change. Otherwise, looks good

pkg/component/etcd/rolebinding/rolebinding_test.go
seshachalam-yv: It is more appropriate to check the creation of the `rolebinding` object based on the input` values.Name` as this ensures that the correct object is being created as per the input. Checking against a static string may not accurately reflect the actual object being created.

pkg/component/etcd/serviceaccount/serviceaccount_suite_test.go
```diff
@@ -21,7 +21,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func TestService(t *testing.T) {
+func TestServiceAccount(t *testing.T) {
```
unmarshall, line 24: see the above comment. This should not be included as part of this PR.

unmarshall:
Thanks for the PR Sesha!
i have added review comments.

pkg/component/etcd/role/role_suite_test.go
```diff
@@ -21,7 +21,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func TestService(t *testing.T) {
+func TestRole(t *testing.T) {
```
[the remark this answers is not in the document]
seshachalam-yv: Thanks Madhav.
Reverted the changes

pkg/component/etcd/serviceaccount/serviceaccount_suite_test.go
seshachalam-yv: Thanks Madhav.
Reverted the changes

pkg/component/etcd/rolebinding/values.go
```diff
@@ -0,0 +1,33 @@
...
+	// Name is the name of the RoleBinding.
+	Name string
+	// Namespace is the namespace of the RoleBinding.
+	Namespace string
+	// RoleName is the role name of the RoleBinding.
+	RoleName string
+	// ServiceAccountName is the service account name of the RoleBinding.
+	ServiceAccountName string
+	// OwnerReferences are the OwnerReferences of the RoleBinding.
+	OwnerReferences []metav1.OwnerReference
```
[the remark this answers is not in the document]
seshachalam-yv: Yes, I agree, as part of this PR https://github.com/gardener/etcd-druid/pull/559, I will make it `OwnerReference` instead of slice for all components

api/v1alpha1/types_etcd.go
```diff
@@ -475,3 +475,8 @@ func (e *Etcd) GetAsOwnerReference() metav1.OwnerReference {
 func (e *Etcd) GetRoleName() string {
 	return fmt.Sprintf("druid.gardener.cloud:etcd:%s", e.Name)
 }
+
+// GetRoleBindingName returns the rolebinding name for the Etcd
+func (e *Etcd) GetRoleBindingName() string {
+	return fmt.Sprintf("druid.gardener.cloud:etcd:%s", e.Name)
```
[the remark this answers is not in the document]
seshachalam-yv: I'm reusing `GroupVersion.Group` to get the group name `druid.gardener.cloud`.
 I will create a separate PR to replace some more occurrences of this in the repository.

unmarshall (approved):
/lgtm

pkg/component/etcd/rolebinding/rolebinding_test.go
abdasgupta: I had an offline discussion with Sesha regarding this. I understood that he is not using GenerateValues() function to generate the values in test cases like what was being done in other components. My suggestion would be, if we think this is the way to test the components going forward, then we can change all other component test cases so that they don't use GenerateValues() as well

abdasgupta (approved):

seshachalam-yv:
Manually triggered E2E test. E2E tests have passed 🟢
[E2E-test.log](https://github.com/gardener/etcd-druid/files/11230044/E2E-test.log)

controllers/etcd/reconciler.go
```diff
@@ -280,6 +279,14 @@ func (r *Reconciler) delete(ctx context.Context, etcd *druidv1alpha1.Etcd) (ctrl
 		}, err
 	}
 
+	roleBindingValues := componentrolebinding.GenerateValues(etcd)
```
shreyas-s-rao, line 282: Just a thought: the order of deletion of resources seems to be the same as order of creation of resources, but maybe inverting it makes more sense semantically. This isn't applicable only for rolebinding, but for all the objects created and managed by etcd controller.

Anyway in execution, we don't need to worry since this simply sets deletion timestamp on the objects and deletion of an object occurs only when all dependent objects are deleted. Maybe something we can improve upon when re-looking at each controller's code.

shreyas-s-rao (approved):
@seshachalam-yv thanks for the well-written PR.

/lgtm

controllers/etcd/reconciler.go
seshachalam-yv: Sure Shreyas

Issue #475, opened 2022-11-28
[Improvement] Remove dependency on gardener/gardener for hack/scripts


unmarshall:
**Describe the enhancement:**

Currently a lot of scripts from g/g are being used inside etcd-druid. This brings in a LOT of dependencies which are not at being used in etcd-druid, resulting in a very fat vendor directory. We would like to only include dependencies that we currently use and nothing more than that.

**Expected behavior:**
Scripts (in /hack folder) that are only needed by druid will be explicitly maintained in this project and importing of all scripts from g/g will be removed.

**How To Reproduce (as minimally and precisely as possible):**

**Logs:**

**Screenshots (if applicable):**

**Environment (please complete the following information):**
 - Etcd version/commit ID :
 - Etcd-druid version/commit ID :
 - Cloud Provider [All/AWS/GCS/ABS/Swift/OSS]:

**Anything else we need to know?**:

unmarshall:
This is now done as part of #988

PR #541, opened 2023-03-01
Installing `kustomize` dependency for `Etcd-druid`

seshachalam-yv:

/area dev-productivity robustness
/kind enhancement

This PR installs the `kustomize` dependency for `Etcd-druid`, which is used by the `make deploy` command [here](https://github.com/gardener/etcd-druid/blob/master/Makefile#L60-L63).

One more thing has been fixed as part of this PR. 
- To address a make warning, the `GINKGO` target has been removed from `etcd-druid/hack/tools.mk`. 
  >**Warning**
  >etcd-druid/vendor/github.com/gardener/gardener/hack/tools.mk:110: warning: overriding commands for target `hack/tools/bin/ginkgo'
  >etcd-druid/hack/tools.mk:32: warning: ignoring old commands for target `hack/tools/bin/ginkgo'

- This is because the `GINKGO` target is always being overridden by `vendor/github.com/gardener/gardener/hack/tools.mk:110`, meaning that the `GINKGO` target from `etcd-druid/hack/tools.mk` is unnecessary

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/hack/tools.mk
+++ b/hack/tools.mk
@@ -13,13 +13,14 @@
 # limitations under the License.
 
 TOOLS_BIN_DIR              := $(TOOLS_DIR)/bin
-GINKGO		               := $(TOOLS_BIN_DIR)/ginkgo
 SKAFFOLD                   := $(TOOLS_BIN_DIR)/skaffold
 GO_ADD_LICENSE             := $(TOOLS_BIN_DIR)/addlicense
+KUSTOMIZE                  := $(TOOLS_BIN_DIR)/kustomize
 
 # default tool versions
 SKAFFOLD_VERSION ?= v1.38.0
 GO_ADD_LICENSE_VERSION ?= latest
+KUSTOMIZE_VERSION ?= v5.0.0
 
 export TOOLS_BIN_DIR := $(TOOLS_BIN_DIR)
 export PATH := $(abspath $(TOOLS_BIN_DIR)):$(PATH)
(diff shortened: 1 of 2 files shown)


hack/tools.mk
```diff
@@ -26,5 +27,5 @@ export PATH := $(abspath $(TOOLS_BIN_DIR)):$(PATH)
 # Tools                                 #
 #########################################
 
-$(GINKGO): go.mod
```
unmarshall, line 29: There are make targets like `test`, `test-cov`, `test-integration` which currently depends upon `GINKGO` dependency. What is the relation between removing Ginkgo and adding Kustomize targets in tools.mk?

seshachalam-yv: @unmarshall There are two things this PR fixes,
- Installs the kustomize dependency for Etcd-druid
- One more thing fixed as part of this PR is to fix a make warning, by removing the `GINKGO` target from etcd-druid/hack/tools.mk.
This is because the `GINKGO` target is always overriding by `vendor/github.com/gardener/gardener/hack/tools.mk:110`. So, we don't need this `GINKGO` target from `etcd-druid/hack/tools.mk`.

seshachalam-yv: Sorry for confusion, I updated the PR description

aaronfern (approved):

hack/tools.mk
shreyas-s-rao: If we plan to eventually remove druid's dependency on gardener/gardener hack scripts, as described by https://github.com/gardener/etcd-druid/issues/475, then we should ideally be retaining the GINKGO target in druid and instead work on removing the dependency on g/g GINKGO target, to avoid this conflict warning.

unmarshall: I agree. @shreyas-s-rao would it make sense to get this change in. Removing dependency on g/g scripts is taking a bit long. I will introduce it again later on once the dependency has been removed.

shreyas-s-rao: Sure, we can remove druid's GINKGO target in this PR, as we anyway plan to move necessary targets from g/g to druid later on.

hack/tools.mk
```diff
@@ -26,5 +27,5 @@ export PATH := $(abspath $(TOOLS_BIN_DIR)):$(PATH)
...
-	go build -o $(GINKGO) github.com/onsi/ginkgo/v2/ginkgo
+$(KUSTOMIZE):
+	@test -s $(TOOLS_BIN_DIR)/kustomize || GOBIN=$(abspath $(TOOLS_BIN_DIR)) go install sigs.k8s.io/kustomize/kustomize/v5@${KUSTOMIZE_VERSION}
```
shreyas-s-rao, line 31: Please add newline

shreyas-s-rao (approved):
@seshachalam-yv thanks for the PR! Please address my minor comment. Other than that, LGTM.
/lgtm

unmarshall (approved):

PR #569, opened 2023-04-05
Added support for compaction metrics in druid.

### LINKED ISSUE #515: [Feature] Expose metrics regarding compaction job
Metrics to gauge compaction job will be exposed.

Currently, we are not exposing metrics regarding compaction job. If compaction jobs are enabled across landscapes, we don't have anything to get an idea about how many times the compaction jobs ran successfully, how many resources the compaction jobs are consuming etc. The metrics will help to keep track of those numbers.

We had couple of rounds of discussions for the approach to implement the solution. @unmarshall summarised that there are four approaches that we can take to implement this:

There were 4 options that viable and 1 option was ruled out.
Option ruled out:
Metric server cannot be used for pushing custom metrics. [README](https://github.com/kubernetes-sigs/metrics-server) mentions the following:
```
Metrics Server is not meant for non-autoscaling purposes.
```

# Viable Options & Considerations:
Option-1
---
Introduce a sleep in the compaction Job allowing prometheus to scrape all metrics out of the Job before completing the job.
This uses the pull approach and no push gateway is required.

Option-2
---
Leverage etcd-druid which is already a long-running pod and expose the metrics out of it which will then be scraped by prometheus.
Case in point: Cron jobs also create Job resources, the controller is then responsible for collecting making job metrics available for push/pull.

Option-3
---
Use push gateway from prometheus which is then provisioned per shoot control namespace. A compaction job which is provisioned within the same namespace can then use the direct service endpoint to push to it.
Istvan has to do some research on what is the best way to setup a push gateway (with(out) disk, thereby making it a sts or a regular deployment).
Other drawback: In a HA setup of the control plane, entire monitoring stack is not HA. This means that in case of a zone outage if the prometheus and push gateway are in the failed zone and push gateway is also setup as a STS then they will not be available.
NOTE: Today access to prometheus is not TLS enabled. However, monitoring team has a mandate to enable TLS to meet the product security requirements. Push gateway will be one more component for which key management (including rotation) will need to be done.

Option 4
---
Use prometheus remote API to directly write to it. Monitoring team is evaluating this for federation to circumvent the firewall restrictions. However, this is not idiomatic way of pushing metrics from a metrics producer.

# Final Approach Selection and Rationale
We had another round of discussion to figure out the best option. 
- **We ruled out option 1** because having a `sleep` command inside a job seemed hack approach as - 
    -  Firstly, it wastes unnecessary CPU cycles.
    -  Secondly, we always have to find a magic number for `sleep` to run, so that Prometheus can scrape it timely.
    - Third, Currently, etcd-backup-restore starts a HTTP server (which also hosts the /metrics endpoint) only under the "server" subcommand, so starting this HTTP server for other subcommands such as compact requires a re-design of the component, as mentioned in https://github.com/gardener/etcd-backup-restore/issues/544#issuecomment-1270339794

- **We ruled out option 3** for now because
  -  If implemented, we have to maintain life-cycle of another new server named pushgateway server. 
  -  We may have to make it HA compatible as well.

- **Option 4 is ruled out** because it's not meant for end consumers to use.

- **Option 2 is looking promising as of now**. Druid needs to open metrics endpoint. Then we need to collect metrics related to the job by druid. 

Next Steps:
---
- [ ] ETCD Druid enables metrics endpoint either by starting up a server or by leveraging metrics endpoint of controller manager.
- [ ] prometheus server in each shoot namespace should collect metrics from respective druid metrics endpoint.
- [ ] Expose the compaction metrics as decided by ETCD Druid team.

So far, following metrics regarding jobs are decided to be exposed by Druid:

1. Increase a metrics counter if the current job is successful.
2. Increase a metrics counter if the current job failed.
3. Gauge CPU consumed by last running job.
4. Gauge Memory consumed by last running job.
5. Time taken by currently running job to complete.
6. Gauge for number of events to be compacted

abdasgupta:

/area disaster-recovery
/kind technical-debt

This PR exposes metrics related to compaction job initiated by druid.

Fixes #515 

As per the issue description, this PR exposes metrices for number of successful jobs, number of failed jobs, time taken to complete last job, number of delta events compacted, but it does not expose CPU and Memory consumption by last successful job. Resource consumption is the gauge that can only be measured from inside the job. Druid doesn't have access to that.

**Release note**:

```feature operator
Druid now exposes metrics related to snapshot compaction, on default port 8080. Please expose the desired metrics port via the etcd-druid service to allow metrics to be scraped by a Prometheus instance.
```

### DIFF
--- a/controllers/compaction/metrics.go
+++ b/controllers/compaction/metrics.go
@@ -1,0 +1,47 @@
+// Copyright 2023 SAP SE or an SAP affiliate company
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compaction
+
+import (
+	druidmetrics "github.com/gardener/etcd-druid/pkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+func init() {
+	// CompactionJobCountTotal
+	CompactionJobCountTotalValues := map[string][]string{
+		druidmetrics.LabelSucceeded: druidmetrics.DruidLabels[druidmetrics.LabelSucceeded],
+	}
+	CompactionJobCounterCombinations := druidmetrics.GenerateLabelCombinations(CompactionJobCountTotalValues)
+	for _, combination := range CompactionJobCounterCombinations {
+		druidmetrics.CompactionJobCounterTotal.With(prometheus.Labels(combination))
+	}
+
+	// CompactionJobDurationSeconds
+	CompactionJobDurationSecondsValues := map[string][]string{
+		druidmetrics.LabelSucceeded: druidmetrics.DruidLabels[druidmetrics.LabelSucceeded],
+	}
+	CompactionJobDurationSecondsCombinations := druidmetrics.GenerateLabelCombinations(CompactionJobDurationSecondsValues)
+	for _, combination := range CompactionJobDurationSecondsCombinations {
+		druidmetrics.CompactionJobDurationSeconds.With(prometheus.Labels(combination))
+	}
+
+	// Metrics have to be registered to be exposed:
+	metrics.Registry.MustRegister(druidmetrics.CompactionJobCounterTotal)
+	metrics.Registry.MustRegister(druidmetrics.CompactionJobDurationSeconds)
+
+	metrics.Registry.MustRegister(druidmetrics.TotalNumberOfEvents)
+}
--- a/controllers/compaction/metrics.go
+++ b/controllers/compaction/metrics.go
@@ -26,8 +26,8 @@ const (
 )
 
 var (
-	// CompactionJobCounterTotal is the metric used to count the total number of compaction jobs initiated by compaction controller.
-	CompactionJobCounterTotal = prometheus.NewCounterVec(
+	// metricJobsTotal is the metric used to count the total number of compaction jobs initiated by compaction controller.
+	metricJobsTotal = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: namespaceEtcdDruid,
 			Subsystem: subsystemCompaction,
@@ -37,8 +37,19 @@ var (
 		[]string{druidmetrics.LabelSucceeded},
 	)
 
-	// CompactionJobDurationSeconds is the metric used to expose the time taken to finish running a compaction job.
-	CompactionJobDurationSeconds = prometheus.NewGaugeVec(
+	// metricJobsCurrent is the metric used to expose currently running comapction job. This metric is important to get a sense of total number of compaction job running in a seed cluster..
+	metricJobsCurrent = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "jobs_current",
+			Help:      "Number of currently running comapction job.",
+		},
+		[]string{druidmetrics.ShootNamespace},
+	)
+
+	// metricJobDurationSeconds is the metric used to expose the time taken to finish running a compaction job.
+	metricJobDurationSeconds = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: namespaceEtcdDruid,
 			Subsystem: subsystemCompaction,
@@ -48,8 +59,8 @@ var (
 		[]string{druidmetrics.LabelSucceeded},
 	)
 
-	// NumDeltaEvents is the metric used to expose the total number of etcd events to be compacted by a compaction job.
-	NumDeltaEvents = prometheus.NewGaugeVec(
+	// metricNumDeltaEvents is the metric used to expose the total number of etcd events to be compacted by a compaction job.
+	metricNumDeltaEvents = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: namespaceEtcdDruid,
 			Subsystem: subsystemCompaction,
@@ -61,27 +72,29 @@ var (
 )
 
 func init() {
-	// compactionJobCountTotal
-	compactionJobCountTotalValues := map[string][]string{
+	// metricJobsTotalValues
+	metricJobsTotalValues := map[string][]string{
 		druidmetrics.LabelSucceeded: druidmetrics.DruidLabels[druidmetrics.LabelSucceeded],
 	}
-	compactionJobCounterCombinations := druidmetrics.GenerateLabelCombinations(compactionJobCountTotalValues)
-	for _, combination := range compactionJobCounterCombinations {
-		CompactionJobCounterTotal.With(prometheus.Labels(combination))
+	metricJobsTotalCombinations := druidmetrics.GenerateLabelCombinations(metricJobsTotalValues)
+	for _, combination := range metricJobsTotalCombinations {
+		metricJobsTotal.With(prometheus.Labels(combination))
 	}
 
-	// compactionJobDurationSeconds
-	compactionJobDurationSecondsValues := map[string][]string{
+	// metricJobDurationSecondsValues
+	metricJobDurationSecondsValues := map[string][]string{
 		druidmetrics.LabelSucceeded: druidmetrics.DruidLabels[druidmetrics.LabelSucceeded],
 	}
-	compactionJobDurationSecondsCombinations := druidmetrics.GenerateLabelCombinations(compactionJobDurationSecondsValues)
+	compactionJobDurationSecondsCombinations := druidmetrics.GenerateLabelCombinations(metricJobDurationSecondsValues)
 	for _, combination := range compactionJobDurationSecondsCombinations {
-		CompactionJobDurationSeconds.With(prometheus.Labels(combination))
+		metricJobDurationSeconds.With(prometheus.Labels(combination))
 	}
 
 	// Metrics have to be registered to be exposed:
-	metrics.Registry.MustRegister(CompactionJobCounterTotal)
-	metrics.Registry.MustRegister(CompactionJobDurationSeconds)
+	metrics.Registry.MustRegister(metricJobsTotal)
+	metrics.Registry.MustRegister(metricJobDurationSeconds)
 
-	metrics.Registry.MustRegister(NumDeltaEvents)
+	metrics.Registry.MustRegister(metricJobsCurrent)
+
+	metrics.Registry.MustRegister(metricNumDeltaEvents)
 }
--- a/controllers/compaction/metrics.go
+++ b/controllers/compaction/metrics.go
@@ -45,7 +45,7 @@ var (
 			Name:      "jobs_current",
 			Help:      "Number of currently running comapction job.",
 		},
-		[]string{druidmetrics.ShootNamespace},
+		[]string{druidmetrics.EtcdNamespace},
 	)
 
 	// metricJobDurationSeconds is the metric used to expose the time taken to finish running a compaction job.
--- a/controllers/compaction/metrics.go
+++ b/controllers/compaction/metrics.go
@@ -34,7 +34,7 @@ var (
 			Name:      "jobs_total",
 			Help:      "Total number of compaction jobs initiated by compaction controller.",
 		},
-		[]string{druidmetrics.LabelSucceeded},
+		[]string{druidmetrics.LabelSucceeded, druidmetrics.EtcdNamespace},
 	)
 
 	// metricJobsCurrent is the metric used to expose currently running compaction job. This metric is important to get a sense of total number of compaction job running in a seed cluster.
@@ -56,7 +56,7 @@ var (
 			Name:      "job_duration_seconds",
 			Help:      "Total time taken in seconds to finish a running compaction job.",
 		},
-		[]string{druidmetrics.LabelSucceeded},
+		[]string{druidmetrics.LabelSucceeded, druidmetrics.EtcdNamespace},
 	)
 
 	// metricNumDeltaEvents is the metric used to expose the total number of etcd events to be compacted by a compaction job.
@@ -67,7 +67,7 @@ var (
 			Name:      "num_delta_events",
 			Help:      "Total number of etcd events to be compacted by a compaction job.",
 		},
-		[]string{},
+		[]string{druidmetrics.EtcdNamespace},
 	)
 )
 
@@ -75,6 +75,7 @@ func init() {
 	// metricJobsTotalValues
 	metricJobsTotalValues := map[string][]string{
 		druidmetrics.LabelSucceeded: druidmetrics.DruidLabels[druidmetrics.LabelSucceeded],
+		druidmetrics.EtcdNamespace:  druidmetrics.DruidLabels[druidmetrics.EtcdNamespace],
 	}
 	metricJobsTotalCombinations := druidmetrics.GenerateLabelCombinations(metricJobsTotalValues)
 	for _, combination := range metricJobsTotalCombinations {
@@ -84,6 +85,7 @@ func init() {
 	// metricJobDurationSecondsValues
 	metricJobDurationSecondsValues := map[string][]string{
 		druidmetrics.LabelSucceeded: druidmetrics.DruidLabels[druidmetrics.LabelSucceeded],
+		druidmetrics.EtcdNamespace:  druidmetrics.DruidLabels[druidmetrics.EtcdNamespace],
 	}
 	compactionJobDurationSecondsCombinations := druidmetrics.GenerateLabelCombinations(metricJobDurationSecondsValues)
 	for _, combination := range compactionJobDurationSecondsCombinations {
--- a/controllers/compaction/metrics.go
+++ b/controllers/compaction/metrics.go
@@ -37,7 +37,7 @@ var (
 		[]string{druidmetrics.LabelSucceeded},
 	)
 
-	// metricJobsCurrent is the metric used to expose currently running comapction job. This metric is important to get a sense of total number of compaction job running in a seed cluster..
+	// metricJobsCurrent is the metric used to expose currently running compaction job. This metric is important to get a sense of total number of compaction job running in a seed cluster.
 	metricJobsCurrent = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: namespaceEtcdDruid,
@@ -48,13 +48,13 @@ var (
 		[]string{druidmetrics.EtcdNamespace},
 	)
 
-	// metricJobDurationSeconds is the metric used to expose the time taken to finish running a compaction job.
-	metricJobDurationSeconds = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
+	// metricJobDurationSeconds is the metric used to expose the time taken to finish a compaction job.
+	metricJobDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
 			Namespace: namespaceEtcdDruid,
 			Subsystem: subsystemCompaction,
 			Name:      "job_duration_seconds",
-			Help:      "Total time taken in seconds to finish running a compaction job.",
+			Help:      "Total time taken in seconds to finish a running compaction job.",
 		},
 		[]string{druidmetrics.LabelSucceeded},
 	)
--- a/controllers/compaction/metrics.go
+++ b/controllers/compaction/metrics.go
@@ -20,28 +20,68 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
+const (
+	namespaceEtcdDruid  = "etcddruid"
+	subsystemCompaction = "compaction"
+)
+
+var (
+	// CompactionJobCounterTotal is the metric used to count the total number of compaction jobs initiated by compaction controller.
+	CompactionJobCounterTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "jobs_total",
+			Help:      "Total number of compaction jobs initiated by compaction controller.",
+		},
+		[]string{druidmetrics.LabelSucceeded},
+	)
+
+	// CompactionJobDurationSeconds is the metric used to expose the time taken to finish running a compaction job.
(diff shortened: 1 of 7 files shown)


controllers/compaction/reconciler.go
```diff
@@ -201,13 +206,18 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
 				RequeueAfter: 10 * time.Second,
 			}, fmt.Errorf("error while deleting failed compaction job: %v", err)
 		}
+		druidmetrics.CompactionJobCounterTotal.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededFalse}).Inc()
```
shreyas-s-rao, line 209: Please also record `druidmetrics.CompactionJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededFalse}).Set(<total time elapsed since the job was started>)`. It is important to record the total time taken by failed jobs as well, in addition to total number of jobs that failed.

Since the job status will not record `completionTime` for non-successful jobs (see [documentation](https://pkg.go.dev/k8s.io/api/batch/v1@v0.26.1#JobStatus.CompletionTime)), you will have to calculate the time elapsed since `job.status.startTime` and record that as the metric value here.

shreyas-s-rao (changes requested):
@abdasgupta thanks for the PR. Please address my review comments and also add documentation about the metrics being exposed by druid, similar to https://github.com/gardener/etcd-druid/blob/master/docs/multinode-metrics.md, under `docs/monitoring/compaction-metrics.md`.

Please also edit the release note to 
```
Druid now exposes metrics related to snapshot compaction, on default port 8080. Please expose the desired metrics port via the etcd-druid service to allow metrics to be scraped by a Prometheus instance.
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,163 @@
...
+	// ValueSucceededTrue is value True for metric label succeeded.
+	ValueSucceededTrue = "true"
+	// ValueSucceededFalse is value False for metric label failed.
+	ValueSucceededFalse = "false"
+	// CompactionJobSubmitted is value for metric of total number of compaction jobs initiated by compaction controller.
+	CompactionJobSubmitted = "compaction_jobs"
+	// ValueRestoreSingleNode is value for metric of single node restoration.
+	ValueRestoreSingleNode = "single_node"
+	// LabelError is a metric error to indicate error occured.
+	LabelError = "error"
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,163 @@
...
+
+	namespaceEtcdDruid  = "etcd_druid"
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,163 @@
...
+	subsystemCompaction = "compaction"
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,163 @@
...
+	// TotalNumberOfEvents is metric to expose the total number of events compacted by last compaction job.
+	TotalNumberOfEvents = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "total_number_of_events",
+			Help:      "Total number of events compacted by last compaction job.",
+		},
+		[]string{},
+	)
```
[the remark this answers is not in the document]
abdasgupta: Done

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,163 @@
...
+// Ex: [p,q,r] -> [[p],[q],[r]]
+func wrapInSlice(s []string) [][]string {
+	output := make([][]string, len(s))
+	for i := 0; i < len(output); i++ {
+		elem := make([]string, 1)
+		elem[0] = s[i]
+		output[i] = elem
+	}
+	return output
+}
```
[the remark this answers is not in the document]
abdasgupta: As discussed, I have defined the functions here. Will raise another PR to use these functions in ETCDBR

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	druidmetrics "github.com/gardener/etcd-druid/pkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+func init() {
+	// CompactionJobCountTotal
+	CompactionJobCountTotalValues := map[string][]string{
```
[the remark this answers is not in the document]
abdasgupta: Removed

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+		druidmetrics.LabelSucceeded: druidmetrics.DruidLabels[druidmetrics.LabelSucceeded],
+	}
+	CompactionJobCounterCombinations := druidmetrics.GenerateLabelCombinations(CompactionJobCountTotalValues)
+	for _, combination := range CompactionJobCounterCombinations {
+		druidmetrics.CompactionJobCounterTotal.With(prometheus.Labels(combination))
+	}
+
+	// CompactionJobDurationSeconds
+	CompactionJobDurationSecondsValues := map[string][]string{
```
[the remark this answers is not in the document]
abdasgupta: Removed

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,163 @@
...
+		},
+		[]string{LabelSucceeded},
+	)
+
+	// TotalNumberOfEvents is metric to expose the total number of events compacted by last compaction job.
+	TotalNumberOfEvents = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "total_number_of_events",
```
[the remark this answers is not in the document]
abdasgupta: Renamed

controllers/compaction/reconciler.go
abdasgupta: Done

controllers/compaction/reconciler.go
```diff
@@ -195,19 +198,27 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
 
 	// Delete job and requeue if the job failed
 	if job.Status.Failed > 0 {
+		if job.Status.StartTime != nil {
+			CompactionJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededFalse}).Set(float64(time.Since(job.Status.StartTime.Time).Seconds()))
```
aaronfern, line 202: Is there a chance that this metric can be recorded multiple times if the job deletion fails and requeues?

In that case I'd suggest moving this block to after the job has been successfully deleted

controllers/compaction/reconciler.go
```diff
@@ -195,19 +198,27 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
...
+		CompactionJobCounterTotal.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededFalse}).Inc()
 		return ctrl.Result{
 			RequeueAfter: 10 * time.Second,
 		}, nil
 	}
 
 	// Delete job and return if the job succeeded
 	if job.Status.Succeeded > 0 {
+		CompactionJobCounterTotal.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededTrue}).Inc()
+		if job.Status.CompletionTime != nil {
```
aaronfern, line 219: Do we not also have to check for `Status.StartTime` here? it's checked above

controllers/compaction/reconciler.go
aaronfern, line 202: Should we not use a `failed` label here? from what I understood, this metric will be stored with a label `succeeded` which would not be accurate

shreyas-s-rao: We actually want it to record a `succeeded: false` metric every time the job fails. That gives us an accurate reading of how many times a job failed. The job can succeed only once, but it can fail multiple times, so the `failed` metric has to captured this many number of times.

controllers/compaction/reconciler.go
shreyas-s-rao: If the job has succeeded, then `completionTime` is sufficient, since it already accurately captures the time taken for the job to complete successfully. So I don't think we need `startTime` when the job has succeeded.

controllers/compaction/reconciler.go
shreyas-s-rao: The code is correct. It sets label `succeeded` to value `false`, as expected.

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "jobs_total",
+			Help:      "Total number of compaction jobs initiated by compaction controller.",
+		},
+		[]string{druidmetrics.LabelSucceeded},
+	)
+
+	// CompactionJobDurationSeconds is the metric used to expose the time taken to finish running a compaction job.
+	CompactionJobDurationSeconds = prometheus.NewGaugeVec(
```
shreyas-s-rao, line 41: Same as above

```suggestion
	metricJobDurationSeconds = prometheus.NewGaugeVec(
```

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "job_duration_seconds",
+			Help:      "Total time taken in seconds to finish running a compaction job.",
+		},
+		[]string{druidmetrics.LabelSucceeded},
+	)
+
+	// NumDeltaEvents is the metric used to expose the total number of etcd events to be compacted by a compaction job.
+	NumDeltaEvents = prometheus.NewGaugeVec(
```
shreyas-s-rao, line 52: Same as above

```suggestion
	metricNumDeltaEvents = prometheus.NewGaugeVec(
```

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	for _, combination := range compactionJobDurationSecondsCombinations {
+		CompactionJobDurationSeconds.With(prometheus.Labels(combination))
+	}
+
+	// Metrics have to be registered to be exposed:
+	metrics.Registry.MustRegister(CompactionJobCounterTotal)
+	metrics.Registry.MustRegister(CompactionJobDurationSeconds)
+
+	metrics.Registry.MustRegister(NumDeltaEvents)
+}
```
shreyas-s-rao, line 87: Same as above for all variables introduced here.

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2023 SAP SE or an SAP affiliate company
```
shreyas-s-rao, line 1: Does it also make sense to introduce another metric `metricJobsCurrentCount` which keeps track of the number of compaction jobs being run currently by druid, so that certain scheduling decisions may be taken based on number of available machines to schedule these compaction jobs.

controllers/compaction/reconciler.go
shreyas-s-rao, line 202:
```suggestion
			CompactionJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededFalse}).Set(time.Since(job.Status.StartTime.Time).Seconds())
```

Redundant `float64` type conversion.

controllers/compaction/reconciler.go
```diff
@@ -195,19 +198,27 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
...
+			CompactionJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededTrue}).Set(float64(job.Status.CompletionTime.Time.Sub(job.Status.StartTime.Time).Seconds()))
```
shreyas-s-rao, line 220: Same as above

shreyas-s-rao (changes requested):
@abdasgupta Thanks for addressing my review comments. Can you please add the documentation for the metrics as requested earlier?

I have added a few more comments based on the changes. PTAL, thanks!

controllers/compaction/reconciler.go
abdasgupta: @aaronfern is right here. `completionTime` is the time when the job completed. We are collecting the duration in the metrics. So, we need to consider `startTime`. I made necessary changes.

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+const (
+	namespaceEtcdDruid  = "etcddruid"
+	subsystemCompaction = "compaction"
+)
+
+var (
+	// CompactionJobCounterTotal is the metric used to count the total number of compaction jobs initiated by compaction controller.
+	CompactionJobCounterTotal = prometheus.NewCounterVec(
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/compaction/metrics.go
abdasgupta: Done

controllers/compaction/metrics.go
abdasgupta: Done

controllers/compaction/reconciler.go
abdasgupta: Done

controllers/compaction/reconciler.go
abdasgupta: Done

controllers/compaction/metrics.go
abdasgupta: Done

controllers/compaction/metrics.go
abdasgupta: Done

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+	// LabelSucceeded is a metric label indicating whether associated metric
+	// series is for success or failure.
+	LabelSucceeded = "succeeded"
+	// ValueSucceededTrue is value True for metric label succeeded.
+	ValueSucceededTrue = "true"
+	// ValueSucceededFalse is value False for metric label failed.
+	ValueSucceededFalse = "false"
+
+	// ShootNamespace is the label for prometheus metrics to indicate shoot namespace
+	ShootNamespace = "shoot_namespace"
```
shreyas-s-rao, line 29: Please make this change since druid must ideally not have any knowledge of whether it's running within Gardener control plane for shoots, or just for any etcds. 
 
```suggestion
	EtcdNamespace = "etcd_namespace"
```

docs/development/metrics.md
```diff
@@ -0,0 +1,36 @@
...
+
+The naming of metrics follows the suggested [Prometheus best practices][prometheus-naming]. All compaction related metrics are put under namespace `etcddruid` and subsystem `compaction`.
+
+### Compaction
+
+These metrics give an idea about the compaction jobs that run after some interval in shoot control planes. Studying the metrices, we can deduce how many compaction job ran successfully, how many failed, how many delta events compacted etc.
+
+| Name | Description | Type |
+|------|-------------|------|
+| jobs_total | Total number of compaction jobs initiated by compaction controller. | Counter |
```
shreyas-s-rao, line 17:
```suggestion
| etcddruid_compaction_jobs_total | Total number of compaction jobs initiated by compaction controller. | Counter |
```

Please specify the full name of the metrics in this table, for the below metrics as well, as it becomes easy for consumption by end-users.

shreyas-s-rao (changes requested):
@abdasgupta Thanks again for making the requested changes and adding the doc. Can you please address my final two comments about cosmetic changes? Other than that, LGTM

shreyas-s-rao (approved):
@abdasgupta thanks for making all the requested changes.
/lgtm

ishan16696 (changes requested):

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "jobs_current",
+			Help:      "Number of currently running comapction job.",
+		},
+		[]string{druidmetrics.EtcdNamespace},
+	)
+
+	// metricJobDurationSeconds is the metric used to expose the time taken to finish running a compaction job.
+	metricJobDurationSeconds = prometheus.NewGaugeVec(
```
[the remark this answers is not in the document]
abdasgupta: @ishan16696 , Histogram needs to observe a parameter. So, if a background function runs after a request has been made, we could use histogram to measure how long the background function took to serve the request. But here, we can't observe the job duration time. We need to get out of the controller once the job starts. Only way to measure job duration is by measuring the start time and completion time in job status. That's why we can only set the duration value, not observe the duration. So, we only can use gauge, not histogram.

docs/development/metrics.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Compaction
+
+These metrics give an idea about the compaction jobs that run after some interval in shoot control planes. Studying the metrices, we can deduce how many compaction job ran successfully, how many failed, how many delta events compacted etc.
+
+| Name | Description | Type |
+|------|-------------|------|
+| etcddruid_compaction_jobs_total | Total number of compaction jobs initiated by compaction controller. | Counter |
+| etcddruid_compaction_jobs_current | Number of currently running comapction job. | Gauge |
+| etcddruid_compaction_job_duration_seconds | Total time taken in seconds to finish running a compaction job. | Gauge |
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		prometheus.GaugeOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "jobs_current",
+			Help:      "Number of currently running comapction job.",
+		},
+		[]string{druidmetrics.EtcdNamespace},
+	)
+
+	// metricJobDurationSeconds is the metric used to expose the time taken to finish running a compaction job.
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		prometheus.CounterOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "jobs_total",
+			Help:      "Total number of compaction jobs initiated by compaction controller.",
+		},
+		[]string{druidmetrics.LabelSucceeded},
+	)
+
+	// metricJobsCurrent is the metric used to expose currently running comapction job. This metric is important to get a sense of total number of compaction job running in a seed cluster..
```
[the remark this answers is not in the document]
abdasgupta: Done

controllers/compaction/metrics.go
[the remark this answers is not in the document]
ishan16696: ok, but can this possible ?
```go
metricJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededFalse}). Observe(time.Since(job.Status.StartTime.Time).Seconds())
```
as I'm also seeing this in backup-restore as well, check this
https://github.com/gardener/etcd-backup-restore/blob/d4e6e462ea5db0bda501f966bcd1308349683e50/pkg/initializer/initializer.go#L85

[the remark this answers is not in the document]
abdasgupta: As discussed, we are already calculating the job duration from the job status. We can't use Observe function from histogram to calculate job duration because the reconcile doesn't actively track the job. @shreyas-s-rao Would you like to share your thoughts on this?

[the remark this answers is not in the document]
shreyas-s-rao: The `Observe` function simply gives Prometheus clients the ability to "record" a value for a given metric. It has no bearing on what kind of metric is being recorded, since both Histogram and Gauge metrics can be set to a specific value using `Observe` and `Set` methods respectively.

Coming to `metricJobDurationSeconds`, I have no string personal opinions on the type of metric to use, but according to what Prometheus recommends, Ishan is right in saying that it should be recorded using a Histogram rather than a Gauge, since durations are generally recorded by Histograms. @ishan16696 thanks for pointing this out.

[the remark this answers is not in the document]
abdasgupta: Made the changes

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// metricNumDeltaEvents is the metric used to expose the total number of etcd events to be compacted by a compaction job.
+	metricNumDeltaEvents = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "num_delta_events",
+			Help:      "Total number of etcd events to be compacted by a compaction job.",
+		},
+		[]string{},
```
ishan16696, line 70: I think labels `EtcdNamespace` should also be added in this `metricNumDeltaEvents ` matric, so that we can able to filter using this namespace labels.

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+var (
+	// metricJobsTotal is the metric used to count the total number of compaction jobs initiated by compaction controller.
+	metricJobsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "jobs_total",
+			Help:      "Total number of compaction jobs initiated by compaction controller.",
+		},
+		[]string{druidmetrics.LabelSucceeded},
```
ishan16696, line 37: I think labels `EtcdNamespace` should also be added in this `metricJobsTotal`  matrix, so that we can able to filter using this namespace labels.

controllers/compaction/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// metricJobDurationSeconds is the metric used to expose the time taken to finish a compaction job.
+	metricJobDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: namespaceEtcdDruid,
+			Subsystem: subsystemCompaction,
+			Name:      "job_duration_seconds",
+			Help:      "Total time taken in seconds to finish a running compaction job.",
+		},
+		[]string{druidmetrics.LabelSucceeded},
```
ishan16696, line 59: same as above

abdasgupta: Done

controllers/compaction/metrics.go
abdasgupta: Done

controllers/compaction/metrics.go
abdasgupta: Done

ishan16696 (approved):

PR #587, opened 2023-05-02
Etcd controller shouldn't removes the scale-up annotation until scale-up succeeds

### LINKED ISSUE #582: [BUG] Etcd-druid removes the scale-up annotation even if scale-up didn't succeed.
**Describe the bug:**
Etcd-druid controller removes the scale-up annotation `gardener.cloud/scaled-to-multi-node` from etcd-statefulset when 2nd member join the cluster but 3rd member fails to join the cluster. 
[clusterScaledUpToMultiNode](https://github.com/gardener/etcd-druid/blob/e30729fec708580f5203de9520a2415c8765a3b5/pkg/component/etcd/statefulset/statefulset.go#L287-L292) wrongly returns false as `sts.Spec.Replicas` is no longer equal to `1`.
```
k get sts etcd-main -oyaml | grep replicas -C 5
spec:
  podManagementPolicy: Parallel
  replicas: 3
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      instance: etcd-main
      name: etcd
--
  collisionCount: 0
  currentReplicas: 1
  currentRevision: etcd-main-64b8bbd7f7
  observedGeneration: 2
  readyReplicas: 2
  replicas: 3
  updateRevision: etcd-main-6479474444
  updatedReplicas: 2
```

**Expected behavior:**
etcd-druid controller shouldn't removes the scale-up annotation `gardener.cloud/scaled-to-multi-node` until all members (defined by `.spec.replicas`) joins the cluster.

1. Start the etcd-cluster deployed by druid with `.spec.replicas` = 1
2. Now, update the `.spec.replicas` to 3 in etcd custom resource
3. Don't let 1 of etcd member to join the cluster and let the other 1 member to join the cluster.
4. you can check the etcd statefulset yaml that `gardener.cloud/scaled-to-multi-node` has been removed.

 - Etcd version/commit ID :

ishan16696:
Etcd controller shouldn't removes the scale-up annotation from etcd statefulset until scale-up succeeds.

Fixes #582 

**Release note**:
```bugfix operator
Added check to ensure that the scale up annotation is removed from the etcd statefulset only when scale-up succeeds
```

### DIFF
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -476,6 +476,16 @@ func getStsAnnotations(val *Values, sts *appsv1.StatefulSet) map[string]string {
 		},
 		val.Annotations,
 	)
+
+	// If `scaleToMultiNodeAnnotationKey` annotation is already present in etcd statefulset
+	// then it is better to check scale-up was successful or not before removing `scaleToMultiNodeAnnotationKey` annotation.
+	if metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) {
+		if (sts.Spec.Replicas != nil && sts.Status.UpdatedReplicas < *sts.Spec.Replicas) || sts.Status.Replicas > sts.Status.UpdatedReplicas {
+			annotations[scaleToMultiNodeAnnotationKey] = ""
+			return annotations
+		}
+	}
+
 	if clusterScaledUpToMultiNode(val, sts) {
 		annotations[scaleToMultiNodeAnnotationKey] = ""
 	}
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -286,7 +286,8 @@ func immutableFieldUpdate(sts *appsv1.StatefulSet, val Values) bool {
 
 func clusterScaledUpToMultiNode(val *Values, sts *appsv1.StatefulSet) bool {
 	if sts != nil && sts.Spec.Replicas != nil {
-		return val.Replicas > 1 && *sts.Spec.Replicas == 1
+		return (val.Replicas > 1 && *sts.Spec.Replicas == 1) ||
+			(metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) && sts.Status.UpdatedReplicas < *sts.Spec.Replicas)
 	}
 	return val.Replicas > 1 && val.StatusReplicas == 1
 }
@@ -476,15 +477,6 @@ func getStsAnnotations(val *Values, sts *appsv1.StatefulSet) map[string]string {
 		},
 		val.Annotations,
 	)
-
-	// If `scaleToMultiNodeAnnotationKey` annotation is already present in etcd statefulset
-	// then it is better to check scale-up was successful or not before removing `scaleToMultiNodeAnnotationKey` annotation.
-	if sts != nil && metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) {
-		if sts.Status.UpdatedReplicas < *sts.Spec.Replicas || sts.Status.Replicas > sts.Status.UpdatedReplicas {
-			annotations[scaleToMultiNodeAnnotationKey] = ""
-			return annotations
-		}
-	}
 
 	if clusterScaledUpToMultiNode(val, sts) {
 		annotations[scaleToMultiNodeAnnotationKey] = ""
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -479,8 +479,8 @@ func getStsAnnotations(val *Values, sts *appsv1.StatefulSet) map[string]string {
 
 	// If `scaleToMultiNodeAnnotationKey` annotation is already present in etcd statefulset
 	// then it is better to check scale-up was successful or not before removing `scaleToMultiNodeAnnotationKey` annotation.
-	if metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) {
-		if (sts.Spec.Replicas != nil && sts.Status.UpdatedReplicas < *sts.Spec.Replicas) || sts.Status.Replicas > sts.Status.UpdatedReplicas {
+	if sts != nil && metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) {
+		if sts.Status.UpdatedReplicas < *sts.Spec.Replicas || sts.Status.Replicas > sts.Status.UpdatedReplicas {
 			annotations[scaleToMultiNodeAnnotationKey] = ""
 			return annotations
 		}


pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -476,6 +476,16 @@ func getStsAnnotations(val *Values, sts *appsv1.StatefulSet) map[string]string {
 		},
 		val.Annotations,
 	)
+
+	// If `scaleToMultiNodeAnnotationKey` annotation is already present in etcd statefulset
+	// then it is better to check scale-up was successful or not before removing `scaleToMultiNodeAnnotationKey` annotation.
+	if metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) {
+		if (sts.Spec.Replicas != nil && sts.Status.UpdatedReplicas < *sts.Spec.Replicas) || sts.Status.Replicas > sts.Status.UpdatedReplicas {
```
abdasgupta, line 483: Should not we also compare with ETCD CR replicas? It would add ultimate safety check

abdasgupta (commented):
Just one comment.

pkg/component/etcd/statefulset/statefulset.go
ishan16696: I don't want any bug in our controllers can lead to add/remove scale-up annotation in etcd sts.
TBH, right now I would like to depend only on statefulset controller.

ishan16696:
Druid e2e tests passed with this PR ✅

pkg/component/etcd/statefulset/statefulset.go
abdasgupta: Okay

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -476,6 +476,16 @@ func getStsAnnotations(val *Values, sts *appsv1.StatefulSet) map[string]string {
...
+			annotations[scaleToMultiNodeAnnotationKey] = ""
+			return annotations
+		}
+	}
+
```
seshachalam-yv, line 488:
``` go
// generateStsAnnotations merges the provided annotations with default annotations and determines if the scaleToMultiNodeAnnotationKey should be added.
func generateStsAnnotations(val *Values, sts *appsv1.StatefulSet) map[string]string {
	// Merge the provided annotations with the default ones.
	annotations := utils.MergeStringMaps(
		map[string]string{
			common.GardenerOwnedBy:   fmt.Sprintf("%s/%s", val.Namespace, val.Name),
			common.GardenerOwnerType: "etcd",
		},
		val.Annotations,
	)

	// Determine if the scaleToMultiNodeAnnotationKey should be added to the annotations.
	shouldAddAnnotation := false
	if sts != nil && sts.Spec.Replicas != nil {
		// If the scaleToMultiNodeAnnotationKey is already present and all pods are not ready,
		// keep the annotation to indicate that scale-up is in progress.
		if metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) &&
			((sts.Status.UpdatedReplicas < *sts.Spec.Replicas) || (sts.Status.Replicas > sts.Status.UpdatedReplicas)) {
			shouldAddAnnotation = true
		} else if val.Replicas > 1 && *sts.Spec.Replicas == 1 {
			// If the cluster has been scaled up to multi-node, add the annotation.
			shouldAddAnnotation = true
		}
	} else if val.Replicas > 1 && val.StatusReplicas == 1 {
		// If the StatefulSet is not available and the cluster has been scaled up to multi-node, add the annotation.
		shouldAddAnnotation = true
	}

	// Add the scaleToMultiNodeAnnotationKey if necessary.
	if shouldAddAnnotation {
		annotations[scaleToMultiNodeAnnotationKey] = ""
	}

	return annotations
}
```

I would like to propose the following changes:
- Remove the `clusterScaledUpToMultiNode` function and inline the conditions for better readability, which also helps avoid duplicate checks.
- Rename the function to `generateStsAnnotations` for improved clarity and to convey the purpose of the function more accurately.
- Update the comments to provide more context and better explanations for the logic behind adding or keeping the `scaleToMultiNodeAnnotationKey` annotation.

I understand that you didn't modify the other parts of the code, but I believe these changes would enhance readability and maintainability. 
What do you think?

abdasgupta: As it's very urgent to push these PRs, I would suggest not to introduce any changes even for readability and maintainability. We can raise another PR to do so

pkg/component/etcd/statefulset/statefulset.go
unmarshall, line 483: This condition is not complete.
I tested and captured some states. What i did was i took a single etcd and deployed it on a KIND cluster and then scaled it directly by using `kubectl scale sts etcd-test --replicas=3` and i saw the following:
**Before the scale-out status looked like:**
```
status:
  availableReplicas: 1
  collisionCount: 0
  currentReplicas: 1
  currentRevision: etcd-test-68496b5c76
  observedGeneration: 1
  readyReplicas: 1
  replicas: 1
  updateRevision: etcd-test-68496b5c76
  updatedReplicas: 1
```
**Just after scale up the status looked like:**
```
status:
  availableReplicas: 1
  collisionCount: 0
  currentReplicas: 2
  currentRevision: etcd-test-68496b5c76
  observedGeneration: 2
  readyReplicas: 1
  replicas: 2
  updateRevision: etcd-test-68496b5c76
  updatedReplicas: 2
```
**When all replicas were launched but they were not ready, the status was:**
```
status:
  availableReplicas: 1
  collisionCount: 0
  currentReplicas: 3
  currentRevision: etcd-test-68496b5c76
  observedGeneration: 2
  readyReplicas: 1
  replicas: 3
  updateRevision: etcd-test-68496b5c76
  updatedReplicas: 3
```
If you see the last status `sts.Status.Replicas = 3` and `sts.Status.UpdatedReplicas=3` but the 2/3 pods are not ready and are not available, this means that k8s will keep restarting these containers. So as per your comment if the intent is to check that the scale-up was successful then the current conditions will not suffice. However if the intent is to check that scale-up has been done to 3 and therefore there are 3 pods (irrespective of their status) then the condition would suffice.

unmarshall: If you wish to ensure that scale has resulted in all pods being ready then you must check availableReplicas/readyReplicas.

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -476,6 +476,16 @@ func getStsAnnotations(val *Values, sts *appsv1.StatefulSet) map[string]string {
 		},
 		val.Annotations,
 	)
+
+	// If `scaleToMultiNodeAnnotationKey` annotation is already present in etcd statefulset
+	// then it is better to check scale-up was successful or not before removing `scaleToMultiNodeAnnotationKey` annotation.
+	if metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) {
```
aaronfern, line 482: There is a chance that the sts could be nil due to a [`getExistingSts()`](https://github.com/gardener/etcd-druid/blob/master/pkg/component/etcd/statefulset/statefulset.go#L155) call and could result in a nil pointer dereference
Please add a `sts != nil` check here

aaronfern (changes requested):

aaronfern:
Can you update the release note to something like the following?
```
Added additional check to ensure that the scale up annotation is removed from the statefulset only when scale-up succeeds
```

pkg/component/etcd/statefulset/statefulset.go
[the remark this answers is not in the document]
ishan16696: What if DoD/SREs scale down the statefulset to `0`, will it be fine as it is pointer points to zero ?

ishan16696: there is a problem with using `availableReplicas or readyReplicas` when scale-up was successful but just after scale-up 1 pod goes to `CrashLoopBackOff`. Let me explain this:
When we marked the cluster for scale-up --> druid adds a sts annotation --> backup-restore detects scale-up and provides config to its corresponding etcd.
so, when other 2 pods starts running fine, 0th pod will always restarted in scale-up scenario,
so in this case **status** of sts etcd is
```
status:
  availableReplicas: 2
  collisionCount: 0
  currentRevision: etcd-main-55c8f5489c
  observedGeneration: 2
  readyReplicas: 2
  replicas: 3
  updateRevision: etcd-main-75989cd994
  updatedReplicas: 3
```
Although scale-up was successful and it is just a pod restart but as you can see if I include `availableReplicas or readyReplicas` then druid will still thinks scale-up is not yet complete and won't remove the annotation, hence wrong config will get provided by backup-restore to etcd.

ishan16696: I will try to simulate your test.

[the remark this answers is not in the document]
unmarshall: A value of zero is still a valid value and that is fine. I created an STS, scaled it down to 0 and the replicas were 0. Then i removed replicas from the spec completely and as soon as i saved it, the value of replicas became 1 as indicated in the doc string. So there is no reason to have a nil check

unmarshall: The scale-up annotation is currently used in 2 places in backup-restore.

- Initializer - here your approach will work.
- httpAPI.go which calls `GetInitialClusterStateIfScaleup` - for discussion i will put the code here:
```go
func GetInitialClusterStateIfScaleup(ctx context.Context, logger logrus.Entry, clientSet client.Client, podName string, podNS string) (*string, error) {
	// Read etcd statefulset to check annotation or updated replicas to toggle `initial-cluster-state`
	etcdSts, err := GetStatefulSet(ctx, clientSet, podNS, podName)
	if err != nil {
		logger.Errorf("unable to fetch statefulset {namespace: %s, name: %s} %v", podNS, podName[:strings.LastIndex(podName, "-")], err)
		return nil, err
	}

	if IsAnnotationPresent(etcdSts, ScaledToMultiNodeAnnotationKey) {
		return pointer.StringPtr(ClusterStateExisting), nil
	}

	if *etcdSts.Spec.Replicas > 1 && *etcdSts.Spec.Replicas > etcdSts.Status.UpdatedReplicas {
		return pointer.StringPtr(ClusterStateExisting), nil
	}
	return nil, nil
}
```
If the annotation is removed then it evaluated the second check which it also does not enter and thus returns nil.
In the caller we have the following code:
```go
if state == nil {
		// Not a Scale-up scenario.
		// Either a multi-node bootstrap or a restoration of single member in multi-node.
		m := member.NewMemberControl(h.EtcdConnectionConfig)

		// check whether a learner is present in the cluster
		// if a learner is present then return `ClusterStateExisting` else `ClusterStateNew`.
		if present, err := m.IsLearnerPresent(ctx); present && err == nil {
			return miscellaneous.ClusterStateExisting, nil
		}
		return miscellaneous.ClusterStateNew, nil

	}
```
If i read the code correctly then this will result in a ClusterStateNew instead of returning Existing, right?

ishan16696: yes

ishan16696: If PVC is intact then I guess etcd don't need the config as it will take the config from db, so may be we can include check of `availableReplicas` then , I will confirm this by trying to run scale-up

ishan16696: Unfortunately, it didn’t work, we can't use `availableReplicas `

ishan16696:
```go
if sts != nil && metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) {
		if sts.Status.UpdatedReplicas < *sts.Spec.Replicas || sts.Status.Replicas > sts.Status.UpdatedReplicas || sts.Status.AvailableReplicas < sts.Status.UpdatedReplicas {
			annotations[scaleToMultiNodeAnnotationKey] = ""
			return annotations
		}
	}
```
I made these changes but as expected druid e2e fails due to this condition `sts.Status.AvailableReplicas < sts.Status.UpdatedReplicas`

aaronfern (approved):
/lgtm

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -286,7 +286,8 @@ func immutableFieldUpdate(sts *appsv1.StatefulSet, val Values) bool {
 
 func clusterScaledUpToMultiNode(val *Values, sts *appsv1.StatefulSet) bool {
 	if sts != nil && sts.Spec.Replicas != nil {
-		return val.Replicas > 1 && *sts.Spec.Replicas == 1
+		return (val.Replicas > 1 && *sts.Spec.Replicas == 1) ||
```
unmarshall, line 289: can you enhance the unit tests to include all cases?

ishan16696: I don't think so there is an already existing unit tests for this.

unmarshall (approved):
/lgtm

Issue #589, opened 2023-05-03
[Feature] Harmonize scaling operations of the etcd cluster


shreyas-s-rao:
**Feature (What you would like to be added):**
Harmonize scaling operations of the etcd cluster through deterministic steps, controlled by etcd-druid.

**Motivation (Why is this needed?):**
The current approach to scaling out an etcd cluster from a single-member etcd to a 3-member etcd cluster is two-fold - it first requires the first replica to be healthy, or in other words to start with a healthy single-member etcd cluster. The second step is to scale the etcd statefulset to 3 replicas, ie, introducing two new etcd pods at once. The two new etcd pods compete with each other to add themselves as learners to the existing cluster. While one of them succeeds, the other continues retrying this operation, although the operation constantly fails because the etcd cluster only supports one learner in the cluster at any given point of time. The one that succeeds then proceeds to wait for its data to catch up with the leader of the etcd cluster, and once its data is in sync with the leader, it promotes itself a voting member of the cluster. Once this is complete, the other pod succeeds in adding itself as a learner and continues with the same procedure of syncing its data with the leader and promoting itself to a voting member.

This entire process relies on a race between the two new members, and has lead to multiple issues with scale-out scenarios of etcd clusters, all of which are detailed in https://github.com/gardener/etcd-druid/issues/584. The current approach also leads to many edge cases that need to be specially handled by either druid or backup-restore.

Druid needs to use a more deterministic approach to scale etcd clusters. This can be achieved by harmonizing the scaling out of the etcd cluster in deterministic steps or states, ie, 0 -> 1 -> 2 -> 3 -> 0. This ensures that druid is able to handle the scaling from either 0 or 1 replicas up to 3 replicas using similar steps/sub-operations, ie, no special handling of different cases is necessary. This provides two main advantages:
1. Makes the code leaner - deterministic steps to achieve well-defined states result in lesser edge cases and eliminate artificial races between different etcd members, thus drastically reducing the chances of failed or hanging scale-out operations
2. Moving between well-defined states allows druid to control the scale-out operation with precision. Since druid is the operator of the etcd cluster, it becomes druid's duty to handle cluster-wide operations such as scale-out, scale-in, etc.

[...]

PR #594, opened 2023-05-05
Fixes for etcd status fields

### LINKED ISSUE #557: [BUG]  Potential improvements in etcd resource status or  in Custom Columns of etcd resource
**Describe the bug:**
When Scale-up scenario fails in middle of scaling up and 1 etcd pod fails to join:
```
etcd-main-0                                   2/2     Running             0             85s
etcd-main-1                                   1/2     Running             0             84s
etcd-main-2                                   2/2     Running             0             83s
```

following etcd status:
```
  status:
    clusterSize: 2
    conditions:
    - lastTransitionTime: "2023-04-25T04:54:29Z"
      lastUpdateTime: "2023-04-25T05:21:42Z"
      message: All members are ready
      reason: AllMembersReady
      status: "True"
      type: AllMembersReady
    - lastTransitionTime: "2023-04-25T04:54:54Z"
      lastUpdateTime: "2023-04-25T05:21:42Z"
      message: Snapshot backup succeeded
      reason: BackupSucceeded
      status: "True"
      type: BackupReady
    - lastTransitionTime: "2023-04-25T04:54:34Z"
      lastUpdateTime: "2023-04-25T05:21:42Z"
      message: The majority of ETCD members is ready
      reason: Quorate
      status: "True"
      type: Ready
    currentReplicas: 1
    etcd:
      apiVersion: apps/v1
      kind: StatefulSet
      name: etcd-main
    members:
    - id: a789d1ab8da2e800
      lastTransitionTime: "2023-04-25T04:54:34Z"
      name: etcd-main-0
      reason: LeaseSucceeded
      role: Leader
      status: Ready
    - id: bda90f379184787e
      lastTransitionTime: "2023-04-25T05:20:52Z"
      name: etcd-main-1
      reason: LeaseSucceeded
      role: Member
      status: Ready
    observedGeneration: 1
    ready: false
    readyReplicas: 2
    replicas: 1
    serviceName: etcd-main-client
    updatedReplicas: 2
```

I observed that etcd `.status.clusterSize: 2` as reason could be scale-up succeeded till `1->2` replicas only and that's why it also sets `AllMembersReady status` as `true` but in `k get etcd -owide` `CLUSTER SIZE` is shown as `3` which IMO is misleading as `ALL MEMBERS READY` column is also set to `True`
```
k get etcd -owide                                                                                                                             
NAME          READY   QUORATE   ALL MEMBERS READY   BACKUP READY   AGE   CLUSTER SIZE   CURRENT REPLICAS   READY REPLICAS
etcd-main     false   True      True                True           3d    3                                 2
```

**Expected behavior:**
Etcd resource status `AllMembersReady` shouldn’t be `true` as 3rd members haven’t join the cluster.

 - Etcd version/commit ID :

unmarshall:

/area control-plane
/kind bug
/kind cleanup

Status.ClusterSize is now marked as deprecated and will not be used hereafter. Logic to compute `AllMembersReady` condition has been fixed to eventually report the correct status. 
> NOTE: Further improvements are required to rename this to `AllMemberLeasesActive` instead as registration and periodic renewal of a lease is no indicator to if etcd is ready and is therefore quite misleading. We already have `etcd.Status.Ready` to correctly indicate readiness.

Fixes #557 

**Release note**:

```other operator
:warning: `etcd.Status.ClusterSize`, `etcd.Status.ServiceName`, `etcd.Status.UpdatedReplicas` have been marked as deprecated and users should refrain from depending on these fields. `AllMemberReady` condition has now been fixed to eventually show the correct overall readiness of an etcd cluster.
```


ishan16696:
Hi @unmarshall ,
Can you please resolve the merge conflicts.

controllers/etcd/reconciler.go
```diff
@@ -400,12 +400,10 @@ func isPeerTLSChangedToEnabled(peerTLSEnabledStatusFromMembers bool, configMapVa
 
 func bootstrapReset(etcd *druidv1alpha1.Etcd) {
 	etcd.Status.Members = nil
-	etcd.Status.ClusterSize = pointer.Int32(etcd.Spec.Replicas)
 }
 
 func clusterInBootstrap(etcd *druidv1alpha1.Etcd) bool {
-	return etcd.Status.Replicas == 0 ||
-		(etcd.Spec.Replicas > 1 && etcd.Status.Replicas == 1)
+	return etcd.Spec.Replicas >= 1 && etcd.Status.Replicas == 0
```
ishan16696, line 406: `func clusterInBootstrap(){...}` should return true if it's the bootstrap case else false. But I found one scenario, in that scenario it is returning false.
Let's take the scenario:
Suppose we are bootstrapping a HA cluster i.e 0->3 and only 1 etcd pod come-up and 2 fails to join may be due to volume attachment issue.
`etcd.Spec.Replicas` is 3 and as one members joins then `etcd.Status.Replicas` should become `1` then your condition will returns false but it is still a bootstrap case.

api/v1alpha1/types_etcd.go
```diff
@@ -397,10 +400,12 @@ type EtcdStatus struct {
 	// +optional
 	Ready *bool `json:"ready,omitempty"`
 	// UpdatedReplicas is the count of updated replicas in the etcd cluster.
+	// Deprecated: this field will be removed in the future.
```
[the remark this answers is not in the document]
unmarshall: There is no need for it because we always get the value of updated replicas from STS.

api/v1alpha1/types_etcd.go
```diff
@@ -376,15 +376,18 @@ type EtcdStatus struct {
...
 	Conditions []Condition `json:"conditions,omitempty"`
 	// ServiceName is the name of the etcd service.
+	// Deprecated: this field will be removed in the future.
 	// +optional
 	ServiceName *string `json:"serviceName,omitempty"`
 	// LastError represents the last occurred error.
 	// +optional
 	LastError *string `json:"lastError,omitempty"`
-	// Cluster size is the size of the etcd cluster.
+	// Cluster size is the current size of the etcd cluster.
```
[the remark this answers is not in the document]
unmarshall: ClusterSize is always set today to replicas. This means  if you change the replicas in the spec to 3 then ClusterSize will always be 3. Also if you want to know how many replicas are there then you can find that out from `CurrentReplicas`. If you want to know how many members are ready please use `ReadyReplicas`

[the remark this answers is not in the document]
ishan16696: > ClusterSize is always set today to replicas. This means if you change the replicas in the spec to 3 then ClusterSize will always be 3. 

If you look this issue: https://github.com/gardener/etcd-druid/issues/557 which I opened it 1 members fails to join cluster during scale-up. I found `etcd.Status.clusterSize` was `2` not equal to `etcd.spec.replicas`  as only two members joined the cluster.

> Also if you want to know how many replicas are there then you can find that out from CurrentReplicas

you mentioned a comment saying we will deprecate this filed `etcd.Status.CurrentReplicas`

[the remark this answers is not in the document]
ishan16696: IMO, either `etcd.Status.CurrentReplicas` or `etcd.Status.ClusterSize` one shouldn't be deprecated.

pkg/health/condition/check_all_members.go
```diff
@@ -35,19 +35,26 @@ func (a *allMembersReady) Check(_ context.Context, etcd druidv1alpha1.Etcd) Resu
...
 
-	for _, member := range etcd.Status.Members {
-		if member.Status != druidv1alpha1.EtcdMemberStatusReady {
-			result.status = druidv1alpha1.ConditionFalse
-			result.reason = "NotAllMembersReady"
-			result.message = "At least one member is not ready"
+	notAllMembersRegistered := int32(len(etcd.Status.Members)) < etcd.Spec.Replicas
+	if notAllMembersRegistered {
+		return result
+	}
```
shreyas-s-rao, line 46:
```suggestion
	if int32(len(etcd.Status.Members)) < etcd.Spec.Replicas {
	        // not all members are registered yet
		return result
	}
```

pkg/health/condition/check_all_members.go
```diff
@@ -35,19 +35,26 @@ func (a *allMembersReady) Check(_ context.Context, etcd druidv1alpha1.Etcd) Resu
...
 
-			return result
-		}
+	// If we are here this means that all members have registered. Check if any member
+	// has a not-ready status. If there is at least one then set the overall status as false.
+	ready := true
+	for _, member := range etcd.Status.Members {
+		ready = ready && member.Status == druidv1alpha1.EtcdMemberStatusReady
```
shreyas-s-rao, line 52:
```suggestion
		ready = ready && member.Status == druidv1alpha1.EtcdMemberStatusReady
		if !ready {
		        break
		}
```

shreyas-s-rao (changes requested):
@unmarshall thanks for the PR. I've left couple of suggestions. PTAL.

I have also taken the liberty to slightly edit the release note in the PR description.

pkg/health/condition/check_all_members.go
unmarshall: only did this for better readability. But if you think this is not necessary and the condition is self-evident then i can accept the change.

controllers/etcd/reconciler.go
shreyas-s-rao: But the current logic will return true even during cluster scale-up from 1 to 3, right? @unmarshall does it make sense to use `return etcd.Spec.Replicas >= 1 && etcd.Status.ReadyReplicas == 0` instead, because the case that Ishan mentioned is quite possible, so even during bootstrap from 0->3, when 1 replica is up and two others are still coming up, the replica will not be ready, hence `ReadyReplicas` will still be `0`, until the cluster is fully formed and ready.

api/v1alpha1/types_etcd.go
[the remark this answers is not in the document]
unmarshall: Removed the deprecation notice from CurrentReplicas.

controllers/etcd/reconciler.go
unmarshall: Consider another case:
Initial state: 1 member (healthy)
Scale up is triggered from 1 -> 3. Now 2 additional pods come up (these are not yet ready).
Now the first pod goes down and post re-start it is unable to come up. So now the `etcd.Spec.Repliacs >=1` and `etcd.Status.ReadyReplicas == 0` however this is a scale up case.

pkg/health/condition/check_all_members.go
shreyas-s-rao: Condition seems self-evident, and extra variable just increases memory usage. Just a small code comment should be fine. If you wish, you can also introduce a new condition reason like:
```
result.reason = "NotAllMembersRegistered"
result.message = "At least one member is not registered"
```

But maybe this is unnecessary.

controllers/etcd/reconciler.go
shreyas-s-rao: Looks like this particular problem might be solved only if druid knows which of the etcd members have already been initialized and part of the cluster. Maybe then we might need another field such as `JoinedReplicas` that holds information about the replicas that have already joined the cluster, and this can be as simple as a string array with non-repeating pod names. Essentially, druid, ie custodian controller, needs to watch for pod readiness of the etcd pods, and as soon as any pod gets ready, it must add it to this particular list, hence maintaining information about which pods have already joined the cluster. So druid etcd controller will simply consult this list when deciding whether it's a scale-up case or bootstrapping case. But again, we don't know how this plays out with quorum loss scenarios.

Anyway, all this complexity is completely avoided when we move to harmonized scaling, because then we would never have a bootstrap case in the first place, so all this complex implementation would be redundant.

unmarshall: Post our discussion today amongst (Shreyas, Aaron, Ishan, Abhishek and me) we think that it might be wise to just remove the `clusterInBootstrap` check.
Reason:
There is no fool proof way to determine if the cluster is in bootstrap. Today the cluster can be in bootstrap in 2 possible cases
- It is a fresh cluster. In this case initially there will not be any STS till etcd-controller reconciles.
- Cluster is woken up from hibernation - in this case STS exists, its replicas are scaled from 0-> <etcd.Spec.Replicas>

Also the only action taken if the cluster is in bootstrap is that it sets `etcd.Status.Members` to nil. We think that this can be removed since custodian controller periodically will recalculate the list of registered members and replaces the existing slice of Members with a new slice. Therefore eventually consistency will be there.

I will wait for a day and if no one has any further objections then i will remove `clusterInBootstrap` and `bootstrapReset` functions.

shreyas-s-rao: After giving it some thought, and I would vote to have `clusterInBootstrap` and `bootstrapReset` removed, simply because these will anyway go away when #589 comes in, and adding any new fields to Etcd status would simply be redundant because those will no longer be needed later and will have to be eventually removed.

unmarshall: I have made the changes as discussed. `clusterInBootstrap` and `bootstrapReset` methods and their usages have been now removed.

shreyas-s-rao (approved):
/lgtm

api/v1alpha1/types_etcd.go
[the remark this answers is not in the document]
ishan16696: Please also update the release note

ishan16696 (approved):
Please update the release note before merging the PR as mentioned in the comment: https://github.com/gardener/etcd-druid/pull/594#discussion_r1241793625

api/v1alpha1/types_etcd.go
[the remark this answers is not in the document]
unmarshall: Done

PR #598, opened 2023-05-16
Fixes scaleup from single-node to multi-node etcd cluster

### LINKED ISSUE #597: [BUG] If peerUrl TLS not enabled for non-HA migrate to HA then druid is recreates the statefulset as well as adds a scale-up annotation
**Describe the bug:**
If non-HA cluster has peerUrl TLS not enabled and it is marked to migrate to HA then `.spec.replicas` is set to `3` and druid starts recreating the statefulset to move from `http` to `https` as well as adding a scale-up annotation `gardener.cloud/scaled-to-multi-node`  to etcd statefulset together.

**Expected behavior:**
It shouldn't add scale-up annotation until first etcd pod becomes `ready/available`. 

1. create a non-HA shoot
2. migrate from non-HA to HA
3. check the etcd statefulset yaml annotation `gardener.cloud/scaled-to-multi-node` 
4. Watch etcd statefulset as it will get recreated.

 - Etcd version/commit ID :

unmarshall:

/area control-plane
/area high-availability
/kind bug

In gardener context, peer URL TLS is disabled for all non-HA etcd clusters. When a single-node etcd cluster is scaled to multi-node cluster, peer URLs are TLS enabled. Druid already has code to handle this but there were a few issues with the earlier implementation (in context to scale-up). This PR fixes these issues.

Fixes #597 
Details of Issues that it fixes:
- A peer URL change requires either 2 restarts of etcd pods or 1 restart of etcd pods + 1 etcd member API to call to update the peer URL. In the earlier solution we sought to restart etcd pods twice and to achieve that we simply deleted STS twice. This has several issues. 
   -  Deletion and re-creation of STS requires volume detach and attach. On platforms like Azure this could be a problem.
   - Partial failure in deleting STS twice results in re-queuing of the event eventually resulting in a bootstrap from 0->3 instead of a scale-up from 1->3 as the STS has been deleted (at least once) and no update to replicas has yet been done. This causes issues in starting the cluster as etcd members do not join.

**Release note**:

```other operator
When scaling from single-node to multi-node etcd cluster, Etcd Druid will now first ensure that any change to the peer URL (e.g TLS enablement)  is seen by the existing etcd process running within the etcd member pod. Once that is confirmed then it will scale up the Etcd StatefulSet and add relevant annotations.
```

### DIFF
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -24,6 +24,7 @@ import (
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 	"github.com/gardener/etcd-druid/pkg/common"
 	"github.com/gardener/etcd-druid/pkg/utils"
+	"github.com/gardener/gardener/pkg/controllerutils"
 	gardenercomponent "github.com/gardener/gardener/pkg/operation/botanist/component"
 	"github.com/gardener/gardener/pkg/utils/flow"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
@@ -53,7 +54,7 @@ type component struct {
 	values Values
 }
 
-// New creates a new statefulset deployer instance.
+// New creates a new StatefulSet deployer instance.
 func New(c client.Client, logger logr.Logger, values Values) Interface {
 	objectLogger := logger.WithValues("sts", client.ObjectKey{Name: values.Name, Namespace: values.Namespace})
 
@@ -101,8 +102,11 @@ const (
 // Wait waits for the deployment of the StatefulSet to finish
 func (c *component) Wait(ctx context.Context) error {
 	sts := c.emptyStatefulset()
-	err := c.waitDeploy(ctx, sts, c.values.Replicas)
+	err := c.waitDeploy(ctx, sts, c.values.Replicas, defaultTimeout)
 	if err != nil {
+		if getErr := c.client.Get(ctx, client.ObjectKeyFromObject(sts), sts); getErr != nil {
+			return err
+		}
 		messages, err2 := c.fetchPVCEventsFor(ctx, sts)
 		if err2 != nil {
 			c.logger.Error(err2, "Error while fetching events for depending PVC")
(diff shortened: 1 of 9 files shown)


ishan16696:
Please update the release note to something like this:
```
etcd controller will now wait for first etcd member to get `ready/available` before adding scale-up annotation to etcd statefulset while migrating from non-HA to HA which internally enables the TLS for peerURL.
```

Please also mention the relevant document of etcd.
you can also mention about the long-term solution as TODO: https://github.com/gardener/etcd-backup-restore/issues/538

ishan16696 (commented):
I had a initial look, overall PR looks good.
few nits from my side. I will review again after you fix integration tests.
I will also run druid e2e tests as well

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -118,15 +119,56 @@ func (c *component) Wait(ctx context.Context) error {
...
 			if apierrors.IsNotFound(err) {
 				return gardenerretry.MinorError(err)
 			}
 			return gardenerretry.SevereError(err)
 		}
-		ready, reason := utils.IsStatefulSetReady(replicas, sts)
+		if updatedSts.Generation < originalSts.Generation {
+			return gardenerretry.MinorError(fmt.Errorf("StatulfulSet generation has not yet been updated in the cache"))
+		}
+		ready, reason := utils.IsStatefulSetReady(replicas, &updatedSts)
```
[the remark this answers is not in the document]
unmarshall: done

unmarshall:
Ran e2e tests with `LocalStack`. It passed results are attached.
[druid-e2e-tests-with-localstack.log](https://github.com/gardener/etcd-druid/files/11486576/druid-e2e-tests-with-localstack.log)

unmarshall:
@ishan16696 I have addressed all the review comments. PTAL

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -286,117 +362,125 @@ func immutableFieldUpdate(sts *appsv1.StatefulSet, val Values) bool {
...
+		}
+		return nil
 	}
 
-	return c.client.Create(ctx, sts)
+	operationResult, err := controllerutils.GetAndCreateOrStrategicMergePatch(ctx, c.client, sts, mutatingFn)
+	if err != nil {
+		return err
+	}
+	c.logger.Info("createOrPath is completed", "Namespace", sts.Namespace, "Name", sts.Name, "operation-result", operationResult)
```
seshachalam-yv, line 482:
```suggestion
	c.logger.Info("createOrPatch is completed", "Namespace", sts.Namespace, "Name", sts.Name, "operation-result", operationResult)
```

seshachalam-yv (commented):
Thanks for Nice PR ❤️ 
/lgtm

unmarshall:
>Please update all of this with current approach of deletion of pod instead recreation of statefulset:

I have updated the doc string. I have not given the link to the issue in backup-restore as druid is not the place to put it.

ishan16696 (approved):

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -118,16 +122,56 @@ func (c *component) Wait(ctx context.Context) error {
 	return err
 }
 
-func (c *component) waitDeploy(ctx context.Context, sts *appsv1.StatefulSet, replicas int32) error {
-	return gardenerretry.UntilTimeout(ctx, defaultInterval, defaultTimeout, func(ctx context.Context) (bool, error) {
-		if err := c.client.Get(ctx, client.ObjectKeyFromObject(sts), sts); err != nil {
+func (c *component) getMostRecentPodCreationTimestamp(ctx context.Context, sts *appsv1.StatefulSet) (time.Time, error) {
```
abdasgupta, line 125: Can we rename the function as `getLatestPodCreationTime` ?

unmarshall: done

ishan16696:
As we found out that g/g e2e tests fails with this PR, so we can't merge this PR until g/g e2e tests passed.
/do-not-merge

ishan16696 (approved):

PR #617, opened 2023-06-15
Optimize test runs by running lint check once

seshachalam-yv:

  /area testing dev-productivity
  /area test

This PR optimizes the way we run our tests by adjusting the `Makefile` so that the lint check is only run once per pull request, instead of with each `test` and `test-integration` execution. The lint check is already handled by a separate job in our CI/CD pipeline job `check`.

### Changes
- Removed `check` from the dependencies of `test` and `test-integration` in the `Makefile`.
- Added `fmt` and `manifests` as dependencies to `check` in the `Makefile`.

Fixes #

cc: @acumino 
**Release note**:

```other operator
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ clean: set-permissions
 
 # Check packages
 .PHONY: check
-check: $(GOLANGCI_LINT) $(GOIMPORTS) set-permissions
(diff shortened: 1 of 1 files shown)


acumino (approved):
/lgtm

PR #619, opened 2023-06-19
Enhance backup ready conditions

### LINKED ISSUE #618: [Enhancement] Improve BackupReady conditions
Enhance BackupReady conditions to provide more meaningful messages and reasons for setting different BackupReady conditions.

Currently not all combinations of full and delta snapshot lease renewal times are handled by custodian controller.

Add additional field state to `EtcdMember.status.snapshots.lastFull` and `EtcdMember.status.snapshots.lastDelta`, in order to allow the etcd member to correctly publish snapshot state information, which druid will use to accurately populate the BackupReady condition, as described in https://github.com/gardener/etcd-druid/pull/619#issuecomment-1831876641

shreyas-s-rao:

/area backup
/area usability
/kind enhancement

- [x] Enhance backup ready conditions to provide more fine-grained condition status messages and reasons based on different states of snapshot leases
  - [x] Set `BackupReady` condition reason to `BackupFailed` no longer depends on previous condition (previously, setting this depended on whether previous condition was failed or unknown, which meant that if previous condition was succeeded, we would never set condition to failed unless either of the leases were recreated. This behavior is now fixed and made fully deterministic)
  - [x] Fine-grained condition messages for operators to infer whether full, delta or both snapshot leases have problems with renewal
  - [x] Improve readability of handling of different cases when setting `BackupReady` condition
  - [x] Remove hardcoding of `24h` for calculating staleness of full snapshot lease, by computing the "schedule duration" (duration between two activations of the cron, assuming activations are equal durations apart) from full snapshot cron schedule

Fixes #618 

**Release note**:

```other operator
Enhance `BackupReady` conditions to allow for more fine-grained condition states, messages and reasons.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ install: manifests
 	kubectl apply -f config/crd/bases
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-.PHONY: deploy
+.PHONY: deploy-via-kustomize
 deploy-via-kustomize: manifests $(KUSTOMIZE)
 	kubectl apply -f config/crd/bases
 	kustomize build config/default | kubectl apply -f -
--- a/pkg/health/condition/builder.go
+++ b/pkg/health/condition/builder.go
@@ -23,7 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// skipMergeConditions contain the list of conditions we dont want to add to the list if not recalculated
+// skipMergeConditions contain the list of conditions we don't want to add to the list if not recalculated
 var skipMergeConditions = map[druidv1alpha1.ConditionType]struct{}{
 	druidv1alpha1.ConditionTypeReady:           {},
 	druidv1alpha1.ConditionTypeAllMembersReady: {},
@@ -111,7 +111,7 @@ func (b *defaultBuilder) Build(replicas int32) []druidv1alpha1.Condition {
 			if condition.Status == "" {
 				condition.Status = druidv1alpha1.ConditionUnknown
 			}
-			condition.Reason = NotChecked
+			condition.Reason = "CLusterScaledToZero"
 			condition.Message = "etcd cluster has been scaled down"
 		} else {
 			condition.Status = res.Status()
--- a/pkg/health/condition/builder.go
+++ b/pkg/health/condition/builder.go
@@ -111,7 +111,7 @@ func (b *defaultBuilder) Build(replicas int32) []druidv1alpha1.Condition {
 			if condition.Status == "" {
 				condition.Status = druidv1alpha1.ConditionUnknown
 			}
-			condition.Reason = ConditionNotChecked
+			condition.Reason = NotChecked
 			condition.Message = "etcd cluster has been scaled down"
 		} else {
 			condition.Status = res.Status()
--- a/pkg/health/condition/check_backup_ready.go
+++ b/pkg/health/condition/check_backup_ready.go
@@ -43,14 +43,6 @@ const (
 )
 
 func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
-	// Default case
-	result := &result{
-		conType: druidv1alpha1.ConditionTypeBackupReady,
-		status:  druidv1alpha1.ConditionUnknown,
-		reason:  Unknown,
-		message: "Cannot determine etcd backup status",
-	}
-
 	// Special case of etcd not being configured to take snapshots
 	// Do not add the BackupReady condition if backup is not configured
 	if etcd.Spec.Backup.Store == nil || etcd.Spec.Backup.Store.Provider == nil || len(*etcd.Spec.Backup.Store.Provider) == 0 {
@@ -58,120 +50,172 @@ func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) R
 	}
 
 	// Fetch snapshot leases
-	var (
-		err            error
-		fullSnapLease  = &coordinationv1.Lease{}
-		deltaSnapLease = &coordinationv1.Lease{}
+	fullSnapshotLease, err := a.fetchLease(ctx, etcd.GetFullSnapshotLeaseName(), etcd.Namespace)
+	if err != nil {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionUnknown, Unknown,
+			fmt.Sprintf("Unable to fetch full snap lease. %s", err.Error()),
+		)
+	}
+
+	deltaSnapSnapshotLease, err := a.fetchLease(ctx, etcd.GetDeltaSnapshotLeaseName(), etcd.Namespace)
+	if err != nil {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionUnknown, Unknown,
+			fmt.Sprintf("Unable to fetch delta snap lease. %s", err.Error()),
+		)
+	}
+
+	deltaSnapshotLeaseRenewTime := deltaSnapSnapshotLease.Spec.RenewTime
+	fullSnapshotLeaseRenewTime := fullSnapshotLease.Spec.RenewTime
+	fullSnapshotLeaseCreationTime := &fullSnapshotLease.ObjectMeta.CreationTimestamp
+	fullSnapshotDuration, err := utils.ComputeScheduleDuration(*etcd.Spec.Backup.FullSnapshotSchedule)
+	if err != nil {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionUnknown, Unknown,
+			fmt.Sprintf("Unable to compute full snapshot duration from schedule. %v", err.Error()),
+		)
+	}
+
+	// Both snapshot leases are not yet renewed
+	if fullSnapshotLeaseRenewTime == nil && deltaSnapshotLeaseRenewTime == nil {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionUnknown, Unknown,
+			"Snapshotter has not started yet",
+		)
+	}
+
+	// Most probable during reconcile of existing clusters if fresh leases are created
+	if fullSnapshotLeaseRenewTime == nil && deltaSnapshotLeaseRenewTime != nil {
+		return handleOnlyDeltaSnapshotLeaseRenewedCase(
+			fullSnapshotLeaseCreationTime.Time, fullSnapshotDuration,
+			deltaSnapshotLeaseRenewTime.Time, 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration,
+		)
+	}
+
+	// Most probable during a startup of a new cluster and only full snapshot has been taken
+	if fullSnapshotLeaseRenewTime != nil && deltaSnapshotLeaseRenewTime == nil {
+		return handleSnapshotterStartupCase(
+			fullSnapshotLease.Spec.RenewTime.Time,
+			5*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration,
+		)
+	}
+
+	// Both snap leases are renewed, ie, maintained. Both are expected to be renewed periodically
+	return handleRenewedSnapshotLeases(
+		fullSnapshotLease.Spec.RenewTime.Time, fullSnapshotDuration,
+		deltaSnapSnapshotLease.Spec.RenewTime.Time, 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration,
 	)
+}
 
-	err = a.cl.Get(ctx, types.NamespacedName{Name: getFullSnapLeaseName(&etcd), Namespace: etcd.ObjectMeta.Namespace}, fullSnapLease)
-	if err != nil {
-		result.message = fmt.Sprintf("Unable to fetch full snap lease. %s", err.Error())
-		return result
+func (a *backupReadyCheck) fetchLease(ctx context.Context, name string, namespace string) (*coordinationv1.Lease, error) {
+	lease := &coordinationv1.Lease{}
+	err := a.cl.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, lease)
+	return lease, err
+}
+
+func createBackupConditionResult(status druidv1alpha1.ConditionStatus, reason string, message string) *result {
+	return &result{
+		conType: druidv1alpha1.ConditionTypeBackupReady,
+		status:  status,
+		reason:  reason,
+		message: message,
 	}
+}
 
-	err = a.cl.Get(ctx, types.NamespacedName{Name: getDeltaSnapLeaseName(&etcd), Namespace: etcd.ObjectMeta.Namespace}, deltaSnapLease)
-	if err != nil {
-		result.message = fmt.Sprintf("Unable to fetch delta snap lease. %s", err.Error())
-		return result
-	}
+func isLeaseStale(renewTime time.Time, renewalGracePeriod time.Duration) bool {
+	return time.Since(renewTime) > renewalGracePeriod
+}
 
-	deltaLeaseRenewTime := deltaSnapLease.Spec.RenewTime
-	fullLeaseRenewTime := fullSnapLease.Spec.RenewTime
-	fullLeaseCreateTime := &fullSnapLease.ObjectMeta.CreationTimestamp
+func wasLeaseCreatedRecently(creationTime time.Time, creationGracePeriod time.Duration) bool {
+	return time.Since(creationTime) < creationGracePeriod
+}
 
-	if fullLeaseRenewTime == nil && deltaLeaseRenewTime == nil {
-		// Both snapshot leases are not yet renewed
-		result.message = "Snapshotter has not started yet"
-		return result
+// handleOnlyDeltaSnapshotLeaseRenewedCase handles cases where only delta snapshot lease is renewed,
+// while the full snapshot lease is not renewed, possibly since it might have been recreated recently.
+// Treats backup as succeeded if delta snapshot lease is renewed within the required time window
+// and full snapshot lease object is not older than the computed full snapshot duration.
+func handleOnlyDeltaSnapshotLeaseRenewedCase(fullSnapshotLeaseCreationTime time.Time, fullSnapshotLeaseCreationGracePeriod time.Duration, deltaSnapshotLeaseRenewTime time.Time, deltaSnapshotLeaseRenewalGracePeriod time.Duration) *result {
+	wasFullSnapshotLeaseCreatedRecently := wasLeaseCreatedRecently(fullSnapshotLeaseCreationTime, fullSnapshotLeaseCreationGracePeriod)
+	isDeltaSnapshotLeaseStale := isLeaseStale(deltaSnapshotLeaseRenewTime, deltaSnapshotLeaseRenewalGracePeriod)
 
-	} else if fullLeaseRenewTime == nil && deltaLeaseRenewTime != nil {
-		// Most probable during reconcile of existing clusters if fresh leases are created
-
-		fullSnapshotDuration, err := utils.ComputeScheduleDuration(*etcd.Spec.Backup.FullSnapshotSchedule)
-		if err != nil {
-			return result
-		}
-
-		// Treat backup as succeeded if delta snap lease renewal happens in the required time window
-		// and full snap lease is not older than the computed full snapshot duration.
-		if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
-			if time.Since(fullLeaseCreateTime.Time) < fullSnapshotDuration {
-				result.reason = BackupSucceeded
-				result.message = "Delta snapshot backup succeeded"
-				result.status = druidv1alpha1.ConditionTrue
-				return result
-			} else {
-				result.status = druidv1alpha1.ConditionFalse
-				result.reason = BackupFailed
-				result.message = "Full snapshot backup failed. Full snapshot lease created long ago, but not renewed"
-				return result
-			}
-		}
-
-	} else if fullLeaseRenewTime != nil && deltaLeaseRenewTime == nil {
-		// Most probable during a startup scenario for new clusters
-		// Special case: return Unknown condition for some time to allow delta backups to start up
-		// Even though full snapshot may have succeeded by the required time, we must still wait
-		// for delta snapshotting to begin to consider the backups as healthy, to maintain the given RPO
-		if time.Since(fullLeaseRenewTime.Time) < 5*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
-			result.message = "Waiting for periodic delta snapshotting to begin"
-			return result
+	if isDeltaSnapshotLeaseStale {
+		if wasFullSnapshotLeaseCreatedRecently {
+			return createBackupConditionResult(
+				druidv1alpha1.ConditionFalse, BackupFailed,
+				"Delta snapshot backup failed. Delta snapshot lease not renewed in a long time",
+			)
 		} else {
-			result.status = druidv1alpha1.ConditionFalse
-			result.reason = BackupFailed
-			result.message = "Delta snapshot backup failed. Delta snapshot lease created long ago, but not renewed"
-			return result
-		}
-
-	} else {
-		// Both snap leases are renewed, ie, maintained. Both are expected to be renewed periodically
-		fullSnapshotDuration, err := utils.ComputeScheduleDuration(*etcd.Spec.Backup.FullSnapshotSchedule)
-		if err != nil {
-			return result
-		}
-
-		// Mark backup as succeeded only if at least one of the last two delta snapshots has been taken successfully
-		// and if the last full snapshot has been taken according to the given full snapshot schedule
-		if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration && time.Since(fullLeaseRenewTime.Time) < fullSnapshotDuration {
-			result.reason = BackupSucceeded
-			result.message = "Snapshot backup succeeded"
-			result.status = druidv1alpha1.ConditionTrue
-			return result
+			return createBackupConditionResult(
+				druidv1alpha1.ConditionFalse, BackupFailed,
+				"Stale snapshot leases. Not renewed in a long time",
+			)
 		}
 	}
 
-	// Cases where snapshot leases are not updated for a long time
-	// If snapshot leases are present and leases aren't updated, it is safe to assume that backup is not healthy
-	if etcd.Status.Conditions != nil {
-		var prevBackupReadyStatus druidv1alpha1.Condition
-		for _, prevBackupReadyStatus = range etcd.Status.Conditions {
-			if prevBackupReadyStatus.Type == druidv1alpha1.ConditionTypeBackupReady {
-				break
-			}
-		}
-
-		// Transition to "False" state only if present state is "Unknown" or "False"
-		if deltaLeaseRenewTime != nil && (prevBackupReadyStatus.Status == druidv1alpha1.ConditionUnknown || prevBackupReadyStatus.Status == druidv1alpha1.ConditionFalse) {
-			if time.Since(deltaLeaseRenewTime.Time) > 3*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
-				result.status = druidv1alpha1.ConditionFalse
-				result.reason = BackupFailed
-				result.message = "Stale snapshot leases. Not renewed in a long time"
-				return result
-			}
-		}
+	if !wasFullSnapshotLeaseCreatedRecently {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionFalse, BackupFailed,
+			"Full snapshot backup failed. Full snapshot lease created long ago, but not renewed",
+		)
 	}
 
-	// Transition to "Unknown" state is we cannot prove a "True" state
-	return result
+	return createBackupConditionResult(
+		druidv1alpha1.ConditionTrue, BackupSucceeded,
+		"Delta snapshot backup succeeded",
+	)
 }
 
-func getDeltaSnapLeaseName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("%s-delta-snap", etcd.ObjectMeta.Name)
+// handleSnapshotterStartupCase handles cases where snapshotter has just started,
+// so only full snapshot has been taken, while delta snapshot is still not taken.
+// Returns `Unknown` condition for some time to allow delta snapshotting to begin, because
+// even though the full snapshot may have succeeded within the required time, we must still wait
+// for delta snapshotting to begin to consider the backups as healthy, to maintain the given RPO.
+func handleSnapshotterStartupCase(fullSnapshotLeaseRenewTime time.Time, deltaSnapshotRenewalGracePeriod time.Duration) *result {
+	if time.Since(fullSnapshotLeaseRenewTime) > deltaSnapshotRenewalGracePeriod {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionFalse, BackupFailed,
+			"Delta snapshot backup failed. Delta snapshot lease not renewed in a long time",
+		)
+	}
+
+	return createBackupConditionResult(
+		druidv1alpha1.ConditionUnknown, Unknown,
+		"Waiting for delta snapshotting to begin",
+	)
 }
 
-func getFullSnapLeaseName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("%s-full-snap", etcd.ObjectMeta.Name)
+// handleRenewedSnapshotLeases checks whether full and delta snapshot leases
+// have been renewed within the required times respectively.
+func handleRenewedSnapshotLeases(fullSnapshotLeaseRenewTime time.Time, fullSnapshotLeaseRenewalGracePeriod time.Duration, deltaSnapshotLeaseRenewTime time.Time, deltaSnapshotLeaseRenewalGracePeriod time.Duration) *result {
+	isFullSnapshotLeaseStale := isLeaseStale(fullSnapshotLeaseRenewTime, fullSnapshotLeaseRenewalGracePeriod)
+	isDeltaSnapshotLeaseStale := isLeaseStale(deltaSnapshotLeaseRenewTime, deltaSnapshotLeaseRenewalGracePeriod)
+
+	if isFullSnapshotLeaseStale && !isDeltaSnapshotLeaseStale {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionFalse, BackupFailed,
+			"Stale full snapshot lease. Not renewed in a long time",
+		)
+	}
+
+	if !isFullSnapshotLeaseStale && isDeltaSnapshotLeaseStale {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionFalse, BackupFailed,
+			"Stale delta snapshot lease. Not renewed in a long time",
+		)
+	}
+
+	if isFullSnapshotLeaseStale && isDeltaSnapshotLeaseStale {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionFalse, BackupFailed,
+			"Stale snapshot leases. Not renewed in a long time",
+		)
+	}
+
+	return createBackupConditionResult(
+		druidv1alpha1.ConditionTrue, BackupSucceeded,
+		"Snapshot backup succeeded",
+	)
 }
 
 // BackupReadyCheck returns a check for the "BackupReady" condition.
--- a/pkg/health/condition/check_backup_ready.go
+++ b/pkg/health/condition/check_backup_ready.go
@@ -38,8 +38,6 @@ const (
 	BackupFailed string = "BackupFailed"
 	// Unknown is a constant that means that the etcd backup status is currently not known
 	Unknown string = "Unknown"
-	// NotChecked is a constant that means that the etcd backup status has not been updated or rechecked
-	NotChecked string = "ConditionNotChecked"
 )
 
 func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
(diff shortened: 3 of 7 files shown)


pkg/health/condition/check_backup_ready.go
```diff
@@ -56,53 +57,92 @@ func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) R
...
+		if err != nil {
 			return result
 		}
-	} else if deltaLeaseRenewTime != nil && fullLeaseRenewTime != nil {
-		//Both snap leases are maintained. Both are expected to be renewed periodically
-		if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration && time.Since(fullLeaseRenewTime.Time) < 24*time.Hour {
+
+		// Mark backup as succeeded only if at least one of the last two delta snapshots has been taken successfully
+		// and if the last full snapshot has been taken according to the given full snapshot schedule
+		if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration && time.Since(fullLeaseRenewTime.Time) < fullSnapshotDuration {
```
seshachalam-yv, line 136: @shreyas-s-rao ,

**In my opinion**:

- The present conditions seem to complicate the readability of the code. I propose simplifying these as in the below code snippet.
- Additionally, it appears we can leverage  `etcd.GetFullSnapshotLeaseName()` and `etcd.GetDeltaSnapshotLeaseName()` in place of `getFullSnapLeaseName` and `getDeltaSnapLeaseName`respectively. This change would allow us to remove these redundant functions.

What are your thoughts on this?

<details> <summary> <b>Please find the actual diff here </b> </summary>

```diff
diff --git a/pkg/health/condition/check_backup_ready.go b/pkg/health/condition/check_backup_ready.go
index 44451f5e..b3700bfe 100644
--- a/pkg/health/condition/check_backup_ready.go
+++ b/pkg/health/condition/check_backup_ready.go
@@ -23,6 +23,7 @@ import (
 	"github.com/gardener/etcd-druid/pkg/utils"
 	coordinationv1 "k8s.io/api/coordination/v1"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -42,105 +43,70 @@ const (
 	NotChecked string = "ConditionNotChecked"
 )
 
-func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
-	// Default case
-	result := &result{
-		conType: druidv1alpha1.ConditionTypeBackupReady,
-		status:  druidv1alpha1.ConditionUnknown,
-		reason:  Unknown,
-		message: "Cannot determine etcd backup status",
-	}
+func (a *backupReadyCheck) fetchLease(ctx context.Context, name string, namespace string) (*coordinationv1.Lease, error) {
+	lease := &coordinationv1.Lease{}
+	err := a.cl.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, lease)
+	return lease, err
+}
 
+func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
 	// Special case of etcd not being configured to take snapshots
 	// Do not add the BackupReady condition if backup is not configured
 	if etcd.Spec.Backup.Store == nil || etcd.Spec.Backup.Store.Provider == nil || len(*etcd.Spec.Backup.Store.Provider) == 0 {
 		return nil
 	}
 
-	// Fetch snapshot leases
-	var (
-		err            error
-		fullSnapLease  = &coordinationv1.Lease{}
-		deltaSnapLease = &coordinationv1.Lease{}
-	)
-
-	err = a.cl.Get(ctx, types.NamespacedName{Name: getFullSnapLeaseName(&etcd), Namespace: etcd.ObjectMeta.Namespace}, fullSnapLease)
+	fullSnapLease, err := a.fetchLease(ctx, etcd.GetFullSnapshotLeaseName(), etcd.Namespace)
 	if err != nil {
-		result.message = fmt.Sprintf("Unable to fetch full snap lease. %s", err.Error())
-		return result
+		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, fmt.Sprintf("Unable to fetch full snap lease. %s", err.Error()))
 	}
 
-	err = a.cl.Get(ctx, types.NamespacedName{Name: getDeltaSnapLeaseName(&etcd), Namespace: etcd.ObjectMeta.Namespace}, deltaSnapLease)
+	deltaSnapLease, err := a.fetchLease(ctx, etcd.GetDeltaSnapshotLeaseName(), etcd.Namespace)
 	if err != nil {
-		result.message = fmt.Sprintf("Unable to fetch delta snap lease. %s", err.Error())
-		return result
+		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, fmt.Sprintf("Unable to fetch delta snap lease. %s", err.Error()))
 	}
 
 	deltaLeaseRenewTime := deltaSnapLease.Spec.RenewTime
 	fullLeaseRenewTime := fullSnapLease.Spec.RenewTime
 	fullLeaseCreateTime := &fullSnapLease.ObjectMeta.CreationTimestamp
 
+	// Handle snapshot leases not yet renewed
 	if fullLeaseRenewTime == nil && deltaLeaseRenewTime == nil {
-		// Both snapshot leases are not yet renewed
-		result.message = "Snapshotter has not started yet"
-		return result
-
-	} else if fullLeaseRenewTime == nil && deltaLeaseRenewTime != nil {
-		// Most probable during reconcile of existing clusters if fresh leases are created
-
-		fullSnapshotDuration, err := utils.ComputeScheduleDuration(*etcd.Spec.Backup.FullSnapshotSchedule)
-		if err != nil {
-			return result
-		}
+		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, "Snapshotter has not started yet")
+	}
 
-		// Treat backup as succeeded if delta snap lease renewal happens in the required time window
-		// and full snap lease is not older than the computed full snapshot duration.
-		if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
-			if time.Since(fullLeaseCreateTime.Time) < fullSnapshotDuration {
-				result.reason = BackupSucceeded
-				result.message = "Delta snapshot backup succeeded"
-				result.status = druidv1alpha1.ConditionTrue
-				return result
-			} else {
-				result.status = druidv1alpha1.ConditionFalse
-				result.reason = BackupFailed
-				result.message = "Full snapshot backup failed. Full snapshot lease created long ago, but not renewed"
-				return result
-			}
-		}
+	// Handle fresh leases
+	if fullLeaseRenewTime == nil && deltaLeaseRenewTime != nil {
+		return handleFreshLeases(&etcd, deltaLeaseRenewTime, fullLeaseCreateTime)
+	}
 
-	} else if fullLeaseRenewTime != nil && deltaLeaseRenewTime == nil {
-		// Most probable during a startup scenario for new clusters
-		// Special case: return Unknown condition for some time to allow delta backups to start up
-		// Even though full snapshot may have succeeded by the required time, we must still wait
-		// for delta snapshotting to begin to consider the backups as healthy, to maintain the given RPO
-		if time.Since(fullLeaseRenewTime.Time) < 5*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
-			result.message = "Waiting for periodic delta snapshotting to begin"
-			return result
-		} else {
-			result.status = druidv1alpha1.ConditionFalse
-			result.reason = BackupFailed
-			result.message = "Delta snapshot backup failed. Delta snapshot lease created long ago, but not renewed"
-			return result
-		}
+	// Handle startup scenario
+	if fullLeaseRenewTime != nil && deltaLeaseRenewTime == nil {
+		return handleStartupScenario(&etcd, fullLeaseRenewTime)
+	}
+	// Handle periodically maintained leases
+	return handleMaintainedLeases(&etcd, deltaLeaseRenewTime, fullLeaseRenewTime)
+}
 
-	} else {
-		// Both snap leases are renewed, ie, maintained. Both are expected to be renewed periodically
-		fullSnapshotDuration, err := utils.ComputeScheduleDuration(*etcd.Spec.Backup.FullSnapshotSchedule)
-		if err != nil {
-			return result
-		}
+// Handles the case where snapshot leases are periodically maintained
+func handleMaintainedLeases(etcd *druidv1alpha1.Etcd, deltaLeaseRenewTime, fullLeaseRenewTime *metav1.MicroTime) *result {
+	// Both snap leases are renewed, ie, maintained. Both are expected to be renewed periodically
+	fullSnapshotDuration, err := utils.ComputeScheduleDuration(*etcd.Spec.Backup.FullSnapshotSchedule)
+	if err != nil {
+		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, "Cannot compute full snapshot schedule duration")
+	}
 
-		// Mark backup as succeeded only if at least one of the last two delta snapshots has been taken successfully
-		// and if the last full snapshot has been taken according to the given full snapshot schedule
-		if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration && time.Since(fullLeaseRenewTime.Time) < fullSnapshotDuration {
-			result.reason = BackupSucceeded
-			result.message = "Snapshot backup succeeded"
-			result.status = druidv1alpha1.ConditionTrue
-			return result
-		}
+	// Mark backup as succeeded only if at least one of the last two delta snapshots has been taken successfully
+	// and if the last full snapshot has been taken according to the given full snapshot schedule
+	if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration && time.Since(fullLeaseRenewTime.Time) < fullSnapshotDuration {
+		return createBackupResult(druidv1alpha1.ConditionTrue, BackupSucceeded, "Snapshot backup succeeded")
 	}
 
+	return handleStaleSnapshotLeases(etcd, deltaLeaseRenewTime)
+}
+
+// Handles the case when snapshot leases are not updated for a long time
+func handleStaleSnapshotLeases(etcd *druidv1alpha1.Etcd, deltaLeaseRenewTime *metav1.MicroTime) *result {
 	// Cases where snapshot leases are not updated for a long time
 	// If snapshot leases are present and leases aren't updated, it is safe to assume that backup is not healthy
 	if etcd.Status.Conditions != nil {
@@ -154,24 +120,54 @@ func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) R
 		// Transition to "False" state only if present state is "Unknown" or "False"
 		if deltaLeaseRenewTime != nil && (prevBackupReadyStatus.Status == druidv1alpha1.ConditionUnknown || prevBackupReadyStatus.Status == druidv1alpha1.ConditionFalse) {
 			if time.Since(deltaLeaseRenewTime.Time) > 3*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
-				result.status = druidv1alpha1.ConditionFalse
-				result.reason = BackupFailed
-				result.message = "Stale snapshot leases. Not renewed in a long time"
-				return result
+				return createBackupResult(druidv1alpha1.ConditionFalse, BackupFailed, "Stale snapshot leases. Not renewed in a long time")
 			}
 		}
 	}
+	// Transition to "Unknown" state if we cannot prove a "True" state
+	return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, "Cannot determine etcd backup status")
+}
 
-	// Transition to "Unknown" state is we cannot prove a "True" state
-	return result
+// Handles the case of a startup scenario for new clusters
+func handleStartupScenario(etcd *druidv1alpha1.Etcd, fullLeaseRenewTime *metav1.MicroTime) *result {
+	// Special case: return Unknown condition for some time to allow delta backups to start up
+	// Even though full snapshot may have succeeded by the required time, we must still wait
+	// for delta snapshotting to begin to consider the backups as healthy, to maintain the given RPO
+	if time.Since(fullLeaseRenewTime.Time) < 5*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
+		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, "Waiting for periodic delta snapshotting to begin")
+	}
+	return createBackupResult(druidv1alpha1.ConditionFalse, BackupFailed, "Delta snapshot backup failed. Delta snapshot lease created long ago, but not renewed")
 }
 
-func getDeltaSnapLeaseName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("%s-delta-snap", etcd.ObjectMeta.Name)
+// Handles the case of fresh leases being created
+func handleFreshLeases(etcd *druidv1alpha1.Etcd, deltaLeaseRenewTime *metav1.MicroTime, fullLeaseCreateTime *metav1.Time) *result {
+	// Most probable during reconcile of existing clusters if fresh leases are created
+	fullSnapshotDuration, err := utils.ComputeScheduleDuration(*etcd.Spec.Backup.FullSnapshotSchedule)
+	if err != nil {
+		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, "Cannot determine etcd backup status")
+	}
+
+	// Treat backup as succeeded if delta snap lease renewal happens in the required time window
+	// and full snap lease is not older than the computed full snapshot duration.
+	if time.Since(deltaLeaseRenewTime.Time) < 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration {
+		if time.Since(fullLeaseCreateTime.Time) < fullSnapshotDuration {
+			return createBackupResult(druidv1alpha1.ConditionTrue, BackupSucceeded, "Delta snapshot backup succeeded")
+		} else {
+			return createBackupResult(druidv1alpha1.ConditionFalse, BackupFailed, "Full snapshot backup failed. Full snapshot lease created long ago, but not renewed")
+		}
+	}
+
+	return handleStaleSnapshotLeases(etcd, deltaLeaseRenewTime)
 }
 
-func getFullSnapLeaseName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("%s-full-snap", etcd.ObjectMeta.Name)
+// Creates a backup result
+func createBackupResult(status druidv1alpha1.ConditionStatus, reason string, message string) *result {
+	return &result{
+		conType: druidv1alpha1.ConditionTypeBackupReady,
+		status:  status,
+		reason:  reason,
+		message: message,
+	}
 }
 
 // BackupReadyCheck returns a check for the "BackupReady" condition.

```

</details>


```go 
func (a *backupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
	// Special case of etcd not being configured to take snapshots
	// Do not add the BackupReady condition if backup is not configured
	if etcd.Spec.Backup.Store == nil || etcd.Spec.Backup.Store.Provider == nil || len(*etcd.Spec.Backup.Store.Provider) == 0 {
		return nil
	}

	fullSnapLease, err := a.fetchLease(ctx, etcd.GetFullSnapshotLeaseName(), etcd.Namespace)
	if err != nil {
		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, fmt.Sprintf("Unable to fetch full snap lease. %s", err.Error()))
	}

	deltaSnapLease, err := a.fetchLease(ctx, etcd.GetDeltaSnapshotLeaseName(), etcd.Namespace)
	if err != nil {
		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, fmt.Sprintf("Unable to fetch delta snap lease. %s", err.Error()))
	}

	deltaLeaseRenewTime := deltaSnapLease.Spec.RenewTime
	fullLeaseRenewTime := fullSnapLease.Spec.RenewTime
	fullLeaseCreateTime := &fullSnapLease.ObjectMeta.CreationTimestamp

	// Handle snapshot leases not yet renewed
	if fullLeaseRenewTime == nil && deltaLeaseRenewTime == nil {
		return createBackupResult(druidv1alpha1.ConditionUnknown, Unknown, "Snapshotter has not started yet")
	}

	// Handle fresh leases
	if fullLeaseRenewTime == nil && deltaLeaseRenewTime != nil {
		return handleFreshLeases(&etcd, deltaLeaseRenewTime, fullLeaseCreateTime)
	}

	// Handle startup scenario
	if fullLeaseRenewTime != nil && deltaLeaseRenewTime == nil {
		return handleStartupScenario(&etcd, fullLeaseRenewTime)
	}
	// Handle periodically maintained leases
	return handleMaintainedLeases(&etcd, deltaLeaseRenewTime, fullLeaseRenewTime)
}
```

Makefile
```diff
@@ -72,7 +72,7 @@ install: manifests
 	kubectl apply -f config/crd/bases
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-.PHONY: deploy
+.PHONY: deploy-via-kustomize
```
seshachalam-yv, line 75: Thanks, Nice catch 😍

seshachalam-yv (changes requested):

pkg/health/condition/check_backup_ready.go
shreyas-s-rao: Thanks @seshachalam-yv . So you agree with the changes to the actual conditions, but you would like to improve code readability, correct? That's definitely a valid point, and I'll make the changes. Thanks for pointing it out! 👍

seshachalam-yv: Overall conditions looks to me

shreyas-s-rao:
@seshachalam-yv thanks for your review. I've addressed your comment, in a slightly different way than you suggested. But overall, readability has improved. PTAL.

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,184 @@ const (
...
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionFalse, BackupFailed,
+			"Full snapshot backup failed. Full snapshot lease created long ago, but not renewed",
+		)
+	}
+
+	return createBackupConditionResult(
+		druidv1alpha1.ConditionTrue, BackupSucceeded,
+		"Delta snapshot backup succeeded",
+	)
```
seshachalam-yv, line 166: @Shreyas-s-rao,

Nitpick: Rather than using nested conditions, what do you think about flattening them into a switch case structure as follows?

```go
	switch {
	case !deltaSnapshotLeaseRenewedRecently && fullSnapshotLeaseCreatedRecently:
		// Delta snapshot lease is stale, but full snapshot lease was created recently
		return createBackupConditionResult(druidv1alpha1.ConditionFalse, BackupFailed, "Delta snapshot backup failed. Delta snapshot lease not renewed in a long time")
	case !deltaSnapshotLeaseRenewedRecently && !fullSnapshotLeaseCreatedRecently:
		// Both leases are stale
		return createBackupConditionResult(druidv1alpha1.ConditionFalse, BackupFailed, "Stale snapshot leases. Not renewed in a long time")
	case !fullSnapshotLeaseCreatedRecently:
		// Full snapshot lease is stale, but delta snapshot lease was renewed recently
		return createBackupConditionResult(druidv1alpha1.ConditionFalse, BackupFailed, "Full snapshot backup failed. Full snapshot lease created long ago, but not renewed")
	default:
		// Delta snapshot lease was renewed recently and full snapshot lease is not stale
		return createBackupConditionResult(druidv1alpha1.ConditionTrue, BackupSucceeded, "Delta snapshot backup succeeded")
	}
```

Alternatively, for consistency, we could consider maintaining similar flattened conditions as those in the `handleRenewedSnapshotLeases` function. [Link to Code](https://github.com/gardener/etcd-druid/pull/619/files#diff-9cc4496b69f9d0fb63a6e6fff5162bfbccae4e70edf42fff483f9da523751ca2R194-R218)

seshachalam-yv (approved):
Apart from one small nitpick, overall, this PR looks great to me. I appreciate the time and effort you've put into addressing all the comments and suggestions. I am impressed with the changes you've made in this PR. It not only enhances the code's readability but also provides a clear and logical flow. - excellent job! 😍

pkg/health/condition/check_backup_ready.go
shreyas-s-rao: Problem with switch cases is that even though we specify a `default` case at the end which will surely be executed if none of the previous cases are satisfied, golang expects a return statement at the end of the function in order to keep things absolutely deterministic. So we will be forced to either remove the default case from the switch construct and move it out to the end of the function, or to return a dummy backupReady result, which is semantically incorrect.

I would like to retain the if-else construct way of handling the different cases, but I'll flatten it out as you requested.

shreyas-s-rao:
@seshachalam-yv I've addressed your follow-up suggestion as well. Thanks for the detailed suggestions!

pkg/health/condition/check_backup_ready.go
seshachalam-yv: Thanks Shreyas

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
-	return fmt.Sprintf("%s-delta-snap", string(etcd.ObjectMeta.Name))
+// handleSnapshotterStartupCase handles cases where snapshotter has just started,
+// so only full snapshot has been taken, while delta snapshot is still not taken.
+// Returns `Unknown` condition for some time to allow delta snapshotting to begin, because
+// even though the full snapshot may have succeeded within the required time, we must still wait
+// for delta snapshotting to begin to consider the backups as healthy, to maintain the given RPO.
+func handleSnapshotterStartupCase(fullSnapshotLeaseRenewTime time.Time, deltaSnapshotRenewalGracePeriod time.Duration) *result {
+	if time.Since(fullSnapshotLeaseRenewTime) > deltaSnapshotRenewalGracePeriod {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionFalse, BackupFailed,
```
unmarshall, line 187: I am not entirely convinced if this results in `BackupFailed`. Can you semantically define this term to better gauge if this is the correct reason code? The reason is that you still have a full-snapshot that has been successfully backed-up. Should you call this `BackupFailed` or `DeltaSnapshotBackupFailed`?

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
+			"Delta snapshot backup failed. Delta snapshot lease not renewed in a long time",
+		)
+	}
+
+	return createBackupConditionResult(
+		druidv1alpha1.ConditionUnknown, Unknown,
```
unmarshall, line 193: Not clear on why the reason code should be `Unknown`. Also using the same condition for full and delta is quite confusing to correctly determine the status of the overall backup. If we now see this condition in the etcd status.Conditions then one must always remember that this condition means that a full snapshot was indeed taken and we are waiting for delta snapshot backups to be taken. This is not clear from this condition as we have decided to overload one condition for 2 different snapshots.

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
+			fullSnapshotLeaseCreationTime.Time, fullSnapshotDuration,
+			deltaSnapshotLeaseRenewTime.Time, 2*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration,
+		)
+	}
+
+	// Most probable during a startup of a new cluster and only full snapshot has been taken
+	if fullSnapshotLeaseRenewTime != nil && deltaSnapshotLeaseRenewTime == nil {
+		return handleSnapshotterStartupCase(
+			fullSnapshotLease.Spec.RenewTime.Time,
+			5*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration,
```
unmarshall, line 100: Can we have create functions for delta and full snapshot grace period? These can be methods on etcd resource itself, if you do not like that then these can just be standalone functions. One single place to compute these preventing duplicating these multipliers in more than one place in future code changes.

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
 
-func getDeltaSnapLeaseName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("%s-delta-snap", string(etcd.ObjectMeta.Name))
+// handleSnapshotterStartupCase handles cases where snapshotter has just started,
+// so only full snapshot has been taken, while delta snapshot is still not taken.
+// Returns `Unknown` condition for some time to allow delta snapshotting to begin, because
+// even though the full snapshot may have succeeded within the required time, we must still wait
+// for delta snapshotting to begin to consider the backups as healthy, to maintain the given RPO.
+func handleSnapshotterStartupCase(fullSnapshotLeaseRenewTime time.Time, deltaSnapshotRenewalGracePeriod time.Duration) *result {
+	if time.Since(fullSnapshotLeaseRenewTime) > deltaSnapshotRenewalGracePeriod {
```
unmarshall, line 185: you could have used `isLeaseStale` passing in `fullSnapshotLeaseRenewTime` as the first argument and `deltaSnapshotRenewalGracePeriod` as the second argument

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
 
-func getFullSnapLeaseName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("%s-full-snap", string(etcd.ObjectMeta.Name))
+// handleRenewedSnapshotLeases checks whether full and delta snapshot leases
+// have been renewed within the required times respectively.
+func handleRenewedSnapshotLeases(fullSnapshotLeaseRenewTime time.Time, fullSnapshotLeaseRenewalGracePeriod time.Duration, deltaSnapshotLeaseRenewTime time.Time, deltaSnapshotLeaseRenewalGracePeriod time.Duration) *result {
+	isFullSnapshotLeaseStale := isLeaseStale(fullSnapshotLeaseRenewTime, fullSnapshotLeaseRenewalGracePeriod)
+	isDeltaSnapshotLeaseStale := isLeaseStale(deltaSnapshotLeaseRenewTime, deltaSnapshotLeaseRenewalGracePeriod)
+
+	if isFullSnapshotLeaseStale && !isDeltaSnapshotLeaseStale {
```
unmarshall, line 204: if you create separate conditions for full and delta snapshots then these checks would get simplified and it will let you capture these 2 conditions independently. BackupsReady can then be a derived condition which could look at delta and full snapshot conditions if at all you require a single condition for all backups.

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionFalse, BackupFailed,
+			"Stale full snapshot lease. Not renewed in a long time",
```
unmarshall, line 207: `Not renewed in a long time` - do you wish to include the time as well since it was not renewed? You already have this information, why not just make it part of the message?

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
+// while the full snapshot lease is not renewed, possibly since it might have been recreated recently.
+// Treats backup as succeeded if delta snapshot lease is renewed within the required time window
+// and full snapshot lease object is not older than the computed full snapshot duration.
+func handleOnlyDeltaSnapshotLeaseRenewedCase(fullSnapshotLeaseCreationTime time.Time, fullSnapshotLeaseCreationGracePeriod time.Duration, deltaSnapshotLeaseRenewTime time.Time, deltaSnapshotLeaseRenewalGracePeriod time.Duration) *result {
+	// wasFullSnapshotLeaseCreatedRecently indicates whether full snapshot was created within the given grace period. If it was,
+	// then it can be renewed only upon the next scheduled full snapshot event, so until then we cannot assume the status of the last
+	// full snapshot. But if it was created before the grace period and not been renewed, then full snapshot lease can be considered stale.
+	wasFullSnapshotLeaseCreatedRecently := wasLeaseCreatedRecently(fullSnapshotLeaseCreationTime, fullSnapshotLeaseCreationGracePeriod)
+
+	// isDeltaSnapshotLeaseStale indicates whether the delta snapshot lease was renewed within the given grace period.
```
unmarshall, line 144: comment says `indicates whether the delta snapshot lease was renewed within the given grace period.` whereas what you actually check is if the delta snapshot lease is stale (opposite of what the comment says). Can you correct the comment?

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
+	isDeltaSnapshotLeaseStale := isLeaseStale(deltaSnapshotLeaseRenewTime, deltaSnapshotLeaseRenewalGracePeriod)
+
+	// Delta snapshot lease is stale, while staleness of full snapshot lease cannot be determined yet
+	if isDeltaSnapshotLeaseStale && wasFullSnapshotLeaseCreatedRecently {
```
unmarshall, line 148: This has become quite complicated. Can be simplified by just having 2 conditions. Then we just require 2 functions overall - one to properly update full snapshot lease status and another to update delta snapshot lease status

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
-	return result
 }
 
-func getDeltaSnapLeaseName(etcd *druidv1alpha1.Etcd) string {
-	return fmt.Sprintf("%s-delta-snap", string(etcd.ObjectMeta.Name))
+// handleSnapshotterStartupCase handles cases where snapshotter has just started,
+// so only full snapshot has been taken, while delta snapshot is still not taken.
+// Returns `Unknown` condition for some time to allow delta snapshotting to begin, because
+// even though the full snapshot may have succeeded within the required time, we must still wait
+// for delta snapshotting to begin to consider the backups as healthy, to maintain the given RPO.
```
aaronfern, line 183: I'm not too sure I agree with this
I think that if a full snapshot has been taken and that is within the `deltaSnapshotRenewalGracePeriod` then backup status should be `BackupSucceeded` as this still maintains our RPO for that instant and makes semantic sense
wdyt?

pkg/health/condition/builder.go
```diff
@@ -111,7 +111,7 @@ func (b *defaultBuilder) Build(replicas int32) []druidv1alpha1.Condition {
 			if condition.Status == "" {
 				condition.Status = druidv1alpha1.ConditionUnknown
 			}
-			condition.Reason = ConditionNotChecked
+			condition.Reason = NotChecked
```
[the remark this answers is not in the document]
shreyas-s-rao: I am not fully sure about this. All I did here was refactor the variable name from `ConditionNotChecked` to `NotChecked`. But from the look of it, yes, looks like `Reason` needs to be set to something other than "NotChecked", like "ClusterScaledToZero" maybe.

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
 	BackupFailed string = "BackupFailed"
 	// Unknown is a constant that means that the etcd backup status is currently not known
 	Unknown string = "Unknown"
-	// ConditionNotChecked is a constant that means that the etcd backup status has not been updated or rechecked
-	ConditionNotChecked string = "ConditionNotChecked"
+	// NotChecked is a constant that means that the etcd backup status has not been updated or rechecked
+	NotChecked string = "ConditionNotChecked"
```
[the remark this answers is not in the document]
shreyas-s-rao: We no longer need this variable at all, since its only usage was when `etcd.Replicas==0`, and we decided to use a different reason string for that. So I'll remove this for now. Even semantically it's ok to do this, since `condition` package sets the conditions only after checking them, so `NotChecked` reason isn't possible anymore.

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
 
-	//Fetch snapshot leases
-	var (
-		fullSnapErr, incrSnapErr error
-		fullSnapLease            = &coordinationv1.Lease{}
-		deltaSnapLease           = &coordinationv1.Lease{}
+	// Fetch snapshot leases
+	fullSnapshotLease, err := a.fetchLease(ctx, etcd.GetFullSnapshotLeaseName(), etcd.Namespace)
+	if err != nil {
+		return createBackupConditionResult(
```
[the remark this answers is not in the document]
shreyas-s-rao: `conType: druidv1alpha1.ConditionTypeBackupReady` is common to all backupReady condition result, so it made sense to pull it out into a separate function, just to avoid adding the `conType` every single time when returning. Of course, the previous method was to create a default result at the beginning of the function and then simply change the values when returning, but @seshachalam-yv pointed out that it was not the most readable, since one has to check the default result as well as the changed values to figure out the final result being returned.

pkg/health/condition/check_backup_ready.go
```diff
@@ -37,101 +38,194 @@ const (
...
+			druidv1alpha1.ConditionUnknown, Unknown,
+			fmt.Sprintf("Unable to fetch full snap lease. %s", err.Error()),
+		)
+	}
+
+	deltaSnapSnapshotLease, err := a.fetchLease(ctx, etcd.GetDeltaSnapshotLeaseName(), etcd.Namespace)
+	if err != nil {
+		return createBackupConditionResult(
+			druidv1alpha1.ConditionUnknown, Unknown,
+			fmt.Sprintf("Unable to fetch delta snap lease. %s", err.Error()),
```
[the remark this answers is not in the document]
shreyas-s-rao: >do you wish to include err.Error() as part of the message or just log it using logger.Error? How large is err.Error()

It's safer to add the error to the condition, just so that it's visible to an operator without having to sift through logs. Even gardener shoot conditions for instance store the error message in the condition, which are printed to the dashboard as well. It's quite helpful for operators and users alike.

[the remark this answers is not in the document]
shreyas-s-rao: > If you see the reason clearly indicates what that condition is for. So having a reason as Unknown would be un-qualified and therefore very hard to reason or disambiguate or even process later. In your proposal one has to look at the message to learn more about the condition which is a departure from the existing set of conditions.

I'll try to add more meaningful reason strings then.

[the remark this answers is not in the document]
shreyas-s-rao: >Check method returns a single Result. If there are problems fetching both the leases then you will always return the condition with message for delta, thereby masking full-snapshot lease condition message. Would it make sense to have different conditions - one for delta and another for full snapshot?

I've handled such cases specifically so that the full snapshot lease error does not get masked by delta snapshot lease error. If both leases have errors, both are captured in the condition, such as [this](https://github.com/shreyas-s-rao/etcd-druid/blob/44cc1ae3558cabff57298039666ba5423dc444a3/pkg/health/condition/check_backup_ready.go#L159) and [this](https://github.com/shreyas-s-rao/etcd-druid/blob/44cc1ae3558cabff57298039666ba5423dc444a3/pkg/health/condition/check_backup_ready.go#L221). 
Looks like only the case of failing to fetch the leases needs to be handled more robustly. I'll handle this then, thanks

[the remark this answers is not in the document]
shreyas-s-rao: >What happens when lease itself is NotFound? Should you not have a different message indicating that the lease itself is missing?

Right now, it's a blanket message of `Unable to fetch full/delta snap lease: <error-string>`, which is still technically correct. The error string holds the reason as to why the fetch failed, and it will specify that lease not found. If you want, I can separate out the lease-not-found case and use a separate `Reason` string for that like `LeaseNotFound`. WDYT?

pkg/utils/miscellaneous.go
```diff
@@ -77,3 +79,19 @@ func Max(x, y int) int {
...
+// TODO: write a new function to accurately compute the previous activation time from the cron schedule
+// in order to compute when the previous activation of the cron schedule was supposed to have occurred.
+func ComputeScheduleDuration(cronSchedule string) (time.Duration, error) {
+	schedule, err := cron.ParseStandard(cronSchedule)
+	if err != nil {
+		return 0, err
+	}
+
+	nextScheduledTime := schedule.Next(time.Now())
+	nextNextScheduledTime := schedule.Next(nextScheduledTime.Add(time.Second))
```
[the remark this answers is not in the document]
shreyas-s-rao: > any reason to add 1 sec?

Just to ensure that the Next() function really returns the next scheduled time and not the same time. But I do see from the [documentation](https://pkg.go.dev/github.com/robfig/cron#SpecSchedule.Next) that Next always returns time greater than the given time, so this extra 1 second isn't necessary. I'll remove this, thanks!

> Can you add a unit test for this function?
Will add this.

pkg/health/condition/check_backup_ready.go
unmarshall: I generally have an issue with having a single condition for delta + full snapshot backup. It will become a LOT easier if we have separate conditions.

pkg/health/condition/check_backup_ready.go
shreyas-s-rao: That again now depends on how granular we want our `Reason` codes to be. `BackupFailed` denotes that either of the backups failed, and either case is dangerous. Without a full snapshot taken on time, we risk longer restoration time and hence a longer RTO. And without a delta snapshot taken on time, we violate SLAs since RPOs is affected. So differentiating between which snapshot failed in the `Reason` does not provide much benefit to operators. If they want more info on why `BackupFailed` was set, they can look into the `Message`.

pkg/health/condition/check_backup_ready.go
shreyas-s-rao: Agreed that something like `PartiallyUnknown` sounds better than `Unknown`. But it simply denotes that there is an "unknown" in the cluster that druid does not know about, hence it sets this reason. It provides more detailed information in the message.

pkg/health/condition/check_backup_ready.go
shreyas-s-rao: If you notice, the grace period isn't same across functions. Grace period for renewing delta snapshot lease that was already renewed previously is 2x delta snapshot period, while grace period for delta snapshotting to start once full snapshot has been taken is 5x delta snapshot period. So it's not achievable by one single function. Also, this is condition-specific information, and may not make sense to be part of the API, because the API has no knowledge of which conditions are set on the etcd status by druid, and it shouldn't as well.

pkg/health/condition/check_backup_ready.go
shreyas-s-rao: I thought about it, but it's semantically incorrect. Ideally `isLeaseStale` checks whether a particular lease is stale based on its own renewal time and grace period. But in this particular case, we're simply checking if full snapshot was renewed more than delta grace period ago. This does not mean full snapshot lease is stale, since full snapshot was uploaded at the expected time. It simply means that delta snapshot has not been taken, so technically delta snapshot lease is the stale one. That's the reason I simply used a check here instead of the `isLeaseStale` function.

pkg/health/condition/check_backup_ready.go
shreyas-s-rao: Good idea, thanks for the suggestion!

pkg/health/condition/check_backup_ready.go
shreyas-s-rao: That's the reason we pass the `deltaSnapshotRenewalGracePeriod` as `5*etcd.Spec.Backup.DeltaSnapshotPeriod.Duration`, to allow backup sidecar that much time to start delta snapshotting. It depends on how we define RPO, and right now RPO loosely means the delta snapshot period (1x). I've removed the mention of `RPO` in the comment to avoid any ambiguity, since we still don't define an official RPO for etcds managed by druid, so it doesn't make sense to bake that into the code now, until we have more clarity.

shreyas-s-rao:
/hold
To be re-looked at from the perspective of having two separate conditions for `FullBackupReady` and `DeltaBackupReady` as suggested by @unmarshall .
/milestone v0.20.0

shreyas-s-rao:
After an out-of-band discussion amongst myself, @unmarshall , @seshachalam-yv and @aaronfern , we concluded that it is not simple to handle all cases of successful, failed, skipped, missed snapshots by etcd-backup-restore, as well as missed renewals of the snapshot lease. Instead, we will solve this holistically as part of #702 , where the EtcdMember `status.snapshots.last[Full|Delta]` will also include an additional field `state`, with possible values `Succeeded`, `Failed` or `Skipped`, to correctly reflect the state of the snapshot.

This PR will be closed in favour of #729 , which focuses on fixing the problem of hardcoded value of `24h` for checking full snapshot staleness.
/close

Issue #643, opened 2023-07-14
[Feature] Introduce feature gates concept, starting with `UseEtcdWrapper`


shreyas-s-rao:
**Feature (What you would like to be added):**
Introduce feature gates or feature flags for etcd-druid to control switching on or off certain large or high-impact or experimental features, such as `UseEtcdWrapper`. These feature flags can be set using CLI flags passed to the etcd-druid command, and passed to specific controllers as required.

**Motivation (Why is this needed?):**
Similar to how gardener defines [feature gates](https://github.com/gardener/gardener/blob/06f3b19981f6ee4fdfb2471b87a1f2cb5bb9f84e/docs/deployment/feature_gates.md?plain=1) for its components, it is necessary for etcd-druid too to define certain feature gates which allow enabling or disabling druid-specific features.

**Approach/Hint to the implement solution (optional):**
- [x] Introduce framework for feature gates - https://github.com/gardener/etcd-druid/pull/646
  - [x] Add `pkg/features` package
  - [x] Allow controllers to specify features relevant to their own operation, via their configs
  - [x] Controller manager should parse feature gates from CLI flags and pass them to each controller as required
- [x] Introduce `UseEtcdWrapper` feature - https://github.com/gardener/etcd-druid/pull/646
  - [x] Set to `false` by default
  - [x] When set to `true`, should make the following changes:
    - [x] Switch to different image vector containing `etcd-wrapper:v0.1.0` and `etcdbrctl:v0.25.0`
    - [x] Add init container to change permissions on etcd volume
    - [x] Add security context to containers as required
    - [x] Set readiness probe specific to etcd-wrapper image
- [x] Allow setting `--feature-gates` CLI flag for druid command via helm chart - https://github.com/gardener/etcd-druid/pull/646
- [x] Introduce new prow job that sets feature gate `UseEtcdWrapper=true` for running druid e2e tests - https://github.com/gardener/ci-infra/pull/798
  - [x] Enhance skaffold to be able to run a `feature-gates` flavour for enabling feature gates
- [x] Add operator-relevant documentation for using feature gates - https://github.com/gardener/etcd-druid/pull/646
- [ ] #655

/area usability
/assign @shreyas-s-rao @aaronfern 
/milestone v0.19.0

PR #646, opened 2023-07-19
Introduced feature gates in etcd druid and added `UseEtcdWrapper` feature gate

aaronfern:

/area disaster-recovery
/kind enhancement

This PR achieves the following
1. Introduce feature gates in `etcd-druid`. Now large or experimental changes can be made available via a feature gate to only those who opt for them

2. A new feature gate `UseEtcdWrapper` has been added. Enabling this feature gates instructs to deploy an `etcd-wrapper` container along with a distroless `etcd-backup-restore` conatainer via the statefulset instead of the `etcd-custom-image` container along with a previous version of the `etcd-backup-restore` container based on the alpine distro

Fixes #

This PR is to be merged only after a release of `etcd-custom-image` is made

**Release note**:

```other operator
feature gates have been introduced in etcd druid
```
```other operator
`UseEtcdWrapper` feature gate has been added to allow users to opt for the new `etcd-wrapper` image
```


controllers/config.go
```diff
@@ -102,6 +112,54 @@ func InitFromFlags(fs *flag.FlagSet, cfg *ManagerConfig) {
 
 	cfg.SecretControllerConfig = &secret.Config{}
 	secret.InitFromFlags(fs, cfg.SecretControllerConfig)
+
+	return nil
+}
+
+// InitFeatureGates initializes feature gates from the provided CLI flag set.
+func InitFeatureGates(fs *flag.FlagSet, featureFlags *featuregate.MutableFeatureGate) error {
```
shreyas-s-rao, line 120:
```suggestion
func initFeatureGates(fs *flag.FlagSet, featureFlags *featuregate.MutableFeatureGate) error {
```

Since this is not being used in any other package, please make it a private function.

controllers/config.go
```diff
@@ -102,6 +112,54 @@ func InitFromFlags(fs *flag.FlagSet, cfg *ManagerConfig) {
...
+	if err := (*featureFlags).Add(features.GetDefaultFeatures()); err != nil {
+		return fmt.Errorf("error adding features to the featuregate: %v", err)
+	}
+	(*featureFlags).AddFlag(fs)
+
+	return nil
+}
+
+// PopulateControllerFeatureGates adds relevant feature gates to every controller configuration
+func PopulateControllerFeatureGates(config *ManagerConfig) {
```
shreyas-s-rao, line 131:
```suggestion
func populateControllersFeatureGates(config *ManagerConfig) {
```

Same as above, with additional plural (`Controllers`) since this function adds feature gates for all controllers.

controllers/config.go
```diff
@@ -102,6 +112,54 @@ func InitFromFlags(fs *flag.FlagSet, cfg *ManagerConfig) {
 
 	cfg.SecretControllerConfig = &secret.Config{}
 	secret.InitFromFlags(fs, cfg.SecretControllerConfig)
+
+	return nil
+}
+
+// InitFeatureGates initializes feature gates from the provided CLI flag set.
+func InitFeatureGates(fs *flag.FlagSet, featureFlags *featuregate.MutableFeatureGate) error {
+	*featureFlags = featuregate.NewFeatureGate()
```
shreyas-s-rao, line 121: Why are you resetting the `featureGates` variable to a new feature gate after passing it as a parameter from the function call? Instead, you should simply pass manager config to this function, and use a separate variable to initialize the feature gates flag and then run `AddFlag` and then assign it to `cfg.FeatureGates`.

controllers/utils/reconciler.go
```diff
@@ -28,16 +28,21 @@ import (
...
 	// DefaultImageVector is a constant for the path to the default image vector file.
 	DefaultImageVector = "images.yaml"
+	// EtcdWrapperImageVector is a constant for the path to the etcd wrapper image vector file.
+	EtcdWrapperImageVector = "images-wrapper.yaml"
 )
 
-// GetDefaultImageYAMLPath returns the path to the default images.yaml file.
-func GetDefaultImageYAMLPath() string {
+// GetImageYAMLPath returns the path to the default images.yaml file.
+func GetImageYAMLPath(useEtcdWrapperImages bool) string {
```
shreyas-s-rao, line 36:
```suggestion
func GetImageYAMLPath(useEtcdWrapperImageVector bool) string {
```

controllers/utils/reconciler.go
```diff
@@ -28,16 +28,21 @@ import (
 const (
 	// DefaultImageVector is a constant for the path to the default image vector file.
 	DefaultImageVector = "images.yaml"
+	// EtcdWrapperImageVector is a constant for the path to the etcd wrapper image vector file.
+	EtcdWrapperImageVector = "images-wrapper.yaml"
 )
 
-// GetDefaultImageYAMLPath returns the path to the default images.yaml file.
-func GetDefaultImageYAMLPath() string {
+// GetImageYAMLPath returns the path to the default images.yaml file.
```
shreyas-s-rao, line 35:
```suggestion
// GetImageYAMLPath returns the path to the image vector YAML file.
// The path to the default image vector YAML path is returned, unless `useEtcdWrapperImageVector`
// is set to true, in which case the path to the etcd wrapper image vector YAML is returned.
```

controllers/utils/reconciler.go
```diff
@@ -28,16 +28,21 @@ import (
...
+		return filepath.Join(common.ChartPath, EtcdWrapperImageVector)
+	}
 	return filepath.Join(common.ChartPath, DefaultImageVector)
 }
 
-// CreateImageVector creates an image vector from the default images.yaml file.
-func CreateImageVector() (imagevector.ImageVector, error) {
-	imageVector, err := imagevector.ReadGlobalImageVectorWithEnvOverride(GetDefaultImageYAMLPath())
+// CreateImageVector creates an image vector from the default images.yaml file or the images-wrapper.yaml file.
+func CreateImageVector(useEtcdWrapperImages bool) (imagevector.ImageVector, error) {
```
shreyas-s-rao, line 44:
```suggestion
func CreateImageVector(useEtcdWrapperImageVector bool) (imagevector.ImageVector, error) {
```

main.go
```diff
@@ -41,10 +41,15 @@ func main() {
...
+	if err := controllers.InitFromFlags(flag.CommandLine, &mgrConfig); err != nil {
+		logger.Error(err, "failed to initialize from flags")
+		os.Exit(1)
+	}
 
 	flag.Parse()
-	printFlags(logger)
+
+	flags := getFlags()
+	printFlags(logger, flags)
```
shreyas-s-rao, line 52: This change is no longer necessary, since we have no use case now for storing `flags` and using it anywhere else other than in `printFlags`. You can revert this to the previous state where flags were fetched and printed within `printFlags` function.

main.go
```diff
@@ -71,12 +76,20 @@ func printVersionInfo() {
...
+	return flags
+}
 
+func printFlags(logger logr.Logger, flags map[string]flag.Value) {
+	var flagKVs []interface{}
+	for k, v := range flags {
+		flagKVs = append(flagKVs, k, v.String())
+	}
 	logger.Info("Running with flags", flagKVs...)
 }
```
shreyas-s-rao, line 94: See above comment.

controllers/etcd/config.go
```diff
@@ -15,9 +15,11 @@
 package etcd
 
 import (
-	"flag"
+	flag "github.com/spf13/pflag"
+	"k8s.io/component-base/featuregate"
 
 	"github.com/gardener/etcd-druid/controllers/utils"
+	"github.com/gardener/etcd-druid/pkg/features"
```
shreyas-s-rao, line 22: Reorder imports - first group should contain standard libraries, next group should contain packages from within the project, next group should contain external packages.

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -508,6 +489,29 @@ func (c *component) createOrPatch(ctx context.Context, sts *appsv1.StatefulSet,
...
 		}
+		if c.values.UseEtcdWrapper {
+			// sections to add only when using etcd wrapper
+			// TODO: @aaronfern add this back to sts.Spec when UseEtcdWrapper becomes GA
+			sts.Spec.Template.Spec.InitContainers = []corev1.Container{
+				{
+					Name:         "change-permissions",
+					Image:        "alpine:3.18.2",
+					Command:      []string{"sh", "-c", "--"},
+					Args:         []string{"chown -R 65532:65532 /var/etcd/data"},
```
shreyas-s-rao, line 500: Shouldn't the args be in separate elements within the string array?
```suggestion
					Args:         []string{"chown", "-R", "65532:65532", "/var/etcd/data"},
```

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -119,36 +119,47 @@ func GenerateValues(
...
-	} else {
-		command = append(command, "--backup-restore-tls-enabled=false")
-		command = append(command, fmt.Sprintf("--backup-restore-host-port=%s-local:8080", val.Name))
-		command = append(command, fmt.Sprintf("--etcd-server-name=%s-local", val.Name))
+
+		return command
 	}
 
-	return command
+	return []string{}
```
shreyas-s-rao, line 162: Can you add a comment here specifying that druid now relies on the default entrypoint from `etcd-custom-image:v3.4.13-bootstrap-12`, which already calls `bootstrap.sh`?

shreyas-s-rao (changes requested):
@aaronfern thanks for the PR! I've tested the changes and it works well.
Please rename all occurrences of `feature flags` to `feature gates` since we no longer use individual CLI flags for setting each feature, but rather use a single `--feature-gates` flag for this, similar to upstream k8s feature gates.

Change release notes to:

feature operator
```
Feature gates have been introduced in etcd-druid, and can be specified using CLI flag `--feature-gate`.
```

feature operator

```
`UseEtcdWrapper` feature gate has been introduced to allow users to opt for the new [etcd-wrapper](https://github.com/gardener/etcd-wrapper) image.
```


Also update PR description to say that this PR partially fixes #643

pkg/component/etcd/statefulset/statefulset.go
aaronfern: Not really, it shouldn't make a difference. However the suggestion reads better, so I've added it

charts/druid/templates/controller-deployment.yaml
```diff
@@ -25,6 +25,13 @@ spec:
         - --enable-leader-election=true
         - --ignore-operation-annotation={{ .Values.ignoreOperationAnnotation }}
         - --workers=3
+        {{ if .Values.featureGates }}
+        {{- $featuregates := "" -}}
+        {{- range $feature, $value := $.Values.featureGates }}
+        {{- $featuregates = printf "%s%s=%t," $featuregates $feature $value }}
+        {{- end }}
+        - --feature-gates={{ $featuregates }}
```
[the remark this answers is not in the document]
aaronfern: I've added this change to the commit
However, it's not really required as feature gates can end with a comma. While parsing feature gates, it considers this case and handles it

aaronfern:
Thanks for your review @shreyas-s-rao!
I've addressed your comments. PTAL whenever possible

controllers/config.go
```diff
@@ -102,6 +112,56 @@ func InitFromFlags(fs *flag.FlagSet, cfg *ManagerConfig) {
 
 	cfg.SecretControllerConfig = &secret.Config{}
 	secret.InitFromFlags(fs, cfg.SecretControllerConfig)
+
+	return nil
+}
+
+// initFeatureGates initializes feature gates from the provided CLI flag set.
+func initFeatureGates(fs *flag.FlagSet, mgrcfg *ManagerConfig) error {
```
shreyas-s-rao, line 120:
```suggestion
func initFeatureGates(fs *flag.FlagSet, mgrCfg *ManagerConfig) error {
```

Using `mgrCfg` or just `cfg` should be absolutely fine. `mgrcfg` does not follow correct nomenclature because `mgr` and `cfg` are abbreviations of two different words, so the first letter of the second word needs to be capitalized.

controllers/etcd/config.go
```diff
@@ -15,9 +15,11 @@
 package etcd
 
 import (
-	"flag"
-
 	"github.com/gardener/etcd-druid/controllers/utils"
+	"github.com/gardener/etcd-druid/pkg/features"
+
+	flag "github.com/spf13/pflag"
```
shreyas-s-rao, line 21: Can you clarify why the default `flag` package is replaced with `pflag` package for this as well as other controller configs?

shreyas-s-rao (changes requested):
@aaronfern thanks for addressing my comments. I have a few nits, but overall LGTM

controllers/etcd/config.go
aaronfern: `pflag` was needed in order to add `--feature-gates` to the flagset as can be seen from the implementation of [`featuregate.AddFlag(fs)`](https://github.com/kubernetes/component-base/blob/v0.27.4/featuregate/feature_gate.go#L321)
Since `pflag` was being used, it made sense to use the same package throughout the repo instead of having multiple packages

shreyas-s-rao: Got it, thanks!

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -119,36 +119,48 @@ func GenerateValues(
...
-		command = append(command, "--backup-restore-tls-enabled=false")
-		command = append(command, fmt.Sprintf("--backup-restore-host-port=%s-local:8080", val.Name))
-		command = append(command, fmt.Sprintf("--etcd-server-name=%s-local", val.Name))
+
+		return command
 	}
 
-	return command
+	// safe to return an empty string array here since etcd-custom-image:v3.4.13-bootstrap-12 now uses an entry point that calls bootstrap.sh
+	return []string{}
```
unmarshall, line 163: have this at the top.
```go
if !val.UseEtcdWrapper {
   return []string{}
}
```
Helps with the readability of the code.

pkg/component/etcd/statefulset/values_helper.go
```diff
@@ -119,36 +119,48 @@ func GenerateValues(
...
 }
 
 func getEtcdCommand(val Values) []string {
-	command := []string{"" + "start-etcd"}
-
-	if val.ClientUrlTLS != nil {
-		dataKey := "ca.crt"
-		if val.ClientUrlTLS.TLSCASecretRef.DataKey != nil {
-			dataKey = *val.ClientUrlTLS.TLSCASecretRef.DataKey
+	if val.UseEtcdWrapper {
```
unmarshall, line 138: Some code duplication can be avoided, by doing the following:

```go
	if val.UseEtcdWrapper {
		//TODO @aaronfern: remove this feature gate when UseEtcdWraper becomes GA
		command := []string{"" + "start-etcd"}
		command = append(command, fmt.Sprintf("--backup-restore-host-port=%s-local:8080", val.Name))
		command = append(command, fmt.Sprintf("--etcd-server-name=%s-local", val.Name))

		if val.ClientUrlTLS == nil {
			command = append(command, "--backup-restore-tls-enabled=false")
		} else {
			dataKey := "ca.crt"
			if val.ClientUrlTLS.TLSCASecretRef.DataKey != nil {
				dataKey = *val.ClientUrlTLS.TLSCASecretRef.DataKey
			}
			command = append(command, "--backup-restore-tls-enabled=true")
			command = append(command, "--etcd-client-cert-path=/var/etcd/ssl/client/client/tls.crt")
			command = append(command, "--etcd-client-key-path=/var/etcd/ssl/client/client/tls.key")
			command = append(command, fmt.Sprintf("--backup-restore-ca-cert-bundle-path=/var/etcd/ssl/client/ca/%s", dataKey))
		}
		return command
	}

```

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -508,6 +489,29 @@ func (c *component) createOrPatch(ctx context.Context, sts *appsv1.StatefulSet,
...
 		}
+		if c.values.UseEtcdWrapper {
+			// sections to add only when using etcd wrapper
+			// TODO: @aaronfern add this back to sts.Spec when UseEtcdWrapper becomes GA
+			sts.Spec.Template.Spec.InitContainers = []corev1.Container{
+				{
+					Name:         "change-permissions",
+					Image:        "alpine:3.18.2",
+					Command:      []string{"sh", "-c", "--"},
+					Args:         []string{"chown", "-R", "65532:65532", "/var/etcd/data"},
```
shreyas-s-rao, line 500: @aaronfern , this does not actually work. It throws error:
```
> k logs etcd-test-2 -c change-permissions -f
BusyBox v1.36.1 (2023-06-02 00:42:02 UTC) multi-call binary.

Usage: chown [-RhLHPcvf]... USER[:[GRP]] FILE...

Change the owner and/or group of FILEs to USER and/or GRP

	-h	Affect symlinks instead of symlink targets
	-L	Traverse all symlinks to directories
	-H	Traverse symlinks on command line only
	-P	Don't traverse symlinks (default)
	-R	Recurse
	-c	List changed files
	-v	Verbose
	-f	Hide errors
```

You'll have to revert back to `Args: []string{"chown -R 65532:65532 /var/etcd/data"},`. Apologies for requesting this change without testing it.

controllers/config.go
```diff
@@ -61,6 +65,8 @@ type ManagerConfig struct {
 	DisableLeaseCache bool
 	// IgnoreOperationAnnotation specifies whether to ignore or honour the operation annotation on resources to be reconciled.
 	IgnoreOperationAnnotation bool
+	// FeatureFlags contains the feature gates to be used by etcd-druid.
+	FeatureFlags featuregate.MutableFeatureGate
```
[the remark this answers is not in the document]
shreyas-s-rao: I think you've reviewed an older commit of the PR. This has already been changed to `FeatureGates` in this PR.

charts/images.yaml
```diff
@@ -4,8 +4,8 @@ images:
...
   sourceRepository: github.com/gardener/etcd-backup-restore
   repository: eu.gcr.io/gardener-project/gardener/etcdbrctl
-  tag: "v0.25.0"
+  tag: "v0.24.2"
 - name: etcd
-  sourceRepository: github.com/gardener/etcd-wrapper
-  repository: eu.gcr.io/gardener-project/gardener/etcd-wrapper
-  tag: "v0.1.0"
+  sourceRepository: github.com/gardener/etcd-custom-image
+  repository: aaronfernandes/etcd-custom
```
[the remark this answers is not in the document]
aaronfern: Yes, this will be updated as soon as `etcd-custom-image:v3.4.13-bootstrap-12` is released
I will keep this comment open until then

controllers/config.go
```diff
@@ -15,7 +15,11 @@
 package controllers
 
 import (
-	"flag"
+	"fmt"
+
+	"github.com/gardener/etcd-druid/pkg/features"
+	flag "github.com/spf13/pflag"
+	"k8s.io/component-base/featuregate"
```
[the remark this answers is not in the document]
aaronfern: Yes, you're right
the cli argument `--feature-gates` is added [here](https://github.com/gardener/etcd-druid/pull/646/files#diff-d8ab9d4826fdce6b2b7f0fca1806011edb4c5a1f81c311cbc935f3b79252024bR125)

go.mod
```diff
@@ -12,12 +12,14 @@ require (
 	github.com/onsi/ginkgo/v2 v2.6.1
 	github.com/onsi/gomega v1.24.2
 	github.com/prometheus/client_golang v1.14.0
+	github.com/spf13/pflag v1.0.5
```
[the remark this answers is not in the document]
aaronfern: `pflag` was needed in order to add `--feature-gates` to the flagset as can be seen from the implementation of [`featuregate.AddFlag(fs)`](https://github.com/kubernetes/component-base/blob/v0.27.4/featuregate/feature_gate.go#L321)

main.go
```diff
@@ -25,6 +24,7 @@ import (
 
 	"github.com/gardener/etcd-druid/controllers"
 	"github.com/gardener/etcd-druid/pkg/version"
+	flag "github.com/spf13/pflag"
```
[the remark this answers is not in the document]
aaronfern: https://github.com/gardener/etcd-druid/pull/646#discussion_r1272558271

aaronfern:
Thanks for your reviews @unmarshall and @shreyas-s-rao 
I've addressed your comments. PTAL

controllers/etcd/config.go
```diff
@@ -51,3 +60,18 @@ func (cfg *Config) Validate() error {
 	}
 	return nil
 }
+
+// GetFeatureList returns feature gates relevant to the Etcd controller
+func (cfg *Config) GetFeatureList() []featuregate.Feature {
```
unmarshall, line 65: This method is no longer required.

controllers/etcd/config.go
```diff
@@ -51,3 +60,18 @@ func (cfg *Config) Validate() error {
...
+	return featureList
+}
+
+// CaptureFeatureActivations captures all feature gates required by the controller into controller config
+func (cfg *Config) CaptureFeatureActivations(fg featuregate.FeatureGate) {
+	if cfg.FeatureGates == nil {
+		cfg.FeatureGates = make(map[string]bool)
+	}
+	for _, feature := range cfg.GetFeatureList() {
```
unmarshall, line 74:
```suggestion
	for _, feature := range featureList {
```

controllers/etcdcopybackupstask/config.go
```diff
@@ -45,3 +54,18 @@ func (cfg *Config) Validate() error {
 	}
 	return nil
 }
+
+// GetFeatureList returns feature gates relevant to the EtcdCopyBackupTask controller
+func (cfg *Config) GetFeatureList() []featuregate.Feature {
```
unmarshall, line 59: This method is no longer required.

controllers/etcdcopybackupstask/config.go
```diff
@@ -45,3 +54,18 @@ func (cfg *Config) Validate() error {
...
+	return featureList
+}
+
+// CaptureFeatureActivations captures all feature gates required by the controller into controller config
+func (cfg *Config) CaptureFeatureActivations(fg featuregate.FeatureGate) {
+	if cfg.FeatureGates == nil {
+		cfg.FeatureGates = make(map[string]bool)
+	}
+	for _, feature := range cfg.GetFeatureList() {
```
unmarshall, line 68:
```suggestion
	for _, feature := range featureList {
```

controllers/compaction/config.go
```diff
@@ -70,3 +80,18 @@ func (cfg *Config) Validate() error {
 	}
 	return nil
 }
+
+// GetFeatureList returns feature gates relevant to the Compaction controller
+func (cfg *Config) GetFeatureList() []featuregate.Feature {
```
unmarshall, line 85: This method is no longer required.

controllers/compaction/config.go
```diff
@@ -70,3 +80,18 @@ func (cfg *Config) Validate() error {
...
+	return featureList
+}
+
+// CaptureFeatureActivations captures all feature gates required by the controller into controller config
+func (cfg *Config) CaptureFeatureActivations(fg featuregate.FeatureGate) {
+	if cfg.FeatureGates == nil {
+		cfg.FeatureGates = make(map[string]bool)
+	}
+	for _, feature := range cfg.GetFeatureList() {
```
unmarshall, line 94:
```suggestion
	for _, feature := range featureList {
```

unmarshall (approved):
Thanks for addressing all comments. Changes look good now.
/lgtm

shreyas-s-rao (approved):
/lgtm

shreyas-s-rao:
@aaronfern I ran the e2e tests locally with feature gates enabled, and I see that some of the tests fail since they are timing out. I've attached the test logs here.

[e2e-test-feature-gates.log](https://github.com/gardener/etcd-druid/files/12196896/e2e-test-feature-gates.log)

I see that the issue is resolved by doubling the values for timeouts `singleNodeEtcdTimeout` and `multiNodeEtcdTimeout` in the e2e tests. After this I ran the e2e test again and had a flake as well. Attaching logs from two runs of e2e tests, where one passed and one failed. PTAL

[e2e-test-feature-gates-changed-passed.log](https://github.com/gardener/etcd-druid/files/12197801/e2e-test-feature-gates-changed-passed.log)
[e2e-test-feature-gates-changed-failed.log](https://github.com/gardener/etcd-druid/files/12197802/e2e-test-feature-gates-changed-failed.log)

Please make the changes in this PR before we merge it. Thanks!

shreyas-s-rao:
/hold until e2e tests when feature-gates are enabled, are fixed.
I suspect it could be due to the additional init container that takes some time to run, in addition to the image pull time.

Issue #588, opened 2023-05-03
[Feature] Druid-controlled updates to the pods in the etcd cluster


shreyas-s-rao:
**Feature (What you would like to be added):**
Druid-controlled updates to the pods in the etcd cluster.

**Motivation (Why is this needed?):**
Currently, druid deploys the etcd cluster as a statefulset with number of replicas set to the desired number of members in the etcd cluster. The `spec.updateStrategy` of this statefulset is set to `RollingUpdate`, which allows the statefulset controller to  roll the etcd pods one after the other in a rolling fashion. The order of updating each of pod is deterministic - from largest ordinal to the smallest, as per the [documentation](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates). This works fine for a perfectly healthy etcd cluster, but poses a risk for a multi-node etcd cluster with an unhealthy pod.

Consider a 3-member etcd cluster `etcd-main`, with pods `etcd-main-0`, `etcd-main-1` and `etcd-main-2` running, in the pink of health. At this point, if `etcd-main-0` (or `etcd-main-1`) becomes unhealthy due to multiple reasons (network connectivity issues, zone outages, node failure, or simply an etcd issue which might be resolvable by restarting the pod), then the etcd cluster still maintains quorum with the other two healthy members, but is now only one step away from losing quorum. What happens now if there's an update to the etcd statefulset spec, like a change in `etcd-backup-restore` image version or a configuration change to the `etcd` or `etcdbrctl` processes? The statefulset controller starts rolling the pods starting with `etcd-main-2`. As soon as it deletes this pod to make room for the updated pod, the cluster loses quorum. This leads to a downtime of the etcd, subsequently causing a downtime to the kube-apiserver whom the etcd is backing, until the updated `etcd-main-2` pod comes back up.

This is an artificially introduced quorum loss scenario, which can be entirely avoided if druid takes control of the order of updation of the etcd pods.

**Approach/Hint to the implement solution (optional):**

[...]

PR #647, opened 2023-07-20
Improved scale-up detection.

ishan16696:
/area high-availability
/kind technical-debt

While testing this PR: https://github.com/gardener/etcd-backup-restore/pull/649 I found out that when third member of etcd cluster won't come up but pod spec has been updated, so etcd statefulset's status `.status.updatedReplicas` becomes `3`:
```
status:
  availableReplicas: 2
  collisionCount: 0
  currentReplicas: 1
  currentRevision: etcd-main-79f88bd678
  observedGeneration: 2
  readyReplicas: 2
  replicas: 3
  updateRevision: etcd-main-86b48fdf9c
  updatedReplicas: 3
```
druid remove the scale-up annotation: `gardener.cloud/scaled-to-multi-node` from etcd statefulset as 
[func clusterScaledUpToMultiNode(){...}](https://github.com/ishan16696/etcd-druid/blob/b4d722f84abb6d9fa83e7c0a8675bcd49e6e514e/pkg/component/etcd/statefulset/statefulset.go#L391) has returned false which is not correct.
more info please refer to this: https://github.com/gardener/etcd-druid/pull/587#discussion_r1183231059, now it won't be an issue as backup-restore will skip checking scale-up conditions for first member

Fixes # https://github.com/gardener/etcd-druid/pull/587#discussion_r1182771744

**Release note**:
```other operator
None
```

### DIFF
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -391,7 +391,8 @@ func immutableFieldUpdate(sts *appsv1.StatefulSet, val Values) bool {
 func clusterScaledUpToMultiNode(val *Values, sts *appsv1.StatefulSet) bool {
 	if sts != nil && sts.Spec.Replicas != nil {
 		return (val.Replicas > 1 && *sts.Spec.Replicas == 1 && sts.Status.AvailableReplicas == 1) ||
-			(metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) && sts.Status.UpdatedReplicas < *sts.Spec.Replicas)
+			(metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) &&
+				(sts.Status.UpdatedReplicas < *sts.Spec.Replicas || sts.Status.AvailableReplicas < sts.Status.UpdatedReplicas))
 	}
 	return val.Replicas > 1 && val.StatusReplicas == 1
 }


pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -391,7 +391,8 @@ func immutableFieldUpdate(sts *appsv1.StatefulSet, val Values) bool {
 func clusterScaledUpToMultiNode(val *Values, sts *appsv1.StatefulSet) bool {
 	if sts != nil && sts.Spec.Replicas != nil {
 		return (val.Replicas > 1 && *sts.Spec.Replicas == 1 && sts.Status.AvailableReplicas == 1) ||
-			(metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) && sts.Status.UpdatedReplicas < *sts.Spec.Replicas)
+			(metav1.HasAnnotation(sts.ObjectMeta, scaleToMultiNodeAnnotationKey) &&
+				(sts.Status.UpdatedReplicas < *sts.Spec.Replicas || sts.Status.AvailableReplicas < sts.Status.UpdatedReplicas))
```
shreyas-s-rao, line 395: I did not fully get this.

During scale-up, this is the sequence (consider `minReadySeconds=0`, which is the default in druid. This means that a pod that is `ready` is also considered as `available`):
1. Pod `etcd-0` is ready. Statefulset `spec.replicas=1`, `status.availableReplicas=1`, `status.updatedReplicas=1`
2. Statefulset `spec.replicas=3`:
    a. Pods `etcd-1` and `etcd-2` are unready. Statefulset `status.availableReplicas=1` (only `etcd-0` is available), `status.updatedReplicas=2` (since both `etcd-1` and `etcd-2` have the latest spec from the sts template)
    b. Pod `etcd-1` becomes ready. Statefulset `status.availableReplicas=2` (both `etcd-0` and `etcd-1` are available), `status.updatedReplicas=2` (does not change yet)
    c. Pod `etcd-2` becomes ready. Statefulset `status.availableReplicas=3` (All 3 pods are available), `status.updatedReplicas=2` (does not change yet)
    d. Pod `etcd-0` is updated with latest sts spec. Briefly, statefulset `status.availableReplicas=2` (pod `etcd-0` is not available), `status.updatedReplicas=3` (since all 3 pods have the updated spec from sts template)
    e. All 3 pods are ready. Statefulset `status.availableReplicas=3` (all 3 pods are available), `status.updatedReplicas=3`

In this entire process, the condition `status.availableReplicas < status.updatedReplicas` holds true for both **2a** and **2d**. I understand that you would like druid to not add the scaled-to-multinode annotation when the pods `etcd-1` and `etcd-2` are available. But this logic here may not really help since you're using an OR (`||`) for the condition, meaning it will never be evaluated in such a case since `status.updatedReplicas` is already lesser than `spec.replicas` during a scale-up case before pod `etcd-0` is updated, and if `status.updatedReplicas` ever equals `spec.replicas`, it means that all the pods have been updated with latest spec changes, so it's already too late to remove the sts annotation now. Rather, relying on this will only lead to race conditions. We need a better way for druid to be able to deterministically remove the annotation. I feel [druid-controlled pod updates](https://github.com/gardener/etcd-druid/issues/588) will certainly solve this problem, since druid will have complete control on when to restart the pod, and can ensure that it has removed the sts annotation before restarting the pod `etcd-0`. 
cc @unmarshall another possible use-case using druid-controlled pod updates.

@ishan16696 can you think of a more deterministic solution for removing the sts annotation before `etcd-0` pod is restarted?

shreyas-s-rao (changes requested):
@ishan16696 thanks for the PR. I feel this may not help solve the corner-case issue we see with scale-up today. Can you please address my comment? Thanks.

pkg/component/etcd/statefulset/statefulset.go
ishan16696: > I feel this may not help solve the corner-case issue we see with scale-up today. 

yes, this is not to fix those edge cases. This is just a technical debt which should be there to cover every scenario.

> relying on this will only lead to race conditions

what makes you think so, you can checked the deployment code or statefulset code they also do the similar check inorder to check the rolling update.  
https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/rollout_status.go#L86


> But this logic here may not really help since you're using an OR (||) for the condition, meaning it will never be evaluated in such a case since status.updatedReplicas is already lesser than spec.replicas


Yes, I agree this will not help in the scenario which you have mentioned as etcd pod-0 spec needs to be updated at the end of scale-up  but consider the case when etcd pod-0 spec doesn't need to be updated eg:
In case of etcd cluster already running with `replicas=3` then permanent quorum loss happens, then to recover from permanent quorum loss scale down the etcd statefulset `spec.replica` to  `0` then scale up to  `1` then to `3` , in this case of recovering from quorum loss while scaling up from replicas `1 -> 3` we are actually using scale-up feature but at the same time `etcd pod-0` spec doesn't need to be updated as it is already updated as it was already running with `cluster size=3`.

As I mentioned in the PR please refer to this comment : https://github.com/gardener/etcd-druid/pull/587#discussion_r1182771744 pointed out by @unmarshall

ishan16696:
@unmarshall can you also take a look as I'm addressing your comments in this PR. Thanks

pkg/component/etcd/statefulset/statefulset.go
shreyas-s-rao: >  deployment code or statefulset code they also do the similar check inorder to check the rolling update.
https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/rollout_status.go#L86

Looks like they're only checking this condition to update the status, and not take any action. Again, since the pods are rolled by the statefulset controller, this controller has full control on when to roll the pods, based on evaluated conditions. Since druid is simply reacting to this, druid currently has no control on the individual pod updates, and has to manage to deterministically perform an action in the middle of a rollout, which I feel is not easy to achieve deterministically. Not unless druid has full control over the pod rollout process.

> in this case of recovering from quorum loss while scaling up from replicas 1 -> 3 we are actually using scale-up feature but at the same time etcd pod-0 spec doesn't need to be updated as it is already updated as it was already running with cluster size=3.

Not entirely true. When the statefulset is scaled down to 0 and then scaled to 1 replica, the etcd cluster has only 1 member, hence it is treated as a single-node etcd **until** it is scaled up to 3 replicas. So the exact same flow applies for quorum loss recovery case as well, and `pod-0` will be restarted at the end, after `pod-1` and `pod-2` come up and become ready.

ishan16696: > Looks like they're only checking this condition to update the status, and not take any action. 

we are also not any taking action, we are interested in checking whether the scale-up rollout was successful or not, so that we can decide whether to remove scale-up annotation or keep the scale-up annotation in etcd statefulset as scale-up might not succeeded yet.

> Not entirely true. When the statefulset is scaled down to 0 and then scaled to 1 replica, the etcd cluster has only 1 member, hence it is treated as a single-node etcd until it is scaled up to 3 replicas. So the exact same flow applies for quorum loss recovery case as well, and pod-0 will be restarted at the end, after pod-1 and pod-2 come up and become ready.

nopes, it's not the scale. Scale-up a non-HA cluster and scale-up used in recovering permanent quorum loss (provided cluster was healthy before) is little bit different. Let me explain this:
1. When you are scaling-up a non-HA cluster, `etcd pod-0` spec needs to updated hence it restart at the end (after `etcd pod-1` and `etcd pod-2` becomes healthy)
2. In case of recovery from permanent quorum loss provided etcd cluster was formed earlier, we actually add ignore-reconciliation to etcd CR, and then scale down the etcd statefulset `spec.replica` to `0` then scale up to `1` then to `3` , in this case of recovering from quorum loss while scaling up from replicas `1 -> 3`  here `etcd pod-0` spec doesn't need to be updated as it is already updated as it was already running with cluster size=3. I have simulated this in my shoot. You can also try you won't see `etcd pod-0` restarting.

```bash
~ ❯ k get pods -w | grep etcd-main       ○ garden--aws-ha-external/shoot--i539698--test-e2e 07:43:57 PM
etcd-main-0                                           0/2     Pending            0               1s
etcd-main-0                                           0/2     Pending            0               5s
etcd-main-0                                           0/2     ContainerCreating   0               5s
etcd-main-0                                           0/2     ContainerCreating   0               9s
etcd-main-0                                           1/2     Running             0               10s
etcd-main-0                                           2/2     Running             0               2m46s
etcd-main-1                                           0/2     Pending             0               1s
etcd-main-2                                           0/2     Pending             0               0s
etcd-main-1                                           0/2     Pending             0               5s
etcd-main-1                                           0/2     ContainerCreating   0               6s
etcd-main-2                                           0/2     Pending             0               4s
etcd-main-2                                           0/2     ContainerCreating   0               4s
etcd-main-2                                           0/2     ContainerCreating   0               12s
etcd-main-1                                           0/2     ContainerCreating   0               14s
etcd-main-2                                           1/2     Running             0               13s
etcd-main-1                                           1/2     Running             0               16s
etcd-main-2                                           2/2     Running             0               42s
etcd-main-1                                           2/2     Running             0               69s
```

there is no restart of `etcd-main-0` in recovery from permanent quorum loss.

shreyas-s-rao: > we are also not any taking action, we are interested in checking whether the scale-up rollout was successful or not, so that we can decide whether to remove scale-up annotation or keep the scale-up annotation in etcd statefulset as scale-up might not succeeded yet.

The fact that *we can decide whether to remove scale-up annotation* itself indicates that druid might take some functional action based on this condition. So being deterministic is key here.

> etcd pod-0 spec doesn't need to be updated as it is already updated as it was already running with cluster size=3

But does this mean that when only `pod-0` is up (replicas=1), at that point etcd cluster size = 3? That means there's still no quorum unless either of `pod-1` or `pod-2` come up, right?

unmarshall (approved):
/lgtm

shreyas-s-rao:
MoM of out-of-band call between @ishan16696 ,@unmarshall and myself:
- this PR is **not** for fixing or mitigating https://github.com/gardener/etcd-backup-restore/issues/646 
- it is a result of [this discussion](https://github.com/gardener/etcd-druid/pull/587#discussion_r1182771744)
- this PR simply adds an extra layer of protection to prevent wrongful detection of scale-up in a multinode etcd cluster
- we agreed that this code will go away once we introduce deterministic scale-up of the etcd cluster, either via #588 or #206 (possibly `EtcdMember.Spec.InitialClusterState`), which needs further discussion

shreyas-s-rao (approved):
/lgtm

pkg/component/etcd/statefulset/statefulset.go
shreyas-s-rao: Review comment can be closed. Refer to https://github.com/gardener/etcd-druid/pull/647#issuecomment-1661489035

Issue #649, opened 2023-07-29
[Feature] Extend `Etcd` CRD to Include `DeltaSnapshotRetentionPeriod` Field


seshachalam-yv:
**Feature (What you would like to be added):**
Add the `DeltaSnapshotRetentionPeriod` field to the `Etcd` CRD's `BackupSpec` struct.

**Motivation (Why is this needed?):**

The `etcd-backup-restore` component has recently introduced a `--delta-snapshot-retention-period` flag ([PR #640](https://github.com/gardener/etcd-backup-restore/pull/640)). This flag allows users to customize the retention period for delta snapshots, providing more flexibility for both Exponential and Limit-Based GC policies. Currently, the `Etcd` CRD in `etcd-druid` does not support this new flag. By adding this field, we allow this configuration to be propagated from the `Etcd` CRD to the `etcd-backup-restore` component.

**Approach/Hint to the implement solution (optional):**
- Add the `DeltaSnapshotRetentionPeriod` field to the `BackupSpec` struct in the `etcd_types.go` file.
- Update the CRD validation schema to include the new field.

PR #651, opened 2023-07-29
Extend `Etcd` CRD and `etcd-druid` to Support Configurable Delta Snapshot Retention

seshachalam-yv:
/area backup
/kind api-change

**What this PR does / why we need it:**
This PR extends the Etcd CRD to include a `DeltaSnapshotRetentionPeriod` field. This new field allows users to define the duration for retaining old delta snapshots, providing more flexibility in managing backup retention. Additionally, this PR modifies etcd-druid to propagate the --delta-snapshot-retention-period flag from the Etcd CRD to the etcd-backup-restore component.

This change allows the etcd-backup-restore to have access to the `DeltaSnapshotRetentionPeriod` configuration which was not previously propagated. As such, it provides greater flexibility in the configuration of backup parameters.

**Which issue(s) this PR fixes:**
Fixes https://github.com/gardener/etcd-druid/issues/649

**Special notes for your reviewer:**

**Release note:**
```feature operator
Introduce `Spec.Backup.DeltaSnapshotRetentionPeriod` in the `Etcd` resource to allow configuring retention period for delta snapshots.
```

### DIFF
--- a/api/v1alpha1/types_etcd.go
+++ b/api/v1alpha1/types_etcd.go
@@ -179,9 +179,9 @@ type BackupSpec struct {
 	// +optional
 	DeltaSnapshotMemoryLimit *resource.Quantity `json:"deltaSnapshotMemoryLimit,omitempty"`
 	// DeltaSnapshotRetentionPeriod defines the duration for which delta snapshots will be retained, excluding the latest snapshot set.
-	// The value should be a string formatted as a duration (e.g., '1s', '2m', '3h', '4d') with a minimum value of 1 second.
+	// The value should be a string formatted as a duration (e.g., '1s', '2m', '3h', '4d')
 	// +kubebuilder:validation:Type=string
-	// +kubebuilder:validation:Pattern="^([1-9][0-9]*([.][0-9]+)?(s|m|h|d))+$"
+	// +kubebuilder:validation:Pattern="^([0-9][0-9]*([.][0-9]+)?(s|m|h|d))+$"
 	// +optional
 	DeltaSnapshotRetentionPeriod *metav1.Duration `json:"deltaSnapshotRetentionPeriod,omitempty"`
 
--- a/api/v1alpha1/types_etcd.go
+++ b/api/v1alpha1/types_etcd.go
@@ -178,9 +178,10 @@ type BackupSpec struct {
 	// DeltaSnapshotMemoryLimit defines the memory limit after which delta snapshots will be taken
 	// +optional
 	DeltaSnapshotMemoryLimit *resource.Quantity `json:"deltaSnapshotMemoryLimit,omitempty"`
-	// DeltaSnapshotRetentionPeriod defines the duration to retain old delta snapshots. Delta snapshots within this
-	// duration will not be deleted. This provides more flexibility for backup retention by allowing older delta snapshots
-	// to be preserved.
+	// DeltaSnapshotRetentionPeriod defines the duration for which delta snapshots will be retained, excluding the latest snapshot set.
+	// The value should be a string formatted as a duration (e.g., '1s', '2m', '3h', '4d') with a minimum value of 1 second.
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([1-9][0-9]*([.][0-9]+)?(s|m|h|d))+$"
 	// +optional
 	DeltaSnapshotRetentionPeriod *metav1.Duration `json:"deltaSnapshotRetentionPeriod,omitempty"`
 
--- a/api/v1alpha1/types_etcd.go
+++ b/api/v1alpha1/types_etcd.go
@@ -178,6 +178,12 @@ type BackupSpec struct {
 	// DeltaSnapshotMemoryLimit defines the memory limit after which delta snapshots will be taken
 	// +optional
 	DeltaSnapshotMemoryLimit *resource.Quantity `json:"deltaSnapshotMemoryLimit,omitempty"`
+	// DeltaSnapshotRetentionPeriod defines the duration to retain old delta snapshots. Delta snapshots within this
+	// duration will not be deleted. This provides more flexibility for backup retention by allowing older delta snapshots
+	// to be preserved.
+	// +optional
+	DeltaSnapshotRetentionPeriod *metav1.Duration `json:"deltaSnapshotRetentionPeriod,omitempty"`
+
 	// SnapshotCompression defines the specification for compression of Snapshots.
 	// +optional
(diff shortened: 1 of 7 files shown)


shreyas-s-rao (changes requested):
@seshachalam-yv thanks for the PR. While the PR adds the new field to the `Etcd` CR, druid does not use the value anywhere. Please ensure that druid creates the etcd statefulset with this field's value correctly passed to the `etcdbrctl` command.

api/v1alpha1/types_etcd.go
```diff
@@ -178,6 +178,13 @@ type BackupSpec struct {
 	// DeltaSnapshotMemoryLimit defines the memory limit after which delta snapshots will be taken
 	// +optional
 	DeltaSnapshotMemoryLimit *resource.Quantity `json:"deltaSnapshotMemoryLimit,omitempty"`
+	// DeltaSnapshotRetentionPeriod defines the duration for which delta snapshots will be retained, excluding the latest snapshot set.
+	// The value should be a string formatted as a duration (e.g., '1s', '2m', '3h', '4d') with a minimum value of 1 second.
```
shreyas-s-rao, line 182: Why do we set minimum value as `1s`? If this field is not set at all (since it's `optional`), then the default value for this would be 0 in etcd-backup-restore, according to https://github.com/gardener/etcd-backup-restore/blob/master/doc/usage/garbage_collection.md. So why would we want to disallow setting `0s` in druid then?

api/v1alpha1/types_etcd.go
```diff
@@ -178,6 +178,13 @@ type BackupSpec struct {
...
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([1-9][0-9]*([.][0-9]+)?(s|m|h|d))+$"
```
shreyas-s-rao, line 184: Thanks for the additional validation for the value.
Let's look at the following cases:
1. `0s`, `0d` (zero-values) - not allowed. Please check previous comment for why these values should be allowed.
2. `5d0h`, `2h0m` (non-zero values with zero-value items) - not allowed. These are perfectly valid durations, according to [metav1.Duration code](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/duration_test.go), and should be allowed.
3. `0.5d`, `0.8h` (fractional values less than 1) - not allowed. These could be valid usecases where user would like to set fractional values.
4. `5s6s7s` - allowed. Does this result in final duration value of `18s`, ie, are the elements summed up as is without checking if a certain time-unit is repeated?

Overall, I would prefer if zero-value elements are allowed in the value, to conform with standard metav1.Duration. And since etcd-backup-restore also allows the value to be zero (by default too), it's ok to allow setting this in the `Etcd` resource as well.

pkg/component/etcd/statefulset/statefulset_test.go
```diff
@@ -927,3 +930,50 @@ func checkLocalProviderVaues(etcd *druidv1alpha1.Etcd, sts *appsv1.StatefulSet,
...
+		fmt.Sprintf("%s=%s", "--etcd-snapshot-timeout", values.EtcdSnapshotTimeout.Duration.String()):         Equal(fmt.Sprintf("%s=%s", "--etcd-snapshot-timeout", values.EtcdSnapshotTimeout.Duration.String())),
+		fmt.Sprintf("%s=%s", "--etcd-defrag-timeout", values.EtcdDefragTimeout.Duration.String()):             Equal(fmt.Sprintf("%s=%s", "--etcd-defrag-timeout", values.EtcdDefragTimeout.Duration.String())),
+		fmt.Sprintf("%s=%s", "--delta-snapshot-lease-name", values.DeltaSnapLeaseName):                        Equal(fmt.Sprintf("%s=%s", "--delta-snapshot-lease-name", values.DeltaSnapLeaseName)),
+		fmt.Sprintf("%s=%s", "--full-snapshot-lease-name", values.FullSnapLeaseName):                          Equal(fmt.Sprintf("%s=%s", "--full-snapshot-lease-name", values.FullSnapLeaseName)),
+	}
+
+	if values.DeltaSnapshotRetentionPeriod != nil {
+		elements[fmt.Sprintf("--delta-snapshot-retention-period=%s", values.DeltaSnapshotRetentionPeriod.Duration.String())] = Equal(fmt.Sprintf("--delta-snapshot-retention-period=%s", values.DeltaSnapshotRetentionPeriod.Duration.String()))
+		println("Adding the check.....")
```
shreyas-s-rao, line 976: Remove

shreyas-s-rao (changes requested):
@seshachalam-yv thanks for your PR. I have a few comments. PTAL.

Please also change the release note to something like:
```
Introduce `Spec.Backup.DeltaSnapshotRetentionPeriod` in the `Etcd` resource to allow configuring retention period for delta snapshots.
```
(Release note does not need to have implementation details of the API, but just the general usage)

api/v1alpha1/types_etcd.go
seshachalam-yv: Yeah Shreyas, permitting zero values does make sense. Thank you for the suggestion.
 Fixed as part of 4f0adfd3eac2c2011c5fb34ace6c83ffc128adf9

pkg/component/etcd/statefulset/statefulset_test.go
seshachalam-yv: Removed

api/v1alpha1/types_etcd.go
```diff
@@ -178,6 +178,13 @@ type BackupSpec struct {
 	// DeltaSnapshotMemoryLimit defines the memory limit after which delta snapshots will be taken
 	// +optional
 	DeltaSnapshotMemoryLimit *resource.Quantity `json:"deltaSnapshotMemoryLimit,omitempty"`
+	// DeltaSnapshotRetentionPeriod defines the duration for which delta snapshots will be retained, excluding the latest snapshot set.
+	// The value should be a string formatted as a duration (e.g., '1s', '2m', '3h', '4d')
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([0-9][0-9]*([.][0-9]+)?(s|m|h|d))+$"
```
shreyas-s-rao, line 184: @seshachalam-yv one more nit here. The current pattern allows values with multiple decimal points, such as `2.5.8.4d`, which is invalid. Please limit the number of decimal points to 0 or 1, with 1 or more digits following it only if the decimal point exists.

shreyas-s-rao (approved):
/lgtm

Issue #361, opened 2022-06-27
[BUG] One member of Multi-node Etcd start its own cluster as its data-dir structure is invalid which triggers the restoration of single member.


timuthy:
**Describe the bug:**
When creating a new `etcd` multi-node cluster it can happen that **multiple clusters** are created instead of one cluster with multiple members. As a result, consumers get a single `etcd-client` service which is backed by distinct clusters.

Etcd-Druid and all involved components use a static bootstrapping procedure as described [here](https://etcd.io/docs/v3.4/op-guide/clustering/#static).
In combination with the StatefulSet's [parallel podManagementPolicy](https://github.com/gardener/etcd-druid/blob/4da816d7247971c5f47cb529b7df2894cb21e1bd/charts/etcd/templates/etcd-statefulset.yaml#L19) it sometimes happen that two or more pods don't detect an existing cluster which makes them to create a new cluster (due to `--initial-cluster-state new`) at the same time.

In this example we saw it for two pods which belonged to the same StatefulSet (please also note the version difference):

- etcd-main-0
```
2022-06-24T15:08:15.87671756Z stderr F 2022-06-24 15:08:15.876585 I | etcdserver: setting up the initial cluster version to 3.0
```

- etcd-main-2
```
2022-06-24T15:08:23.391710822Z stderr F 2022-06-24 15:08:23.391621 I | etcdserver: setting up the initial cluster version to 3.4
```

After all the pods were running we constantly observed logs like:

```
2022-06-24T15:08:25.488303476Z stderr F 2022-06-24 15:08:25.488156 E | rafthttp: request sent was ignored (cluster ID mismatch: peer[1e2acb0b00e637d5]=cdf818194e3a8c32, local=dd673093ade40ad6)
2022-06-24T15:08:25.488463226Z stderr F 2022-06-24 15:08:25.488363 E | rafthttp: request sent was ignored (cluster ID mismatch: peer[1e2acb0b00e637d5]=cdf818194e3a8c32, local=dd673093ade40ad6)
2022-06-24T15:08:25.565269748Z stderr F 2022-06-24 15:08:25.564967 E | rafthttp: request sent was ignored (cluster ID mismatch: remote[1e2acb0b00e637d5]=cdf818194e3a8c32, local=dd673093ade40ad6)
[...]

PR #658, opened 2023-08-07
[DEP-04] EtcdMember Custom Resource

### LINKED ISSUE #206: ☂️ [Feature] Add EtcdMember resource

shreyas-s-rao:

/area quality robustness
/kind enhancement

This PR introduces a new `EtcdMember` Custom Resource via DEP-04.

Fixes #206

**Release note**:

```doc operator
Introduce DEP-04 [EtcdMember Custom Resource](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/04-etcd-member-custom-resource.md).
```


docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+Every etcd-member has a unique `memberID` and it is part of an etcd cluster which has a unique `clusterID`. In a well-formed etcd cluster every member must have the same `clusterID`. Publishing this information to druid helps in identifying issues when one or more etcd-members form their own individual clusters, thus resulting in multiple clusters where only one was expected. Issues [Issue#361](https://github.com/gardener/etcd-druid/issues/361) and [Issue#419](https://github.com/gardener/etcd-druid/issues/419) are some such occurrences.
+
+Today, this information is published by using a member [lease](https://kubernetes.io/docs/concepts/architecture/leases/). Both these fields are populated in the leases' `Spec.HolderIdentity` by the backup-sidecar container. 
+
+The authors propose to publish member metadata information in `EtcdMember` resource.
+
+```yaml
+id: <etcd-member id>
+clusterID: <etcd cluster id>
```
timuthy, line 102:
```suggestion
status
  id: <etcd-member id>
  clusterID: <etcd cluster id>
```

I suppose it was meant to be part of the status?!

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+At present, these states are not captured and only the final state is known - i.e the etcd-member either fails to come up (all re-attempts to bring up the pod via the StatefulSet controller has exhausted) or it comes up. Getting an insight into all its state transitions would help in diagnostics.
+
+The status of an etcd-member at any given point in time can be best categorized as a combination of a top-level `State` and a `SubState`. The authors propose to introduce the following states and sub-states:
+
+#### States and Sub-States
+
+> **NOTE**: Abbreviations have been used wherever possible, only to represent sub-states. These representations are chosen only for brevity and will have proper longer names.
```
timuthy, line 119: What does that mean, will they be part of the states? TBH, I find the listed abbreviations rather weird and doesn't help the operator much.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+Following DFA represents top level state transitions (without any representation of sub-states). As described in the table above there are 4 top level states:
+
+* `New`- this is a start state for all newly created etcd-members
+
+* `Initializing` - In this state backup-restore will perform pre-requisite actions before it triggers the start of an etcd process. DB validation and optionally restoration is done in this state. Possible sub-states are: `DBValidationSanity`, `DBValidationFully` and `Restoration`
+
+* `Starting` - Once the optional initialization is done backup-restore will trigger the start of an etcd process. It can either directly go to `Learner` sub-state or wait for getting added as a learner and therefore be in `PendingLearner` sub-state.
+
+* `Started` - In this state the etcd-member is a full voting member. It can either be in `Leader` or `Follower` sub-states.
```
timuthy, line 144: WDYT of `Running` instead?

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+### Etcd Member State Transitions
+
+Each etcd-member goes through different `States` during its lifetime. `State` is a derived high-level summary of where an etcd-member is in its lifecycle. A `SubState` gives additional information about the state. This proposal extends the concept of states with the notion of a `SubState`, since `State` indicates a top-level state of an `EtcdMember` resource, which can have one or more `SubState`s.
+
+While `State` is sufficient for many human operators, the notion of a `SubState` provides operators with an insight about the discrete stage of an etcd-member in its lifecycle. For example, consider a top-level `State: Starting`, which indicates that an etcd-member is starting. `Starting` is meant to be a transient state for an etcd-member. If an etcd-member remains in this `State` longer than expected, then an operator would require additional insight, which the authors propose to provide via `SubState` (in this case, the possible `SubStates` could be `PendingLearner` and `Learner`, which are detailed in the following sections).
```
timuthy, line 111: Is the `SubState` meant for operators only? If yes, how about providing a field `description` instead?

timuthy (commented):
Some first feedback, as I didn't make it through yet 🙂

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: We have deliberately not added this under status. Each section (per use case) only captured what information we wish to capture. Then in the section `Spec Vs Status` we make it clear the reasoning to include all of the above as part of `Status` and not `Spec`.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: This is going to be part of the state machine that we are going to create. It is not only meant for developers but also will be consumed by druid.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: In the diagrams it was easier to just use some abbreviations and these will never be used in the code. If we had to use the full-form then each node in the DFA would become very big. So this is only for making the representation a bit easier. If it is confusing then we can change that or add some additional comment to make it absolutely clear that these will never be used elsewhere other than their limited scope in the diagrams.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: We chose `Started` as it was a natural succession state after `Starting` but there is no hard and fast rule. `Running` is semantically same as `Started` and could be considered.

docs/proposals/04-etcd-member-custom-resource.md
timuthy: Okay, thanks.

docs/proposals/04-etcd-member-custom-resource.md
timuthy: Nice. My point is that it's visually nicer to tell the difference:
- `starting` - `started`
vs.
- `starting` - `running`

docs/proposals/04-etcd-member-custom-resource.md
timuthy: Got it. While this is valid I think that `state`  or especially `substate` is not very intuitive and consistent compared to other `status` subresources of Gardener or Kubernetes objects. Maybe you find some field names that are closer to our ecosystem?

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+It has been agreed that the current solution to [this]([[BUG] Restart of first cluster member can lead to quorum loss while scaling up a non-HA etcd cluster. · Issue #646 · gardener/etcd-backup-restore · GitHub](https://github.com/gardener/etcd-backup-restore/issues/646)) issue is a quick and dirty fix and needs to be revisited to be uniformly applied to all etcd-members. The authors propose to provide a more deterministic approach to scale-up using the `EtcdMember` resource.
+
+**New approach**
+
+Instead of adding an annotation `gardener.cloud/scaled-to-multi-node` on the `StatefulSet`, a new annotation `druid.gardener.cloud/create-as-learner` should be added by druid on an `EtcdMember` resource. This annotation will only be added to newly created members during scale-up.
+
+Each etcd-member should look at the following to deterministically compute the `bootstrap information` specified above:
+
+* `druid.gardener.cloud/create-as-learner` annotation on its respective `EtcdMember` resource. This new annotation will be honored in the following cases:
```
timuthy, line 198: Can you explain why we need this `druid.gardener.cloud/create-as-learner` annotation? From reading the text I got that a single member cluster has one `EtcdMember` resource and when scaling up, one could simply create further `EtcdMember` resources to trigger the scale-up.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+At present, information about backed-up snapshot (*only* `latest-revision-number`) is published by leading-backup-sidecar container by updating `Spec.HolderIdentity` of the delta-snapshot and full-snapshot [lease](https://kubernetes.io/docs/concepts/architecture/leases/)s.
+
+Druid maintains `conditions` in the `Etcd` resource status, which include but are not limited to maintaining information on whether backups being taken for an etcd cluster are healthy (up-to-date) or stale (outdated in context to a configured schedule). Druid computes these conditions using information from full/delta snapshot leases.
+
+In order to provide a holistic view of the health of backups to human operators, druid requires additional information about the snapshots that are being backed-up. The authors propose to not misuse [lease](https://kubernetes.io/docs/concepts/architecture/leases/)s and instead publish the following snapshot information as part `EtcdMember` custom resource:
+
+```yaml
+snapshots:
```
timuthy, line 234: IIRC, only the leader backs up the data and since a leader change over time is possible, different `EtcdMember` resources have different status information about snapshots.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+The authors propose to enhance each etcd-member to regularly publish information about the `dbSize` and ` dbSizeInUse` so that druid may trigger defragmentation for the etcd cluster. 
+
+```yaml
+dbSize: <db-size> # e.g 6Gi
+dbSizeInUse: <db-size-in-use> # e.g 3.5Gi
+```
+
+Difference between `dbSize` and `dbSizeInUse` gives a clear indication of how much storage space would be freed up if a defragmentation is performed. If the difference is not significant (based on a configurable threshold provided to druid), then no defragmentation should be performed. This will ensure that druid does not perform frequent defragmentations that do not yield much benefit. Effectively it is to maximise the benefit of defragmentation since this operations involves transient downtime for each etcd-member.
```
timuthy, line 292: Will the `defragmentationSchedule` still be considered for defragmentation? If not, I'm a bit worried about single-node clusters.
While mutli-node clusters it's possible to execute defragmentation in a rolling fashion and more precautionary, it's a difficult balance for single-node clusters because of the involved downtime.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+As discussed in the previous section, every etcd-member is defragmented periodically, and can also be defragmented based on the DB size reaching a certain threshold. It is beneficial for druid to have knowledge of this data from each etcd-member for the following reasons:
+
+* [**Diagnostics**] Druid can aggregate data about every member's defragmentations and calculate the frequency of defragmentations on an etcd cluster. Combined with details about the triggers for these defragmentations, druid can derive insights and raise alerts about the cause for frequent defragmentations, such as possible DB spikes.
+
+* [**Operational**] Derive status of defragmentation at etcd cluster level. In case of partial failures for a subset of etcd-members druid can potentially re-trigger defragmentation only for those etcd-members.
+
+The authors propose to capture this information as part of `defragmentations` section in the `EtcdMember` resource.
+
+```yaml
+defragmentations:
```
timuthy, line 305: How long will this information be retained and will there be some garbage collection for older defragmentation runs? I'm wondering, this "historical data" is usually more something that is exposed by metrics and scraped by Prometheus.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+  - apiVersion: druid.gardener.cloud/v1alpha1
+    blockOwnerDeletion: true
+    controller: true
+    kind: Etcd
+    name: <name of the parent Etcd resource>
+    uid: <UID of the parent Etcd resource> 
+status:
+  id: <etcd-member id>
+  clusterID: <etcd cluster id>
+  peerTLSEnabled: <bool>
```
timuthy, line 475: Isn't `peerTLSEnabled` a good candidate for a `spec` too, e.g. having it in both places?

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: we were divided between using `phase` and `subPhase` vs `state` and `subState` and we chose the latter.
Pods have a top level phase which essentially is derived by looking at conditions.  So we could align to that terminology as well.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+2. backup-sidecar attempts to fetch etcd cluster member-list and checks if this etcd-member is already part of the cluster.
+
+3. Size of the cluster by checking `initial-cluster` in the etcd config.
+
+Druid adds an annotation `gardener.cloud/scaled-to-multi-node` on the `StatefulSet` which is then shared by all etcd-members irrespective of the starting state of an etcd-member (as `Learner` or `Voting-Member`). This especially creates an issue for the current leader (often pod with index 0) during the scale-up of an etcd cluster as described in [this](https://github.com/gardener/etcd-backup-restore/issues/646) issue. 
+
+It has been agreed that the current solution to [this]([[BUG] Restart of first cluster member can lead to quorum loss while scaling up a non-HA etcd cluster. · Issue #646 · gardener/etcd-backup-restore · GitHub](https://github.com/gardener/etcd-backup-restore/issues/646)) issue is a quick and dirty fix and needs to be revisited to be uniformly applied to all etcd-members. The authors propose to provide a more deterministic approach to scale-up using the `EtcdMember` resource.
```
ishan16696, line 190:
```suggestion
It has been agreed that the current solution to [this issue](https://github.com/gardener/etcd-backup-restore/issues/646) issue is a quick and dirty fix and needs to be revisited to be uniformly applied to all etcd-members. The authors propose to provide a more deterministic approach to scale-up using the `EtcdMember` resource.
```

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+The leading-backup-sidecar performs periodic defragmentations of the DBs of all the etcd-members in the cluster, controlled via a defragmentation cron schedule provided to each backup-sidecar. Defragmentation is a costly maintenance operation and causes a brief downtime to the etcd-member being defragmented, due to which the leading-backup-sidecar defragments each etcd-member sequentially. This ensures that only one etcd-member would be unavailable at any given time, thus avoiding an accidental quorum loss in the etcd cluster.
+
+The authors propose to move the responsibility of orchestrating these individual defragmentations to druid due to the following reasons:
+
+- Since each backup-sidecar only has knowledge of the health of its own etcd, it can only determine whether its own etcd can be defragmented or not, based on etcd-member health. Trying to defragment a different healthy etcd-member while another etcd-member is unhealthy would lead to a transient quorum loss.
```
ishan16696, line 280: > Since each backup-sidecar only has knowledge of the health of its own etcd, it can only determine whether its own etcd can be defragmented or not, based on etcd-member health

That's absolutely wrong assumption, in actual leading sidecar backup-restore has cluster wide information about health of all etcd cluster members, please check this code: 
https://github.com/gardener/etcd-backup-restore/blob/5bb95a80183d7b31c284cfcc4b9c0d5bdb50c581/pkg/defragmentor/defrag.go#L81


> Trying to defragment a different healthy etcd-member while another etcd-member is unhealthy would lead to a transient quorum loss.

sorry, I didn't get this part. We already thought through this kind-of scenario while working on multi-node etcd design.  please check this [doc of leader_election](https://github.com/gardener/etcd-backup-restore/blob/master/doc/concepts/leader_election.md#defragmentation)
We only trigger defragmentation when all of the members of etcd cluster returns `true` of `etcd status API call` i.e all members are healthy else leading sidecar backup-restore keep checking the health of all members of etcd cluster until all etcd cluster members become healthy.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+- Each backup-sidecar is only a `sidecar` to its own etcd-member, and by good design principles, it must not be performing any cluster-wide maintenance operations, and this responsibility should remain with the etcd cluster operator.
```
ishan16696, line 281: >  and by good design principles, it must not be performing any cluster-wide maintenance operations, and this responsibility should remain with the etcd cluster operator.

AFAIU, you are suggesting that defragmentation should be triggered by etcd-druid itself ... Is it ?
But previous statement mentioned etcd-druid should be just orchestrate the defragmention. Please correct me if I understood in wrong way.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: During a scale-up, new EtcdMember resources will be created, that is correct. However, every backup-sidecar container (which today plays an active role in bootstrapping an etcd peer container), needs to deterministically ascertain two things:

1. Should it create a new cluster or join an existing cluster.
2. Should it join as a learner (when it is started for the first time) or should it directly join as a voting member which could happen either in case of a restart (during scale-up or otherwise) or for a single-member etcd cluster.

Today druid annotates the STS with an annotation which tell any member that a scale-up is currently in progress. This has some issues as this annotation is now shared by all members (even the 0th member). We now propose to remove that annotation and instead define another annotation per `EtcdMember`, which then tells the backup-sidecar how it should bootstrap. We have also mentioned that when the first-ever bootstrap is completed then backup-sidecar will remove this annotation.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: That is right, only the leading backup-sidecar will take snapshots, however since it is possible that due to re-election there is a leadership change any EtcdMember whose backup-sidecar is a leader can then start to record this information. Since all this information has a timestamp associated to it, it helps analyse the following:

- Druid can check the leader and inspect the published snapshot meta-info to find if its been published regularly.
- Since all EtcdMember resources will have this, you can get a holistic and near-historic information about earlier snapshots taken by other EtcdMember's. This is only meant for diagnostic.

The structure of `EtcdMember` will be kept consistent and we will keep a limit to the number of snapshot information that we capture. This is not yet decided.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Yes defragmentation will now happen in two ways:

- As per the configured schedule. Today it is set to once in 24 hrs. This will always be in place.
- Threshold based defragmentations - this will be triggered as explained and will ensure that etcd DB size is kept in check even during sudden/long peaks which causes the etcd DB to increase in size rapidly. There is auto-compaction but that is not sufficient to release memory and must be followed by defragmentation.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: We will only maintain last X run data. As and when newer data arrives, oldest entry will get removed. `X` is not yet been decided and will be decided during implementation.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: We initially thought about it but then went against it. What we `define` as part of the spec is a desired state that must be constantly reconciled and eventually met and thereafter maintained. `peerTLSEnabled` is an information that is published by an `EtcdMember` depicting the as-is state of TLS enablement of peer URL as seen by the etcd process running within `etcd-wrapper` container within this etcd-member and thus belongs in status.

docs/proposals/04-etcd-member-custom-resource.md
shreyas-s-rao: The question is not whether leading backup sidecar currently has knowledge of the entire cluster, but whether it should, going forward. If leading backup "sidecar" has knowledge of the entire cluster, then it is no longer a sidecar but an orchestrator, and it interferes with the responsibilities of the actual cluster orchestrator, aka, etcd-druid.

> We only trigger defragmentation when all of the members of etcd cluster returns true of etcd status API call i.e all members are healthy else leading sidecar backup-restore keep checking the health of all members of etcd cluster until all etcd cluster members become healthy

Got it, thanks. So this means that the leading backup sidecar simply waits until the cluster is fully healthy before it can trigger a defragmentation. But the operator who can likely recover the cluster from an unhealthy state is still druid, so it doesn't do any harm if druid is the one who initiates a cluster-wide defragmentation operation, again because druid should be the one who has overall knowledge of all the members, not any one backup sidecar.

docs/proposals/04-etcd-member-custom-resource.md
shreyas-s-rao: Defragmentation will be triggered, ie initiated by druid. This means that druid issues a defragmentation call to a particular etcd member. The member will then perform the actual defragmentation by calling the defragment API on the etcd.

So while druid is still not directly talking to any etcd process, it still decides the order in which to defragment the members, and "orchestrates" the defragmentation.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: In addition, druid can pause the entire defragmentation operation if there is a transient quorum loss. This can happen if lets say it is currently issued a defragmentation request to etcd-member-1. While etcd-member-1 is running defragmentation, etcd-member-2 goes down and does not come up for some time. Now we have a quorum loss. Druid will know this and pause any further defragmentation. Even after the current defragmentation is over and quorum is restored, if the etcd-member-2 still is down then it will not continue. Once all members are up, then it can resume from where it left (provided we are still far away from the scheduled defragmentation schedule). This is a responsibility of an orchestrator/operator and not of a side-car.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: k8s native way would be for druid to signal intent to etcd-member via well defined API objects, which then reacts to the requested operation and updates the status of the said API object (in this case that would be `EtcdMember`). So there is no direct communication between druid and any etcd-member.

docs/proposals/04-etcd-member-custom-resource.md
ishan16696: >  If leading backup "sidecar" has knowledge of the entire cluster, then it is no longer a sidecar but an orchestrator, and it interferes with the responsibilities of the actual cluster orchestrator, aka, etcd-druid.

yes, I agree that `backup-restore` is now no longer just a sidecar as it used to be, may be in initial days of backup-restore. May be we need to redefine some of our terminologies as we are just keep using the old terminologies.
Why I'm saying this because `backup-restore` also provides the configuration to corresponding etcd, by sidecar logic this feature of `providing config` should also change ... right ?
IMO, we shouldn't label backup-restore as sidecar now and we should redefine/revisit functionalities performed by etcd-backup-restore, so it become easy to decide which way we want to go. WDYT ?

unmarshall: >Why I'm saying this because backup-restore also provides the configuration to corresponding etcd, by sidecar logic this feature of providing config should also change ... right ?

And this will also change. It will return to just being a sidecar which provides backup and restore functionality.

docs/proposals/04-etcd-member-custom-resource.md
shreyas-s-rao: Yes. That's one of the main objectives of this proposal, to provide a way where each etcd member can reliably publish its state to etcd-druid. Etcd-druid will be able to derive each member's health via the pod's readiness probe check, the liveness heartbeats via each member lease, and also the individual EtcdMember state transitions (to determine the state and subState of each member). By seeing the `defragmentations`, `dbSize` and `dbSizeInUse` fields maintained in the EtcdMember, druid will additionally know the status of previous defragmentations as well as triggers on when the next defragmentation should be performed.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+## Proposal
+
+This proposal introduces a new custom resource `EtcdMember`, and in the following sections describes different sets of information that should be captured as part of the new resource.
+
+### Etcd Member Metadata
+
+Every etcd-member has a unique `memberID` and it is part of an etcd cluster which has a unique `clusterID`. In a well-formed etcd cluster every member must have the same `clusterID`. Publishing this information to druid helps in identifying issues when one or more etcd-members form their own individual clusters, thus resulting in multiple clusters where only one was expected. Issues [Issue#361](https://github.com/gardener/etcd-druid/issues/361) and [Issue#419](https://github.com/gardener/etcd-druid/issues/419) are some such occurrences.
+
+Today, this information is published by using a member [lease](https://kubernetes.io/docs/concepts/architecture/leases/). Both these fields are populated in the leases' `Spec.HolderIdentity` by the backup-sidecar container.
```
ishan16696, line 96:
```suggestion
Today, this information is published by using a member [lease](https://kubernetes.io/docs/concepts/architecture/leases/). Both these fields are populated in the lease's `Spec.HolderIdentity` by the backup-sidecar container. 
```

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+* `druid.gardener.cloud/create-as-learner` annotation on its respective `EtcdMember` resource. This new annotation will be honored in the following cases:
+  * When an etcd-member is created for the very first time.
+
+  * An etcd-member is restarted while it is in `Starting` state (`PendingLearner` and `Learner` sub-states).
+
+* Etcd-cluster member list.  to check if it is already part of the cluster. 
+
+* Existing etcd data directory and its validity.
+
+> **NOTE:** When the etcd-member gets promoted to a voting-member, then it should remove the annotation on its respective `EtcdMember` resource.
```
ishan16696, line 207: what if it get restarted and failed to remove this annotation `druid.gardener.cloud/create-as-learner`

shreyas-s-rao: Well, then when the member comes back up, if its data dir is valid, it doesn't need to see the annotation anymore, and will start as a follower as expected. But if its data dir is corrupt, then it will have to re-join the cluster as a learner. It's the same case today - if etcd-member-2 restarts just after it joined the cluster and its data dir gets corrupted, then it will rejoin the cluster as a learner, doesn't it?

docs/proposals/04-etcd-member-custom-resource.md
shreyas-s-rao: The original is correct. `leases'` means `of many leases`, while `lease's` means `of one lease`. In this line we speka about the `Spec.HolderIdentity` of all the snapshot leases.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+* Auto-recovery from quorum loss or cluster-split due to network partitioning.
+* Auto-recovery of an etcd-member due to volume mismatch.
+
+## Proposal
+
+This proposal introduces a new custom resource `EtcdMember`, and in the following sections describes different sets of information that should be captured as part of the new resource.
+
+### Etcd Member Metadata
+
+Every etcd-member has a unique `memberID` and it is part of an etcd cluster which has a unique `clusterID`. In a well-formed etcd cluster every member must have the same `clusterID`. Publishing this information to druid helps in identifying issues when one or more etcd-members form their own individual clusters, thus resulting in multiple clusters where only one was expected. Issues [Issue#361](https://github.com/gardener/etcd-druid/issues/361) and [Issue#419](https://github.com/gardener/etcd-druid/issues/419) are some such occurrences.
```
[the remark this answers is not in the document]
shreyas-s-rao: We clearly mention that `Issues Issue#361 and Issue#419 are some such occurrences.`, and this is still not wrong. We have still been seeing few cases of multiple clusters being formed, which may or may not occur in the future. Hence `clusterID` will provide a deterministic way of determining whether multiple clusters have been formed.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+There is a need to have a clear distinction between an etcd-member state and etcd cluster state since most of an etcd cluster state is often derived by looking at individual etcd-member states. In addition, actors which update each of these states should be clearly identified so as to prevent multiple actors updating a single resource holding the state of either an etcd cluster or an etcd-member. As a consequence, etcd-members should not directly update the `Etcd` resource status and would therefore need a new custom resource allowing each member to publish detailed information about its latest state.
+
+### Goals
+
+* Introduce `EtcdMember` custom resource.
+
+* Enhance [etcd-backup-restore](https://github.com/gardener/etcd-backup-restore) to publish member specific information by updating `EtcdMember` resource. 
+
+* Today [leases](https://kubernetes.io/docs/concepts/architecture/leases/) are misused to share member-specific information with etcd-druid. Their usage to share member state [leader, follower, learner], member-id, snapshot revisions etc should be removed.
```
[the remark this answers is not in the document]
shreyas-s-rao: Yes. The proposal wishes to eliminate the use of snapshot leases only, since that was not used in the right way. Member leases will still remain, but the annotation on it for peer-tls-enabled will go away and will be moved to `EtcdMember.Status` instead.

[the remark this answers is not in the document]
shreyas-s-rao: No. Compaction will only be further enhanced to also consider the accumulated size of delta snapshots so that druid can trigger compaction job early if it sees that the snapshots are growing too large.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,545 @@
...
+
+### Non-Goals
```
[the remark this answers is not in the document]
shreyas-s-rao: That's correct. Usually proposals, including [GEPs](https://github.com/gardener/gardener/blob/master/docs/proposals/00-template.md), define Goals and Non-Goals. Goals are the objectives that the proposal wishes to achieve, while non-goals are those that the proposers explicitly state that the proposal does not target, to remove any ambiguity from the proposal's targeted goals.

docs/proposals/04-etcd-member-custom-resource.md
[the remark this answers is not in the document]
ishan16696: > We clearly mention that Issues Issue#361 and Issue#419 are some such occurrences., and this is still not wrong

yes, I'm agreeing with the intent of this but I'm saying mentioning `Issue#361` as it was occurred due to we didn't had a single restoration feature back then.

>  We have still been seeing few cases of multiple clusters being formed

please mention those occurrences if possible, if not then you can just mention `some occurrences in landscapes`.

docs/proposals/04-etcd-member-custom-resource.md
timuthy: > Should it create a new cluster or join an existing cluster.

Alternatively, you could have a `spec.clusterId` or a similar spec field that instructs the member to join a existing cluster.

> Should it join as a learner (when it is started for the first time) or should it directly join as a voting member which could happen either in case of a restart (during scale-up or otherwise) or for a single-member etcd cluster.

The text already says that the side-car needs to be able to figure out if it is a member of the cluster, i.e. if it is join as voting member, else join as learner?

docs/proposals/04-etcd-member-custom-resource.md
timuthy: Sounds good!

docs/proposals/04-etcd-member-custom-resource.md
timuthy: It's understandable to save historic data in the status if it's read again by Druid to determine if something goes south. However for operators and diagnosis it would be better to expose metrics (in addition).

docs/proposals/04-etcd-member-custom-resource.md
timuthy: I'm not sure if I got the point but if I read the DEP again then the conclusion would be that if TLS is enabled or not is derived from the `advertised peer URL(s)` and thus it doesn't make sense to have yet another spec field to configure `peerTLSEnabled`. Is that correct?

docs/proposals/04-etcd-member-custom-resource.md
timuthy: You mean to have `phase` for top level and `conditions` (alternatively `lastOperation`) to determine more information about the phase?

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+| New           | -                          | Every newly created etcd-member will start in this state and is termed as the initial state or the start state. |
+| Initializing  | DBV-S (DBValidationSanity) | This state denotes that backup-restore container in etcd-member pod has started initialization. Sub-State `DBV-S` which is an abbreviation for `DBValidationSanity` denotes that currently sanity etcd DB validation is in progress. |
+| Initializing  | DBV-F (DBValidationFull)   | This state denotes that backup-restore container in etcd-member pod has started initialization. Sub-State `DBV-F` which is an abbreviation for `DBValidationFull` denotes that currently full etcd DB validation is in progress. |
+| Initializing  | R (Restoration)            | This state denotes that backup-restore container in etcd-member pod has started initialization. Sub-State `R` which is an abbreviation for `Restoration` denotes that DB validation failed and now backup-restore has commenced restoration of etcd DB from the backup (comprising of full snapshot and delta-snapshots). An etcd-member will transition to this sub-state only when it is part of a single-node etcd-cluster. |
```
abdasgupta, line 126: I think, we are classifying too many states and sub-states. We should be very minimalistic here, otherwise backup-restore needs to connect to the Kubernetes client for too many times. This, in turn, would widen the scope of failures.

docs/proposals/04-etcd-member-custom-resource.md
abdasgupta: Are we going to update the ETCD member states even for the sub-states? I think, we can cut those substates and only keep Two main states, named Initializing and Running.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+
+<img title="" src="assets/04-state-transitions-scaleup-1-n.excalidraw.png" alt="" data-align="center">
+
+#### Restart of a Voting Etcd-Member in a Multi-Node Etcd Cluster
+
+Following DFA represents the states, sub-states and transitions when a voting etcd-member in a multi-node etcd cluster restarts.
+
+<img title="" src="assets/04-state-transitions-restart-member.excalidraw.png" alt="" data-align="center">
+
+> NOTE: If the DB validation fails then data directory of the etcd-member is removed and etcd-member is removed from cluster membership, thus transitioning it to `New` state. The state transitions from `New` state are depicted by [this section](#addition-of-a-new-etcd-member-in-a-multi-node-etcd-cluster).
```
abdasgupta, line 166: We are not doing anything different for failure of DVF-S and DVF-F. In both the cases, we are just restarting.

docs/proposals/04-etcd-member-custom-resource.md
abdasgupta: Druid is going to add the annotation to each member resource as per writeup. How does it know which member is going start? Also, there are too many edge cases in this approach

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+
+### Lifecycle of an EtcdMember
+
+#### Creation
+
+Druid creates an `EtcdMember` resource for every replica in `etcd.Spec.Replicas` during reconciliation of an etcd resource. For a fresh etcd cluster this is done prior to creation of the StatefulSet resource and for an existing cluster which has now been scaled-up, it is done prior to updating the StatefulSet resource.
+
+#### Updation
+
+All fields in `EtcdMember.Status` are *only* updated by the corresponding etcd-member. Druid only consumes the information published via `EtcdMember` resources.
```
abdasgupta, line 525: If we are going ahead with the status updation for a myriad number of reasons, a single Druid controller requires to orchestrate multiple of things. What worries me is suppose the druid controller is orchestrating defragmentation of a member and at the same time status is updated for restoration for another member. Can druid handle both the scenarios?

abdasgupta (changes requested):

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: We evaluated using conditions but it was getting too complicated to represent state transitions and therefore we did not go down that path and instead stuck to using simple, easily interpreted state transitions. So for `EtcdMember` resource we do not envision any conditions as of now. In future if there is a need we could introduce it.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: The total number of states are 4 which is quite small number. Coming to sub-states, during initialisation a member will only enter into 1 state (DB validation) and Restoration will only be when there is just 1 member (single node etcd cluster). Coming to starting -> there are a fixed 2 states (only when the member is going to be added as a learner), there are cases when it directly jumps to `Started` state. In voting member state there are 2 states and any member which finishes learning always goes to just `Follower` state and only upon leader election changes the state would change.

Just like in any k8s - all well-written controllers are level-triggered and not edge-triggered. Therefore entities which are changing the status to which a controller reacts need to be eventually consistent only. I do not see an issue at this point.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: >Also, there are too many edge cases in this approach

@abdasgupta can you list down the edge cases so that it helps us evaluate the current proposal in more light?

> Alternatively, you could have a spec.clusterId or a similar spec field that instructs the member to join a existing cluster.

`spec.clusterId` - yes we can think of this as a possibility. Thanks for the suggestion. Now we have one possibility for using the spec :)

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Ok you are right, maybe we do not even need this any longer. Reason, recently we changed the handling of peer TLS URL changes and we do the following:
> _Assumption:_ Initial single-node etcd cluster does not have TLS enabled. When scale is requested then g/g enables TLS for peer communication

- We update the existing etcd-member and mount the TLS secret, this cases one restart of the pod.
- Once the update is over then we delete the pod, causing a second restart. This will now ensure that the existing pod's etcd process will deterministically always start with TLS enabled for peer URL.
- Now we scale etcd STS from 1-3.  At this point, 0th pod already has TLS enabled and therefore no other properly indicating and tracking TLS is required for other 2 pods that are going to be launched.

We can remove this property and also the section. Thanks for pointing it out.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Yes. Is this an observation or is there a question?

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Ok we will discuss this point again.

docs/proposals/04-etcd-member-custom-resource.md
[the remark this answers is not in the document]
unmarshall: I have now removed [Issue#361](https://github.com/gardener/etcd-druid/issues/361) and i have added only references to 2 canary issues (no links are provided). I have checked with Ashwani and if no links are provided and only numbers it is acceptable.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: So lets take case by case. 
Take a case that defragmentation is started. Typically as it happens today the current leader will be the last one to be selected for defragmentation to prevent multiple leader elections during the entire process. Let us assume that pod-0 is the current leader and let us assume that pod-1 and pod-2 are heathy and there is cluster readiness (quorum is established).

Druid picks up pod-1, while pod-1 is defragmenting there is a leadership change and the new leader is now pod-2. When pod-1's defragmentation finishes, then druid will re-inspect the cluster state. If there is a quorum and there are still pending members to be defragmented then druid will choose any non-leader that has not yet been defragmented. It will requested that member to initiate the defragmentation. Since druid does defragmentation one member at a time, it will always get the latest state of each member.

Now consider a case that there was a leadership change but the status update got delayed. Max what will happen is that it will cause a re-election. We will need to observe how frequent would this be (leadership change) in real world.

Can you list down other cases and we can go over them and see which cases are problematic.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+    message: "Now in sync with leader, promoted as voting member"
+```
+
+##### Reason Codes
+
+The authors propose the following list of possible reason codes for transitions. This list is not exhaustive, and can be further enhanced to capture any new transitions in the future.
+
+| Reason                                | Transition From State (SubState)                             | Transition To State (SubState)         |
+| ------------------------------------- | ------------------------------------------------------------ | -------------------------------------- |
+| `ClusterCreated`                      | nil                                                          | New                                    |
```
ashwani2k, line 440: `ClusterCreated` as the reason is a little misleading as it depicts a completed activity more in line with final state. 
I was wondering if it will be better to change it to something in present or present-continuous tense like `CreateCluster` or `CreatingCluster`.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+
+2. Scale down of an etcd cluster to 0 replicas due to hibernation of the k8s control plane.
+
+3. Transient scale down of an etcd cluster to 0 replicas to recover from a quorum loss.
+
+Authors found no reason to retain EtcdMember resources when the etcd cluster is scale down to 0 replicas since the information contained in each EtcdMember resource would no longer represent the current state of each member and would thus be stale. Any controller in druid which acts upon the `EtcdMember.Status` could potentially take incorrect actions.
+
+#### Reconciliation
+
+Authors propose to introduce a new controller (lets call it `etcd-member-controller`) which watches for changes to the `EtcdMember` resource(s). If a reconciliation of an `Etcd` resource is required as a result of change in `EtcdMember` status then this controller should enqueue an event and force a reconciliation via existing `etcd-controller`, thus preserving the single-actor-principal constraint which ensures deterministic changes to etcd cluster resources.
```
ashwani2k, line 541: Just curious what happens to the[ EtcdMemberStatus](https://github.com/gardener/etcd-druid/blob/9a598d05e639099ddf404803f87376852261a052/api/v1alpha1/types_etcd.go#L358) which is currently managed by druid in the `etcd` resource. 
Do we continue to maintain it there as well, or will  the new `EtcdMember` be the only place to check this. 
May be a few sentences here might help to gain some context and if there is an impact of this proposal on the existing functionality as currently it focusses only on `EtcdMember`

ashwani2k (commented):
Thanks a lot @shreyas-s-rao and @unmarshall  for such a detailed and well written proposal. 
For someone like me with limited view of the functionality it was easy to follow the intent and its benefits. 

I've raised some minor pointers for your consideration. They might not be valid as I lack the big picture or the depth of implement, so please bear with me here.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+```yaml
+transitions:
+- state: <name of the state that the etcd-member has transitioned to>
+  subState: <name of the sub-state if any>
+  reason: <reason code for the transition>
+  transitionTime: <time of transition to this state>
+  message: <detailed message if any>
+```
+
+As an example, consider the following transitions which represent addition of an etcd-member during scale-up of an etcd cluster, followed by a restart of the etcd-member which detects a corrupt DB:
```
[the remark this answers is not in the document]
unmarshall: It can happen, but we will never store more than 'n' last state transitions. This number can be decided during implementation.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Yes this was also the comment during an earlier discussion as well. I have now proposed a change in the name. PTAL.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: No, it will only have references to `EtcdMember` resources that are created for an `Etcd` resource as part of its status. I have now added proposed changes to `Etcd` resource in a new section under `API`.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+volumeMismatches:
+- identifiedAt: <time at which wrong volume mount was identified>
+  fixedAt: <time at which correct volume was mounted>
+  volumeID: <volume ID of wrong volume that got mounted>
+  numRestarts: <num of etcd-member restarts that were attempted>
+```
+
+Each entry under `volumeMismatches` will be for a unique `volumeID`. If there is a pod restart and it results in yet another unexpected `volumeID` (different from the already captured volumeIDs) then a new entry will get created. `numRestarts` denotes the number of restarts seen by the etcd-member for a specific `volumeID`.
+
+Based on information from the `volumeMismatches` section, druid *may* choose to perform rudimentary remediatory actions as simple as restarting the member pod to force a possible rescheduling of the pod to a different node which could potentially force the correct volume to be mounted to the member.
```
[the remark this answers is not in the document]
unmarshall: It should be technically possible via [downwardAPI](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/). However, you can already find this information in the etcd pod today - just check `spec.nodeName`.  Should we duplicate this information in `EtcdMember` resource? If yes, then it should not be added under `volumeMismatches` but directly under `status` as a top level field.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Thinking again when the first member is created then the `clusterID` is not known to druid. Therefore it must be part of the `EtcdMember.Status`. Do you think it is wise to include it as part of spec (nil for first member and non-nil for rest of the members) and status?

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: what is the rationale?

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+
+Following DFA represents the states, sub-states and transitions of an etcd cluster which starts with having a single member (Leader) and then one or more new members are added which represents a scale-up of an etcd cluster from 1 -> n, where n is [odd](https://etcd.io/docs/v3.5/faq/#why-an-odd-number-of-cluster-members).
+
+<img title="" src="assets/04-state-transitions-scaleup-1-n.excalidraw.png" alt="" data-align="center">
+
+#### Restart of a Voting Etcd-Member in a Multi-Node Etcd Cluster
+
+Following DFA represents the states, sub-states and transitions when a voting etcd-member in a multi-node etcd cluster restarts.
+
+<img title="" src="assets/04-state-transitions-restart-member.excalidraw.png" alt="" data-align="center">
```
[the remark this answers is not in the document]
ashwani2k: Got it. Thanks for clarification.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+    endRevision: <end revision of etcd db captured in the snapshot>
+  lastDelta:
+    timestamp: <time of delta snapshot>
+    name: <name of the file that is uploaded>
+    size: <size of the un-compressed snapshot file uploaded>
+    startRevision: <start revision of etcd db captured in the snapshot>
+    endRevision: <end revision of etcd db captured in the snapshot>
+```
+
+While this information will primarily help druid compute accurate conditions regarding backup health from snapshot information and publish this to human operators, it could be further utilised by human operators to take remediatory actions (e.g. manually triggering a full or delta snapshot or further restarting the leader if the issue is still not resolved) if backup is unhealthy.
```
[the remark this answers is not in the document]
ashwani2k: Thanks, then i've got it right :)

docs/proposals/04-etcd-member-custom-resource.md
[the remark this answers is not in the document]
ashwani2k: Good point, this might be over engineering if we can anyways fetch it from pod spec.

docs/proposals/04-etcd-member-custom-resource.md
ashwani2k: Thanks looks complete now with specifics of etcd api changes.

docs/proposals/04-etcd-member-custom-resource.md
ashwani2k: Nice I missed that clarity of `ClusterScaleup` and `NewSingleNodeClusterCreated`. This looks more apt.

docs/proposals/04-etcd-member-custom-resource.md
```diff
@@ -0,0 +1,561 @@
...
+
+There is a need to have a clear distinction between an etcd-member state and etcd cluster state since most of an etcd cluster state is often derived by looking at individual etcd-member states. In addition, actors which update each of these states should be clearly identified so as to prevent multiple actors updating a single resource holding the state of either an etcd cluster or an etcd-member. As a consequence, etcd-members should not directly update the `Etcd` resource status and would therefore need a new custom resource allowing each member to publish detailed information about its latest state.
+
+### Goals
+
+* Introduce `EtcdMember` custom resource.
+
+* Enhance [etcd-backup-restore](https://github.com/gardener/etcd-backup-restore) to publish member specific information by updating `EtcdMember` resource.
```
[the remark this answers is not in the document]
ashwani2k: Looks good now and more crisp then what I had proposed.

docs/proposals/04-etcd-member-custom-resource.md
abdasgupta: The question is why do we need to separate states then?

docs/proposals/04-etcd-member-custom-resource.md
abdasgupta: Okay. I was worried about the amount of network traffics. Moreover as we are critically depending on these state changes, I was afraid that there might be many edge cases. The edge cases are combination of start-restart-delay in volume attachment-temporary network failures. You can include the edge cases in this doc itself and we can contribute if there are any edge cases yet to be added.

docs/proposals/04-etcd-member-custom-resource.md
abdasgupta: What I am suggesting here is let Druid not handle the defragmentation or restoration. I think it would be less complex and easier to debug in case of fault if individual members handle these. I am still not buying the fact that druid should handle everything for ETCD BR. Let's not increase the complexities and make things difficult to debug and resolve. It would be a real problem, in case of a failure, to pore over the logs of druid and figure out which steps happened when and after what status change.

docs/proposals/04-etcd-member-custom-resource.md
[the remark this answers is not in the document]
abdasgupta: delta snapshots leases are updated frequently. Very very frequently in a busy cluster. So far it worked well with leases. Are we planning to update ETCD Member resources that frequently on top of updating ETCD member resources with other critical things like restoration, leader election etc? On top of that, it's adding complexities in compaction. Compaction control now has to figure out the member resource of leader and then take action after that!

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: >What I am suggesting here is let Druid not handle the defragmentation or restoration. I think it would be less complex and easier to debug in case of fault if individual members handle these

On the contrary, this is much more complex. A sidecar should only serve its respective main container within the pod and not engage in cluster wide operations like defragmentations. Reasons:
1. Leadership can change at any given point of time.
2. Any member can go down at any given point of time.
Due to above two reasons its important that the defragmentation be stopped. Its actually quite easy to implement, maintain and orchestrate when you do this from druid. It is not only simple but it is completely transparent of the state of defragmentation for the entire cluster.

>Let's not increase the complexities and make things difficult to debug and resolve. It would be a real problem

I would like to understand the reason why you think it is more complex and what is difficult to debug. This statement is quite generic and i would not be able to respond properly unless i understand what problems you see.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: The state changes are not going to be that frequent and therefore network traffic should not be a worry IMHO.
As explained above druid will never react based on edges. In any distributed environment there is always a possibility of late update or failure to update. Reconcilers will always be designed to ensure that deal with these situations.

>You can include the edge cases in this doc itself and we can contribute if there are any edge cases yet to be added.

Since you pointed out that there are edge case then may be it would be better that you list at least a few down as i am sure you have already thought through this a bit and its fresh in your mind.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Diagnostics and transparency is one such reason.

docs/proposals/04-etcd-member-custom-resource.md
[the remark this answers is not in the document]
unmarshall: If leases were updated that frequently then how is it going to be different when we update the EtcdMember resource?
Leases are never meant for as a carrier of domain specific information. In all upstream components leases are only used either for heartbeats or leader election. 

>On top of that, it's adding complexities in compaction. Compaction control now has to figure out the member resource of leader and then take action after that!

What is the complexity that you refer to here? It is very easy to figure out who is the leader. Today member leases are used for that. What is the main concern, its not clear?

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Ok as discussed offline - for the first member `spec.clusterID` will be nil/empty. For all the other members, it will be set by druid. This then removes the need to use any annotation. Every member will also update `status.clusterID`.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: As discussed during the meeting on Wed 23 Aug:
We agreed that during the implementation care should be taken to handle the following:
- Failure or delay to record state changes from backup-sidecar should be handled by druid.
- Keep an eye on cases where repeated restarts are happening causing frequent change of state/sub-state. Re-look at the current handling where containers are repeatedly restarted.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: As discussed during the meeting on Wed 23 Aug:
For now we will retain the same set of state and sub-states as described in this proposal. Also see [comment](https://github.com/gardener/etcd-druid/pull/658/files#r1303818726)

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: >And this will also change. It will return to just being a sidecar which provides backup and restore functionality and anything around it.

This will not be addressed as part of this proposal but there will be a separate proposal which details out how the responsibilities will change between etcd-wrapper and backup-restore containers.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: In the meeting on 23rd Aug we agreed on the following:
During the implementation, care should be taken that only one actor makes changes to the `EtcdMember` resource in backup-restore container. This prevents issues of conflict during updates. It is also understood that there is always a possibility that these updates get delayed or fail. These situations should be properly handled.

docs/proposals/04-etcd-member-custom-resource.md
ishan16696: thanks @unmarshall,
I'm just letting this comment thread open for future references.

docs/proposals/04-etcd-member-custom-resource.md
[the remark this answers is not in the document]
abdasgupta: We have kept default threshold for compaction of 10 million events because in the performance taste we had a cluster that created that much events in 30 minutes. Compaction job runs for 30 minutes to compact 10 million events. So in that cluster, compaction jobs were running in each 30 minute for 30 minutes. For 10 million events, frequency of delta snapshots depends on the size of the events. If the accumulated events are more than 100 mb then delta snapshots are taken. Anyway, my point is, if with this huge number, we can guarantee the seamless updation of ETCD member resources along with other type of updates, I am fine.

docs/proposals/04-etcd-member-custom-resource.md
ishan16696: > if its data dir is valid, it doesn't need to see the annotation anymore

right now, we first check the scale-up condition first after data-dir validation is checked.
https://github.com/gardener/etcd-backup-restore/blob/64995e097fd75e9838887302cc41ff454557d511/pkg/initializer/initializer.go#L69-L91

shreyas-s-rao: @ishan16696 don't worry about what is implemented currently. We have an opportunity to change the behavior and correct a few wrongs here, so I would implore you to look at it from that perspective.

If the data-sir is present and valid (revision matches the leader revision), do we really need to check the scale-up annotation? Scale-up annotation is required *only* when we need to determine whether the member needs to be added as a learner or not. If the data-dir is valid, then we have no need to add as a learner, hence no need to check the annotation. If the data-dir is invalid (corrupt), then we anyway need to clean up the data-dir and add as a learner so that it catches up with the leader, in the case of a multi-node cluster.

ishan16696 (approved):
LGTM!!

docs/proposals/04-etcd-member-custom-resource.md
ishan16696: > so I would implore you to look at it from that perspective.

yes, I already thought about this while working on scale-up fixes. I discussed this with Aaron sometime back that why we are first checking the scale-up then data-dir validation but right now I forgot what we discussed and the reason behind this.

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: As per the proposal we capture:
```yaml
 - startTime: <start time of defragmentation>
    endTime: <end time of defragmentation>
    reason: 
    message:
    initialDBSize: <size of etcd DB prior to defragmentation>
    finalDBSize: <size of etcd DB post defragmentation>
```
Using metrics it would not be very clean to capture all of the above information using a single metric. So you will actually need 2 metrics - one for `initialDBSize` and another for `finalDBSize` and then you need to correlate the two metrics as well. Also message/reason can only be captured via labels (ok for reason as it will be a code, but message will be a description which is not really nice for labels as indexing is done for all defined labels on metrics)

Do you have an alternative way to capture this information via metrics?

docs/proposals/04-etcd-member-custom-resource.md
unmarshall: Discussion amongst: @shreyas-s-rao, @unmarshall and @seshachalam-yv 

@timuthy we can get rid of both the annotation `druid.gardener.cloud/create-as-learner` and the proposed `spec.clusterID`. 
When a member comes up for the first time or is restarted then it should do the following:

> NOTE: Data directory for any member is considered valid if the data is not corrupt. It does not consider revision mismatch with the leader as a criterion for checking validity.

- Check the data directory. 
  - If the data directory is valid and its last state transition capture is either `PendingLearner` or `Learner` then rejoins as `PendingLearner`. If it was `New` then it rejoins as `New`. If it was a voting member then it always rejoins as `Follower` in case of a multi-node cluster and `Leader` if it is a single node etcd cluster.
  - If the data directory is not valid, then anyways data directory will get removed and it will be added as a `PendingLearner` in case of a multi-node cluster. In case of a single-node cluster it will go into `Restoration` from the backup.

The above proposal is only going to work when we make it mandatory to update state/sub-state till the member is learning and is not yet a voting member. Once it becomes a voting member then state transitions (Leader <-> Follower) is done with eventual consistency.

seshachalam-yv (approved):
/lgtm

docs/proposals/04-etcd-member-custom-resource.md
timuthy: Fine for me, thanks 🙂

docs/proposals/04-etcd-member-custom-resource.md
timuthy: You are right, you can't map this concept to metrics 1:1. Instead we'd need different metrics, e.g. DB size (already there), maybe a defragmentation failure count and similar. Details for a particular error can still be extracted from logs.

shreyas-s-rao: Another reason for not pushing these as metrics is because we want druid to take certain actions (mainly remediatory) based on this information, and we'd need to make druid prometheus-aware, which is not a very nice thing to do.

shreyas-s-rao: > Details for a particular error can still be extracted from logs.

That doesn't seem like a very clean approach, given that druid will then need to fetch etcd pod logs, parse them and extract error information and the exact error codes too. Not to mention that if the etcd pod gets restarted, the logs are lost to druid, unless it then relies on loki for the logs, which means druid will then have to be loki-aware too.

timuthy: I think this is a misunderstanding. Druid should never parse logs or metrics in order to perform some action, see my [previous comment](https://github.com/gardener/etcd-druid/pull/658#discussion_r1298071505):

> It's understandable to save historic data in the status if it's read again by Druid to determine if something goes south. However for operators and diagnosis it would be better to expose metrics (in addition) - same for snapshot information.

unmarshall: We can surely look at enhancing metrics for etcd clusters and we will do this during the implementation. I think we will have a much better idea on whether it makes sense to only capture the last defragmentation in the status of an etcd-member vs having an array of X number of entries as proposed currently. This will depend upon what druid would need in the end to take an action.

unmarshall: @timuthy we internally discussed and decided that we will only capture last defragmentation and last restoration as part of the `EtcdMember` spec. We have made the changes to these sections. PTAL.

PR #669, opened 2023-08-16
Minor fixes in error handling and typos

shreyas-s-rao:

/area quality
/kind cleanup

Minor fixes in error handling and typos:
1. Better error handling in `pkg/component/etcd/statefulset` package
2. `pkg/component/etcd/statefulset`: rename `Values.EtcdBackupCommand` to `Values.EtcdBackupRestoreCommand` since it is semantically more correct (because the command not only sets backup config but also restoration config)

Fixes #

**Release note**:

```other operator
NONE
```


unmarshall (approved):
/lgtm

pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -448,7 +448,7 @@ func (c *component) createOrPatch(ctx context.Context, sts *appsv1.StatefulSet,
 							Name:            "backup-restore",
 							Image:           c.values.BackupImage,
 							ImagePullPolicy: corev1.PullIfNotPresent,
-							Args:            c.values.EtcdBackupCommand,
+							Args:            c.values.EtcdBackupRestoreCommand,
```
aaronfern, line 451: I think it will be better to call this `EtcdBackupRestoreArguments` since we're now passing arguments instead of a command

shreyas-s-rao: Yes, that makes sense. I've made the change for `EtcdCommand` as well now. PTAL

aaronfern (approved):
Thanks for the changes!
/lgtm

Issue #648, opened 2023-07-25
[Feature] Improvements for ETCD Druid to accommodate compaction job better with the compaction dashboard.


abdasgupta:
**Feature (What you would like to be added):**
Three short term improvements have been identified in ETCD Druid so that compaction metrics are more meaningful for the compaction dashboard.
- [x]  Compaction Jobs follow a difficult naming convention when they are created by ETCD Druid. The current format is like - `<ETCD UID>-compact-job`. Due to this convention, the pods created by the job have names like `<ETCD UID>-compact-job-<POD UID>`. Due to this the regex expression, which is required to scrape POD resource usage, needs to include `*` infront and afterwards of the string `compact-job`. The regex expression takes the form of `*-compact-job-"`. If we can change the naming convention of compaction job while creating by druid to `compact-job-<ETCD UID`, the regex expression would need only the `*` at the end of the string `compact-job`.
- [x] Currently, we update the metrics `metricsJobCurrent` only when we create a new compaction job in druid and delete the earlier successful/failed compaction job. So, the dashboard project not so accurate completion time for a compaction job. See, we create a new job only when the difference between delta snapshot revision and full snapshot revision cross a certain threshold, configured by end user. It is the same time when we update `metricsJobCurrent` if the last job is completed. To accurately capture  `metricsJobCurrent`, we should actually update it everytime when we check the difference between delta snapshot revision and full snapshot revision. Though we can't be exactly accurate with `metricsJobCurrent` as we are not actively monitoring compaction job. #685 
- [x] Currently, `metricsJobDuration` histogram show average duration of compaction jobs. But the range for average duration of the jobs are captured wrong in the graph of the dashboard. If we follow the attached image, we would know that we don't get a clear idea of the average job duration as all of them falling from 10s to +Inf boundary. The image is also showing unnecessary breakups for milliseconds where no job duration is posted. This is happening because we have not set boundary for the `metricsJobDuration` histogram in ETCD druid. So, Set the expected bucket boundaries for the histogram `metricJobDurationSeconds` so that the compaction job dashboard show proper and meaningful breakup of the duration of the compaction jobs in the Left Y axis of the graph.
      <img width="400" height="200" alt="image" src=https://github.com/gardener/etcd-druid/assets/15626057/8feb1ed9-6649-49f9-ab8b-7be3bc54a0d6>

[...]

PR #671, opened 2023-08-18
Optimize Registration Logic for Snapshot Compaction controller

seshachalam-yv:
**How to categorize this PR?**
/area storage
/kind enhancement

**What this PR does / why we need it**:
This PR optimizes the logic behind registering snapshot compaction-controller. Instead of always running the compaction-controller and checking the CLI flag `--enable-backup-compaction` just before deploying the compaction job, the new approach registers the compaction-controller based on the presence of the CLI flag `--enable-backup-compaction`. 

**Which issue(s) this PR fixes**:
Fixes https://github.com/gardener/etcd-druid/issues/655#issuecomment-1696786117

**Special notes for your reviewer**:

**Release note**:
```other operator
None
```

PR #672, opened 2023-08-21
Use better naming for compaction jobs

abdasgupta:

/area control-plane
/kind enhancement

Adds some improvements for betterment of compaction dashboard.

Fixes #
#648

We need to address the issue https://github.com/gardener/gardener/issues/8576 after this PR is merged

**Release note**:

```noteworthy operator
Etcd snapshot compaction jobs will now be named `<etcd-name>-compactor` for better readability for human operators.
```

### DIFF
--- a/api/v1alpha1/types_etcd.go
+++ b/api/v1alpha1/types_etcd.go
@@ -444,7 +444,7 @@ func (e *Etcd) GetConfigmapName() string {
 
 // GetCompactionJobName returns the compaction job name for the Etcd.
 func (e *Etcd) GetCompactionJobName() string {
-	return fmt.Sprintf("%s-compact-job", string(e.UID[:6]))
+	return fmt.Sprintf("compact-job-%s", string(e.UID[:6]))
 }
 
 // GetOrdinalPodName returns the Etcd pod name based on the ordinal.
--- a/api/v1alpha1/types_etcd.go
+++ b/api/v1alpha1/types_etcd.go
@@ -444,7 +444,7 @@ func (e *Etcd) GetConfigmapName() string {
 
 // GetCompactionJobName returns the compaction job name for the Etcd.
 func (e *Etcd) GetCompactionJobName() string {
-	return fmt.Sprintf("compact-job-%s", string(e.UID[:6]))
+	return fmt.Sprintf("%s-compactor", e.Name)
 }
 
 // GetOrdinalPodName returns the Etcd pod name based on the ordinal.
--- a/api/v1alpha1/types_etcd_test.go
+++ b/api/v1alpha1/types_etcd_test.go
@@ -95,7 +95,7 @@ var _ = Describe("Etcd", func() {
 
 	Context("GetCompactionJobName", func() {
 		It("should return the correct compaction job name", func() {
-			Expect(created.GetCompactionJobName()).To(Equal("123456-compact-job"))
+			Expect(created.GetCompactionJobName()).To(Equal("compact-job-123456"))
 		})
 	})
 
--- a/api/v1alpha1/types_etcd_test.go
+++ b/api/v1alpha1/types_etcd_test.go
@@ -95,7 +95,7 @@ var _ = Describe("Etcd", func() {
 
 	Context("GetCompactionJobName", func() {
 		It("should return the correct compaction job name", func() {
-			Expect(created.GetCompactionJobName()).To(Equal("compact-job-123456"))
+			Expect(created.GetCompactionJobName()).To(Equal("foo-compactor"))
 		})
 	})
 
--- a/controllers/compaction/metrics.go
+++ b/controllers/compaction/metrics.go
@@ -56,6 +56,8 @@ var (
 		prometheus.HistogramOpts{
 			Namespace: namespaceEtcdDruid,
 			Subsystem: subsystemCompaction,
+			Name:      "job_duration_seconds",
+			Help:      "Total time taken in seconds to finish a running compaction job.",
 			Buckets: []float64{baseDuration.Seconds(),
 				5 * baseDuration.Seconds(),
 				10 * baseDuration.Seconds(),
@@ -63,8 +65,6 @@ var (
 				20 * baseDuration.Seconds(),
 				30 * baseDuration.Seconds(),
 			},
-			Name: "job_duration_seconds",
-			Help: "Total time taken in seconds to finish a running compaction job.",
 		},
 		[]string{druidmetrics.LabelSucceeded, druidmetrics.EtcdNamespace},
 	)
--- a/controllers/compaction/metrics.go
+++ b/controllers/compaction/metrics.go
@@ -15,6 +15,8 @@
 package compaction
 
 import (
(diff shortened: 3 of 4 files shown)


ashwani2k:
Please update the TODO's in description.

abdasgupta:
cc @istvanballok

api/v1alpha1/types_etcd.go
```diff
@@ -444,7 +444,7 @@ func (e *Etcd) GetConfigmapName() string {
 
 // GetCompactionJobName returns the compaction job name for the Etcd.
 func (e *Etcd) GetCompactionJobName() string {
-	return fmt.Sprintf("%s-compact-job", string(e.UID[:6]))
+	return fmt.Sprintf("compact-job-%s", string(e.UID[:6]))
```
istvanballok, line 447: Looks good to me!

Once this PR is merged and released with a new version of etcd druid, we should adjust the alert definition here:

https://github.com/gardener/gardener/pull/8361/files#diff-e5831943a4ab78c34cc948a519ab72334ec2d651c7621b2cd4f897488a4d9c4dR38

We could change the regex prefix `(.+)` to `(.*)` even now to support the case when the prefix is empty,
and we can remove it altogether when we know that there is no prefix. Then the new regex will become `pod!~"compact-job(.+)"`.

abdasgupta: Yes. I will do it after this PR is merged and released

shreyas-s-rao, line 447: If we're changing the compaction job name in this PR, I would vote to change it simply to `etcd-<role>-compaction-job`. We don't really need the etcd UID here since it doesn't add any information that can ease a human operator's life. Rather, calling it `etcd-main-compaction-job` makes more sense since:
1. It directly references the `etcd-main` instance, making it easier to locate the corresponding Etcd resource
2. It uses `etcd-` prefix, allowing us to group all etcd pods together when viewing pods in the control plane of the shoot cluster

PS: each Etcd instance can have utmost 1 compaction job running at any given time, so there would be no naming conflicts, same as today with the usage of the etcd UID rather than name.

api/v1alpha1/types_etcd_test.go
```diff
@@ -95,7 +95,7 @@ var _ = Describe("Etcd", func() {
 
 	Context("GetCompactionJobName", func() {
 		It("should return the correct compaction job name", func() {
-			Expect(created.GetCompactionJobName()).To(Equal("123456-compact-job"))
+			Expect(created.GetCompactionJobName()).To(Equal("compact-job-123456"))
```
shreyas-s-rao, line 98: See above comment

controllers/compaction/reconciler.go
```diff
@@ -154,6 +154,19 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if err != nil {
+		logger.Info("Could not fetch any running compaction job")
+	}
+
+	if job != nil && job.Name != "" {
+		metricJobsCurrent.With(prometheus.Labels{druidmetrics.EtcdNamespace: etcd.Namespace}).Set(1)
+	} else {
+		metricJobsCurrent.With(prometheus.Labels{druidmetrics.EtcdNamespace: etcd.Namespace}).Set(0)
+	}
+
```
shreyas-s-rao, line 169: Seems like a lot of code is jumbled up in the current code for compaction-controller:
The current sequence (without this PR) looks like this:
- if diff in full and delta events has crossed configured threshold, then:
  - try and fetch running compaction job
  - if it's in deletion, requeue after 10 seconds
  - if no compaction job exists, and if snapshot compaction is enabled, then deploy a new compaction job
  - if it's failed, delete and requeue the job recreation
  - if it's succeeded, delete and exit without requeueing

This logic is flawed due to the following reasons:
1. when a compaction job succeeds, the diff between full and delta events drops to near-zero, so when the requeue from the previous reconciler runs, it can never run `reconcileJob` method since the diff is no longer greater than the threshold. This means that a successful compaction job run can never be recorded in prometheus.
2. when a compaction job was previously deployed but is no longer running now, and in the meantime a regular scheduled full snapshot was taken by backup sidecar, the diff between full and delta events becomes near-zero, and thus `reconcileJob` method never runs. Due to this, a failed/successful/completed compaction job run might not be recorded in prometheus. This condition is fixed by you in the current PR, so thank you for that.
3. checking whether snapshot compaction is enabled or not should happen even before registering the compaction-controller with druid manager, and this fix would be taken care of in #671 by @seshachalam-yv 

@abdasgupta I would request you to discuss this with @seshachalam-yv so that you both can come up with a holistic solution that ensures that the logic is sound and the exported metrics are precise.

controllers/compaction/metrics.go
```diff
@@ -53,8 +56,15 @@ var (
...
 			Subsystem: subsystemCompaction,
-			Name:      "job_duration_seconds",
-			Help:      "Total time taken in seconds to finish a running compaction job.",
+			Buckets: []float64{baseDuration.Seconds(),
+				5 * baseDuration.Seconds(),
+				10 * baseDuration.Seconds(),
+				15 * baseDuration.Seconds(),
+				20 * baseDuration.Seconds(),
+				30 * baseDuration.Seconds(),
+			},
```
shreyas-s-rao, line 65: Just a nit. It's easier for readability if the `Buckets` field comes towards the end of the struct, so that the name and purpose of the metric is stated first, followed by other parameters such as buckets for histogram.

shreyas-s-rao (changes requested):
@abdasgupta thanks for your PR. I have requested for a few changes. PTAL.

/cc @seshachalam-yv

abdasgupta:
I had an out-of-band discussion on this with @shreyas-s-rao and @seshachalam-yv and we came to conclusion that we need to take care of `metricJobsCurrent` with more accuracy. So we decided that, instead of dealing `metricJobsCurrent` in this same PR which is only handling small improvements, I will raise another PR to handle `metricJobsCurrent` and update this PR only with the rest of the changes.

api/v1alpha1/types_etcd_test.go
abdasgupta: Done

api/v1alpha1/types_etcd.go
abdasgupta: Done

controllers/compaction/metrics.go
abdasgupta: Done

controllers/compaction/reconciler.go
abdasgupta: Raised https://github.com/gardener/etcd-druid/issues/685 in favour of the discussion

abdasgupta:
Raised https://github.com/gardener/etcd-druid/issues/685 to take care of `metricsjobcurrent` metric updation correctly.

shreyas-s-rao (approved):
@abdasgupta thanks for making the requested changes and removing the contentious code changes. 
/lgtm

PR #673, opened 2023-08-21
[Bug] Moved all default images to a single file

aaronfern:

/kind bug

We found a bug where the `etcd-wrapper` images were not considered by the landscape because it was specified in a different file instead of the standard expected `images.yaml` file

This PR moves the `etcd-wrapper` images so that now all images are in a single `images.yaml` file

Fixes #

**Release note**:

```other operator
All default images are now present in `images.yaml`
```


controllers/compaction/reconciler.go
```diff
@@ -260,7 +260,7 @@ func (r *Reconciler) delete(ctx context.Context, logger logr.Logger, etcd *druid
 func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd) (*batchv1.Job, error) {
 	activeDeadlineSeconds := r.config.ActiveDeadlineDuration.Seconds()
 
-	_, etcdBackupImage, err := utils.GetEtcdImages(etcd, r.imageVector)
+	_, etcdBackupImage, err := utils.GetEtcdImages(etcd, r.imageVector, r.config.FeatureGates[string(features.UseEtcdWrapper)])
```
unmarshall, line 263: Not directly related to this PR but since you are making related changes i will propose one small change.
In each Config we have `FeatureGates map[string]bool`. Instead you can use `FeatureGates map[featuregate.Feature]bool` This then allows you do just do this for the above line of code:
```go
utils.GetEtcdImages(etcd, r.imageVector, r.config.FeatureGates[features.UseEtcdWrapper])
```
Casting to string is quite unnecessary specially when `featuregate.Feature` is itself is of type string.

controllers/etcd/reconciler.go
```diff
@@ -77,7 +77,7 @@ type Reconciler struct {
 
 // NewReconciler creates a new reconciler for Etcd.
 func NewReconciler(mgr manager.Manager, config *Config) (*Reconciler, error) {
-	imageVector, err := utils.CreateImageVector(config.FeatureGates[string(features.UseEtcdWrapper)])
+	imageVector, err := utils.CreateImageVector()
```
unmarshall, line 80: can we unify the import alias name and call it `ctrlutils` as we have done it in compaction reconciler?

pkg/utils/image.go
```diff
@@ -25,15 +25,27 @@ import (
 // It will give preference to images that are set in the etcd spec and only if the image is not found in it should
 // it be picked up from the image vector if it's set there.
 // A return value of nil for either of the images indicates that the image is not set.
-func GetEtcdImages(etcd *druidv1alpha1.Etcd, iv imagevector.ImageVector) (*string, *string, error) {
-	etcdImage, err := chooseImage(common.Etcd, etcd.Spec.Etcd.Image, iv)
+func GetEtcdImages(etcd *druidv1alpha1.Etcd, iv imagevector.ImageVector, useEtcdWrapper bool) (*string, *string, error) {
+	var err error
+	var etcdImage, etcdBackupRestoreImage *string
```
unmarshall, line 30: A more cleaner way to do it would be:
```go
func getEtcdImageKeys(useEtcdWrapper bool) (etcdImageKey string, etcdbrImageKey string) {
	switch useEtcdWrapper {
	case true:
		etcdImageKey = common.EtcdWrapper
		etcdbrImageKey = common.BackRestoreDistroless
	default:
		etcdImageKey = common.Etcd
		etcdbrImageKey = common.BackupRestore
	}
	return
}

func GetEtcdImages(etcd *druidv1alpha1.Etcd, iv imagevector.ImageVector, useEtcdWrapper bool) (*string, *string, error) {
	etcdImageKey, etcdbrImageKey := getEtcdImageKeys(useEtcdWrapper)
	etcdImage, err := chooseImage(etcdImageKey, etcd.Spec.Etcd.Image, iv)
	if err != nil {
		return nil, nil, err
	}
	etcdBackupRestoreImage, err := chooseImage(etcdbrImageKey, etcd.Spec.Backup.Image, iv)
	if err != nil {
		return nil, nil, err
	}
	return etcdImage, etcdBackupRestoreImage, nil
}
```

pkg/utils/image.go
```diff
@@ -52,6 +64,10 @@ func chooseImage(key string, specImage *string, iv imagevector.ImageVector) (*st
 }
 
 // GetEtcdBackupRestoreImage returns the image for backup-restore from the given image vector.
-func GetEtcdBackupRestoreImage(iv imagevector.ImageVector) (*string, error) {
-	return chooseImage(common.BackupRestore, nil, iv)
+func GetEtcdBackupRestoreImage(iv imagevector.ImageVector, useEtcdWrapper bool) (*string, error) {
```
unmarshall, line 67: you can also use the above proposed function `getEtcdImageKeys` like follows:
```go
_, etcdbrImageKey := getEtcdImageKeys(useEtcdWrapper)
return chooseImage(etcdbrImageKey, nil, iv)
```

aaronfern:
Thanks for your review @unmarshall 
I've addressed your review comments. PTAL

seshachalam-yv (approved):
Thanks @aaronfern 
/lgtm
/approve

shreyas-s-rao:
@aaronfern thanks for this fix! I do see one slight problem in terms of maintenance of the `images.yaml` file. Since all new changes (features) for etcd-backup-restore are developed on the distroless branches (v0.25+), any time we make a release of etcd-backup-restore, the automated image update PR on druid will update the version for the entry for `etcd-backup-restore` instead of `etcd-backup-restore-distroless`. This means that every time such an automated PR is raised, one of the us maintainers will have to consciously edit the PR or re-open it with the correct change.

It's not a major inconvenience, but definitely an FYI that everyone needs to keep in mind when making releases for the components.

aaronfern:
Ah, you're right
Ideally it should've been the other was round with the distroless image in the entry for `etcd-backup-restore`, that would've saved us a bit of trouble

unmarshall:
>any time we make a release of etcd-backup-restore, the automated image update PR on druid will update the version for the entry for etcd-backup-restore instead of etcd-backup-restore-distroless. This means that every time such an automated PR is raised, one of the us maintainers will have to consciously edit the PR or re-open it with the correct change.

Is it not possible to then change the name of the docker image for `etcd-backup-restore` to be in sync with what we have on  druid. So for versions v0.25+ use `etcd-backup-restore-distroless` and for all earlier versions use `etcd-backup-restore`. Will this not resolve the issue?

PR #681, opened 2023-09-15
Drop support for `PodDisruptionBudget`s with API version `policy/v1beta1`

aaronfern:

/area control-plane
/kind api-change
/kind cleanup

`PodDisruptionBudget`s from API version `policy/v1beta1` were deprecated from k8s `v1.21.x` onwards and were available as part of API version `policy/v1`
With #430, `etcd-druid` supported `PodDisruptionBudget`s from both the API versions depending on the k8s version used and this feature was eventually meant to be dropped 

This PR drops support for `PodDisruptionBudget` `policy/v1beta1`

Fixes #

**Release note**:

```breaking operator
Etcd druid will now not support `policy/v1beta1` for `PodDisruptionBudget`s and will only use `policy/v1` for `PodDisruptionBudget`s
```


pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -17,111 +17,68 @@ package poddisruptionbudget
...
-			pdb.Spec.MinAvailable = &intstr.IntOrString{
-				IntVal: c.values.MinAvailable,
-				Type:   intstr.Int,
-			}
-			pdb.Spec.Selector = &metav1.LabelSelector{
-				MatchLabels: c.values.SelectorLabels,
-			}
-			return nil
-		})
-		return err
```
shreyas-s-rao, line 115: I'm surprised that this code checks the API version before creating the PDB, there isn't any substantial difference between the two objects except the annotations which were added for `policyv1beta1.PodDisruptionBudget`. Was there a reason for this? I'm curious.

shreyas-s-rao (changes requested):
@aaronfern thanks a lot for this PR. Can you please add a doc stating that druid supports k8s versions > 1.21? A new doc will be required.

I have one question about the previous code. PTAL, thanks

pkg/component/etcd/poddisruptionbudget/poddisruptionbudget.go
aaronfern: Ah, you're right. That is something that was left off from the old commit that should've been removed

If you look at an older commit form the PR that added the PDB API version change [here](https://github.com/gardener/etcd-druid/compare/b18830640140b09215af0d00f4a40185582e2cdd..adae70f3566af9e6c08fbaa29c2b6f7ec013fcb6#diff-8999169794a10a59d0e1e6a3ad6e6b8d23c718da652687ad670931127ca811f7R140-R174), I was trying to create a `DeepCopy()` of the object, update the copy and then apply that. In that context, checking the API version made sense. Later however, this was deemed not required, and I just missed removing the unneeded case statement

docs/usage/supported_k8s_versions.md
```diff
@@ -0,0 +1,10 @@
+# Supported Kubernetes Versions
+
+We strongly recommend using `etcd-druid` with the kubernetes version for which it was meant to work with
+The following is a list of kubernetes versions supported by `etcd-druid` versions
+
+| Etcd-druid version | Kubernetes version |
+|------|------|
+| >=1.20.X | >=1.25.X |
+| >=1.14.X && <=1.19.X | All versions supported |
+| <=1.13.X | < 1.25.X |
```
shreyas-s-rao, line 10:
```suggestion
We strongly recommend using `etcd-druid` with the supported kubernetes versions, published in this document.

The following is a list of kubernetes versions supported by the respective `etcd-druid` versions.

| Etcd-druid version | Kubernetes version |
|------|------|
| >=1.21.0 | >=1.25.X |
| >=1.14.0 && <1.21.0 | All versions supported |
| <1.14.0 | <1.25.X |
```

.docforge/manifest.yaml
```diff
@@ -51,3 +51,12 @@ structure:
         - Developers
   nodesSelector:
     path: /docs/proposals
+- name: usage
+  properties:
+    frontmatter:
+      title: Usage
+      weight: 6
+      categories:
+        - Operators
```
shreyas-s-rao, line 60:
```suggestion
        - Users
```

shreyas-s-rao (changes requested):
@aaronfern thanks for adding the doc. Please check and make these minor changes. Otherwise, lgtm

shreyas-s-rao (approved):
/lgtm

docs/usage/supported_k8s_versions.md
```diff
@@ -0,0 +1,10 @@
+# Supported Kubernetes Versions
+
+We strongly recommend using `etcd-druid` with the supported kubernetes versions, published in this document.
+The following is a list of kubernetes versions supported by the respective `etcd-druid` versions.
+
+| Etcd-druid version | Kubernetes version |
+|------|------|
+| >=0.20 | >=1.25 |
```
acumino, line 8: This says druid of version >=0.20 doesn't support Kubernetes version 1.24. We are planning to support 1.24 in `Gardener` till the end of the year or later. Will the 0.20 release of the druid will come after the end of this year?

PR #682, opened 2023-09-24
Add local provider support for compaction

abdasgupta:

/area control-plane
/kind bug

This PR fixes the bug that was not letting the compaction job use local storage. Without this PR, compaction jobs were not running for gardener local setup with local provider.

Fixes gardener/etcd-druid#709

**Release note**:

```bugfix operator
Local storage provider for backups is now supported for snapshot compaction jobs.
```

### DIFF
--- a/controllers/compaction/reconciler.go
+++ b/controllers/compaction/reconciler.go
@@ -180,6 +180,7 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd 
 		logger.Info("Creating etcd compaction job", "namespace", etcd.Namespace, "name", etcd.GetCompactionJobName())
 		job, err = r.createCompactionJob(ctx, logger, etcd)
 		if err != nil {
+			metricJobsTotal.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededFalse, druidmetrics.EtcdNamespace: etcd.Namespace}).Inc()
 			return ctrl.Result{
 				RequeueAfter: 10 * time.Second,
 			}, fmt.Errorf("error during compaction job creation: %v", err)
@@ -298,13 +299,37 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
 						Image:           *etcdBackupImage,
 						ImagePullPolicy: v1.PullIfNotPresent,
 						Args:            getCompactionJobArgs(etcd, r.config.MetricsScrapeWaitDuration.String()),
-						VolumeMounts:    getCompactionJobVolumeMounts(etcd, logger),
-						Env:             getCompactionJobEnvVar(etcd, logger),
 					}},
-					Volumes: getCompactionJobVolumes(ctx, r.Client, logger, etcd),
 				},
 			},
 		},
+	}
+
+	if vms, err := getCompactionJobVolumeMounts(etcd); err != nil {
+		return nil, fmt.Errorf("error while creating compaction job in %v for %v : %v",
+			etcd.Namespace,
+			etcd.Name,
+			err)
+	} else {
+		job.Spec.Template.Spec.Containers[0].VolumeMounts = vms
+	}
+
+	if env, err := getCompactionJobEnvVar(etcd); err != nil {
+		return nil, fmt.Errorf("error while creating compaction job in %v for %v : %v",
+			etcd.Namespace,
+			etcd.Name,
+			err)
+	} else {
+		job.Spec.Template.Spec.Containers[0].Env = env
+	}
+
+	if vm, err := getCompactionJobVolumes(ctx, r.Client, r.logger, etcd); err != nil {
+		return nil, fmt.Errorf("error creating compaction job in %v for %v : %v",
+			etcd.Namespace,
+			etcd.Name,
+			err)
+	} else {
+		job.Spec.Template.Spec.Volumes = vm
 	}
 
 	if etcd.Spec.Backup.CompactionResources != nil {
@@ -331,7 +356,7 @@ func getLabels(etcd *druidv1alpha1.Etcd) map[string]string {
 		"networking.gardener.cloud/to-public-networks":  "allowed",
 	}
 }
-func getCompactionJobVolumeMounts(etcd *druidv1alpha1.Etcd, logger logr.Logger) []v1.VolumeMount {
+func getCompactionJobVolumeMounts(etcd *druidv1alpha1.Etcd) ([]v1.VolumeMount, error) {
 	vms := []v1.VolumeMount{
 		{
 			Name:      "etcd-workspace-dir",
@@ -339,14 +364,9 @@ func getCompactionJobVolumeMounts(etcd *druidv1alpha1.Etcd, logger logr.Logger) 
 		},
 	}
 
-	if etcd.Spec.Backup.Store == nil {
-		return vms
-	}
-
 	provider, err := utils.StorageProviderFromInfraProvider(etcd.Spec.Backup.Store.Provider)
 	if err != nil {
-		logger.Error(err, "Storage provider is not recognized. Compaction job will not mount any volume with provider specific credentials", "namespace", etcd.Namespace, "name", etcd.Name)
-		return vms
+		return vms, fmt.Errorf("storage provider is not recognized while fetching volume mounts")
 	}
 	switch provider {
 	case utils.Local:
@@ -366,10 +386,10 @@ func getCompactionJobVolumeMounts(etcd *druidv1alpha1.Etcd, logger logr.Logger) 
 		})
 	}
 
-	return vms
+	return vms, nil
 }
 
-func getCompactionJobVolumes(ctx context.Context, cl client.Client, logger logr.Logger, etcd *druidv1alpha1.Etcd) []v1.Volume {
+func getCompactionJobVolumes(ctx context.Context, cl client.Client, logger logr.Logger, etcd *druidv1alpha1.Etcd) ([]v1.Volume, error) {
 	vs := []v1.Volume{
 		{
 			Name: "etcd-workspace-dir",
@@ -379,22 +399,16 @@ func getCompactionJobVolumes(ctx context.Context, cl client.Client, logger logr.
 		},
 	}
 
-	if etcd.Spec.Backup.Store == nil {
-		return vs
-	}
-
 	storeValues := etcd.Spec.Backup.Store
 	provider, err := utils.StorageProviderFromInfraProvider(storeValues.Provider)
 	if err != nil {
-		logger.Error(err, "Storage provider is not recognized. Compaction job will fail as no storage could be configured", "namespace", etcd.Namespace, "name", etcd.Name)
-		return vs
+		return vs, fmt.Errorf("could not recognize storage provider while fetching volumes")
 	}
 	switch provider {
 	case "Local":
 		hostPath, err := utils.GetHostMountPathFromSecretRef(ctx, cl, logger, storeValues, etcd.Namespace)
 		if err != nil {
-			logger.Error(err, "host mount path could not be determined from provided secrets", "namespace", etcd.Namespace, "name", etcd.Name)
-			return vs
+			return vs, fmt.Errorf("could not determine host mount path for local provider")
 		}
 
 		hpt := v1.HostPathDirectory
@@ -409,9 +423,7 @@ func getCompactionJobVolumes(ctx context.Context, cl client.Client, logger logr.
 		})
 	case utils.GCS, utils.S3, utils.OSS, utils.ABS, utils.Swift, utils.OCS:
 		if storeValues.SecretRef == nil {
-			logger.Info("No secretRef is configured for backup store. Compaction job will fail as no storage could be configured.",
-				"namespace", etcd.Namespace, "name", etcd.Name)
-			return vs
+			return vs, fmt.Errorf("could not configure secretRef for backup store %v", provider)
 		}
 
 		vs = append(vs, v1.Volume{
@@ -424,14 +436,11 @@ func getCompactionJobVolumes(ctx context.Context, cl client.Client, logger logr.
 		})
 	}
 
-	return vs
+	return vs, nil
 }
 
-func getCompactionJobEnvVar(etcd *druidv1alpha1.Etcd, logger logr.Logger) []v1.EnvVar {
+func getCompactionJobEnvVar(etcd *druidv1alpha1.Etcd) ([]v1.EnvVar, error) {
 	var env []v1.EnvVar
-	if etcd.Spec.Backup.Store == nil {
-		return env
-	}
 
 	storeValues := etcd.Spec.Backup.Store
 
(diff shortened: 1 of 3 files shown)


controllers/compaction/reconciler.go
```diff
@@ -364,7 +369,7 @@ func getCompactionJobVolumeMounts(etcd *druidv1alpha1.Etcd, logger logr.Logger)
 	return vms
 }
 
-func getCompactionJobVolumes(etcd *druidv1alpha1.Etcd, logger logr.Logger) []v1.Volume {
+func getCompactionJobVolumes(ctx context.Context, etcd *druidv1alpha1.Etcd, cl client.Client, logger logr.Logger) []v1.Volume {
```
shreyas-s-rao, line 372:
```suggestion
func getCompactionJobVolumes(ctx context.Context, cl client.Client, logger logr.Logger, etcd *druidv1alpha1.Etcd) []v1.Volume {
```

Standard coding practice is to first pass method-related meta parameters, such as context, clients and loggers, followed by helper objects .

controllers/compaction/reconciler.go
```diff
@@ -384,8 +389,25 @@ func getCompactionJobVolumes(etcd *druidv1alpha1.Etcd, logger logr.Logger) []v1.
 		logger.Error(err, "Storage provider is not recognized. Compaction job will fail as no storage could be configured", "namespace", etcd.Namespace, "name", etcd.Name)
 		return vs
 	}
+	switch provider {
+	case "Local":
+		hostPath, err := utils.GetHostMountPathFromSecretRef(ctx, cl, logger, storeValues, etcd.Namespace)
+		if err != nil {
+			logger.Error(err, "Host mount path could not be detremined from provided secrets", "namespace", etcd.Namespace, "name", etcd.Name)
```
shreyas-s-rao, line 396:
```suggestion
			logger.Error(err, "host mount path could not be determined from provided secrets", "namespace", etcd.Namespace, "name", etcd.Name)
```

controllers/compaction/reconciler.go
```diff
@@ -384,8 +389,25 @@ func getCompactionJobVolumes(etcd *druidv1alpha1.Etcd, logger logr.Logger) []v1.
...
+			return vs
```
shreyas-s-rao, line 397: If there is an error in getting host mount path, then you should return the error to the caller of `getCompactionJobVolumes()`, instead of returning `vs` without the correct volumes. You will need to enhance `getCompactionJobVolumes()` to also return an error, which is correctly handled in `createCompactionJob()`.

controllers/compaction/reconciler.go
```diff
@@ -384,8 +389,25 @@ func getCompactionJobVolumes(etcd *druidv1alpha1.Etcd, logger logr.Logger) []v1.
...
+		}
 
-	if provider == utils.GCS || provider == utils.S3 || provider == utils.OSS || provider == utils.ABS || provider == utils.Swift || provider == utils.OCS {
+		hpt := v1.HostPathDirectory
+		vs = append(vs, v1.Volume{
+			Name: "host-storage",
+			VolumeSource: v1.VolumeSource{
+				HostPath: &v1.HostPathVolumeSource{
+					Path: hostPath + "/" + pointer.StringDeref(storeValues.Container, ""),
```
shreyas-s-rao, line 405:
```suggestion
					Path: path.Join(hostPath, *storeValues.Container),
```

controllers/compaction/reconciler.go
```diff
@@ -348,13 +348,18 @@ func getCompactionJobVolumeMounts(etcd *druidv1alpha1.Etcd, logger logr.Logger)
...
 		return vms
 	}
-
-	if provider == utils.GCS {
+	switch provider {
+	case utils.Local:
+		vms = append(vms, v1.VolumeMount{
+			Name:      "host-storage",
+			MountPath: pointer.StringDeref(etcd.Spec.Backup.Store.Container, ""),
+		})
```
shreyas-s-rao, line 356: When using EtcdWrapper, special handling is required to mount the volume to nonroot directory. Please refer to https://github.com/gardener/etcd-druid/blob/0477a279db70d42a83cd94fa0de8caad7b70a990/controllers/etcdcopybackupstask/reconciler.go#L476-L487 as an example

shreyas-s-rao (changes requested):
@abdasgupta thanks for the PR. I have a few comments. Please address them. Additionally, you might also need to use an init container to change file permissions for the store provider directories, as seen here: https://github.com/gardener/etcd-druid/blob/0477a279db70d42a83cd94fa0de8caad7b70a990/controllers/etcdcopybackupstask/reconciler.go#L365-L383

You may consult with @aaronfern for help with this.

controllers/compaction/reconciler.go
abdasgupta: I think it is better to not handle the error separately. I am raising the error in logger so the error will be visible in container log. Apart from that, if the mount path is not determined and an empty `vs` is returned, the compaction job will also raise another error of not able to recognise the volume.  I think, we are good here.

controllers/compaction/reconciler.go
abdasgupta: What's wrong here? `StringDeref` checks nil poiinter error as well

controllers/compaction/reconciler.go
shreyas-s-rao: Errors always need to be well-handled. Druid should make no assumption about how backup-restore would handle the misconfiguration. If druid is unable to get the volume path, then it should mandatorily return error, rather than knowingly return wrong volumes. Logging it is not sufficient, since that requires a human operator to dig through the logs to figure out what went wrong.

controllers/compaction/reconciler.go
shreyas-s-rao: If it's possible to avoid a function call with such simple syntax for pointer dereference, then why not use it?

abdasgupta: This function call is doing a nil pointer check as well. I would prefer to persist with it

controllers/compaction/reconciler.go
abdasgupta: See, you can see throughout the same file that errors are handled the same way as in this function. I would not like to change the behaviour for this particular case. If a compaction job fails, the operator needs to check logs anyway. If you have another suggestion to handle this error in better way, we can work that in further enhancement but not in this PR

shreyas-s-rao: > you can see throughout the same file that errors are handled the same way

As a maintainer of the project, it would be advisable to fix the behaviour for the other error handling as well, rather than repeating previous mistakes again. If there are other instances in this file where errors are handled in a wrong way, I would actually advise to fix those as well.

> If a compaction job fails, the operator needs to check logs anyway

With well-defined and well-exposed error codes, an operator has no necessity to go through the logs of any component. As you already know, we are heading towards this direction for druid, and the first step is to ensure that errors are atleast returned correctly to the top level caller of functions.

> If you have another suggestion to handle this error in better way

I have already suggested a way to correctly handle the error, in my [initial comment] (https://github.com/gardener/etcd-druid/pull/682#discussion_r1341300449). It's a simple enough change. I would prefer if PRs don't introduce technical debt that needs to be handled later. As a maintainer, you would also realise the pain and effort it takes to correct past mistakes in code bases. Please ensure that this PR has correct error handling. For other instances of mishandling of errors, please raise an issue so that someone else might pick it up as a good first issue.

controllers/compaction/reconciler.go
shreyas-s-rao: Ok. But please use path.Join(), since that's the safe and recommended way of generating path strings, especially if we intend to run druid on a different platform/architecture tomorrow.

aaronfern:
Compaction job would not need an init container as it used an `emptyDir` volume, and that's fresh for every run, so no permission change is needed

controllers/compaction/reconciler.go
abdasgupta: We  won't need it because https://github.com/gardener/etcd-druid/pull/682#issuecomment-1744249958

controllers/compaction/reconciler.go
abdasgupta: Done

controllers/compaction/reconciler.go
abdasgupta: I considered your suggestion. But I won't apply your suggestion for error handling in this PR. Also, the suggestion looks not ideal to me.

I also want to mention that error is handled in current design as well. If `utils.GetHostMountPathFromSecretRef` returns an error, compaction controller will print the error in its log. But the controller will also allow compaction jobs to be created with no volume. So, the compaction job will eventually fail after creation and post an error in it's description. It will be helpful this way because an operator, who looks into compaction dashboard, will easily see that the compaction jobs are failing and take action on that. Otherwise, if the compaction jobs were not created at all due to error in `utils.GetHostMountPathFromSecretRef`, an operator would not know the problem easily as he needs to find the log in etcd-druid in that case.

Considering the above-mentioned reasoning, I will not implement any other way of error handling in this PR. Thank you for your review.

controllers/compaction/reconciler.go
abdasgupta: Done

controllers/compaction/reconciler.go
shreyas-s-rao: I don't agree with this reasoning. It's rudimentary to properly handle errors in the code. In this case, if the function encounters an error, it becomes its duty to pass that error back to the caller of the function to decide what action to take based on that information. But passing back a nil error and a wrong return value, ie, incomplete `vs` (volumes) for the compaction job makes no sense to me. Between catching the error within druid itself, and allowing it to deploy a compaction job (which takes up time and resources) and *then* catching error, I don't think there's any question that the former is preferable. I'm not sure why you still don't see that.

shreyas-s-rao: @abdasgupta as agreed, can you please update the details of the out-of-band discussion that we had couple of days ago?

abdasgupta: In an out-of-band discussion with @shreyas-s-rao , I agreed to make the changes for stricter error handling in Druid itself. Druid won't create a compaction job if it identifies an error in the job spec. Though I still like Druid to be just a provisioner of compaction job and the flexibility of compaction job to run independently with minimum supervision of Druid, but our future design changes are intended to give more control on Druid for the objects it spawn. So, I agreed to change this PR to take care of error handling for volume mounts and for other scenarios.

abdasgupta: Changes are done

controllers/compaction/reconciler.go
```diff
@@ -366,10 +386,10 @@ func getCompactionJobVolumeMounts(etcd *druidv1alpha1.Etcd, logger logr.Logger)
 		})
 	}
 
-	return vms
+	return vms, nil
 }
 
-func getCompactionJobVolumes(ctx context.Context, cl client.Client, logger logr.Logger, etcd *druidv1alpha1.Etcd) []v1.Volume {
+func getCompactionJobVolumes(ctx context.Context, cl client.Client, logger logr.Logger, etcd *druidv1alpha1.Etcd) ([]v1.Volume, error) {
```
shreyas-s-rao, line 392: You can now remove `logger` from these params as well, similar to what you did for `getCompactionJobVolumeMounts()`

controllers/compaction/reconciler.go
```diff
@@ -298,15 +299,39 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
 			},
 		},
 	}
 
+	if vms, err := getCompactionJobVolumeMounts(etcd); err != nil {
+		return nil, fmt.Errorf("error while creating compaction job in %v for %v : %v",
+			etcd.Namespace,
+			etcd.Name,
+			err)
+	} else {
```
shreyas-s-rao, line 313: `else` is not necessary since you are returning in the `if` block, but I understand that the scope of the condition evaluation, ie, `vms` variable gets carried over to the else block as well. Usually in any of the golang code in gardener/kubernetes, it's done like this. Not a compulsion, but you can consider it:
```
            vms, err := getCompactionJobVolumeMounts(etcd)
            if err != nil {
                return nil, fmt.Errorf("error while creating compaction job in %v for %v : %v",
			etcd.Namespace,
			etcd.Name,
			err)
	    }
            job.Spec.Template.Spec.Containers[0].VolumeMounts = vms
```

controllers/compaction/reconciler.go
```diff
@@ -298,15 +299,39 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+		job.Spec.Template.Spec.Containers[0].VolumeMounts = vms
+	}
+
+	if env, err := getCompactionJobEnvVar(etcd); err != nil {
+		return nil, fmt.Errorf("error while creating compaction job in %v for %v : %v",
+			etcd.Namespace,
+			etcd.Name,
+			err)
+	} else {
```
shreyas-s-rao, line 322: Same as above

controllers/compaction/reconciler.go
```diff
@@ -298,15 +299,39 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+		job.Spec.Template.Spec.Containers[0].Env = env
+	}
+
+	if vm, err := getCompactionJobVolumes(ctx, r.Client, r.logger, etcd); err != nil {
+		return nil, fmt.Errorf("error creating compaction job in %v for %v : %v",
+			etcd.Namespace,
+			etcd.Name,
+			err)
+	} else {
```
shreyas-s-rao, line 331: Same as above

shreyas-s-rao (changes requested):
@abdasgupta thanks for making the requested changes. Overall looks good, except for couple of nits. PTAL, thanks

test/utils/etcd.go
```diff
@@ -361,6 +373,15 @@ func getBackupStore(name string, provider druidv1alpha1.StorageProvider) *druidv
...
 
+func getBackupStoreForLocal(name string) *druidv1alpha1.StoreSpec {
+	provider := druidv1alpha1.StorageProvider("local")
+	return &druidv1alpha1.StoreSpec{
+		Container: &container,
+		Prefix:    name,
+		Provider:  &provider,
+	}
+}
+
```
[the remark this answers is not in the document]
abdasgupta: Why to add an unnecessary secret ref when case for local provider can be handled separately. I thought it more cleaner this way. I tested it well and it was working fine. I don't know what error may get in due to adding extra `SecretRef`

controllers/compaction/reconciler.go
abdasgupta: `GetHostMountPathFromSecretRef` function used in this method needs the logger

controllers/compaction/reconciler.go
abdasgupta: Two reasons:
1. I got to use the nice feature of go lang `if vms, err := getCompactionJobVolumeMounts(etcd); err != nil {` to put assignment and condition in same line. It always looks more appealing and cleaner to me
2. As the scope is only limited to the if condition, garbage collector does not need to preserve the scope of the variable `vms` till the end of the function but only within the if block. So, an extra variable is free just after the `if` block.

shreyas-s-rao (approved):
/lgtm

PR #697, opened 2023-10-04
Updated recovery from permanent quorum loss operation guide.

ishan16696:
With this PR: https://github.com/gardener/etcd-backup-restore/pull/661 the operator needs to perform one more step in case of recovery from permanent quorum loss. 
This PR updates permanent quorum loss operation guide with that additional step which needs to be performed in recovery from permanent quorum loss.

Fixes #

**Release note**:

```doc operator
Updated the recovery from permanent quorum loss ops guide.
```

### DIFF
--- a/docs/operations/recovery-from-permanent-quorum-loss-in-etcd-cluster.md
+++ b/docs/operations/recovery-from-permanent-quorum-loss-in-etcd-cluster.md
@@ -1,7 +1,7 @@
 # Recovery from Permanent Quorum Loss in an Etcd Cluster
 
 ## Quorum loss in Etcd Cluster
-[Quorum loss](https://etcd.io/docs/v3.4/op-guide/recovery/) means when majority of Etcd pods(greater than or equal to n/2 + 1) are down simultaneously for some reason. 
+[Quorum loss](https://etcd.io/docs/v3.4/op-guide/recovery/) means when majority of Etcd pods(greater than or equal to n/2 + 1) are down simultaneously for some reason.
 
 There are two types of quorum loss that can happen to [Etcd multinode cluster](https://github.com/gardener/etcd-druid/tree/master/docs/proposals/multi-node) :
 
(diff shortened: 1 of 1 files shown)


shreyas-s-rao (approved):
/lgtm

docs/operations/recovery-from-permanent-quorum-loss-in-etcd-cluster.md
```diff
@@ -47,8 +48,23 @@ There are two [Etcd clusters](https://github.com/gardener/etcd-druid/tree/master
       Delete all PVCs that are attached to `etcd-main` cluster.
 
       `kubectl delete pvc -l instance=etcd-main`
-   5. Edit the `etcd-main` cluster's configmap (ex: `etcd-bootstrap-4785b0`) as follows:
-      
+
+   5. Check the etcd's member leases. There should be leases starting with `etcd-main` as many as `etcd-main` replicas.
+   One of those leases will have holder identity as `<etcd-member-id>:Leader` and rest of etcd member leases have holder identities as `<etcd-member-id>:Member`.
```
abdasgupta, line 53: Would it be better to re-write the sentence as:
```suggestion
   One of those leases will have holder identity as `<16 digit etcd member id>:Leader` and rest of etcd member leases have holder identities as `<16 digit etcd member id>:Member`.
```

I suggested this because at the first look I was trying figure out the meaning of `etcd`, `member`, `id`, instead of considering them as part of a single phrase.

abdasgupta (changes requested):
Looks good. I just requested a minor change. Please, make the change only if you think it valid.

docs/operations/recovery-from-permanent-quorum-loss-in-etcd-cluster.md
ishan16696: should we mention about `16 digit` 🤔 ?

abdasgupta: We can mention that if we are sure that all the member ids are 16 digits. Are  we limiting the holder identity to only contain 16 digits? If so, we can mention `16 digit` in the doc.

ishan16696: but lease holder identity is already mentioned ? `16 digit` might cause an confusion and it is a miscellaneous information which need not to be present in doc

abdasgupta: Okay. I agree. We don't need to mention `16 digit` as it may add further confusion . I am resolving this conversation .

abdasgupta (approved):
Looks good.

Issue #698, opened 2023-10-04
[BUG] Improve error handling and reconciliation logic for compaction controller


shreyas-s-rao:
**Describe the bug:**
Compaction controller's purpose is to deploy a compaction job if it detects that a compaction job is necessary, due to possible triggers such as delta snapshot revision exceeding full snapshot revision by a given threshold (`etcd-events-threshold`), and in the future even upon the total delta snapshots' size crossing a certain threshold. Information about snapshot revisions is available today from snapshot leases maintained by the leading backup sidecar, and in the future will be published via [EtcdMember resource status](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/04-etcd-member-custom-resource.md#etcdmember).

Once the compaction job has been deployed, the controller requeues reconciliation. Until the compaction job succeeds or fails, the reconciliation must keep requeueing and checking compaction job status. Once compaction job succeeds/fails, the snapshot revision information will be updated by the compaction job itself, and compaction controller will know not to deploy a new compaction job until the next time a snapshot compaction trigger is received.

All the logic for reconciling the compaction job lies within [reconcileJob()](https://github.com/gardener/etcd-druid/blob/00274d0929b34e41486236946ec1488344212495/controllers/compaction/reconciler.go#L165). The issue with the current logic in compaction controller is that this method is called only [when the compaction threshold is reached](https://github.com/gardener/etcd-druid/blob/00274d0929b34e41486236946ec1488344212495/controllers/compaction/reconciler.go#L158-L160). But if a new full (regular scheduled) snapshot is uploaded by the leading backup sidecar while a compaction job is still running, then compaction controller never runs `reconcileJob()` because the snapshot revision threshold is no longer reached. This prevents the controller from reconciling the currently running compaction job to either cut it short, or atleast wait for completion and export the correct metrics for it.

[...]

PR #711, opened 2023-10-16
Reconciles compaction job on job status change along with snapshot lease change

### LINKED ISSUE #685: [BUG] Make the metric metricsjobcurrent capture accurate job end time
**Describe the bug:**
Currently the metric `metricsjobcurrent` is updated only when the delta event threshold crosses the limit and the compaction-controller in druid tries to create a new job or follow up the old job. But ideally, the metric should be updated even when the delta events are successfully compacted and the job is successful.

**Expected behavior:**
After a compaction job is run successfully and the delta events are compacted in backup, metric `metricsjobcurrent` should be updated to mark the end of the currently running job correctly in the dashboard. We can watch on the running compaction job status and as soon as the job is successful, we can update the `metricsjobcurrent` in the reconciler.

Compaction job dashboard does not show the graph for `metricsjobcurrent` with correct end time of any successfully run compaction job.

**Logs:**

 - Etcd version/commit ID :

abdasgupta:

/area control-plane
/kind bug

This PR solves a bug due to which compaction job related metrics were not having exactly accurate values. This PR actually reconciles compaction job on job status changes along with the holder identity changes in snapshot leases.

Fixes #685 

**Release note**:

```noteworthy operator
Compaction job now reconciles on Job Status changes along with the holder identity changes in snapshot leases.
```

### DIFF
--- a/controllers/compaction/reconciler.go
+++ b/controllers/compaction/reconciler.go
@@ -111,18 +111,18 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd 
 	job := &batchv1.Job{}
 	if err := r.Get(ctx, types.NamespacedName{Name: etcd.GetCompactionJobName(), Namespace: etcd.Namespace}, job); err != nil {
 		if errors.IsNotFound(err) {
-			logger.Info("Currently, no compaction job is running")
+			logger.Info("Currently, no compaction job is running in the namespace ", etcd.Namespace)
 		} else {
 			// Error reading the object - requeue the request.
 			return ctrl.Result{
 				RequeueAfter: 10 * time.Second,
-			}, err
+			}, fmt.Errorf("error while fetching compaction job with the name %v, in the namespace %v: %v", etcd.GetCompactionJobName(), etcd.Namespace, err)
 		}
 	}
 
 	if job != nil && job.Name != "" {
 		if !job.DeletionTimestamp.IsZero() {
-			logger.Info("Job is already in deletion. A new job will be created only if the previous one has been deleted.", "namespace", job.Namespace, "name", job.Name)
+			logger.Info("Job is already in deletion. A new job will be created only if the previous one has been deleted.", "namespace: ", job.Namespace, "name: ", job.Name)
 			return ctrl.Result{
 				RequeueAfter: 10 * time.Second,
 			}, nil
--- a/controllers/compaction/reconciler.go
+++ b/controllers/compaction/reconciler.go
@@ -109,9 +109,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd) (ctrl.Result, error) {
 	// Update metrics for currently running compaction job, if any
 	job := &batchv1.Job{}
-	err := r.Get(ctx, types.NamespacedName{Name: etcd.GetCompactionJobName(), Namespace: etcd.Namespace}, job)
-	if err != nil {
-		logger.Info("Could not fetch any running compaction job")
+	if err := r.Get(ctx, types.NamespacedName{Name: etcd.GetCompactionJobName(), Namespace: etcd.Namespace}, job); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("Currently, no compaction job is running")
+		} else {
+			// Error reading the object - requeue the request.
+			return ctrl.Result{
+				RequeueAfter: 10 * time.Second,
+			}, err
+		}
 	}
 
 	if job != nil && job.Name != "" {
@@ -135,7 +141,7 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd 
 			if job.Status.CompletionTime != nil {
 				metricJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededTrue, druidmetrics.EtcdNamespace: etcd.Namespace}).Observe(job.Status.CompletionTime.Time.Sub(job.Status.StartTime.Time).Seconds())
 			}
-			if err = r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
 				logger.Error(err, "Couldn't delete the successful job", "namespace", etcd.Namespace, "name", etcd.GetCompactionJobName())
 				return ctrl.Result{
 					RequeueAfter: 10 * time.Second,
@@ -150,7 +156,7 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd 
 			if job.Status.StartTime != nil {
 				metricJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededFalse, druidmetrics.EtcdNamespace: etcd.Namespace}).Observe(time.Since(job.Status.StartTime.Time).Seconds())
 			}
-			err = r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground))
+			err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground))
 			if err != nil {
 				return ctrl.Result{
 					RequeueAfter: 10 * time.Second,
--- a/controllers/compaction/reconciler.go
+++ b/controllers/compaction/reconciler.go
@@ -82,7 +82,7 @@ func NewReconcilerWithImageVector(mgr manager.Manager, config *Config, imageVect
 
 // Reconcile reconciles the compaction job.
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	r.logger.Info("Lease controller reconciliation started")
+	r.logger.Info("Compaction job reconciliation started")
 	etcd := &druidv1alpha1.Etcd{}
 	if err := r.Get(ctx, req.NamespacedName, etcd); err != nil {
 		if errors.IsNotFound(err) {
@@ -107,12 +107,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	logger := r.logger.WithValues("etcd", kutil.Key(etcd.Namespace, etcd.Name).String())
 
+	// Update metrics for currently running compaction job, if any
+	job := &batchv1.Job{}
+	err := r.Get(ctx, types.NamespacedName{Name: etcd.GetCompactionJobName(), Namespace: etcd.Namespace}, job)
+	if err != nil {
+		logger.Info("Could not fetch any running compaction job")
+	}
+
+	if job != nil && job.Name != "" {
+		// Check if there is one active job or not
+		if job.Status.Active > 0 {
+			metricJobsCurrent.With(prometheus.Labels{druidmetrics.EtcdNamespace: etcd.Namespace}).Set(1)
+			// Don't need to requeue if the job is currently running
+			return ctrl.Result{}, nil
+		}
+
+		// Delete job if the job succeeded
+		if job.Status.Succeeded > 0 {
+			metricJobsTotal.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededTrue, druidmetrics.EtcdNamespace: etcd.Namespace}).Inc()
+			metricJobsCurrent.With(prometheus.Labels{druidmetrics.EtcdNamespace: etcd.Namespace}).Set(0)
+			if job.Status.CompletionTime != nil {
+				metricJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededTrue, druidmetrics.EtcdNamespace: etcd.Namespace}).Observe(job.Status.CompletionTime.Time.Sub(job.Status.StartTime.Time).Seconds())
+			}
+			if err = r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+				logger.Error(err, "Couldn't delete the succeful job", "namespace", etcd.Namespace, "name", etcd.GetCompactionJobName())
+			}
+		}
+	}
+
 	// Get full and delta snapshot lease to check the HolderIdentity value to take decision on compaction job
 	fullLease := &coordinationv1.Lease{}
 
 	if err := r.Get(ctx, kutil.Key(etcd.Namespace, etcd.GetFullSnapshotLeaseName()), fullLease); err != nil {
 		logger.Error(err, "Couldn't fetch full snap lease", "namespace", etcd.Namespace, "name", etcd.GetFullSnapshotLeaseName())
-
 		return ctrl.Result{
 			RequeueAfter: 10 * time.Second,
 		}, err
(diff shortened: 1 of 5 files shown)


controllers/compaction/reconciler.go
```diff
@@ -107,12 +107,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	logger := r.logger.WithValues("etcd", kutil.Key(etcd.Namespace, etcd.Name).String())
 
+	// Update metrics for currently running compaction job, if any
+	job := &batchv1.Job{}
```
shreyas-s-rao, line 111: Why are we fetching the job here and checking the status (active, succeeded, failed) when we're already doing it in `reconcileJob()`? This seems redundant.

The PR may be correcting the inaccuracy in metrics exported, but in the process introduces redundancies and inefficiency in the reconciliation loop. I would ask you to please take a more holistic approach for this PR - if done right, it can solve both #685 as well as #698, by correcting the reconciliation loop flow (`reconcileJob()` should be called even if the revision difference crosses the threshold, unlike how it is done today). Ideally, reconciliation loop logic should be sound, and only then metrics should be exposed based on the operations and status information from the job.

shreyas-s-rao (changes requested):
@abdasgupta thanks for your PR. PTAL at the comment and incorporate it in the PR. Thanks

controllers/compaction/reconciler.go
abdasgupta: We are checking the job status here because `reconcileJob()` is only called when the revision difference crosses the threshold. If the Job status is active, it also don't check the revision difference unnecessarily.

I don't understand what do you mean by `reconcileJob() should be called even if the revision difference crosses the threshold, unlike how it is done today` ! Currently, reconcileJob() is called upon revision difference crossing the threshold ..

What do you mean by holistic approach? You can point out which case is not handled in https://github.com/gardener/etcd-druid/issues/685 and https://github.com/gardener/etcd-druid/issues/698. I will add the case in this PR, if not added already.

shreyas-s-rao: > We are checking the job status here because reconcileJob() is only called when the revision difference crosses the threshold. If the Job status is active, it also don't check the revision difference unnecessarily

This is exactly what needs to be changed. reconcileJob() needs to run irrespective of the difference in revisions. Meaning, the revision difference needs to be checked within reconcileJob(), which then determines whether to deploy a new job or not. But the check for the currently running job needs to happen irrespective of the revision check.

```code
reconcile() {
  ...
  reconcileJob()
}

reconcileJob() {
  getJob()
  if job exists {
    checkRunningJobStatus()
    updateJobStatusMetrics()
    if jobStatus == Running {
      return
    }
    if jobStatus == Succeeded || jobStatus == Failed {
      deleteJob()
    }
  }

  if deltaSnapRev - fullSnapRev > threshold {
    deployJob()
  }
}
```

This ensures that the reconciliation loop logic is sound, and also takes care of metrics being correctly exported.

abdasgupta: Okay. Now I understand why do you think it's redundancy !

See, deployJob() can also fail instantly (for some reason). That's why we need to check job status even after deployJob() so that the controller is not run again from start. Only redundant check for  `job` is in line number 195. I think that is required after fetching leases and checking diff. One extra job status check won't be too much.

Otherwise , it is just moving codes from one block to other.

shreyas-s-rao: > deployJob() can also fail instantly (for some reason)

For what reason exactly? And why would that have any effect on checking the job status?

> That's why we need to check job status even after deployJob() so that the controller is not run again from start

Once the reconciler deploys the job, it must requeue and exit, so that in the next reconciliation it can again check the job status and take appropriate actions. 

> One extra job status check won't be too much

It's not about whether it's too much. Even though the network impact of this may me minimal, the fact that the reconciler checks the same job status twice seems unnecessary to me, when in fact we can simplify the reconciliation flow logic to ensure that the job status is fetched irrespective of the revision threshold breach, and deploy a new job only if the threshold is breached and a job is not already currently running. I think this is a valid ask.

abdasgupta: I tested repeatedly and after facing various error conditions under various circumstances , I decided this flow. I can't list them now as I don't remember all of them. What you suggesting is movement of code block from one place to another. I had my rationale and I tried to save a few loop reconciliation. But one or two redundant check of job status won't do any harm, specially when the job status comes from informer cache. Current flow goes like this:

1. In the Reconciler, a job is first checked for status. If it's succeeded, a delete call is issued on job.
2. The flow then enters to fetching leases and check the diff between leases.  Ideally, for most cases, when the reconciler is called upon status change of job for succeeded, diff check would not cross threshold because the delta events are just compacted. But, there might be some busy cluster where a job status change to successful and diff threshold cross may happen simultaneously.
3. Then we need to check if the job is really deleted or not and so we need the "redundant" job check of line number 195.
4. After the `createCompactionJob` call, I am no more issuing `Get` call on job, other than while deleting.
5. You may think line number 248 is redundant but I placed that for a reason. This line will only execute when a job is instantaneously succeeded, may be if there is no full snapshot in the storage at all or may be for some unknown reason which we should always be prepared for in a distributed system.
6. Action upon `job.Status.Failed` is executing only once, it's not redundant.

As I said, I won't feel confident if I need to change the flow that I wrote because I came to it after lots of testing in real scenarios upon various favourable and erroneous conditions. But if you just wish to move the codes from line number 110 inside the function `reconcileJob()`, that I can do. Do you wish that?

shreyas-s-rao: > What you suggesting is movement of code block from one place to another

Not really. I suggested not adding the code block in this PR as it already exists in the code, within `reconcileJob()` method. I simply suggested calling `reconcileJob()` irrespective of the revision threshold breach, because reconciliation of a job is not just about deploying the job, but also fetching the status of the job and deleting the job, and these tasks need to performed irrespective of whether the revision threshold has been breached or not. Only the task of deploying a new job should be dependent on the revision threshold breach.

regarding the flow, I would request a second opinion from @unmarshall , because to me the flow you mentioned still seems to be repeated certain code like getJob() and checkStatus() partially and deleteJob(), and seems to be trying to handle specific corner cases, rather than being generic enough to work for any case.

/needs second-opinion
/assign @unmarshall

controllers/predicate/predicate.go
```diff
@@ -179,8 +149,9 @@ func EtcdReconciliationFinished(ignoreOperationAnnotation bool) predicate.Predic
 	}
 }
 
-// IsSnapshotLease is a predicate that is `true` if the passed lease object is a snapshot lease.
-func IsSnapshotLease() predicate.Predicate {
+// SnapshotLeaseChanged is a predicate that is `true` if the passed lease object is a snapshot lease and if the lease
+// object's holderIdentity is updated.
+func SnapshotLeaseChanged() predicate.Predicate {
```
unmarshall, line 154: The lease holder identity currently holds the revision number. What you compare is only the revision. Why not just name it - `SnapshotRevisionChanged` which is semantically correct name for the function?

controllers/predicate/predicate.go
```diff
@@ -205,3 +189,33 @@ func IsSnapshotLease() predicate.Predicate {
 		},
 	}
 }
+
+// JobStatusChanged is a predicate that is `true` if the status of a job changes.
+func JobStatusChanged() predicate.Predicate {
```
unmarshall, line 194: Why is this predicate required?

controllers/predicate/predicate_test.go
```diff
@@ -114,13 +115,99 @@ var _ = Describe("Druid Predicate", func() {
 		var pred predicate.Predicate
 
 		JustBeforeEach(func() {
-			pred = LeaseHolderIdentityChange()
+			pred = SnapshotLeaseChanged()
```
unmarshall, line 118: Should be renamed to `SnapshotRevisionChanged`

controllers/predicate/predicate.go
```diff
@@ -190,12 +161,25 @@ func IsSnapshotLease() predicate.Predicate {
...
+
 	return predicate.Funcs{
 		CreateFunc: func(event event.CreateEvent) bool {
 			return isSnapshotLease(event.Object)
 		},
 		UpdateFunc: func(event event.UpdateEvent) bool {
-			return isSnapshotLease(event.ObjectNew)
+			return isSnapshotLease(event.ObjectNew) && holderIdentityChange(event.ObjectOld, event.ObjectNew)
 		},
 		GenericFunc: func(event event.GenericEvent) bool {
```
unmarshall, line 184: Why do you need to return true for `GenericFunc`? You do not do that in `JobStatusChanged` predicate.
Also as per the definition of `GenericEvent` I do not see why a `true` should ever be returned.
Refer: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.3/pkg/event#GenericEvent
Currently we do not define any `Source` which generates events of this type. Please adjust the unit tests as well.

controllers/predicate/predicate.go
```diff
@@ -205,3 +189,33 @@ func IsSnapshotLease() predicate.Predicate {
 		},
```
unmarshall, line 189: Why do we need to react to delete event of a snapshot lease?

controllers/compaction/reconciler.go
```diff
@@ -107,12 +107,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if job != nil && job.Name != "" {
+		// Check if there is one active job or not
+		if job.Status.Active > 0 {
+			metricJobsCurrent.With(prometheus.Labels{druidmetrics.EtcdNamespace: etcd.Namespace}).Set(1)
+			// Don't need to requeue if the job is currently running
+			return ctrl.Result{}, nil
+		}
+
+		// Delete job if the job succeeded
+		if job.Status.Succeeded > 0 {
```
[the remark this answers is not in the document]
abdasgupta: I think we really need a call to clarify the doubts.. I deliberately left failed situation unattended here

controllers/predicate/predicate.go
abdasgupta: I think it is easier for anyone going through the code for the first time to understand that this function is dealing with Snapshot leases if we name it as `SnapshotLeaseChanged` .. if we name it `SnapshotRevisionChanged`, the person needs to understand quite a lot about backup-restore internals.

controllers/predicate/predicate.go
abdasgupta: To trigger the reconciler when a job status is changed

controllers/predicate/predicate.go
abdasgupta: Yes.. We can return false here

controllers/predicate/predicate.go
abdasgupta: Is it not a good idea to check the revision diff for one last time and save snapshot using compaction?

controllers/predicate/predicate.go
unmarshall: Well if someone is looking at etcd druid code base then first confusion point will be why do you even need a lease for a full and delta snapshot. Second confusing point will be what is this weird looking number mean when i look at the holderIdentity which is meant to hold something completely different. 

So from your own arguments it now becomes very clear to me when the predicate is named as `SnapshotRevisionChanged`. Any code reader would then realize that this predicate will check that the snapshot (full or delta) revision has changed and how it checks is the implementation detail. When he/she looks at how its done then they will come to realize that the holderIdentity holds the revision.

controllers/predicate/predicate.go
unmarshall: In which even will a lease be deleted?

controllers/predicate/predicate.go
unmarshall: When a Job is created then a re-queue is already done by the compaction reconciler. There is only one case which is etcd-druid restarts after spawning a Job. In that case also it will still get an event for the Job. Previously this predicate was not there, therefore my question on why is it required now

abdasgupta: Mainly to update the job related metrics correctly

controllers/compaction/reconciler.go
abdasgupta: I made the changes

controllers/predicate/predicate_test.go
abdasgupta: Done

controllers/predicate/predicate.go
abdasgupta: Done

controllers/predicate/predicate.go
abdasgupta: added changes to not react to delete event

controllers/predicate/predicate.go
abdasgupta: We discussed this in the meeting on why do we need it

controllers/predicate/predicate.go
abdasgupta: Renamed

controllers/compaction/reconciler.go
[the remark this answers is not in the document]
abdasgupta: I made the changes after an offline meeting where we decided that we exploit the reconciler to reconcile again in case of some immediate error just after creation of the job by the controller.

shreyas-s-rao (approved):
@abdasgupta thanks a lot for making the requested changes.
/lgtm

controllers/compaction/reconciler.go
```diff
@@ -96,23 +96,78 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	logger := r.logger.WithValues("etcd", kutil.Key(etcd.Namespace, etcd.Name).String())
+
+	return r.reconcileJob(ctx, logger, etcd)
+}
+
+func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd) (ctrl.Result, error) {
+	// Update metrics for currently running compaction job, if any
+	job := &batchv1.Job{}
+	err := r.Get(ctx, types.NamespacedName{Name: etcd.GetCompactionJobName(), Namespace: etcd.Namespace}, job)
+	if err != nil {
```
unmarshall, line 113: We are treating NotFound error and other errors as same. If there is an error getting a Job should we not re-queue instead of just logging it? In case it is not found then we move ahead.

abdasgupta: Made the change

controllers/compaction/reconciler.go
```diff
@@ -96,23 +96,84 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+
+	return r.reconcileJob(ctx, logger, etcd)
+}
+
+func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd) (ctrl.Result, error) {
+	// Update metrics for currently running compaction job, if any
+	job := &batchv1.Job{}
+	if err := r.Get(ctx, types.NamespacedName{Name: etcd.GetCompactionJobName(), Namespace: etcd.Namespace}, job); err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("Currently, no compaction job is running")
```
unmarshall, line 114: Can you please improve this log a bit by also logging the namespace and job name? Since druid runs for multiple etcd clusters, this log will not be very useful without additional information.

abdasgupta: Done. Though we shouldn't print name of the job when there is no job running.

shreyas-s-rao: I think namespace and etcd name should be sufficient, and rather necessary to identify which job does not exist according to the log. It's easier to simply add a `logger.WithValues()` at the beginning of reconciliation, so that all logs already have the etcd name and namespace, and the same logger is passed around within the reconciler.

unmarshall (approved):
/lgtm

PR #735, opened 2023-12-12
switch GCR -> Artifact-Registry

ccwienk:
GCR has been deprecated [0] in favour of Artifact-Registry.

Thus, change push-targets for OCI-Images:

- europe-docker.pkg.dev/gardener-project/snapshots for snapshots
- europe-docker.pkg.dev/gardener-project/releases for releases
- europe-docker.pkg.dev/gardener-project/public for combined view of snapshots + releases

[0]
https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr

**Release note**:

```breaking operator
Change OCI Image Registry from GCR (`eu.gcr.io/gardener-project`) to Artifact-Registry (`europe-docker.pkg.dev/gardener-project/releases`). Users should update their references.

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ spec:
     fullSnapshotSchedule: 0 */24 * * *
     garbageCollectionPeriod: 43200s
     garbageCollectionPolicy: Exponential
-    imageRepository: eu.gcr.io/gardener-project/gardener/etcdbrctl
+    imageRepository: europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl
     imageVersion: v0.25.0
     port: 8080
     resources:
@@ -67,7 +67,7 @@ spec:
     clientPort: 2379
     defragmentationSchedule: 0 */24 * * *
     enableTLS: false
-    imageRepository: eu.gcr.io/gardener-project/gardener/etcd-wrapper
+    imageRepository: europe-docker.pkg.dev/gardener-project/public/gardener/etcd-wrapper
     imageVersion: v0.1.0
     initialClusterState: new
     initialClusterToken: new
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -3,22 +3,22 @@ images:
   resourceId:
     name: 'etcdbrctl'
   sourceRepository: github.com/gardener/etcd-backup-restore
-  repository: eu.gcr.io/gardener-project/gardener/etcdbrctl
+  repository: europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl
   tag: "v0.24.7"
 - name: etcd
   sourceRepository: github.com/gardener/etcd-custom-image
-  repository: eu.gcr.io/gardener-project/gardener/etcd
+  repository: europe-docker.pkg.dev/gardener-project/public/gardener/etcd
   tag: "v3.4.26-3"
 - name: etcd-backup-restore-distroless
   resourceId:
     name: 'etcdbrctl'
   sourceRepository: github.com/gardener/etcd-backup-restore
-  repository: eu.gcr.io/gardener-project/gardener/etcdbrctl
+  repository: europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl
   tag: "v0.27.0"
 - name: etcd-wrapper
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -51,7 +51,7 @@ spec:
     fullSnapshotSchedule: 0 */24 * * *
     garbageCollectionPeriod: 43200s
     garbageCollectionPolicy: Exponential
-    imageRepository: eu.gcr.io/gardener-project/gardener/etcdbrctl
+    imageRepository: europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl
```
shreyas-s-rao, line 54: Have the older etcdbrctl images been copied over to artifact registry? Atleast `europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl:v0.25.0` can't be pulled. Will artifact registry be used only for newer releases which explicitly push to it?

README.md
```diff
@@ -67,7 +67,7 @@ spec:
     clientPort: 2379
     defragmentationSchedule: 0 */24 * * *
     enableTLS: false
-    imageRepository: eu.gcr.io/gardener-project/gardener/etcd-wrapper
+    imageRepository: europe-docker.pkg.dev/gardener-project/public/gardener/etcd-wrapper
```
shreyas-s-rao, line 70: Same as above

shreyas-s-rao (commented):
@ccwienk Thanks for the PR. Main question here is whether the older etcdbrctl, etcd-wrapper, and etcd-custom-image images have been copied over from GCR to artifact registry.

README.md
ccwienk: @shreyas-s-rao : yes, I copied `europe-docker.pkg.dev/gardener-project/public/gardener/etcdbrctl:v0.25.0` from GCR -> AR prior to having opened this pull-request (I just double-checked: the images does exist and can be retrieved e.g. using docker).

So for special-case of v0.25.0 there should not be an issue. For future versions, new images will be published into AR.

charts/images.yaml
```diff
@@ -3,22 +3,22 @@ images:
   resourceId:
     name: 'etcdbrctl'
   sourceRepository: github.com/gardener/etcd-backup-restore
-  repository: eu.gcr.io/gardener-project/gardener/etcdbrctl
+  repository: europe-docker.pkg.dev/gardener-project/releases/gardener/etcdbrctl
```
shreyas-s-rao, line 6: @ccwienk Any specific reason to use `releases` and not `public` registry here? Because the PR uses `public` registry in other files, such as `api/v1alpha1/types_etcd_test.go`, `test/utils/etcd.go` and `test/utils/etcdcopybackupstask.go`.

shreyas-s-rao:
@ccwienk please also rebase the PR on latest master. Just a small change in Makefile (`deploy` target)

ccwienk:
> @Christian Cwienk Thanks for the PR. Main question here is whether the older etcdbrctl, etcd-wrapper, and etcd-custom-image images have been copied over from GCR to artifact registry.

I copied currently referenced images as a one-off manual operation

charts/images.yaml
ccwienk: two reasons:
- assuming helmcharts are used to install release-versions, I preferred configuring `releases` over `public`, because the latter will also show snapshot-versions (which are also subject to automatic cleanup)
- image-references from images.yaml are used in conjunction with those configured in .ci/pipeline_definitions to add mapping-labels (done by "component-cli"). I wanted to avoid surprises and preferred to keep those image-references in sync

shreyas-s-rao: Helm charts could also be used to install snapshots, especially when testing. Wouldn't it then be better to keep the repository as `public`, so that it can use either release or snapshot images?

ccwienk: @shreyas-s-rao : okay, changed to `public`.

ccwienk:
related pull-requests:
- https://github.com/gardener/ci-infra/pull/1067
- https://github.com/gardener/ci-infra/pull/1066

shreyas-s-rao (approved):
@ccwienk thanks for making the requested changes. 
/lgtm

Issue #556, opened 2023-03-16
[Feature] Enable automatic cleanup of finished compaction jobs and make TTL configurable


unmarshall:
**Feature (What you would like to be added):**

Druid creates compaction jobs for every etcd cluster based on crossing a defined threshold of accumulated events since the last compaction run. These jobs run to completion but are not automatically removed. An operator today has to manually delete these jobs. 

Since k8s version 1.21 [spec.ttlSecondsAfterFinished](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) can be set to configure after how much time the job resource should be removed. 
Introduce a new flag that the consumer can pass onto druid to configure the value for the TTL. Assume a default value (> 0) of `ttlSecondsAfterFinished` and use it if none is explicitly passed by the consumer for all compaction jobs across all etcd clusters.

**Motivation (Why is this needed?):**
To prevent additional work for an operator to manually remove completed jobs.

renormalize:
#711 introduced functionality which deletes all compaction jobs - after they enter the `Completed` or `Failed` state.

https://github.com/gardener/etcd-druid/blob/fec5c099bed8e058a3ff34a6d07db9141ca68d93/controllers/compaction/reconciler.go#L140-L171

The controller reconciles whenever there is a job status change event, or when the snapshot lease gets updated. These events would make the controller:

- Delete the old job which enters the `Completed` state
- Delete the old job which enters the `Failed` state and then create a new job.

Even for a job which fails when `etcd-druid` is down, the next snapshot lease update would cause the controller to reconcile, and delete the old job which it now sees to be in the `Failed` state.

Halting work on this for now, until further brainstorming reveals edge cases which might not be covered in #711.

shreyas-s-rao:
We will need to relook at the compaction controller design and re-think garbage collection of completed compaction jobs.

PR #747, opened 2024-01-06
Add missing delete permission for jobs, record events for completed compaction jobs

shreyas-s-rao:

/area quality
/kind bug

Add missing delete permission for jobs in druid helm chart, required for deleting completed/failed compaction jobs and etcdcopybackupstask jobs. Also, remove unused cronjobs permissions from clusterrole, since druid no longer uses cronjobs for snapshot compaction.

Fixes #

/invite @seshachalam-yv @ishan16696 @abdasgupta 

The first iteration of this PR ([first commit](https://github.com/gardener/etcd-druid/pull/747/commits/fb47fb56481495ad60bc66c95a3791fe1717cd05)) caused a regression in e2e tests, where the test code was not able to check whether the job was successful or not, because the job resource could not be found. Upon a closer look, I realised that the compaction e2e tests used to previously succeed because of the missing delete permission for jobs, due to which druid could not delete the completed job, and hence the test could look at the job status and check whether it succeeded or failed. But once the missing permission was added, druid was able to immediately delete the completed job, hence the test code could not check the job status.

The [second commit](https://github.com/gardener/etcd-druid/pull/747/commits/d625cdb026b759cc543d19d0946aad1df8ea343b) enhances compaction controller to record events after the compaction job completes. This allows druid to simply record an event and delete the compaction job after it succeeds/fails. The e2e test now checks for this event to be present, for determining whether the compaction job succeeded or not.

**Release note**:

```other operator
Compaction controller now records events for indicating completed compaction jobs.
```

### DIFF
--- a/charts/druid/templates/druid-clusterrole.yaml
+++ b/charts/druid/templates/druid-clusterrole.yaml
@@ -138,7 +138,6 @@ rules:
 - apiGroups:
   - batch
   resources:
-  - cronjobs
   - jobs
   verbs:
   - get
@@ -147,3 +146,4 @@ rules:
   - create
   - patch
   - update
+  - delete
--- a/controllers/compaction/reconciler.go
+++ b/controllers/compaction/reconciler.go
@@ -22,6 +22,7 @@ import (
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 	ctrlutils "github.com/gardener/etcd-druid/controllers/utils"
+	"github.com/gardener/etcd-druid/pkg/common"
 	"github.com/gardener/etcd-druid/pkg/features"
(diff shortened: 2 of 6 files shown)


shreyas-s-rao:
After an out-of-band discussion with @unmarshall and @seshachalam-yv , we decided that recording events for completed compaction jobs is an overkill just to fix the behavior of the e2e test for compaction job. Instead, we realised that #556 will solve this issue with e2e tests, because adding a TTL on the job resource will ensure that k8s garbage collects the job at a later point in time, and will eliminate the need for druid to actively delete the job. This will also ensure the the test, which currently checks for job status to be succeeded (which is semantically correct) can continue to be used as is without any changes.

I will be closing this PR now, in favour of a new PR that fixes #556 instead.

/close

PR #764, opened 2024-02-08
Add a new configurable field `fullSnapshotLeaseUpdateInterval ` in spec.backup section of Etcd CR

### LINKED ISSUE #763: Add a new configurable field `fullSnapshotLeaseUpdateInterval` in spec.backup section of Etcd CR

Add `fullSnapshotLeaseUpdateInterval` field to the spec.backup section of Etcd yaml.

* The backup-restore [PR#711](https://github.com/gardener/etcd-backup-restore/pull/711) introduces a new flag `full-snapshot-lease-update-interval` to configure the retry interval for updating the full snapshot lease. Adding this new `fullSnapshotLeaseUpdateInterval` field to Etcd CR allows user to control the behaviour of retrying to update full snapshot lease 

Note: It will be an optional field, and when not set, backup-restore takes care of setting a default value to it.

Add the `fullSnapshotLeaseUpdateInterval` field to the spec.backup section of Etcd yaml, and define the `FullSnapshotLeaseUpdateInterval` field in the BackupSpec struct of Etcd CRD

anveshreddy18:

/area usability
/kind enhancement

This PR adds a new field `fullSnapshotLeaseUpdateInterval` in the `spec.backup` section of Etcd yaml and makes necessary changes, which allows to configure `full-snapshot-lease-update-interval` parameter used to configure the interval to retry updating full snapshot lease.

* The backup-restore [PR#711](https://github.com/gardener/etcd-backup-restore/pull/711) introduces a new flag `full-snapshot-lease-update-interval` to configure the retry interval for updating the full snapshot lease. Adding this new `fullSnapshotLeaseUpdateInterval` field to Etcd CR allows user to control the behaviour of retrying to update full snapshot lease 

**Note:** It will be an optional field, and when not set, backup-restore takes care of setting a default value to it. 

Fixes #763

**Release note**:

```other operator
Enabling the configurability of `full-snapshot-lease-update-interval` flag through the etcd resource spec `.spec.backup.fullSnapshotLeaseUpdateInterval`.

```

### DIFF
--- a/api/v1alpha1/etcd.go
+++ b/api/v1alpha1/etcd.go
@@ -159,6 +159,9 @@ type BackupSpec struct {
 	// All full snapshots beyond this limit will be garbage collected.
 	// +optional
 	MaxBackupsLimitBasedGC *int32 `json:"maxBackupsLimitBasedGC,omitempty"`
+	// FullSnapshotLeaseUpdateInterval defines the interval for retrying to update the full snapshot lease.
+	// +optional
+	FullSnapshotLeaseUpdateInterval *metav1.Duration `json:"fullSnapshotLeaseUpdateInterval,omitempty"`
 	// GarbageCollectionPeriod defines the period for garbage collecting old backups
 	// +optional
 	GarbageCollectionPeriod *metav1.Duration `json:"garbageCollectionPeriod,omitempty"`
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -62,6 +62,11 @@ func (in *BackupSpec) DeepCopyInto(out *BackupSpec) {
 		*out = new(int32)
 		**out = **in
 	}
+	if in.FullSnapshotLeaseUpdateInterval != nil {
+		in, out := &in.FullSnapshotLeaseUpdateInterval, &out.FullSnapshotLeaseUpdateInterval
+		*out = new(v1.Duration)
+		**out = **in
+	}
 	if in.GarbageCollectionPeriod != nil {
 		in, out := &in.GarbageCollectionPeriod, &out.GarbageCollectionPeriod
 		*out = new(v1.Duration)
--- a/charts/druid/charts/crds/templates/crd-druid.gardener.cloud_etcds.yaml
+++ b/charts/druid/charts/crds/templates/crd-druid.gardener.cloud_etcds.yaml
@@ -162,6 +162,10 @@ spec:
                     description: EtcdSnapshotTimeout defines the timeout duration
                       for etcd FullSnapshot operation
                     type: string
+                  fullSnapshotLeaseUpdateInterval:
+                    description: FullSnapshotLeaseUpdateInterval defines the interval
+                      for retrying to update the full snapshot lease.
+                    type: string
                   fullSnapshotSchedule:
                     description: FullSnapshotSchedule defines the cron standard schedule
                       for full snapshots.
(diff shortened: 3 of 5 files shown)


shreyas-s-rao (changes requested):
Thanks for the PR @anveshreddy18 ! 

Couple of comments from my side:
1. PTAL at the points I have mentioned in https://github.com/gardener/etcd-druid/issues/763#issuecomment-1934432942 
2. Please use a PR-built image of etcdbr in this PR (at `charts/images.yaml`) so that this PR can be easily tested. You can obtain the PR-built image from the concourse `publish` step from https://github.com/gardener/etcd-backup-restore/pull/711

api/v1alpha1/etcd.go
```diff
@@ -159,6 +159,9 @@ type BackupSpec struct {
 	// All full snapshots beyond this limit will be garbage collected.
 	// +optional
 	MaxBackupsLimitBasedGC *int32 `json:"maxBackupsLimitBasedGC,omitempty"`
+	// FullSnapshotLeaseUpdateInterval defines the interval for retrying to update the full snapshot lease.
+	// +optional
+	FullSnapshotLeaseUpdateInterval *metav1.Duration `json:"fullSnapshotLeaseUpdateInterval,omitempty"`
```
unmarshall, line 164: Why is a change in API required, especially when we are no longer going to use snapshot leases when we work on steward. Introducing something now in the API which is anyways going to be removed is not so nice.
Also the original issue was to retry update of full snapshot lease from backup-restore if the first attempt failed. Why does this now result in a ETCD API change?

unmarshall, line 164: Why is a change in API required, especially when we are no longer going to use snapshot leases when we work on steward. Introducing something now in the API which is anyways going to be removed is not so nice.
Also the original issue was to retry update of full snapshot lease from backup-restore if the first attempt failed. Why does this now result in a ETCD API change?

unmarshall (changes requested):

api/v1alpha1/etcd.go
anveshreddy18: I agree `etcdbr` has a default interval for doing this but we can also provide this option to configure from the `druid` to give a choice for users when this default value doesn't suit their purposes. 

Regarding API change, since we have done a lot of api changes in 777, maybe we can push everything together and later when steward comes we can remove this as the changes are very minimal. The reason I say this is because I presume we don't have a near future plan for steward as priorities have changed to improve druid & other security issues. So I'm guessing it will take some time to get steward running, till then we can provide this option. 

Also if snapshot leases will be removed in the `steward`, we can say the same for the PR #820 which decouples the ready condition for snapshot leases. But I think it's important to let the users configure & have better knowledge about the conditions even if they'll be present for a few months. WDYT?

unmarshall: Can you explain why do we need a consumer to define when and at what frequency a retry to update a snapshot lease should be done? Using snapshot leases is an implementation detail and should not be exposed anyways as part of the druid API.

anveshreddy18: Sorry, by user I meant Gardener here. The plan was to enable gardener to modify this as per the needs. Just that If they feel 3 min default is not necessary ( as it fills the logs with retry ) and want to increase it, then they have an option to. But yeah I agree it's not a necessary configuration which requires a change to API. I'll close this

unmarshall: >we can say the same for the PR https://github.com/gardener/etcd-druid/pull/820 which decouples the ready condition for snapshot leases.

These are not the same things. What #820 does is quite different. It changes what goes into the status which anyways cannot be influenced by any consumer of druid. The intent of #820 is to improve monitoring of full and delta snapshots that are taken by an etcd cluster. We do not leak any implementation detail into the API.

Issue #476, opened 2022-11-28
[Enhancement] Simplify etcd configuration and use proper URL instead of using @ as separator


unmarshall:
**Feature (What you would like to be added):**

When creating configuration to be used subsequently by etcd-backup-restore and etcd containers two properties, namely:
`initial-advertise-peer-urls` and `advertise-client-urls` use @ character to separate different parts of the URL. This is not really required. Other properties which also contain URLs use proper HTTP URL format namely: `initial-cluster`, `listen-[client/peer]-urls`. Using @ only complicates the usage with no added benefit.

**Motivation (Why is this needed?):**
Reduce the complexity where it is not needed.

**Approach/Hint to the implement solution (optional):**

Issue #728, opened 2023-11-28
☂  Druid Refactor to Address Multiple Controller Conflicts


seshachalam-yv:
**Feature (What you would like to be added):**

1. **Avoid Race Conditions:** To eliminate race conditions, the redesign proposes all etcd status updates be managed solely by the etcd controller. This includes not waiting for `StatefulSet (sts)` readiness for spec reconciliation and capturing Persistent Volume Claim (`PVC`) errors in a dedicated `etcd` condition and get rid of `custodian` controller (or repurpose it to a care controller).

2. **Codebase Refactoring:** To prevent downstream projects from vendoring all internal code from druid and encountering cyclic dependency issues, the solution involves moving packages into `internal`.

3. **Boilerplate Code Removal:** To streamline the codebase, the redesign proposes replacing the component model with the `ResourceOperator` model, which would eliminate the need for boilerplate ocomponents like `values_helper.go` files.

4. **Webhook for Safeguarding Owned Resources:** Validating webhook is proposed to safeguard against unintentional updates or deletions of owned resources. This webhook is crucial as it allows changes to these resources exclusively when etcd reconciliation or deletion activities are underway. The introduction of this webhook effectively renders the custodian (or care) controller redundant, streamlining the system’s management.

5. **Improve Observability and Debuggability** 
    - Introduce `LastOperation` and `LastErrors` as part of etcd status. Ensure that etcd reconciler captures relevant operations and errors thus improving visibility of reconciliation process.
    - Add `RunID` for every reconciler run which improves debuggability. On can now filter logs based on runID thus providing all logs for all actions/errors done/encountered in a single reconciler run.

6. **Streamline Labels on resources** Ensure that we use k8s recommended labels for all resources that are provisioned by druid.

7. **Streamline error handling** - Introduce custom error type to capture contextual information and start to use it in operators and controllers. Druid errors captured will then be use to populate the etcd resource status LastErrors as well.

8. **Fixes the following bugs:**
   1. In utils/lease.go - function [isPeerURLTLSEnabled](https://github.com/gardener/etcd-druid/blob/master/pkg/utils/lease.go#L29-L47) does not treat absence of annotation as false and currently ignores it. This results in an incorrect state of the overall status of peer URL TLS enablement.

9. **Enhancements to peer TLS handling:**
[...]

PR #771, opened 2024-02-27
Enhance Etcd configMap to provide consistency and controllability

### LINKED ISSUE #717: ☂️  Etcd ConfigMap changes
One `ConfigMap` is created per etcd (etcd-main and etcd-events) today. We propose to do/relook at the following:

- [ ] Revive https://github.com/gardener/etcd-druid/pull/467 and ensure that the name of the ConfigMap's are consistent with all other resources that are created for an etcd resource. While doing this ensure that the new version of `druid` which now uses the new name properly reconciles the older configmap's having the older name.
- [ ] Make `snapshot-count` configurable See impact [here](https://etcd.io/docs/v3.4/op-guide/maintenance/#raft-log-retention). We can retain the default of 75000. Upstream default changes from 10000 to 100,000. We can revisit the current default if need be.
- [ ] #476

anveshreddy18:

/area quality
/kind cleanup
/kind enhancement

This PR is to cleanup the Etcd ConfigMap to :

- [ ] Rename the configMap with the prefix of `etcd.Name` instead of just having `etcd`. This is in consistent with the other resources that gets created as part of Etcd Reconciliation, like leases, role, rolebinding, etc. This also involves cleanup logic for the existing configMap with older name. 
- [ ] Make `snapshot-count` configurable with the flag `snapshotCount` from `spec.etcd` section of etcd CR thus the operator is able to control the no.of transactions before committing a snapshot to disk if they want to.
- [ ] Use proper URLs for `initial-advertise-peer-urls` and `advertise-client-urls` properties of the configMap and thereby fixing #476

Fixes #717

**Release note**:

```other operator
Enhances Etcd configuration by organizing ConfigMap naming convention, enabling snapshot-count configuration, and rectifying URL issues for improved functionality and consistency
```

### DIFF
--- a/api/v1alpha1/types_etcd.go
+++ b/api/v1alpha1/types_etcd.go
@@ -446,7 +446,7 @@ func (e *Etcd) GetServiceAccountName() string {
 
 // GetConfigmapName returns the name of the configmap for the Etcd.
 func (e *Etcd) GetConfigmapName() string {
-	return fmt.Sprintf("etcd-bootstrap-%s", string(e.UID[:6]))
+	return fmt.Sprintf("%s-bootstrap-%s", e.Name, (e.UID[:6]))
 }
 
 // GetCompactionJobName returns the compaction job name for the Etcd.
--- a/api/v1alpha1/types_etcd.go
+++ b/api/v1alpha1/types_etcd.go
@@ -208,7 +208,7 @@ type EtcdConfig struct {
 	// Quota defines the etcd DB quota.
 	// +optional
 	Quota *resource.Quantity `json:"quota,omitempty"`
-	// SnapshotCount defines the number of committed transactions before a snapshot is created.
+	// SnapshotCount defines the number of committed transactions to trigger a snapshot to disk
 	// +optional
 	SnapshotCount *int `json:"snapshotCount,omitempty"`
 	// DefragmentationSchedule defines the cron standard schedule for defragmentation of etcd.
--- a/api/v1alpha1/types_etcd.go
+++ b/api/v1alpha1/types_etcd.go
@@ -208,6 +208,9 @@ type EtcdConfig struct {
 	// Quota defines the etcd DB quota.
 	// +optional
 	Quota *resource.Quantity `json:"quota,omitempty"`
+	// SnapshotCount defines the number of committed transactions before a snapshot is created.
+	// +optional
+	SnapshotCount *int `json:"snapshotCount,omitempty"`
 	// DefragmentationSchedule defines the cron standard schedule for defragmentation of etcd.
 	// +optional
 	DefragmentationSchedule *string `json:"defragmentationSchedule,omitempty"`
--- a/config/crd/bases/10-crd-druid.gardener.cloud_etcds.yaml
+++ b/config/crd/bases/10-crd-druid.gardener.cloud_etcds.yaml
@@ -563,6 +563,10 @@ spec:
                   serverPort:
                     format: int32
                     type: integer
+                  snapshotCount:
+                    description: SnapshotCount defines the number of committed transactions
+                      before a snapshot is created.
+                    type: integer
                 type: object
               labels:
                 additionalProperties:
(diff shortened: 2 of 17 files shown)


api/v1alpha1/types_etcd.go
```diff
@@ -443,7 +446,7 @@ func (e *Etcd) GetServiceAccountName() string {
 
 // GetConfigmapName returns the name of the configmap for the Etcd.
 func (e *Etcd) GetConfigmapName() string {
-	return fmt.Sprintf("etcd-bootstrap-%s", string(e.UID[:6]))
+	return fmt.Sprintf("%s-bootstrap-%s", e.Name, (e.UID[:6]))
```
unmarshall, line 449: Maybe we should just name this as `<etcd-name>-config-<etcd UID suffix>`. There is no need of including `bootstrap` as part of the name IMHO.

api/v1alpha1/types_etcd.go
```diff
@@ -208,6 +208,9 @@ type EtcdConfig struct {
 	// Quota defines the etcd DB quota.
 	// +optional
 	Quota *resource.Quantity `json:"quota,omitempty"`
+	// SnapshotCount defines the number of committed transactions to trigger a snapshot to disk
+	// +optional
+	SnapshotCount *int `json:"snapshotCount,omitempty"`
```
unmarshall, line 213: 1. once we include this as a configurable parameter, we will also have to check what happens when this parameter is updated since this is now going to be part of an etcd spec.
2. The description can simply be: `configures the number of applied Raft entries to hold in-memory before compaction` and also provide a link to https://etcd.io/docs/v3.4/op-guide/maintenance/#raft-log-retention

config/crd/bases/10-crd-druid.gardener.cloud_etcds.yaml
```diff
@@ -563,6 +563,10 @@ spec:
                   serverPort:
                     format: int32
                     type: integer
+                  snapshotCount:
+                    description: SnapshotCount defines the number of committed transactions
```
unmarshall, line 567: adjust the description.

unmarshall (commented):
A lot of these changes we will have to be re-done once the https://github.com/gardener/etcd-druid/issues/728 is opened up for review. This is a bit unfortunate. What i would suggest is either to include these changes in the refactor PR OR do it after that gets merged.

What happens for configmaps that are still having the old naming convention for all existing etcd-clusters created for shoots which are either active or hibernated?

anveshreddy18:
@unmarshall thanks for the review. 
> either to include these changes in the refactor PR OR do it after that gets merged

That's fine, if that PR decides to incorporate these, then I'll close this. Else, we can merge this after that.

> What happens for configmaps that are still having the old naming convention for all existing etcd-clusters created for shoots which are either active or hibernated?

For that, some temporary code has been added [here](https://github.com/gardener/etcd-druid/pull/771/files#diff-c9bb5a35d8e8bb96b51ab4965ee929c3b50afa3c5d6eacd29545396c67674d22R51-R55) to delete the old configmaps during the reconciliation. That can be removed once all the configmaps of every shoot namespace are deleted, probably after a few releases.

api/v1alpha1/types_etcd.go
anveshreddy18: > we will also have to check what happens when this parameter is updated since this is now going to be part of an etcd spec.

Sorry, I didn't get this. If the param is updated, then in the next reconciliation it will be reflected in the configMap. Is your question about how etcd container reacts to this change? I've tried this locally and it didn't affect the cluster in anyway. Pls correct me if this is about something else. 

> The description can simply be: configures the number of applied Raft entries to hold in-memory before compaction

Sure, this is more readable. I'll update.

api/v1alpha1/types_etcd.go
anveshreddy18: The reason I left it like that is because the closed [PR](https://github.com/gardener/etcd-druid/pull/467) tried to do the same and since this is a revival of it, I went with what they proposed in it. But yeah that is good, I'll update it.

anveshreddy18:
This PR was closed automatically by Github due to diff becoming null at some point while rebasing the PR on #777. I didn't think about reopening it and instead created a new PR after rebasing. Link to new [PR](#812) for reference.

PR #781, opened 2024-04-03
Map `stackit` infra provider to S3

maboehm:

/area control-plane
/kind enhancement

Adds a new constant for the infrastructure provider `stackit` and maps it to the StorageProvider `S3`, since they offer S3 compatible storage, see https://www.stackit.de/de/produkt/stackit-object-storage/

**Release note**:
```feature user
Added support for new backup store provider `stackit` which is an alias for `S3`.
```

### DIFF
--- a/pkg/utils/store.go
+++ b/pkg/utils/store.go
@@ -30,6 +30,7 @@ const (
 	openstack = "openstack"
 	dell      = "dell"
 	openshift = "openshift"
+	stackit   = "stackit"
 )
 
 const (
@@ -78,7 +79,7 @@ func StorageProviderFromInfraProvider(infra *druidv1alpha1.StorageProvider) (str
 	}
 
 	switch *infra {
-	case aws, S3:
+	case aws, stackit, S3:
 		return S3, nil
 	case azure, ABS:
 		return ABS, nil
--- a/pkg/utils/store.go
+++ b/pkg/utils/store.go
@@ -79,7 +79,10 @@ func StorageProviderFromInfraProvider(infra *druidv1alpha1.StorageProvider) (str
 	}
 
 	switch *infra {
-	case aws, stackit, S3:
+	case aws, S3:
+		return S3, nil
+	// S3 compatible providers
+	case stackit:
 		return S3, nil
 	case azure, ABS:
 		return ABS, nil


gardener-robot-ci-3:
Thank you @maboehm for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

maboehm:
/hold 
we are mapping to `openshift` currently, but STACKIT's Object Storage is supposed to be S3 compatible. I'll check tomorrow if that would work as well.

maboehm:
/unhold
unsure, why we had OCS before. I tested S3 and it works just as well. 

Should not be too surprising though, because OCS is essentially just an alias for S3 anyways in etcd-backup-restore (see https://github.com/gardener/etcd-backup-restore/blob/master/pkg/snapstore/ocs_s3_snapstore.go#L39 ), and the keys of the credential secret are also the same.

pkg/utils/store.go
```diff
@@ -78,7 +79,7 @@ func StorageProviderFromInfraProvider(infra *druidv1alpha1.StorageProvider) (str
 	}
 
 	switch *infra {
-	case aws, S3:
+	case aws, stackit, S3:
 		return S3, nil
 	case azure, ABS:
 		return ABS, nil
```
renormalize, line 85: Could you add another `case` to the `switch` instead of adding it to the `aws, S3` case, since it would make it cleaner for all S3 compatible providers to be in their own case sepearate from AWS S3?
Not a strong opinion whatsoever, just a style preference. Totally up to you.
```suggestion
	switch *infra {
	case aws, S3:
		return S3, nil
	// S3 compatible providers
	case stackit:
        return S3, nil
	case azure, ABS:
		return ABS, nil
```

renormalize (commented):
Thanks for your contribution @maboehm! I have a very tiny nit. Just a stylistic preference.
/lgtm

maboehm:
It's not too time sensitive -  we are a few gardener / etcd-druid releases behind anyways, and will just continue using our fork with this small change instead.

shreyas-s-rao:
@maboehm since #777 makes changes to the package structure, I have taken the liberty of pulling your change from this PR into that one, to avoid making you rebase this PR later. Please track #777 to check for your change being merged and released (as part of etcd-druid:v0.23.0).

If you're ok with this, you can go ahead and close this PR in favour of #777.

maboehm:
sounds good, thank you.

PR #785, opened 2024-04-17
Refresh indirect dependencies in `go.mod`

### LINKED ISSUE #788: Clashing VPA dependencies between etcd-druid and gardener/gardener

/area dev-productivity
/kind impediment

**What happened**:
As a developer, I work on both etcd-druid as well as gardener/gardener on my local machine. With the latest change in druid from #748 , the vendor folder is removed, and dependencies are vendored to the go mod cache. Since master branch of g/g uses dependency `k8s.io/autoscaler/vertical-pod-autoscaler v1.0.0` and etcd-druid uses indirect dependency `k8s.io/autoscaler v0.0.0-20190805135949-100e91ba756e // indirect` from an older version of g/g, there seems to be an issue with my GoLand IDE in resolving the correct vpa dependency for g/g. Instead of using `k8s.io/autoscaler/vertical-pod-autoscaler@v1.0.0` from the go mod cache, it seems to be resolving to `k8s.io/autoscaler v0.0.0-20190805135949-100e91ba756e/vertical-pod-autoscaler` by default, which causes linter errors for g/g.

I would like to bring the indirect vpa dependency in etcd-druid to the same version as the g/g dependnecy it vendors, ie, `k8s.io/autoscaler/vertical-pod-autoscaler v1.0.0`.

Checkout current [etcd-druid master](https://github.com/gardener/etcd-druid/commit/dcc9984f010f868b09f1a84c996a522b7abdc2bf) and [g/g master](https://github.com/gardener/gardener/commit/f3e32fd8514d818ec2a373f5306d1c3b4adb77ba) and run `make tidy` on both.

shreyas-s-rao:

/area quality
/kind impediment

Indirect dependency `k8s.io/autoscaler` was at an old version, and was not compatible with gardener/gardener dependencies, causing improper imports when checking out both etcd-druid and gardener repos on the same machine. This has now been fixed, by refreshing all indirect dependencies, which caused `k8s.io/autoscaler` to be updated to `k8s.io/autoscaler/vertical-pod-autoscaler v1.0.0`.

Fixes #788 

**Release note**:

```other operator
NONE
```


go.mod
```diff
@@ -132,7 +132,7 @@ require (
 	istio.io/api v1.19.2-0.20231011000955-f3015ebb5bd4 // indirect
 	istio.io/client-go v1.19.3 // indirect
 	k8s.io/apiextensions-apiserver v0.28.3 // indirect
-	k8s.io/autoscaler v0.0.0-20190805135949-100e91ba756e // indirect
+	k8s.io/autoscaler/vertical-pod-autoscaler v1.0.0 // indirect
```
renormalize, line 135: https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-1.0.0 was tagged as Pre-release.
Would it be possible to directly bump up to v1.1.0 without causing issues with the Gardener dependency?

renormalize (commented):
Just wondering if this might be better. Thanks.

go.mod
shreyas-s-rao: @renormalize If you observe, [all releases](https://github.com/kubernetes/autoscaler/releases?q=vertical-pod-autoscaler&expanded=true) of `k8s.io/autoscaler/vertical-pod-autoscaler` other than the latest version are marked as pre-release. I'm guessing this might have something to do with the fact that one single repo https://github.com/kubernetes/autoscaler has multiple components which have their own releases and tags, so the tagging of these releases may not be accurate, and more of a workaround, hence all previous releases are marked as pre-release, which doesn't really make sense if you see it practically. Also, if you observe, the other components like `Cluster Autoscaler` and `Addon Resizer` don't have any release tags like `latest` or `pre-release` at all.

Also, since this is an indirect dependency for etcd-druid, we need not worry about its exact version, as long as it corresponds to the direct dependencies that are vendored, like gardener/gardener v1.86.0, which is `k8s.io/autoscaler/vertical-pod-autoscaler:v1.0.0`, as you can see [here](https://github.com/gardener/gardener/blob/cee120166b7b8a857dea90fc0217d86c2e41850f/go.mod#L54).

renormalize: Thanks for the explanation.

unmarshall (approved):
/lgtm

shreyas-s-rao:
/close
Superseded by #789

PR #748, opened 2024-01-07
Update gardener/gardener to v1.86.0

shreyas-s-rao:
**How to categorize this PR?**

/area dev-productivity
/kind cleanup

**What this PR does / why we need it**:
This PR updates the gardener/gardener dependency to v1.86.0, bringing in multiple other changes:
- [x] Remove vendor directory, in line with gardener, similar to https://github.com/gardener/gardener/pull/8775
- [x] Bump dependencies, required for g/g v1.86.0
- [x] Update paths in Makefile, CI scripts and hack scripts to remove usage of vendor dir for g/g hack scripts, and find the g/g dir in mod cache instead
- [x] Remove `-mod=vendor` from builds, runs and test runs
- [x] Run `make tidy`, which pulls in new CI hack scripts and issue/PR templates
- [x] Adapt and run `make generate`
- [x] Remove unused chart renderer code from g/g, which was deprecated and now removed in the latest g/g version
- [x] Adapt manager and controller registration code to new controller-runtime version, both in main code and integration test utils
- [x] Replace `github.com/golang/mock/gomock` with `go.uber.org/mock/gomock`, as stated in https://github.com/golang/mock 
- [x] Adapt etcdcopybackupstask deletion unit tests to new controller-runtime fake client behavior. There is a change in behavior of [fake client handling of deletion timestamps](https://github.com/kubernetes-sigs/controller-runtime/pull/2316), as well as [handling updates to status subresource of CRDs](https://github.com/kubernetes-sigs/controller-runtime/pull/2259), due to which the tests needed to be adapted 
- [x] Downgrade kustomize from v5.0.0 to v4.5.7 due to [this bug](https://github.com/kubernetes-sigs/kustomize/issues/4731), which we and many other users used as a feature. #765 details out the issue and a recommendation to properly fix this in the long-term, but this PR should go ahead with the checked in fix.
- [x] Update github issue templates with new ones from g/g.

**Which issue(s) this PR fixes**:
Fixes #734 

**Special notes for your reviewer**:
/cc @afritzler @shafeeqes 
/invite @seshachalam-yv @ishan16696 @unmarshall 

[...]

PR #786, opened 2024-04-19
Add the `vendor` directory to `.gitignore`

renormalize:
/area dev-productivity
/kind enhancement

#748 removed the `vendor` directory from `etcd-druid` and makes the project rely on the module cache set up by the `go` tool for finding dependencies.

However, there are use cases where for better developer experience, it would make sense to have the dependencies reside in the same project folder as `etcd-druid`, so that tokens can be searched for, using project wide search in any text editor. Results can be fetched from the dependencies, without having to rely on the LSP. For example, this would make looking at all usages of a constant that is defined by a dependency easier.

The developer can simply run `go mod vendor` without it affecting the modified build process which doesn't use `vendor`, as defined in #748.

However, to not break the current build process which doesn't use the `vendor` directory anymore, rename the `vendor` directory to anything else while running `make run` or `make docker-build` (`mv vendor .vendor` for example).

/hold till #777 merges.

```other developer
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@
 /dev
 /vendor
 *.test
+default.etcd*
+default.bkp*
 
 *.coverprofile
 /cover.out
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /.kube-secrets
 /tmp/*
 /dev
+/vendor
 *.test
 
 *.coverprofile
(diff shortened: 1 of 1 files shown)


renormalize:
Can be merged once #777 merges.

shreyas-s-rao (approved):
/lgtm

PR #805, opened 2024-06-05
Use localstack-s3 image instead of localstack

anveshreddy18:

/area dev-productivity
/area storage
/kind test

This PR modifies the localstack image used for performing tests in druid. The current `localstack` image contains a lot of other services provided by `AWS` apart from `S3` such as Lambda, DynamoDB, etc and has a compressed size of ~450 MB

The change proposes to use a much light weight version of this which provides only S3 services and has a compressed size of ~100 MB, which is a drastic reduction in size. This helps in faster setup for tests both locally and on pipelines as well. 

Fixes #

**Release note**:

```other developer
use `localstack:s3-latest` image instead of `localstack:latest` in tests for faster setup
```

### DIFF
--- a/hack/e2e-test/infrastructure/localstack/localstack.yaml
+++ b/hack/e2e-test/infrastructure/localstack/localstack.yaml
@@ -24,7 +24,8 @@ spec:
               value: "1"
             - name: SERVICES
               value: s3
-          image: localstack/localstack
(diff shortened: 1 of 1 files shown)


renormalize (approved):
Thanks for this find!

PR #812, opened 2024-06-24
Enhance Etcd configMap to provide consistency and controllability

### LINKED ISSUE #717: ☂️  Etcd ConfigMap changes

anveshreddy18:

/area quality
/kind cleanup
/kind enhancement

This PR is to cleanup the Etcd ConfigMap to :

- [x] Rename the configMap in the format `{etcd.Name}-config` instead of `etcd-bootstrap-<UID>`. This is for consistency with other resources that gets created as part of Etcd Reconciliation, like leases, role, rolebinding, etc and also have a proper name `config` for the configMap instead of `bootstrap`. The cleanup logic for the existing configMap with older name will be taken up at a later stage after 2 releases. This [issue](https://github.com/gardener/etcd-druid/issues/931) tracks that.
- [x] Make `snapshot-count` configurable with the flag `snapshotCount` from `spec.etcd` section of etcd CR thus the operator is able to control the no.of transactions before committing a snapshot to disk if they want to.
- [x] Use proper URLs for `initial-advertise-peer-urls` and `advertise-client-urls` properties of the configMap and thereby fixing #476

Previously the fields `initial-advertise-peer-urls` and `advertise-client-urls` of the etcd ConfigMap were constructed using an `@` separator in the format `scheme@peerSvcName@namespace@port` which is then parsed by the `backup-restore` container to construct the full URL to be used by the `etcd` container as its configuration. 

This PR aims to modify that to have the fully qualified URL(s) for each `etcd` member in the ConfigMap itself which then `backup-restore` parses and sends only it's corresponding etcd member's URL(s) to the etcd process. 

The following is the format in the ConfigMap for the fields `initial-advertise-peer-urls` & `advertise-client-urls`. 

```
initial-advertise-peer-urls:
  etcd-main-0:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
  etcd-main-1:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
  etcd-main-2:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
advertise-client-urls:
  etcd-main-0:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
  etcd-main-1:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
  etcd-main-2:
    - main in-cluster url
    - additional urls if required (e.g. live CPM)
```

The `additional urls` above is used for e.g. in CPM in the context of Gardener use case. For now from code we only populate the main in-cluster URL, but in future we will modify the API to allow passing these additional urls from the Etcd Spec which will get populated here and be subsequently parsed and sent to etcd process. 

Fixes #717

**Release note**:

```other operator
Enhances Etcd configuration by organizing ConfigMap naming convention, enabling snapshot-count configuration, and rectifying URL issues for improved functionality and consistency
```
```noteworthy operator
Etcd ConfigMap Naming Update: The naming convention has changed to {etcd.Name}-config for consistency, replacing etcd-bootstrap-<UID>. Unused old ConfigMaps will be removed in v0.27.0
```


renormalize:
@anveshreddy18 can you please add https://github.com/gardener/etcd-backup-restore/pull/715's merge and release to the PR description as a prerequisite? Just so we don't accidentally merge this PR once it's approved.

anveshreddy18:
> @anveshreddy18 can you please add https://github.com/gardener/etcd-backup-restore/pull/715 merge and release as a prerequisite to this PR's description? Just so we don't accidentally merge this PR before the changes are etcd-backup-restore are released.

Sure, I hope your comment itself is sufficient. But I can add that once this PR is reviewed as we may lose any comment that we put right now because of the reviews. And for now the `do-not-merge` label is present, I will make sure to keep that label till `etcdbr` PR is merged and released.

ishan16696 (changes requested):

anveshreddy18:
@ishan16696, 
- I've now rebased on master
- The doc has been updated already previously.

PTAL. Thanks

ishan16696:
> The doc has been updated already previously.

I'm not sure what you are referring to, doc of recovery from permanent quorum loss is not updated with new name of config map as I'm still able to find the `etcd-bootstrap` naming convention written  in steps of doc.

anveshreddy18:
> above command was written for etcd-bootstrap, will it work with new convention as well

It works for the new config map name as well as we fetch the name from the `.spec.volumes`. I've also tested this and it works i.e gives the name of the new config map currently in use.

internal/component/configmap/etcdconfig.go
```diff
@@ -22,27 +22,22 @@ const (
 	defaultInitialClusterToken     = "etcd-cluster"
 	defaultInitialClusterState     = "new"
 	// For more information refer to https://etcd.io/docs/v3.4/op-guide/maintenance/#raft-log-retention
-	// TODO: Ideally this should be made configurable via Etcd resource as this has a direct impact on the memory requirements for etcd container.
-	// which in turn is influenced by the size of objects that are getting stored in etcd.
-	defaultSnapshotCount = 75000
+	defaultSnapshotCount = int64(75000)
+	commTypePeer         = "peer" // communication type
+	commTypeClient       = "client"
```
seshachalam-yv, line 27:
```suggestion
	advURLTypePeer   = "peer"
	advURLTypeClient = "client"
```

test/e2e/etcd_backup_test.go
```diff
@@ -247,7 +247,7 @@ func checkEtcdReady(ctx context.Context, cl client.Client, logger logr.Logger, e
 
 	logger.Info("Checking configmap")
 	cm := &corev1.ConfigMap{}
-	ExpectWithOffset(2, cl.Get(ctx, client.ObjectKey{Name: "etcd-bootstrap-" + string(etcd.UID[:6]), Namespace: etcd.Namespace}, cm)).To(Succeed())
+	ExpectWithOffset(2, cl.Get(ctx, client.ObjectKey{Name: etcd.Name + "-config-" + string(etcd.UID[:6]), Namespace: etcd.Namespace}, cm)).To(Succeed())
```
seshachalam-yv, line 250:
```suggestion
	ExpectWithOffset(2, cl.Get(ctx, client.ObjectKey{Name: "etcd-config-" + string(etcd.UID[:6]), Namespace: etcd.Namespace}, cm)).To(Succeed())
```

test/e2e/etcd_backup_test.go
```diff
@@ -280,7 +280,7 @@ func deleteAndCheckEtcd(ctx context.Context, cl client.Client, logger logr.Logge
 	ExpectWithOffset(1,
 		cl.Get(
 			ctx,
-			client.ObjectKey{Name: "etcd-bootstrap-" + string(etcd.UID[:6]), Namespace: etcd.Namespace},
+			client.ObjectKey{Name: etcd.Name + "-config-" + string(etcd.UID[:6]), Namespace: etcd.Namespace},
```
seshachalam-yv, line 283:
```suggestion
			client.ObjectKey{Name: "etcd-config-" + string(etcd.UID[:6]), Namespace: etcd.Namespace},
```

internal/component/configmap/configmap_test.go
```diff
@@ -377,8 +370,20 @@ func matchClientTLSRelatedConfiguration(g *WithT, etcd *druidv1alpha1.Etcd, actu
 	}
 }
 
+func getAdvertiseURLsInterface(etcd *druidv1alpha1.Etcd, commType, scheme string) map[string]interface{} {
+	advertiseUrlsMap := getAdvertiseURLs(etcd, commType, scheme, druidv1alpha1.GetPeerServiceName(etcd.ObjectMeta))
```
seshachalam-yv, line 374: `getAdvertiseURLsInterface` should not rely on `getAdvertiseURLs` directly, as this does not help in testing `getAdvertiseURLs` independently.

seshachalam-yv (changes requested):

api/v1alpha1/etcd.go
```diff
@@ -194,6 +194,10 @@ type EtcdConfig struct {
 	// Quota defines the etcd DB quota.
 	// +optional
 	Quota *resource.Quantity `json:"quota,omitempty"`
+	// SnapshotCount defines the number of applied Raft entries to hold in-memory before compaction.
```
unmarshall, line 197: have you run into a case where you needed this as configurable? Can you please elaborate that case?

anveshreddy18: I've never run into that personally but this PR solves the issue [717](https://github.com/gardener/etcd-druid/issues/717) where the second point is to make `snapshot-count` configurable, so went ahead with it as that made sense. I can remove this if it's not good to have.

internal/component/configmap/configmap_test.go
```diff
@@ -377,8 +370,20 @@ func matchClientTLSRelatedConfiguration(g *WithT, etcd *druidv1alpha1.Etcd, actu
 	}
 }
 
+func getAdvertiseURLsInterface(etcd *druidv1alpha1.Etcd, commType, scheme string) map[string]interface{} {
```
unmarshall, line 373: A few comments:

1. What is the need to create a type alias `type advertiseURLs map[string][]string`? There is no method defined on this type. If its only for readability then that is a not a strong argument to have it.
2. Can you please rename this function to `convertAdvertiseURLsValuesToInterface`

internal/component/configmap/etcdconfig.go
unmarshall: can we just use clear names and just call it `advertiseURLTypePeer` and `advertiseURLTypeClient` instead?

test/e2e/etcd_backup_test.go
unmarshall: Actually what he has done is right. Please check the naming convention in `v1alpha1.GetConfigMapName`

api/v1alpha1/helper.go
```diff
@@ -31,7 +31,7 @@ func GetServiceAccountName(etcdObjMeta metav1.ObjectMeta) string {
 
 // GetConfigMapName returns the name of the configmap for the Etcd.
 func GetConfigMapName(etcdObjMeta metav1.ObjectMeta) string {
-	return fmt.Sprintf("etcd-bootstrap-%s", string(etcdObjMeta.UID[:6]))
+	return fmt.Sprintf("%s-config-%s", etcdObjMeta.Name, (etcdObjMeta.UID[:6]))
```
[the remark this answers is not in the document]
anveshreddy18: I think historically, the UID was used to identify if the configMap is in use by the running `Etcd`. Having UID will allow us to make cleaning operations of any orphan configMap's present in the cluster easily, as it's easy to find whether that configMap is currently being used or not. Although this isn't a strict requirement i.e we can still find without this but I feel having it would be better.

Yeah, I'm not sure why we haven't followed this UID suffix for other resources, but I'll just let them be for now.

> If you really want to keep the suffix then there is no need to enclose it in parenthesis.

Sure, somehow missed it.

internal/component/configmap/configmap_test.go
anveshreddy18: 1. Yes it was only for readability. Will remove it.
2. Sure, this is more readable.
Thanks

anveshreddy18: Done

api/v1alpha1/etcd.go
unmarshall: ok then lets keep it.

api/v1alpha1/helper.go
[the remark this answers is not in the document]
unmarshall: >the UID was used to identify if the configMap is in use by the running Etcd

I did not understand this reasoning. You can easily look at the owner reference, right? I would like that we have on convention and since we are changing the name, lets avoid suffixing a partial UID to it.

api/v1alpha1/helper.go
```diff
@@ -31,7 +31,7 @@ func GetServiceAccountName(etcdObjMeta metav1.ObjectMeta) string {
 
 // GetConfigMapName returns the name of the configmap for the Etcd.
 func GetConfigMapName(etcdObjMeta metav1.ObjectMeta) string {
-	return fmt.Sprintf("etcd-bootstrap-%s", string(etcdObjMeta.UID[:6]))
+	return fmt.Sprintf("%s-config-%s", etcdObjMeta.Name, etcdObjMeta.UID[:6])
```
unmarshall, line 34: As stated in another comment lets remove the suffix of UID. One can look at the owner reference and also check the name of the configmap which also has etcd-name in it.

unmarshall (changes requested):

api/v1alpha1/helper.go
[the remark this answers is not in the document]
anveshreddy18: > You can easily look at the owner reference, right?

yes yes, as I mentioned, `this isn't a strict requirement i.e we can still find without this but I feel having it would be better`, but yeah to maintain consistency, I can remove the suffix.

internal/component/configmap/configmap_test.go
```diff
@@ -377,8 +370,38 @@ func matchClientTLSRelatedConfiguration(g *WithT, etcd *druidv1alpha1.Etcd, actu
 	}
 }
 
+func assertAdvertiseURLs(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme string) map[string][]string {
```
shreyas-s-rao, line 373:
```suggestion
func expectedAdvertiseURLs(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme string) map[string][]string {
```

internal/component/configmap/configmap_test.go
```diff
@@ -377,8 +370,38 @@ func matchClientTLSRelatedConfiguration(g *WithT, etcd *druidv1alpha1.Etcd, actu
...
+	}
+	advUrlsMap := make(map[string][]string)
+	for i := 0; i < int(etcd.Spec.Replicas); i++ {
+		podName := druidv1alpha1.GetOrdinalPodName(etcd.ObjectMeta, i)
+		advUrlsMap[podName] = []string{fmt.Sprintf("%s://%s.%s.%s.svc:%d", scheme, podName, druidv1alpha1.GetPeerServiceName(etcd.ObjectMeta), etcd.Namespace, port)}
+	}
+	return advUrlsMap
+}
+
+func convertAdvertiseURLsValuesToInterface(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme string) map[string]interface{} {
```
shreyas-s-rao, line 391:
```suggestion
func expectedAdvertiseURLsAsInterface(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme string) map[string]interface{} {
```

shreyas-s-rao (changes requested):
@anveshreddy18 thank you very much for this PR. Just couple of naming nits from my side from code review. PTAL.

Additionally, the configmap reconciliation logic now seems to create a new configmap with name `<etcd-name>-config`, but does not clean up the old configmap named `etcd-<uid>-bootstrap`. I remember a discussion around this, where it becomes tricky to replace the configmap without causing issues in a running etcd cluster. Is this still the case? Or can the old configmap be safely removed? I would prefer that it be removed, unless there's a strong reason not to.

If at all the old configmap has to be retained, please add that to the release note, that the cleanup will happen in the next 2/3 releases, and create an issue for it as well, with the correct milestone.

shreyas-s-rao:
Also, check-generate seems to be failing. Can you please run `make generate` as well?

Can you also check why the e2e tests are failing? Looks like the infra job fails, but I'm unsure why, since this PR doesn't seem to change anything in the e2e test setup.

anveshreddy18:
> Also, check-generate seems to be failing. Can you please run make generate as well?

Yes. My bad.

> Can you also check why the e2e tests are failing? Looks like the infra job fails, but I'm unsure why, since this PR doesn't seem to change anything in the e2e test setup.

There seems to be some problem with ubuntu pkg archives. The same problem is occurring in the `master` branch  as well where the `aws-infra-job` fails continuously with errors: 

```
Installing awscli...

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Ign:1 http://ports.ubuntu.com/ubuntu-ports mantic InRelease
Ign:2 http://ports.ubuntu.com/ubuntu-ports mantic-updates InRelease
Ign:3 http://ports.ubuntu.com/ubuntu-ports mantic-backports InRelease
Ign:4 http://ports.ubuntu.com/ubuntu-ports mantic-security InRelease
Err:5 http://ports.ubuntu.com/ubuntu-ports mantic Release
  404  Not Found [IP: 185.125.190.39 80]
Err:6 http://ports.ubuntu.com/ubuntu-ports mantic-updates Release
  404  Not Found [IP: 185.125.190.39 80]
Err:7 http://ports.ubuntu.com/ubuntu-ports mantic-backports Release
  404  Not Found [IP: 185.125.190.39 80]
Err:8 http://ports.ubuntu.com/ubuntu-ports mantic-security Release
  404  Not Found [IP: 185.125.190.39 80]
Reading package lists...
E: The repository 'http://ports.ubuntu.com/ubuntu-ports mantic Release' does not have a Release file.
E: The repository 'http://ports.ubuntu.com/ubuntu-ports mantic-updates Release' does not have a Release file.
E: The repository 'http://ports.ubuntu.com/ubuntu-ports mantic-backports Release' does not have a Release file.
E: The repository 'http://ports.ubuntu.com/ubuntu-ports mantic-security Release' does not have a Release file.
Stream closed EOF for e2e-test/aws-infra-job-5f4sq (infra)
```

Not sure what's happening, will try to debug it.

shreyas-s-rao:
e2e test `pull-etcd-druid-e2e-kind-nondistroless-etcd` is expected to fail, since the configmap URL parsing has not been updated for etcd-backup-restore:v0.24.x release branch. This is expected, since support for non-distroless images will be removed by #936 in this release of etcd-druid.

shreyas-s-rao (approved):
@anveshreddy18 thanks for making the requested changes.
/lgtm

seshachalam-yv (approved):

PR #817, opened 2024-06-25
Don't add `StorageClassName` in pvcClaim if StorageClass is empty string

acumino:

/area control-plane
/kind bug

If `StorageClass` is set to an empty string in the `ETCD` object don't set `StorageClassName` in `VolumeClaimTemplates` otherwise PVCs `StorageClassName` will get set to an `empty string` instead of `default` and it will forever be stuck.

Introduced in https://github.com/gardener/etcd-druid/pull/777

Fixes #

**Release note**:

```other operator
NONE
```


shreyas-s-rao (approved):
/lgtm

PR #820, opened 2024-06-26
Decouple backup ready condition for snapshot leases and remove support for deprecated fields in etcd status

### LINKED ISSUE #645: [Enhancement] Streamline etcd resource status
Streamline etcd resource status to make it more meaningful and accurate for human operators.

Currently, etcd status has the following fields:
- `ObservedGeneration`
- `Etcd`
- `Conditions`
- `ServiceName`
- `LastError`
- `ClusterSize`
- `CurrentReplicas`
- `Replicas`
- `ReadyReplicas`
- `Ready`
- `UpdatedReplicas`
- `LabelSelector`
- `Members`
- `PeerUrlTLSEnabled`

As part of https://github.com/gardener/etcd-druid/pull/594, many of these fields have already been marked deprecated, such as `Status.ClusterSize`, `Status.ServiceName` and `Status.UpdatedReplicas`, and will be removed in the future. But there are still many fields which are not really required to deduce the state of the etcd resource, such as `Status.Etcd` (which provides a self-reference to the Etcd object and isn't required since anybody who has access to the etcd status already has a reference to the etcd object itself), `Status.Ready` (readiness is a state, and can be better represented by a `condition` denoting quorum in the cluster, which means that the cluster is "ready" to serve traffic), `Status.Members` (which will probably be replaced by something like `Status.MemberRefs` via https://github.com/gardener/etcd-druid/issues/206), and fields `Status.Replicas`, `Status.ReadyReplicas` and `Status.CurrentReplicas` which are blindly copied over from the statefulset status and provide no meaningful information about the cluster, but rather druid should to the work of deducing such member-specific info and correctly populating conditions.

The `Status.Conditions` types that set today are:

- `Ready`
  
  Denotes readiness of the cluster, but for a human operator this condition would make more sense when renamed to `QuorumReached`, to denote that the cluster has reached quorum, which means it's ready to serve traffic.

- `AllMembersReady`

  Currently, this does not provide any information that `QuorumReached` cannot already provide. We would still require this condition to deduce whether every member is alive, ie, both containers in the pod are up. In such a case, a better name for this condition would be `AllMembersAlive` since readiness of each member already denotes readiness of the cluster, whereas liveness of a member has no bearing on whether the cluster is ready or not. In order to enable this, we will need to enhance the lease renewal logic in the backup sidecar to also take into consideration the liveness of its etcd member, via a serializable GET call to the etcd (which can be achieved via a liveness endpoint on the etcd-wrapper container).

- `BackupReady`

  This denotes whether the backups are uploaded on schedule as expected and are not stale. It considers both full and delta backups, and is computed using a complex logic that considers various different combinations of the states of the full snapshot lease and delta snapshot lease, which becomes quite complex to read and maintain. It would make better sense to split this condition into two simpler conditions for full and delta snapshot staleness respectively, so that druid can individually compute the condition state for both these and populate them individually, providing a cleaner view to human operators.

Proposed status fields to be removed/deprecated, or already deprecated:
- `Etcd`
- `ServiceName`
- `ClusterSize`
- `Replicas`
- `ReadyReplicas`
- `CurrentReplicas`
- `UpdatedReplicas`
- `Ready`
- `LabelSelector`

Proposed conditions:
```
- type: QuorumReached
  status: True|False|Unknown
  reason: MajorityMembersReady | MajorityMembersUnready | QuorumNotChecked
  message: "x/y members ready: <member-id>,<member-id>; z/y members unready: <member-id>" | "Quorum not checked"
- type: AllMembersAlive
  status: True|False|Unknown
  reason: MemberLeasesRenewed | MemberLeasesStale | MemberLeasesNotChecked
  message: "x/y members alive: <member-id>,<member-id>; z/y members not alive: <member-id>" | "Member leases not checked"
- type: FullSnapshotStale/FullSnapshotOutdated
  status: True|False|Unknown
  reason: SnapshotUploadedOnSchedule | SnapshotMissedSchedule | CannotDetermineSnapshotUploadStatus
  message: "Full snapshot uploaded successfully in the last x (time)" | "Cannot determine snapshot upload status"
- type: DeltaSnapshotStale/DeltaSnapshotOutdated
  status: True|False|Unknown
  reason: SnapshotUploadedOnSchedule | SnapshotMissedSchedule | CannotDetermineSnapshotUploadStatus
  message: "Delta snapshot uploaded successfully in the last x (time)" | "Cannot determine snapshot upload status"
```

/area usability

anveshreddy18:

/area usability monitoring
/kind enhancement cleanup

This PR splits the existing `BackupReady` condition which talks about both full and delta snapshot health as a single condition into two new conditions `FullSnapshotBackupReady` and `DeltaSnapshotBackupReady`, one for each type of snapshot i.e `Full` and `Delta` respectively. This helps in better understanding of status of each backup type thus helping debug issues and analyze more effectively.  The existing `BackupReady` condition will still be present, which aggregates both the individual snapshot conditions into one.

This PR also removes the support for few fields in etcd status which are in deprecated state for enough time. The fields that are removed are `ServiceName`, `ClusterSize`, `UpdatedReplicas` and `LabelSelector`.  These fields were essentially taken from `statefulset` and not necessarily needed to be present under `etcd.status` hence they were deprecated and this PR stops the support. 

This PR partially fixes this [issue for streamlining etcd status](https://github.com/gardener/etcd-druid/issues/645)

**Release note**:

```noteworthy operator
The existing `BackupReady` condition in `etcd.status` is split into 2 new specific conditions `FullSnapshotBackupReady` & `DeltaSnapshotBackupReady` to better understand the health of each type of snapshot backup
```
```noteworthy operator
Support for fields `ServiceName`, `ClusterSize`,  `UpdatedReplicas` and `LabelSelector`  in `etcd.status`  are removed after they've been deprecated for 4 minor releases 
```

### DIFF
--- a/internal/health/condition/check_backup_ready.go
+++ b/internal/health/condition/check_backup_ready.go
@@ -10,13 +10,22 @@ import (
 	"time"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
-
+	"github.com/gardener/etcd-druid/internal/utils"
 	coordinationv1 "k8s.io/api/coordination/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type backupReadyCheck struct {
+	cl      client.Client
+	results []Result
+}
+
+type fullSnapshotBackupReadyCheck struct {
+	cl client.Client
+}
+
+type deltaSnapshotBackupReadyCheck struct {
 	cl client.Client
 }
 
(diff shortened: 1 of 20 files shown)


seshachalam-yv (changes requested):
Please remove `Cluster Size` and `Backup Ready`.
https://github.com/gardener/etcd-druid/blob/d0f316396b96ed5182e00256fd8de5772a86c431/api/v1alpha1/etcd.go#L53

https://github.com/gardener/etcd-druid/blob/d0f316396b96ed5182e00256fd8de5772a86c431/api/v1alpha1/etcd.go#L51

Add  `FullSnapshotBackupReady` and `DeltaSnapshotBackupReady` as part of print columns

anveshreddy18:
@seshachalam-yv  Nice find, thanks. Done

go.sum
```diff
@@ -438,6 +438,8 @@ github.com/prometheus/procfs v0.10.1 h1:kYK1Va/YMlutzCGazswoHKo//tZVlFpKYh+Pymzi
 github.com/prometheus/procfs v0.10.1/go.mod h1:nwNm2aOCAYw8uTR/9bWRREkZFxAUcWzPHWJq+XBB/FM=
 github.com/rakyll/gotest v0.0.6 h1:hBTqkO3jiuwYW/M9gL4bu0oTYcm8J6knQAAPUsJsz1I=
 github.com/rakyll/gotest v0.0.6/go.mod h1:SkoesdNCWmiD4R2dljIUcfSnNdVZ12y8qK4ojDkc2Sc=
+github.com/robfig/cron v1.2.0 h1:ZjScXvvxeQ63Dbyxy76Fj3AT3Ut0aKsyd2/tl3DTMuQ=
```
seshachalam-yv, line 441: Why are we adding `github.com/robfig/cron` when `github.com/robfig/cron/v3` is already available?

seshachalam-yv (changes requested):

test/e2e/etcd_backup_test.go
```diff
@@ -243,8 +243,8 @@ func checkEtcdReady(ctx context.Context, cl client.Client, logger logr.Logger, e
 		}
 
 		for _, c := range etcd.Status.Conditions {
-			// skip BackupReady status check if etcd.Spec.Backup.Store is not configured.
-			if etcd.Spec.Backup.Store == nil && c.Type == v1alpha1.ConditionTypeBackupReady {
+			// skip BackupReady, FullSnapshotBackupReady & DeltaSnapshotBackupReady status checks if etcd.Spec.Backup.Store is not configured.
```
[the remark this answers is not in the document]
anveshreddy18: Yes this looks neat. I've moved it to separate function

go.sum
anveshreddy18: My bad. Didn't check that. used only `github.com/robfig/cron/v3` now

internal/health/condition/check_backup_ready.go
```diff
@@ -25,93 +34,198 @@ const (
...
+		} else {
+			result.status = druidv1alpha1.ConditionFalse
+			result.reason = SnapshotMissedSchedule
+			result.message = fmt.Sprintf("Full snapshot missed schedule, last full snapshot was taken %v ago", time.Since(fullLeaseRenewTime.Time))
+			return result
+		}
+	}
+}
+
+func (d *deltaSnapshotBackupReadyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
```
unmarshall, line 156: there is a LOT of code repetition and this code needs to be rewritten.

internal/health/status/check.go
```diff
@@ -36,7 +36,8 @@ var (
 	ConditionChecks = []ConditionCheckFn{
 		condition.AllMembersReadyCheck,
 		condition.ReadyCheck,
-		condition.BackupReadyCheck,
+		condition.FullSnapshotBackupReadyCheck,
```
unmarshall, line 39: you have removed `BackupReadyCheck` here but you have also changed and kept the implementation of `BackupReady` in `check_backup_ready.go`.  Any reason for that?

internal/health/status/check.go
```diff
@@ -90,10 +91,12 @@ func (c *Checker) executeConditionChecks(ctx context.Context, etcd *druidv1alpha
 		wg.Wait()
 	})()
 
-	results := make([]condition.Result, 0, len(ConditionChecks))
+	results := make([]condition.Result, 0, len(ConditionChecks)+1)
```
unmarshall, line 94: why is this conditions checks + 1?

internal/health/status/check.go
```diff
@@ -90,10 +91,12 @@ func (c *Checker) executeConditionChecks(ctx context.Context, etcd *druidv1alpha
...
 	for r := range resultCh {
 		results = append(results, r)
 	}
+	backupReadyChecker := condition.BackupReadyCheck(c.cl, results)
```
unmarshall, line 98: i really do not understand this. You removed it from `ConditionChecks` but introduce this here? why?

unmarshall (changes requested):
I think we need to pair program this. Current code needs a LOT of changes.

PR #823, opened 2024-07-02
Fix statefulset `PreSync` to work for certain cases of unhealthy etcd clusters upon druid upgrade

### LINKED ISSUE #818: Deleting the `etcd-bootstrap` configmap leads to etcd reconciliation to never succeed

/area control-plane
/kind bug

**What happened**:
If the stateful set is in errored state because of some wrong data in the etcd bootstrap configMap then fixing the etcd also doesn't lead to etcd reconciliation getting succeeded.
This is so because we expect `sts` to be ready in the preSync [here](https://github.com/gardener/etcd-druid/blob/7b347f70333ae171846226627910c2682153e6cf/internal/component/statefulset/statefulset.go#L285-L306) which will not be ready because etcd bootstrap configMap will not get updated because configMap is updated in the `syncEtcdResources` which is after `preSyncEtcdResources`.
This lead to reconciliation getting stuck forever [here](https://github.com/gardener/etcd-druid/blob/7b347f70333ae171846226627910c2682153e6cf/internal/component/statefulset/statefulset.go#L108-L115).

Introduced in https://github.com/gardener/etcd-druid/pull/777

/cc @shafeeqes 

**Environment**:

- etcd-druid version:
- Others:

shreyas-s-rao:

/area quality
/kind bug

This PR fixes statefulset `PreSync` behavior, to no longer wait for etcd pods to be updated with the latest sts spec, or to be ready. Instead, `PreSync` will now simply check whether the pods have the expected labels, and continue with the next steps. This fixes certain cases where an etcd cluster is unhealthy due to a wrong reconfiguration of the etcd configmap, causing one of the three etcd pods to fail, which causes the `PreSync` step to get stuck in the pod updated+ready check, and not allow druid to run `Sync` which reconciles and potentially fixes the issue by reverting the configmap to the correct state.

Fixes #818 

While working on this fix, @unmarshall and I found certain cases where PreSync cannot succeed. Consider the following case:
1. Druid is running with old version (v0.22.0)
2. 3-node etcd cluster is deployed and ready
3. Etcd spec is updated with non-existent secret reference (TLS or backup store)
4. Etcd is reconciled, causing the rollout of the statefulset pods to start, updating pod etcd-2 to go into error state since the mentioned secret does not exist.
    1. This unhealthy pod can be caused by any other reason as well, and not necessarily a misconfiguration. We simply use this as an example to show this case.
5. Upgrade druid to v0.23.0 (or current master version)
6. Reconcile the Etcd. This causes the PreSync to get stuck since the statefulset rollout is stuck since the pod never becomes ready.
7. Druid marks the Etcd status.lastOperation as `Operation: Reconcile, State: Failed`.
8. Now, the user sees this and figures out the Etcd spec misconfiguration issue, corrects the Etcd spec with correct configuration and reconciles the Etcd again. Druid will now succeed and the etcd cluster becomes healthy and also updated with new labels.
9. Thus, misconfiguration or transient errors can be overcome by the current druid logic after this fix PR.
10. But, if before the druid upgrade, there is a data corruption in the etcd cluster, causing a quorum loss, then the new druid PreSync will not be able to handle this, and this is a current limitation of the reconciliation logic today. Druid will still mark the `Reconcile` operation as `Failed`, and an operator will need to manually look into it and fix it by performing a recovery from quorum loss using [this guide](https://github.com/gardener/etcd-druid/blob/master/docs/operations/recovery-from-permanent-quorum-loss-in-etcd-cluster.md).

/invite @acumino 

**Release note**:

```breaking user
Before upgrading druid to `v0.23.0+`, please ensure that druid is running with at least `v0.22.3+`. This is required to avoid any downtime during the upgrade of the etcds by the new druid version, as well as to ensure backward compatibility of your etcds, in case you wish to downgrade back to `v0.22.3+`.
```

### DIFF
--- a/api/v1alpha1/helper.go
+++ b/api/v1alpha1/helper.go
@@ -44,6 +44,15 @@ func GetOrdinalPodName(etcdObjMeta metav1.ObjectMeta, ordinal int) string {
 	return fmt.Sprintf("%s-%d", etcdObjMeta.Name, ordinal)
 }
 
+// GetAllPodNames returns the names of all pods for the Etcd.
+func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
+	podNames := make([]string, 0, replicas)
+	for i := 0; i < int(replicas); i++ {
+		podNames = append(podNames, GetOrdinalPodName(etcdObjMeta, i))
+	}
+	return podNames
+}
+
 // GetMemberLeaseNames returns the name of member leases for the Etcd.
 func GetMemberLeaseNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
 	leaseNames := make([]string, 0, replicas)
--- a/api/v1alpha1/helper.go
+++ b/api/v1alpha1/helper.go
@@ -47,7 +47,7 @@ func GetOrdinalPodName(etcdObjMeta metav1.ObjectMeta, ordinal int) string {
 // GetAllPodNames returns the names of all pods for the Etcd.
 func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
 	podNames := make([]string, 0, replicas)
-	for i := 0; i < int(replicas); i++ {
+	for i := range int(replicas) {
 		podNames = append(podNames, GetOrdinalPodName(etcdObjMeta, i))
 	}
 	return podNames
--- a/api/v1alpha1/helper.go
+++ b/api/v1alpha1/helper.go
@@ -46,18 +46,18 @@ func GetOrdinalPodName(etcdObjMeta metav1.ObjectMeta, ordinal int) string {
 
 // GetAllPodNames returns the names of all pods for the Etcd.
 func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
-	podNames := make([]string, 0, replicas)
+	podNames := make([]string, replicas)
 	for i := range int(replicas) {
-		podNames = append(podNames, GetOrdinalPodName(etcdObjMeta, i))
+		podNames[i] = GetOrdinalPodName(etcdObjMeta, i)
 	}
 	return podNames
 }
 
 // GetMemberLeaseNames returns the name of member leases for the Etcd.
 func GetMemberLeaseNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
-	leaseNames := make([]string, 0, replicas)
-	for i := 0; i < int(replicas); i++ {
-		leaseNames = append(leaseNames, fmt.Sprintf("%s-%d", etcdObjMeta.Name, i))
+	leaseNames := make([]string, replicas)
+	for i := range int(replicas) {
+		leaseNames[i] = fmt.Sprintf("%s-%d", etcdObjMeta.Name, i)
 	}
 	return leaseNames
 }
--- a/internal/component/statefulset/statefulset.go
+++ b/internal/component/statefulset/statefulset.go
@@ -17,6 +17,7 @@ import (
 	"github.com/gardener/gardener/pkg/controllerutils"
 	"github.com/gardener/gardener/pkg/utils/imagevector"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


api/v1alpha1/helper.go
```diff
@@ -44,6 +44,15 @@ func GetOrdinalPodName(etcdObjMeta metav1.ObjectMeta, ordinal int) string {
 	return fmt.Sprintf("%s-%d", etcdObjMeta.Name, ordinal)
 }
 
+// GetAllPodNames returns the names of all pods for the Etcd.
+func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
+	podNames := make([]string, 0, replicas)
+	for i := 0; i < int(replicas); i++ {
+		podNames = append(podNames, GetOrdinalPodName(etcdObjMeta, i))
+	}
+	return podNames
```
seshachalam-yv, line 53: The `for` loop in Go 1.22 now supports ranging directly over integers, as detailed here: [Go 1.22 Update](https://tip.golang.org/doc/go1.22#language). I suggest utilizing this feature for cleaner code:
```suggestion
	podNames := make([]string, replicas)
	for i := range int(replicas) {
		podNames[i] = GetOrdinalPodName(etcdObjMeta, i)
	}
	return podNames
```

internal/component/statefulset/statefulset.go
```diff
@@ -282,25 +283,18 @@ func getDesiredPodTemplateLabels(etcd *druidv1alpha1.Etcd) map[string]string {
...
+	// Etcd.spec.replicas is updated but not yet reflected in the etcd cluster
+	podNames := druidv1alpha1.GetAllPodNames(etcd.ObjectMeta, *sts.Spec.Replicas)
+	for _, podName := range podNames {
+		pod := &corev1.Pod{}
+		if err := r.client.Get(ctx, client.ObjectKey{Name: podName, Namespace: etcd.Namespace}, pod); err != nil {
+			return false, err
+		}
+		if !utils.ContainsAllDesiredLabels(pod.Labels, getDesiredPodTemplateLabels(etcd)) {
+			return false, nil
+		}
```
seshachalam-yv, line 297:
```suggestion
	if sts.Spec.Replicas == nil {
		return false, fmt.Errorf("statefulset %s does not have a replicas count defined", sts.Name)
	}
	podNames := druidv1alpha1.GetAllPodNames(etcd.ObjectMeta, *sts.Spec.Replicas)
	desiredLabels := getDesiredPodTemplateLabels(etcd)

	for _, podName := range podNames {
		pod := &corev1.Pod{}
		err := r.client.Get(ctx, client.ObjectKey{Name: podName, Namespace: etcd.Namespace}, pod)
		if err != nil {
			return false, fmt.Errorf("failed to get pod %s: %w", podName, err)
		}
		if !utils.ContainsAllDesiredLabels(pod.Labels, desiredLabels) {
			return false, nil
		}
```

seshachalam-yv (changes requested):

shreyas-s-rao:
@seshachalam-yv thanks for your review. Your comments are addressed now. PTAL

seshachalam-yv (approved):
/lgtm

ishan16696:
> PreSync behaviour to no longer wait for etcd pods to be updated with the latest sts spec, or to be ready

If it's isn't waiting for pod to be `ready/running`, isn't that can cause a transient quorum loss ?

ishan16696:
> Etcd spec is updated with non-existent secret reference (TLS or backup store)

we usually never go from TLS to non-TLS, Is there a case where we want to do this ?

ishan16696:
Had a offline discussion with @shreyas-s-rao and he informed me that he didn't see the transient quorum loss while testing a upgrading from `v0.22.3` to `v0.23.0`, but he observed a transient quorum loss while upgrading `v0.22.0` to `v0.23.0`.
So, we concluded that he will mentioned this in a release note to upgrade the druid to `v0.23.0` only atleast from `v0.22.3`.

ishan16696 (approved):
LGTM!!

shreyas-s-rao:
I've added a `noteworthy` release note informing users to upgrade to v0.22.3+ before upgrading to v0.23.0.

PR #828, opened 2024-07-11
Minimize probability of race condition between etcd and custodian controllers for etcd status updates

shreyas-s-rao:

/area quality
/kind bug

Minimize probability of race condition between etcd and custodian controllers for etcd status updates, by re-fetching the etcd resource whenever etcd controller is about to update the etcd resource status.

Fixes #

This fix is not required on master branch, since there is no custodian controller there, hence no race condition.

**Release note**:

```other operator
NONE
```


seshachalam-yv (approved):
/lgtm

PR #834, opened 2024-07-22
removed g/g scripts dependency, updated gomod dependencies, adapted g…

### LINKED ISSUE #503: Remove the dependency on g/g hack/scripts
**Describe the bug:**
gardener core currently has a LOT of scripts under /hack which brings in a lot of dependencies into the vendor directory. Most of these dependencies are not used by etcd-druid. Via this issue we wish to remove the dependency on g/g /hack/scripts so that we can have leaner vendor directly which should only contain what druid uses.

**Expected behavior:**
Vendor what is used and nothing more than that.

NA

Linked issue: #447 Once we remove the dependency then we will also upgrade the go.mod to move to the latest released version of g/g

unmarshall:

/area quality
/area open-source
/kind cleanup
/kind technical-debt

1. Removes dependency on gardener/gardener hack/scripts.
2. Updates gardener/gardener dependency to 1.99.1
3. Updates controller-runtime k8s.io/*, logr, ginkgo, zap etc dependencies in go.mod
4. Adapt code to fix issues reported by golanglint-ci
5. etcd CRD has been updated to incorporate recent changes in [kep-3633](https://github.com/kubernetes/enhancements/issues/3633)
6. Go native unit tests now run concurrently. This was not completely done earlier.

Fixes #503 

**Release note**:

```other operator
1. Dependency version upgrades done to gardener/gardener, controller-runtime, controller-tools, k8s.io/*, logr, zap, ginkgo, uber mock, uuid dependencies.
2. Adapted golanglint-ci recommendations.
3. Removed dependency on gardener/gardener hack/scripts.

```


seshachalam-yv (approved):
Thanks @unmarshall for removing g/g scripts dependency.
Nice PR ❤️ 
/lgtm

Makefile
```diff
@@ -32,12 +29,10 @@ include $(HACK_DIR)/tools.mk
 .PHONY: tidy
 tidy:
 	@env GO111MODULE=on go mod tidy
-	@GARDENER_HACK_DIR=$(GARDENER_HACK_DIR) bash $(HACK_DIR)/update-github-templates.sh
-	@cp $(GARDENER_HACK_DIR)/cherry-pick-pull.sh $(HACK_DIR)/cherry-pick-pull.sh && chmod +xw $(HACK_DIR)/cherry-pick-pull.sh
 
 .PHONY: clean
 clean:
-	@bash $(GARDENER_HACK_DIR)/clean.sh ./api/... ./internal/...
+	@$(HACK_DIR)/clean.sh ./api/... ./internal/...
```
renormalize, line 35: In most of the make targets we're calling the shell scripts by prepending the script filename with the `bash` command - but in the proposed changes this convention wasn't followed.

We could possibly make all script invocations through the explicit `bash` command to ensure the targets work without any problem in incompatible shells like `fish` as well.

renormalize (commented):
@unmarshall thanks for the PR!

Please take a look at the single comment from me.

Makefile
unmarshall: `fish` is not a posix compliant shell. https://wiki.gentoo.org/wiki/Fish also as the following warning:
```
fish should not be set as the system shell by linking in /bin/sh, this could result in an inoperable system.
```
even upstream projects like k8s (https://github.com/kubernetes/kubernetes/blob/master/build/root/Makefile), etcd (https://github.com/etcd-io/etcd/blob/main/Makefile), our own gardener (https://github.com/gardener/gardener/blob/master/Makefile) also do not care about invoking commands via `bash`

renormalize: All the make targets I've tested indeed work on shells like `fish` because all our scripts indicate the file is a bash script through the `#!/bin/bash` comment at the top.

Since this is the case, it would be cleaner to not invoke bash explicitly in make targets since it is redundant?

unmarshall: yes i would also second that.

Makefile
```diff
@@ -2,8 +2,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-ENSURE_GARDENER_MOD := $(shell go get github.com/gardener/gardener@$$(go list -m -f "{{.Version}}" github.com/gardener/gardener))
-GARDENER_HACK_DIR   := $(shell go list -m -f "{{.Dir}}" github.com/gardener/gardener)/hack
 VERSION             := $(shell cat VERSION)
 REPO_ROOT           := $(shell dirname "$(realpath $(lastword $(MAKEFILE_LIST)))")
 HACK_DIR            := $(REPO_ROOT)/hack
```
renormalize, line 7: I think we can get rid of this `HACK_DIR` variable now that there is no `GARDENER_HACK_DIR`. All "hack" related files/binaries would live in `hack` anyway.

Hardcoding `hack` in the file paths might not seem ideal, but in my opinion it's pretty unlikely this directory would ever be renamed, and having full paths to files like `./hack/run-script.sh` would enable us to directly jump to this file in any text editor which provides the capability to jump to a file with a path; instead of having to search the file tree or type the file name in a command pallette. I personally use this feature of my text editor a lot while navigating.

We can re-evaluate other such parts of the Makefile which could be improved to enhance developer productivity.

renormalize (approved):

PR #835, opened 2024-07-22
Fixes pre-deploy step having incorrect desired labels and adds a temp fix for race condition

### LINKED ISSUE #836: `pre-deploy` step expecting wrong desired labels.
/area high-availability
/kind bug

**What happened**:
In branch `hot-v0.22` a [pre-deploy step](https://github.com/gardener/etcd-druid/blob/876927637414906e86ad3f798de6641e8713d975/pkg/component/etcd/statefulset/statefulset.go#L223) was introduced to ensure that etcd-druid is backward compatible with the changes that are made in master w.r.t pod labels and pre-deploy flow [waits until pods have desired labels](https://github.com/gardener/etcd-druid/blob/876927637414906e86ad3f798de6641e8713d975/pkg/component/etcd/statefulset/statefulset.go#L290-L308). 
When g/g e2e tests are run then in case of upgrading the existing etcd cluster from nonHA to HA, g/g will add additional networking labels on the etcd resource. These labels will never be present on the pod in the PreDeploy step. Hence, this will result in the reconciliation getting stuck at waiting stage. Deploy step is never called, so the new labels that are added by g/g on the etcd resource never make it to the pods.

**Environment**:

- etcd-druid version: v0.22.x
- Others:

unmarshall:

/area control-plane
/kind bug

In hotfix branch `PreDeploy` step was introduced to ensure that etcd-druid is backward compatible with the changes that are made in master w.r.t pod labels and statefulset label-selector. An issue was discovered when running e2e g/g tests.

Desired labels passed to [utils.ContainsAllDesiredLabels](https://github.com/gardener/etcd-druid/blob/876927637414906e86ad3f798de6641e8713d975/pkg/component/etcd/statefulset/statefulset.go#L302) contains the additional labels which are fetched from `etcd.Spec.Labels`. When g/g e2e tests are run then in case we upgrade existing etcd cluster from 1 to 3, g/g will add additional networking labels on the etcd resource. These labels will never be present on the pod in the `PreDeploy` step. This will result in the reconciliation getting stuck at this stage. `Deploy` step is never called, so the new labels that are added by g/g on the etcd resource never make it to the pods.

> NOTE: This is not caught in druid e2e tests because we do not add additional labels to etcd resource when running an upgrade test from 1 to 3 replicas. We should make this change so that these issues can be caught early and not when g/g e2e tests are run.

Additionally we now add one more safeguard in the custodian controller to skip reconciliation if the reconcile annotation is still added on the etcd resource and retry after 10s. This is done in addition to the already present predicate which also does something similar. However the predicate does not prevent requeue of requests due to error in the custodian which can still interfere with the etcd reconciler.

Fixes #836 

**Release note**:

```bugfix operator
Fixes the labels comparison check done in the PreDeploy step which ensures that the pods have both the old and the new labels.
```

### DIFF
--- a/controllers/custodian/reconciler.go
+++ b/controllers/custodian/reconciler.go
@@ -20,6 +20,7 @@ import (
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/v1alpha1"
 
+	ctrlutils "github.com/gardener/etcd-druid/controllers/utils"
 	"github.com/gardener/etcd-druid/pkg/health/status"
 	"github.com/gardener/etcd-druid/pkg/utils"
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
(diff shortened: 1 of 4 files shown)


ishan16696 (approved):
LGTM!!

seshachalam-yv (approved):
/lgtm

PR #841, opened 2024-07-23
Statefulset PreDeploy flow checks for pod readiness if labels were patched

shreyas-s-rao:

/area quality
/kind bug

Statefulset PreDeploy flow checks for pod readiness if labels were patched. This ensures there is no transient quorum loss due to back-to-back double rolling of sts pods during downgrade of etcd-druid from master/v0.23.0+ version to v0.22.x version.

Fixes #

/cc @unmarshall @seshachalam-yv @ishan16696 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/component/etcd/statefulset/statefulset.go
+++ b/pkg/component/etcd/statefulset/statefulset.go
@@ -242,12 +242,14 @@ func (c *component) createPreDeployFlow(ctx context.Context, etcd *druidv1alpha1
 }
 
 func (c *component) addTasksForLabelsAndSelectorUpdation(g *flow.Graph, etcd *druidv1alpha1.Etcd, sts *appsv1.StatefulSet) {
+	var labelsPatched bool
+
 	// If the labels are not as expected, then patch the labels.
 	patchLabelsOpName := "(patch-labels): Patching labels"
 	patchLabelsTaskID := g.Add(flow.Task{
 		Name: patchLabelsOpName,
 		Fn: func(ctx context.Context) error {
-			return c.patchPodTemplateLabels(ctx, sts)
+			return c.patchPodTemplateLabels(ctx, sts, &labelsPatched)
 		},
 		Dependencies: nil,
 	})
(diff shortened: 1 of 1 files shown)


pkg/component/etcd/statefulset/statefulset.go
```diff
@@ -302,11 +308,25 @@ func (c *component) waitUntilPodsHaveDesiredLabels(ctx context.Context, etcd *dr
 			if !utils.ContainsAllDesiredLabels(pod.Labels, c.values.PodLabels) {
 				return false, nil
 			}
+			// If labels were patched in the current pre-deploy flow, then wait for all pods to be ready.
+			// This is to ensure that the double rolling of the sts pods does not lead to transient quorum loss.
+			if *labelsPatched && !isPodReady(pod) {
```
seshachalam-yv, line 313: @shreyas-s-rao 
I'm concerned about the scenario where a transient quorum loss occurs. If the error reconciliation process triggers a retry later, it seems we may not recheck pod readiness if the initial readiness check fails.

seshachalam-yv (changes requested):

pkg/component/etcd/statefulset/statefulset.go
shreyas-s-rao: In this branch of druid, the reconciliation loop does not exit early and retry operation later. Instead, it waits for the operations to succeed or fail and then continue. So this PR fixes this issue of transient quorum loss for the happy path, which is still a large improvement over what we have right now on this branch.

In the happy path of pre-deploy, reconciler patches labels, waits for rolling, then deletes sts within the same reconciliation loop. If the reconciler errors or druid restarts during this period for any reason, then we face a transient quorum loss. But this is a rare case, and with the reconciliation design of this branch (v0.22), there's not much we can do here. Such issues don't occur on master branch since reconciliation is streamlined and made resilient to restarts by #777 .

So I would say we can merge this PR without further changes. After all, the worst case is still a transient quorum loss, which causes 1-2 minutes of downtime. And in the case of gardener, this happens within the shoot maintenance window, so the impact is minimized.

seshachalam-yv: 👍

seshachalam-yv (approved):
/lgtm

ishan16696:
Hi @shreyas-s-rao ,
Can't we use this https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#minimum-ready-seconds to give some buffer to the pods to become `Ready` ?

shreyas-s-rao:
@ishan16696 this is a good suggestion in general, but atleast as of today druid does not set `minReadySeconds` on the sts, and druid is the only actor who creates and manages the sts. So there's no need to consider this field here.

ishan16696 (approved):
LGTM!!

ishan16696:
> @ishan16696 this is a good suggestion in general, but atleast as of today druid does not set minReadySeconds on the sts, and druid is the only actor who creates and manages the sts. So there's no need to consider this field here.

yeah, let's discuss this later . As of now, we can go ahead with this PR.

PR #842, opened 2024-07-24
Fix build information in logs

shreyas-s-rao:

/area logging
/kind bug

Fix build information in logs, due to change in package structure from `pkg` to `internal`, as well as missing ldflags in skaffold KO build manifest.

I have also added 3 new Makefile targets: `docker-clean`, `clean-build-cache`, `clean-mod-cache`. The first two are helpful in cleaning up the cache created by KO builds for druid images, to make it easy to test changes in the PR (since KO reuses caches unless something in the code changes, as specified in the skaffold manifest).

Fixes #

/cc @unmarshall 

**Release note**:

```other developer
Add Make target `make docker-clean` for cleaning up all docker builds related to etcd-druid.
```
```other developer
Add Make targets `make clean-build-cache` and `make clean-mod-cache` for cleaning up Go build and mod caches respectively.
```

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -17,6 +17,10 @@ build:
             - VERSION
         flags:
           - -v
+        ldflags:
+          - -w
+          - -X github.com/gardener/etcd-druid/internal/version.Version={{.VERSION}}
+          - -X github.com/gardener/etcd-druid/internal/version.GitSHA={{.GIT_SHA}}
 deploy:
   helm:
     releases:
--- a/.ci/build
+++ b/.ci/build
@@ -33,5 +33,5 @@ GIT_SHA="${GIT_SHA:-$(git rev-parse --short HEAD || echo "GitNotFound")}"
 CGO_ENABLED=0 GO111MODULE=on go build \
   -v \
   -o "${BINARY_PATH}"/etcd-druid \
-  -ldflags "-w -X ${REPOSITORY}/pkg/version.Version=${VERSION} -X ${REPOSITORY}/pkg/version.GitSHA=${GIT_SHA}" \
+  -ldflags "-w -X ${REPOSITORY}/internal/version.Version=${VERSION} -X ${REPOSITORY}/internal/version.GitSHA=${GIT_SHA}" \
   main.go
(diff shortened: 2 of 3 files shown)


skaffold.yaml
```diff
@@ -17,6 +17,10 @@ build:
             - VERSION
         flags:
           - -v
+        ldflags:
+          - -w
+          - -X github.com/gardener/etcd-druid/internal/version.Version={{.VERSION}}
```
seshachalam-yv, line 22: We are not explicitly setting the VERSION; it must be passed externally when invoking Skaffold. [Prow e2e test](https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_etcd-druid/842/pull-etcd-druid-e2e-kind/1816302820614213632#1:build-log.txt%3A1033-1124) is failing. Since VERSION is not set, it defaults to an empty string, leading to an incorrect configuration:

```
.PHONY: deploy
deploy: $(SKAFFOLD) $(HELM)
	@VERSION=$(VERSION) GIT_SHA=$(GIT_SHA) $(SKAFFOLD) run -m etcd-druid
```

seshachalam-yv (changes requested):
Thank you, @shreyas-s-rao, for the fix. Overall, it looks good to me. Please address my one review comment.

skaffold.yaml
shreyas-s-rao: Ah you're right. Will fix this, thanks

shreyas-s-rao: On second thought, `make deploy` is never called from `make ci-e2e-kind`. Skaffold is called from [here](https://github.com/gardener/etcd-druid/blob/c73cbf9c5715ff157694a919ee73d079d668c0aa/hack/e2e-test/run-e2e-test.sh#L98), via `make test-e2e`. So the change needed to be made there. I have pushed the new changes now. PTAL.

Also, I have added some helper Make targets for easily cleaning up go build cache and docker builds for druid.

seshachalam-yv: 👍

seshachalam-yv (approved):
/lgtm

Issue #580, opened 2023-04-17
☂️ [Enhancement] Create v1beta1 for etcd-druid API


unmarshall:
# How to categorize this issue?
/kind enhancement

## What would you like to be added
Deprecate `v1alpha1` (current version) and introduce `v1beta1` of the API.

# Motivation (Why is this needed?)
druid API is currently v1alpha1 (which technically indicates that it is experimental and should be used with caution). However the API has been stable for quite some time and therefore it should be moved to v1beta1. 
In addition the API in its current form has several problems:
- Inconsistent use of labels and annotations across several resources that are created.
- Semantically incorrect names for structs/fields.
- Sub-optimal structuring of different sets of logically cohesive configuration.
- cluster specific configuration mixed with etcd-member specific configuration

As a consequence v1beta1 will deviate from v1alpha1 incompatibly. It should be ensured that both the versions `v1alpha1` and `v1beta1` be supported with `v1beta1` to be used as a stored version (in  the control plane of druid). An API conversion web-hook should be created to convert between the two versions. 

# Tasks
  - [ ] Define the v1beta1 API and get this reviewed. 
    - [ ] Add a compaction-specific configuration section to the Etcd API, and remove the special handling added in #845 
  - [ ] Till [etcd-druid-api](https://github.com/gardener/etcd-druid-api/) is integrated with etcd-druid, the changes needs to be present in both the places.
  - [ ] Create an API conversion web hook
  - [ ] Add e2e tests to consume both v1alpha1 and v1beta1

PR #845, opened 2024-07-30
Compaction job shouldn't add `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` annotation to compaction pod.

### LINKED ISSUE #766: Improve Node Utilization by Avoiding "safe-to-evict" Annotation for Druid-Managed Pods
# How to categorize this issue?
/kind enhancement

# Stories
- As operator, I want to run my ETCDs efficiently (using the bin-packing kube-scheduler and the cluster-autoscaler down-scaling), but etcd-main(-compactor) pods disallow cluster-autoscaler down-scaling

## What would you like to be added
- [x] Do not add the `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` annotation to the etcd-main-compactor jobs (just because it's in the etcd-main spec)
- [x] Do not add the `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` annotation to the etcd-main pods for the clustered etcd-main statefulset
- [x] ~~Either switch all clusters to clustered etcd-main (except for evaluation clusters) and remove the annotation in general~~ (cost-prohibitive)
- [ ] ...or consider a more elaborate approach how to deal with the singleton etcd-main statefulset:
  - If inside the maintenance time window (with a jitter, so that we do not terminate pods all at once at the full hour, which may lead to prolonged detach/attach times)
  - If the cluster does not have the purpose "production"
  - If the requests utilization (CPU or memory) for the node hosting the pod is below 70% (atm.; ideally configurable)

  Implementation details:
  - Instead of the druid reading the purpose, gardenlet could do that and set a new field at the etcd resource that permits or forbids voluntary evictions.
  - A new controller, let's call it terminator (:-)), could do the above as part of druid.

# Motivation (Why is this needed?)
We have adopters seeing 80-90 % requests utilization on the regular seed cluster nodes, but only 40-50% for their dedicated ETCD seed cluster nodes.

ishan16696:
/area cost
/kind enhancement

Right now, compaction job adds all annotation(s) present in `etcd custom resource spec` to compaction pod template. With this PR, compaction will not add annotation: `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` to compaction pod if its present to improve node utilization.

Fixes Point1 of #766 

**Release note**:

```noteworthy operator|user
To improve node utilization compaction job will not add annotation: `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` to compaction pod.
```

### DIFF
--- a/internal/controller/compaction/reconciler.go
+++ b/internal/controller/compaction/reconciler.go
@@ -38,6 +38,9 @@ import (
 const (
 	// DefaultETCDQuota is the default etcd quota.
 	DefaultETCDQuota = 8 * 1024 * 1024 * 1024 // 8Gi
+
+	// SafeToEvictKey - annotation that ignores constraints to evict a pod if set to "false".
+	SafeToEvictKey = "cluster-autoscaler.kubernetes.io/safe-to-evict"
 )
 
 // Reconciler reconciles compaction jobs for Etcd resources.
@@ -284,7 +287,7 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
 			BackoffLimit:          pointer.Int32(0),
 			Template: v1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Annotations: etcd.Spec.Annotations,
+					Annotations: getEtcdCompactionAnnotations(etcd.Spec.Annotations),
 					Labels:      getLabels(etcd),
 				},
 				Spec: v1.PodSpec{
@@ -492,4 +495,17 @@ func getCompactionJobArgs(etcd *druidv1alpha1.Etcd, metricsScrapeWaitDuration st
 	}
 
 	return command
+}
+
+func getEtcdCompactionAnnotations(etcdAnnotations map[string]string) map[string]string {
+	etcdCompactionAnnotations := make(map[string]string)
+
+	for key, value := range etcdAnnotations {
+		// Do not add annotation: `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` to compaction job's pod template.
+		if key == SafeToEvictKey && value == etcdAnnotations[SafeToEvictKey] {
+			continue
+		}
+		etcdCompactionAnnotations[key] = value
+	}
+	return etcdCompactionAnnotations
 }
--- a/internal/controller/compaction/reconciler.go
+++ b/internal/controller/compaction/reconciler.go
@@ -502,7 +502,7 @@ func getEtcdCompactionAnnotations(etcdAnnotations map[string]string) map[string]
 
 	for key, value := range etcdAnnotations {
 		// Do not add annotation: `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` to compaction job's pod template.
-		if key == SafeToEvictKey && value == etcdAnnotations[SafeToEvictKey] {
+		if key == SafeToEvictKey && value == "false" {
 			continue
 		}
 		etcdCompactionAnnotations[key] = value
--- a/internal/controller/compaction/reconciler.go
+++ b/internal/controller/compaction/reconciler.go
@@ -39,7 +39,8 @@ const (
 	// DefaultETCDQuota is the default etcd quota.
 	DefaultETCDQuota = 8 * 1024 * 1024 * 1024 // 8Gi
 
-	// SafeToEvictKey - annotation that ignores constraints to evict a pod if set to "false".
+	// SafeToEvictKey - annotation that ignores constraints to evict a pod like not being replicated, being on
+	// kube-system namespace or having a local storage if set to "false".
 	SafeToEvictKey = "cluster-autoscaler.kubernetes.io/safe-to-evict"
 )
 
--- a/internal/controller/compaction/register_test.go
+++ b/internal/controller/compaction/register_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/gardener/etcd-druid/internal/utils"
 	. "github.com/onsi/gomega"
 	batchv1 "k8s.io/api/batch/v1"
 	coordinationv1 "k8s.io/api/coordination/v1"
@@ -197,6 +198,20 @@ func TestJobStatusChangedForUpdateEvents(t *testing.T) {
 			g.Expect(predicate.Update(event.UpdateEvent{ObjectOld: oldObj, ObjectNew: obj})).To(Equal(test.shouldAllowUpdateEvent))
 		})
 	}
+}
+
+func TestEtcdCompactionAnnotation(t *testing.T) {
+	test1EtcdAnnotation := map[string]string{
+		"dummy-annotation": "dummy",
+	}
+
+	test2EtcdAnnotation := map[string]string{
+		SafeToEvictKey: "false",
+	}
+
+	g := NewWithT(t)
+	compactionAnnotation := getEtcdCompactionAnnotations(utils.MergeMaps(test1EtcdAnnotation, test2EtcdAnnotation))
+	g.Expect(compactionAnnotation).To(Equal(test1EtcdAnnotation))
 }
 
 func createObjectsForJobStatusChangedPredicate(g *WithT, name string, isJobObj, isStatusChanged bool) (obj client.Object, oldObj client.Object) {


internal/controller/compaction/reconciler.go
```diff
@@ -38,6 +38,9 @@ import (
 const (
 	// DefaultETCDQuota is the default etcd quota.
 	DefaultETCDQuota = 8 * 1024 * 1024 * 1024 // 8Gi
+
+	// SafeToEvictKey - annotation that ignores constraints to evict a pod if set to "false".
+	SafeToEvictKey = "cluster-autoscaler.kubernetes.io/safe-to-evict"
```
renormalize, line 43: Should we use this constant which is originally defined [here](https://github.com/kubernetes/autoscaler/blob/45dfba4026a0d1bc4d5d584a57771107e1663437/cluster-autoscaler/utils/drain/drain.go#L36), by adding the Go module `k8s.io/autoscaler/cluster-autoscaler` as a dependency, instead of re-declaring the string here?

This makes the dependency on cluster autoscaler explicit in etcd-druid.
This would also help us track potential changes to this annotation since we would be forced to keep track of changes to cluster autoscaler if we depend on it.

renormalize (changes requested):
Just a question, I'd like to know what you think. Thanks.

internal/controller/compaction/reconciler.go
ishan16696: we don't directly consume autoscaler, we use forked of autoscaler https://github.com/gardener/autoscaler.
I don't think it's good to create dependency just for 1 constant. Even g/g is not using const via autoscaler: https://github.com/gardener/gardener/blob/8b7e4cc9e29a0bae4201ce118f45db9d08b721e4/pkg/component/etcd/etcd/etcd.go#L224

renormalize: I don't particularly feel strongly about this.
We shouldn't ponder too much and get this in for now. We can always add the dependency later on if we feel that is the better way of doing it.

internal/controller/compaction/reconciler.go
```diff
@@ -493,3 +496,16 @@ func getCompactionJobArgs(etcd *druidv1alpha1.Etcd, metricsScrapeWaitDuration st
 
 	return command
 }
+
+func getEtcdCompactionAnnotations(etcdAnnotations map[string]string) map[string]string {
+	etcdCompactionAnnotations := make(map[string]string)
+
+	for key, value := range etcdAnnotations {
+		// Do not add annotation: `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` to compaction job's pod template.
+		if key == SafeToEvictKey && value == etcdAnnotations[SafeToEvictKey] {
```
renormalize, line 505: Don't `key == SafeToEvictKey` and `value == etcdAnnotations[SafeToEvictKey]` mean the same thing given the fact that `&&` is short-circuit?

`value` is nothing but `etcdAnnotations[key]`, which means:

when `key == SafeToEvictKey`
it is tautologically `value == etcdAnnotations[SafeToEvictKey]`
in the expression `key == SafeToEvictKey && value == etcdAnnotations[SafeToEvictKey]`.

```suggestion
		if key == SafeToEvictKey {
```

renormalize (changes requested):

internal/controller/compaction/reconciler.go
ishan16696: addressed.

renormalize: Thanks.

renormalize (approved):
Thanks for addressing all comments!

renormalize:
Also, @ishan16696; using "Fixes #<issue-number>" would automatically close that issue when this PR merges.
Maybe you can mention the issue number in the PR description instead of the "Fixes:" section. Thanks!

ishan16696:
> Also, @ishan16696; using "Fixes #" would automatically close that issue when this PR merges.

I will re-open it, if it closes ... not a bigger thing 😅

anveshreddy18 (commented):
Thanks for the PR @ishan16696. Just a couple minor questions. Thanks.

internal/controller/compaction/reconciler.go
```diff
@@ -493,3 +496,16 @@ func getCompactionJobArgs(etcd *druidv1alpha1.Etcd, metricsScrapeWaitDuration st
 
 	return command
 }
+
+func getEtcdCompactionAnnotations(etcdAnnotations map[string]string) map[string]string {
+	etcdCompactionAnnotations := make(map[string]string)
+
+	for key, value := range etcdAnnotations {
+		// Do not add annotation: `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` to compaction job's pod template.
+		if key == SafeToEvictKey && value == "false" {
```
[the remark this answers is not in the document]
ishan16696: Right now, I want functionality to be correct, so I'm specifically checking and ignoring this annotation only `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` only.

>Can we just check only key because we don't set SafeToEvictKey:"true" to etcd anyway.

we don't know, may be somebody want to set this in future, as it's kind of depend on consumer of druid as in our case it's g/g which creates Etcd CR and annotation is used by CA.
Right now, I want to be functionality to be exact to what we want.

internal/controller/compaction/reconciler.go
```diff
@@ -38,6 +38,9 @@ import (
 const (
 	// DefaultETCDQuota is the default etcd quota.
 	DefaultETCDQuota = 8 * 1024 * 1024 * 1024 // 8Gi
+
+	// SafeToEvictKey - annotation that ignores constraints to evict a pod if set to "false".
```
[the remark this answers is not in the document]
ishan16696: I don't get your point exactly.
PDB has nothing to do with `SafeToEvictKey`. PDB will and must be respected at all time.
This annotation is used by CA which checks the pod before evicting the pod  while doing node drain or any maintenance operation.

> And when it is set to "true", then I think it means pod can be evicted even if it violates some constraints, again like pdb, or pod affinity constraints.

If this annotation is set to true then CA can evict the pod but PDB still be respected as PDB has maxUnavailable set `1`.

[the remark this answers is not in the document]
ishan16696: And this is done on compaction pod, and there is no PDB on compaction pod.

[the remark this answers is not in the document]
ishan16696: reason behind this is `compaction pod` should not have this annotation `cluster-autoscaler.kubernetes.io/safe-to-evict: "false"` as this will hinder the CA ops as CA won't be able to drain the node.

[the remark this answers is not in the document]
anveshreddy18: > If this annotation is set to true then CA can evict the pod but PDB still be respected as PDB has maxUnavailable set 1

Yes but I meant when it has to evict second pod for some reason, then if we set to "true", then will it allow pod eviction by "ignoring constraints" of pdb ( or some other constraint, not necessarily pdb)? 

my only concern was the word "ignoring constraints" in the comment. Because I thought we "ignore constraints" when it is set to "true" but the comment says we ignore when it is set to "false" that's why I had a doubt.  But I get your point of view. Thanks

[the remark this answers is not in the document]
ishan16696: done.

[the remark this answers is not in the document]
anveshreddy18: > First of all, this PR is related to compaction pod created by compaction job, so there is no PDB or any constraints for compaction pod.

Yeah true, but since the comment was explanation of the annotation, I was thinking from a generic point of view. 

Thanks for improving the comment.

anveshreddy18 (approved):
Thanks for addressing the comment. 
/lgtm

shreyas-s-rao:
/needs cherry-pick
As discussed, to be cherry-picked to `hotfix-v0.28` branch.

internal/controller/compaction/reconciler.go
```diff
@@ -38,6 +38,10 @@ import (
 const (
 	// DefaultETCDQuota is the default etcd quota.
 	DefaultETCDQuota = 8 * 1024 * 1024 * 1024 // 8Gi
+
+	// SafeToEvictKey - annotation that ignores constraints to evict a pod like not being replicated, being on
+	// kube-system namespace or having a local storage if set to "false".
+	SafeToEvictKey = "cluster-autoscaler.kubernetes.io/safe-to-evict"
```
shreyas-s-rao, line 44: Druid should not have any idea about which annotations are being set. Druid should simply never set the Etcd `spec.annotations` on the compaction job pod, because that is something the creator of the Etcd resource has no idea about. I strongly recommend completely removing the logic of adding spec.annotations to the compaction job pod.

shreyas-s-rao (changes requested):
@ishan16696 thanks for the PR. Please address my review comment, thanks.

internal/controller/compaction/reconciler.go
shreyas-s-rao: After an out-of-band discussion with @ishan16696 and @unmarshall , we decided to go ahead with this change, since the current Etcd API does not provide any way for users to specify separate set of annotations for the compaction jobs. We will remove this special handling when working on #580

shreyas-s-rao (approved):
/lgtm

Issue #707, opened 2023-10-12
[Feature] Configurable resource requests/limits for compaction jobs


shreyas-s-rao:
**Feature (What you would like to be added):**
Configurable resource requests/limits for compaction jobs, per etcd resource.

**Motivation (Why is this needed?):**
It would be beneficial to have configurable initial resource requests/limits for compaction jobs, so that it can be configured per etcd resource, and can be completely controlled. 

See #652 for why setting requests & limits are helpful for compaction job

**Approach/Hint to the implement solution (optional):**
- [x] #652

Issue #652, opened 2023-07-31
[Feature] Set Request and Limit for CPU and Memory resources for compaction job


abdasgupta:
**Feature (What you would like to be added):**
Compaction job needs to start with request and limit resources set.

**Motivation (Why is this needed?):**
Currently, compaction jobs have no request and limit set for CPU and Memory resources. A needy compaction job may use up all the resources and can cause starvation for other pods running in the same machine/node. If we consider the scenario of OOM kill, because of a compaction job system memory might be exhausted and OS may trigger OOM kill for any random pod. If we set request and limit resource for a compaction job, kubernetes scheduler might be aware of the needy compaction job and can issue a OOM kill only for the compaction job.

**Approach/Hint to the implement solution (optional):**
It's not much of a change to just set the request and limit for compaction job. But what can take longer is to find out what is the good value for request and limit so that, most of the compaction jobs can be accommodated and least number of pod kill happens by the Kubernetes scheduler 

Tasks: 
- [x] Requests
- [ ] Limits

anveshreddy18:
Closing this in favour of #707 which will track requests & limits going forward.

PR #853, opened 2024-08-08
Set CPU & Memory requests for compaction job

anveshreddy18:

/area backup
/kind enhancement

This PR adds CPU and Memory requests for compaction job. The requests are calculated from seed clusters and heavy load shoot clusters. This is more from a Gardener point of view. To understand why I set these numbers, you can visit the internal backlog issue 14 to know more.

Compaction Jobs originating from druid needs to have Requests & Limits ( limits can be configured after observing the landscapes with just requests set initially ) so as to allow them to get scheduled to nodes with proper resource availability. This will let the scheduler know about the consumption/expectation before hand so that it can schedule optimally so that the compaction jobs can get completed and also not interfere with resources intended for other processes.

Fixes part of #707

**Release note**:

```other operator
set cpu and memory requests for compaction pods
```

### DIFF
--- a/internal/controller/compaction/reconciler.go
+++ b/internal/controller/compaction/reconciler.go
@@ -25,6 +25,7 @@ import (
 	coordinationv1 "k8s.io/api/coordination/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/component-base/featuregate"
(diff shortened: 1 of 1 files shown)


unmarshall (approved):
Looks good. Please remove the link to github.tools.sap. You can just mention the issue number instead.

ishan16696 (approved):
LGTM

ishan16696:
Hi @anveshreddy18 ,
I think this PR is fixing this issue: https://github.com/gardener/etcd-druid/issues/652. Can you please close this issue

anveshreddy18:
Hi @ishan16696, this PR currently only adds requests but we didn't add limits yet. But that issue is both for requests & limits.  I think we can keep it open? or should I edit it to have check list for requests & limits and tick the requests part?

unmarshall:
I am not sure if we need limits. Adding requests are good enough for now.

ishan16696:
> I am not sure if we need limits. Adding requests are good enough for now.

yes, I agree. For now we needed the requests which have been added and right now, requests values are hardcoded so we anyway have to revisit this to make it configurable for which this issue is already in open: https://github.com/gardener/etcd-druid/issues/707 which we aren't closing. IMO, one issue is fine to keep a track of it.

anveshreddy18:
Yeah, I closed that now.

PR #855, opened 2024-08-08
PDB component sets `unhealthyPodEvictionPolicy` based on annotation

### LINKED ISSUE #854: Evaluate and add PDB spec field `unhealthyPodEvictionPolicy: AlwaysAllow`

/area auto-scaling
/kind enhancement

Evaluate and add PDB spec field `unhealthyPodEvictionPolicy: AlwaysAllow`

**Why is this needed**:
Consider using this field to allow components such as cluster-autoscaler to drain nodes by evicting unhealthy etcd pods. 

Cases (considering a 3-member etcd cluster, and our current readiness probe which essentially reflects cluster health rather than member health):
1. All 3 members are healthy: the PDB spec field will have no effect
2. 2 out of 3 members are healthy - quorum is maintained:
    1. the unhealthy member pod's readiness probe returns false due to scheduling errors, issues with network reachability with the other members, backup sidecar is unhealthy, etc: the pod will be allowed to be evicted by the PDB
    2. the unhealthy member pod's readiness probe returns true due to the pod being up and member being able to talk to other members, but the member itself has a local error/problem, due to which it is actually unhealthy: the pod will not be allowed to be evicted by the PDB
3. 1 out of 3 members is healthy - quorum is lost: since the readiness probe is a cluster-wide health check, a quorum loss results in the 1 healthy member also returning a false readiness check. This results in all the 3 pods to be allowed to be evicted. Now, when the readiness probe returns false, it would result in the etcd client service losing its endpoints, and the service becoming unreachable anyway, until quorum is restored. So in my opinion, with the current implementation of readiness probe, we have no issues in setting PDB field `unhealthyPodEvictionPolicy: AlwaysAllow`. 
    Going forward, if and when we change the readiness probe to only consider member health, then we may need to re-evaluate the `unhealthyPodEvictionPolicy` field. For a case of quorum loss, the field may affect how transient and permanent quorum loss cases are handled. IMHO, setting this field should still be fine, because even a transient quorum loss would have already resulted in service degradation/downtime, and allowing an unhealthy pod to be evicted would not cause any disruption to the single healthy member, if at all there is one.

/cc @unmarshall @ishan16696 @seshachalam-yv @renormalize @anveshreddy18
/cc @dguendisch

shreyas-s-rao:

/area quality
/kind enhancement

PDB component sets `unhealthyPodEvictionPolicy` based on annotation

Fixes #854 

/cc @unmarshall @ishan16696 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/internal/component/poddistruptionbudget/poddisruptionbudget.go
+++ b/internal/component/poddistruptionbudget/poddisruptionbudget.go
@@ -114,8 +114,7 @@ func buildResource(etcd *druidv1alpha1.Etcd, pdb *policyv1.PodDisruptionBudget) 
 		MatchLabels: druidv1alpha1.GetDefaultLabels(etcd.ObjectMeta),
 	}
 	if metav1.HasAnnotation(etcd.ObjectMeta, annotationAllowUnhealthyPodEviction) {
-		alwaysAllow := policyv1.AlwaysAllow
-		pdb.Spec.UnhealthyPodEvictionPolicy = &alwaysAllow
+		pdb.Spec.UnhealthyPodEvictionPolicy = utils.PointerOf(policyv1.AlwaysAllow)
 	} else {
 		pdb.Spec.UnhealthyPodEvictionPolicy = nil
 	}
--- a/internal/component/poddistruptionbudget/poddisruptionbudget.go
+++ b/internal/component/poddistruptionbudget/poddisruptionbudget.go
@@ -29,6 +29,12 @@ const (
 	ErrDeletePodDisruptionBudget druidv1alpha1.ErrorCode = "ERR_DELETE_POD_DISRUPTION_BUDGET"
 )
 
+const (
+	// annotationAllowUnhealthyPodEviction is an annotation that can be set on the Etcd resource, to allow unhealthy pod eviction.
+	// WARNING: this annotation may be removed at any point of time, and is NOT meant to be generally used.
+	annotationAllowUnhealthyPodEviction = "resources.druid.gardener.cloud/allow-unhealthy-pod-eviction"
+)
+
 type _resource struct {
 	client client.Client
 }
@@ -106,6 +112,12 @@ func buildResource(etcd *druidv1alpha1.Etcd, pdb *policyv1.PodDisruptionBudget) 
 	}
 	pdb.Spec.Selector = &metav1.LabelSelector{
 		MatchLabels: druidv1alpha1.GetDefaultLabels(etcd.ObjectMeta),
+	}
+	if metav1.HasAnnotation(etcd.ObjectMeta, annotationAllowUnhealthyPodEviction) {
+		alwaysAllow := policyv1.AlwaysAllow
+		pdb.Spec.UnhealthyPodEvictionPolicy = &alwaysAllow
+	} else {
+		pdb.Spec.UnhealthyPodEvictionPolicy = nil
 	}
 }
 


internal/component/poddistruptionbudget/poddisruptionbudget.go
```diff
@@ -107,6 +113,12 @@ func buildResource(etcd *druidv1alpha1.Etcd, pdb *policyv1.PodDisruptionBudget)
 	pdb.Spec.Selector = &metav1.LabelSelector{
 		MatchLabels: druidv1alpha1.GetDefaultLabels(etcd.ObjectMeta),
 	}
+	if metav1.HasAnnotation(etcd.ObjectMeta, annotationAllowUnhealthyPodEviction) {
+		alwaysAllow := policyv1.AlwaysAllow
+		pdb.Spec.UnhealthyPodEvictionPolicy = &alwaysAllow
```
unmarshall, line 118: You can use `utils.PointerOf` instead, then you do not need to declare `alwaysAllow`

shreyas-s-rao: Thanks for the suggestion. Addressed now

unmarshall (approved):
/lgtm
Once the tests are green then feel free to merge

internal/component/poddistruptionbudget/poddisruptionbudget.go
```diff
@@ -107,6 +113,11 @@ func buildResource(etcd *druidv1alpha1.Etcd, pdb *policyv1.PodDisruptionBudget)
 	pdb.Spec.Selector = &metav1.LabelSelector{
 		MatchLabels: druidv1alpha1.GetDefaultLabels(etcd.ObjectMeta),
 	}
+	if metav1.HasAnnotation(etcd.ObjectMeta, annotationAllowUnhealthyPodEviction) {
```
ishan16696, line 116: so, we are going to set `resources.druid.gardener.cloud/allow-unhealthy-pod-eviction` in `.metadata.annotation` but why not in `.spec.annotation` ? I'm asking this as CA `safe-to-evict` annotation for single node etcd-main is present in `.spec.annotation`

ishan16696:
I have tested the PR, it's working fine.
```bash
> k get pdb etcd-main -oyaml | yq .spec 
minAvailable: 2
selector:
  matchLabels:
    app.kubernetes.io/managed-by: etcd-druid
    app.kubernetes.io/part-of: etcd-main
unhealthyPodEvictionPolicy: AlwaysAllow
```

internal/component/poddistruptionbudget/poddisruptionbudget.go
unmarshall: Yes that is another possible place this annotation can be mentioned. However if you look closely then the `spec.annotations` in an `Etcd` resource is used in 2 places - STS component and compaction Job where it is set on the Job resource. If you set this annotation in `spec.annotations` then it also gets set on STS as well which does not make much sense.

Additionally this annotation should not be set directly as in on any resource managed by druid (which is the current purpose of annotations specified via `spec.annotations`). It is therefore correct to set it on etcd resource in `objectmeta.annotations` as it is a indication to druid that if this is set then on the PDB component set `pdb.Spec.UnhealthyPodEvictionPolicy` to `AlwaysAllow`.

ishan16696: make sense. thanks

ishan16696 (approved):
LGTM!!

PR #874, opened 2024-09-06
Fix peer TLS enablement during scale-up of etcd cluster

shreyas-s-rao:

/area quality
/kind bug

The PR fixes a bug in peer TLS enablement during scale-up of etcd cluster, which caused new pods to come up even before first pod had enabled peer TLS for itself. The root cause was the erroneous usage of `druiderrors.Wrap(err,,,)` in `handlePeerTLSChanges()`, where `err` was set to `nil` by the previous `patchOrUpdate(sts)` call, rather than using `druiderrors.New(,,)` in order to retry reconciliation after the default 10 second interval. The PR also simplifies the `handlePeerTLSChanges()` flow.

Thanks @ashwani2k for identifying this bug in the behavior.

The PR also rectifies couple of log messages.

Fixes #

/cc @unmarshall @ashwani2k @ishan16696 

**Release note**:

```other operator
NONE
```


unmarshall (approved):
/lgtm

PR #876, opened 2024-09-09
Fix `make deploy-debug`

shreyas-s-rao:

/area dev-productivity
/kind bug

Fix a bug introduced by #842, which added the `-w` flag to the go build flags. This has now been removed. Additionally, enabling leader election during debugging caused timeouts due to druid not being able to renew its lease while the developer walks through the code via breakpoints. This has now been fixed by removing leader election when deploying druid in debug mode, and setting replicas to `1`.

Fixes #

/cc @unmarshall 

**Release note**:

```other operator
NONE
```


unmarshall (approved):
/lgtm

PR #888, opened 2024-10-08
Add additional latest tag to images during release

anveshreddy18:

/area delivery
/kind enhancement

Add `latest` tag to every image built as part of the release job in the CI pipeline.

This is to consume the helm charts as it is, currently it expects a `latest` tag for the `etcd-druid` image which it cannot find from the `europe-docker` registry as we currently only have `version` specific tags. Adding the `latest` tag would make it better for open-source consumers to quickly test the functionality, as we have seen cases of people reporting the missing tag.

Fixes #

**Release note**:

```other operator
add additional `latest` tag to released images.
```

### DIFF
--- a/.ci/pipeline_definitions
+++ b/.ci/pipeline_definitions
@@ -14,6 +14,7 @@ etcd-druid:
         dockerimages:
           etcd-druid:
             image: europe-docker.pkg.dev/gardener-project/snapshots/gardener/etcd-druid
+            tag_as_latest: True
             dockerfile: 'Dockerfile'
             inputs:
               repos:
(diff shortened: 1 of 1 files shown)


renormalize (approved):
Thanks for this feature which will be very useful @anveshreddy18!
Will save us a lot of pain during testing.

It'd be great if you could reflect these changes on etcd-backup-restore  as well!

anveshreddy18:
> It'd be great if you could reflect these changes on etcd-backup-restore as well!

Sure @renormalize, backup-restore helm charts need to be updated, so will do it then if that's okay.

PR #894, opened 2024-10-23
[hotfix-v0.23] Fixes Label-Selector, Pod Template label updates and TLS changes for client and peer communication (#883)

### LINKED ISSUE #881: Etcd cluster fails to come up when TLS, Pod labels and Replicas are all changed with upgrade from v0.22.x to v0.23.0

/area control-plane
/kind bug

**What happened**:
With etcd-druid version v0.23.0 there is a change to STS Pod labels and label-selector. These are immutable fields and just updating the STS spec would not work as the update will be rejected. A change was introduced in v0.23.0 to orphan delete the STS and then update the STS spec. However if you club this with TLS and replicas change then we ran into a situation where not all members of the etcd cluster could join.

What was observed:

* etcd-test-0 pod gets restarted, but without peer TLS (the pod never gets the new peerTLS spec from sts, so it has no peerTLS secrets mounted at all, to begin with)
* Almost immediately, two new pods etcd-test-1 and etcd-test-2 get spawned, even before etcd-test-0 has become ready, let alone waiting for the peer-tls-enabled annotation on the lease to become true
* The two new pods have peerTLS in their spec (peer TLS secrets are mounted correctly), so they expect etcd-test-0 to have peer TLS enabled. But because etcd-test-0 is still running with peerTLS disabled, the two new pods fail to come up (expected HTTPS response, got HTTP instead)

etcd-cluster should come up and become healthy even if there are simultaneous changes to pod labels, label selector, peer TLS and replicas.

1. checkout etcd-druid v0.22.5
2. run make kind-up and make deploy
3. create an Etcd with replicas=1, no peerTLS enabled, and spec.labels foo=bar  (additional pod labels).
4. annotate with gardener.cloud/operation=reconcile (previous version of druid requires this annotation even upon creation of Etcd resource)
5. wait for reconciliation to succeed and etcd to become ready
6. modify Etcd spec with replicas=3, enable peerTLS, and add a new label foo1=bar1 to spec.labels (trying to simulate changes that gardenlet does, like adding new networking labels to allow peer communication).
7. ensure that the Etcd spec is NOT reconciled (do not annotate with gardener.cloud/operation=reconcile)
8. checkout druid v0.23.0
9. run make deploy again (you might need to run make clean-tools-bin to remove the old skaffold version)
10. once druid v0.23.0 is up, reconcile the Etcd spec using the annotation and observe.

shreyas-s-rao:
* fixes #881
* Added ability to handle unknown CLI args to allow switching between v0.22 and v0.23
* Added use-etcd-wrapper cli arg for etcdbr container
* removed etcd-cluster-size label to be added later with ability to restore while keep etcd.spec.replicas > 1

/area control-plane
/kind bug

Cherry-pick of #883 

Fixes #

**Release note**:

```other user
Fixed the ready condition for the Etcd resource.
```

```other operator
Fixes for handling of pod template labels, label-selector, replicas and TLS changes to Etcd resource. StatefulSet does not allow update of label-selector. v0.23.x changes the label-selector, to get that reflected in the STS, it will be orphan deleted and subsequently created. Similarly for peer TLS and pod label changes an update of pods will be done. For single member etcd clusters this will cause a transient downtime. If replicas, TLS, label-selector are changed together then it will also cause transient quorum loss in multi-node etcd clusters.
```

```noteworthy operator
etcd-backup-restore has been bumped to v0.30.2 and etcd-wrapper has been bumped to v0.2.0.
```

```breaking operator
If you wish to downgrade from druid `v0.23.x` to versions =<`v0.22.7`, please ensure that you change the CLI flags for the druid command to remove the new CLI flags introduced in `v0.23.0`. If you are using the provided helm charts to deploy druid, you may ignore this and simply deploy the helm chart, which takes care of the CLI flag changes for you.
```

```noteworthy operator github.com/gardener/etcd-backup-restore #794 @ishan16696
etcd-backup-restore now triggers a restart of the etcd member after updating etcd's advertise peer URLs if found updated.
```

```noteworthy user github.com/gardener/etcd-backup-restore #794 @ishan16696
Introduced a CLI flag `--use-etcd-wrapper` (default: false) to enable/disable the backup-restore to use etcd-wrapper related functionality. Note: enable this flag only if etcd-wrapper is deployed.
```

```other operator github.com/gardener/etcd-wrapper #31 @ishan16696
Added a capability to stop the etcd-wrapper container by exposing an endpoint `/stop`. 
```

```other developer github.com/gardener/etcd-wrapper #32 @renormalize
Upgrade the Go dependency to `go1.23.1`.
```

```other operator github.com/gardener/etcd-wrapper #18 @unmarshall
ops/print-etcd-cert-paths.sh has been removed and is now replaced with ops/print-etcd-cheatsheet.sh
```

### DIFF
--- a/api/v1alpha1/etcd.go
+++ b/api/v1alpha1/etcd.go
@@ -443,6 +443,8 @@ const (
 	LastOperationStateSucceeded LastOperationState = "Succeeded"
 	// LastOperationStateError indicates that an operation is completed with errors and will be retried.
 	LastOperationStateError LastOperationState = "Error"
+	// LastOperationStateRequeue indicates that an operation is not completed and either due to an error or unfulfilled conditions will be retried.
+	LastOperationStateRequeue LastOperationState = "Requeue"
 )
 
 // LastOperation holds the information on the last operation done on the Etcd resource.
@@ -486,7 +488,8 @@ func (e *Etcd) IsReconciliationInProgress() bool {
 	return e.Status.LastOperation != nil &&
 		e.Status.LastOperation.Type == LastOperationTypeReconcile &&
 		(e.Status.LastOperation.State == LastOperationStateProcessing ||
-			e.Status.LastOperation.State == LastOperationStateError)
+			e.Status.LastOperation.State == LastOperationStateError ||
+			e.Status.LastOperation.State == LastOperationStateRequeue)
 }
 
 // IsDeletionInProgress returns true if the Etcd resource is currently being reconciled, else returns false.
@@ -494,5 +497,6 @@ func (e *Etcd) IsDeletionInProgress() bool {
 	return e.Status.LastOperation != nil &&
 		e.Status.LastOperation.Type == LastOperationTypeDelete &&
 		(e.Status.LastOperation.State == LastOperationStateProcessing ||
-			e.Status.LastOperation.State == LastOperationStateError)
+			e.Status.LastOperation.State == LastOperationStateError ||
+			e.Status.LastOperation.State == LastOperationStateRequeue)
 }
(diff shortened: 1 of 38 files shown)


unmarshall (approved):
/lgtm

PR #895, opened 2024-10-23
Etcd components webhook allows druid to always update its managed resources

shreyas-s-rao:

/area control-plane
/kind enhancement

There are instances when druid's informer cache seems to be stale, and thus it sees older versions of resources. This causes a problem when druid picks up an Etcd resource for reconciliation since it sees that an operation-reconcile annotation has been added, and records the start of the operation on the Etcd status, but this change does not get reflected in the cache immediately, and subsequent update/patch calls to the druid-managed resources such as leases fails, because the etcd-components webhook prevents any updates to these resources unless the Etcd status shows that a reconciliation is in progress (but it doesn't, since the cache is stale at that point).

With this PR, etcd-components webhook allows druid to update its own managed resources at all times, allowing reconciliations to succeed even if the informer cache is stale, drastically reducing the probability of reconciliation failures.

Fixes #

/invite @unmarshall 

**Release note**:

```other operator
Etcd components webhook now allows druid to always update its managed resources, to tackle issues with stale informer cache.
```

### DIFF
--- a/hack/test.sh
+++ b/hack/test.sh
@@ -44,7 +44,7 @@ if ${TEST_COVER:-false}; then
   output_dir=test/output
   coverprofile_file=coverprofile.out
   mkdir -p test/output
-  ginkgo $GINKGO_COMMON_FLAGS --coverprofile ${coverprofile_file} -covermode=set -outputdir ${output_dir} $@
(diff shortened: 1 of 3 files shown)


unmarshall (approved):
/lgtm

PR #995, opened 2025-02-03
CRD Validation for Etcd resource using CEL expressions

### LINKED ISSUE #989: CRD Validation for ETCD
---
/kind enhancement
/area quality

Validation for the fields within the `etcd` resource to be done within `etcd-druid`. This would be done using `x-validations` wherein the validation rules are written using [CEL expressions](https://kubernetes.io/docs/reference/using-api/cel/).

**Why is this needed**:
This would ensure that the validations for etcd are done within etcd-druid. By using CEL expressions, we can validate resource fields without requiring an external `validating webhook`. Since these CEL expressions are compiled and type checked against a CRD's schema when the CRDs are created and updated, Cluster operability improves because CRD validation no longer requires a remote call to a webhook endpoint, eliminating a potential point of failure.

Supersedes #409 . Please check #409  for history of arriving at CEL expressions as a viable solution for validation.
TODO
- [ ] Add CRD validations
- [ ] Add integration tests for each field being validated
- [ ] Add documentation for the list of validations being done

Shreyas-s14:

/area quality
/area control-plane
/area open-source
/kind enhancement

-  Performs validations for the etcd resource via `kubebuilder markers`, using a combination of [CEL Expressions](https://kubernetes.io/docs/reference/using-api/cel/) via the `x-validation` and the `validation` markers.
- Verifies the effectiveness of the validation rules using integration tests.
- To support k8s clusters < 1.29 and >=1.29 it generates 2 versions of CRDs one with CEL expressions and one without.
- `api/core/crds` is changed to now have API methods which takes in a k8s version to provide the appropriate CRD.
- Enhances `make deploy-*` and `make prepare-helm-chart` targets to work with CRDs with and without CEL expressions.
- Updates the documentation to include instructions on how to properly use the CRDs when setting up etcd-druid.
- Adapts IT tests setup to take the correct CRDs.

Fixes #989

**Release note**:

```noteworthy operator
Introduces 'Etcd` custom resource validations using CEL expressions. This will be available for kubernetes clusters with version >= 1.29 only. For kubernetes clusters with version < 1.29, `Etcd` CRD will not contain validations using CEL expressions.
```
```noteworthy developer
Programatic access to the CRDs via the API is available using the `api/core/crds/GetAll()` function, which now accepts k8sVersion as a parameter, in order to return the correct CRD, either with CEL validations for k8s versions v1.29 and above, or without CEL validation for k8s versions below 1.29.
```
```breaking user
The field `etcd.spec.selector` field is now deprecated.
```


gardener-robot-ci-1:
Thank you @Shreyas-s14 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

api/v1alpha1/etcd.go
```diff
@@ -122,13 +122,18 @@ type CompressionSpec struct {
 type LeaderElectionSpec struct {
 	// ReelectionPeriod defines the Period after which leadership status of corresponding etcd is checked.
 	// +optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validationXValidation:message="Invalid duration given for etcd.spec.backup.leaderElection.reelectionPeriod",rule="self.matches('^([0-9]+([.][0-9]+)?h)?([0-9]+([.][0-9]+)?m)?([0-9]+([.][0-9]+)?s)?([0-9]+([.][0-9]+)?d)?$')"
 	ReelectionPeriod *metav1.Duration `json:"reelectionPeriod,omitempty"`
 	// EtcdConnectionTimeout defines the timeout duration for etcd client connection during leader election.
 	// +optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validationXValidation:message="Invalid duration given for etcd.spec.backup.leaderElection.etcdCOnnectionTimeout",rule="self.matches('^([0-9]+([.][0-9]+)?h)?([0-9]+([.][0-9]+)?m)?([0-9]+([.][0-9]+)?s)?([0-9]+([.][0-9]+)?d)?$')"
```
seshachalam-yv, line 131:
```suggestion
	// +kubebuilder:validationXValidation:message="Invalid duration given for etcd.spec.backup.leaderElection.etcdConnectionTimeout",rule="self.matches('^([0-9]+([.][0-9]+)?h)?([0-9]+([.][0-9]+)?m)?([0-9]+([.][0-9]+)?s)?([0-9]+([.][0-9]+)?d)?$')"
```

docs/usage/validating-etcd-clusters.md
```diff
@@ -0,0 +1,35 @@
...
+### Update validations
+These validations are triggered when an update operation is done on the etcd resource.
+* Immutable fields: The fields `etcd.spec.StorageClass` , `etcd.spec.StorageCapacity` and `etcd.spec.VolumeClaimTemplate` are immutable. The immutability is enforced by the CEL expression : `self == oldSelf`.
+
+* The value set for the field `etcd.spec.replicas` can either be decreased to `0` or increased. This is enforced by the CEL expression:
+    `self==0 ? true : self < oldSelf ? false : true`
+
+### Field validations
+- The fields which expect only a particular set of values are checked by using the kubebuilder marker: `+kubebuilder:validation:Enum=<value1>;<value2>`
+    * The `etcd.spec.etcd.metrics` can only be set as either `basic` or `extesnive`.
```
seshachalam-yv, line 24:
```suggestion
    * The `etcd.spec.etcd.metrics` can only be set as either `basic` or `extensive`.
```

test/it/crd-validation/etcd/validation_test.go
```diff
@@ -0,0 +1,849 @@
...
+
+	for _, test := range tests {
+		t.Run(test.name, func (t *testing.T){
+			etcd := utils.EtcdBuilderWithoutDefaults(test.etcdName, testNs).WithReplicas(int32(test.initialReplicas)).Build()
+			cl := itTestEnv.GetClient()
+			ctx := context.Background()
+			g.Expect(cl.Create(ctx, etcd)).To(Succeed())
+
+			etcd.Spec.Replicas = int32(test.updatedReplicas)
+			time.Sleep(30)
```
seshachalam-yv, line 743: Do not use `time.sleep`,  For more detailed guidance, please refer to the
https://github.com/gardener/etcd-druid/blob/6b3b94e2b42c6d0235d6825d3ad689c3fa1c35a9/docs/development/testing.md?plain=1#L25

seshachalam-yv (changes requested):
Thank you, @Shreyas-s14, for adding the validations for the `etcd` CR. I've provided a few initial review comments based on my first glance. I will share more detailed feedback after a thorough review.

docs/usage/validating-etcd-clusters.md
```diff
@@ -0,0 +1,37 @@
+# CRD Validations for etcd
+
+* The validations for the fields within the `etcd` resource are done via [kubebuilder markers for CRD validation](https://book.kubebuilder.io/reference/markers/crd-validation).
```
unmarshall, line 3: Some where in this doc you must mention that this is meant for k8s clusters with version >= 1.30.
What should consumers who are using k8s versions older than 1.30 do? This needs to be clearly mentioned. I think we will in general need to enhance the CRD apis to have a k8s version otherwise with this change etcd-druid can no longer work with k8s clusters with version < 1.30 which is going to be an issue.

So if we wish to have etcd-druid be consumed outside gardener then it is essential that we have some sort of validation for all the CRs. Unfortunately we do not have that today as well for non-gardener use cases. What should be our approach for non-gardener users using k8s version < 1.30? - do we say your CRs will never get validated at the time of submission to KAPI and you will only come to know of any problem during runtime or during provisioning an etcd cluster?

test/it/crd-validation/etcd/specEtcd_test.go
```diff
@@ -0,0 +1,180 @@
+package etcd
```
unmarshall, line 1: camel casing is not used in file names. please correct them

config/crd/bases/crd-druid.gardener.cloud_etcds.yaml
```diff
@@ -0,0 +1,2106 @@
+---
```
unmarshall, line 1: CRDs are no longer present in `config/crd/bases` - please ensure that these are removed

api/core/crds/druid.gardener.cloud_etcds.yaml
```diff
@@ -162,12 +162,13 @@ spec:
                   deltaSnapshotPeriod:
                     description: DeltaSnapshotPeriod defines the period after which
                       delta snapshots will be taken
+                    pattern: ^([0-9]+([.][0-9]+)?h)?([0-9]+([.][0-9]+)?m)?([0-9]+([.][0-9]+)?s)?([0-9]+([.][0-9]+)?d)?$
```
unmarshall, line 165: Have you not introduced CEL expressions for `EtcdCopyBackupsTask` CRD? There should be uniformity in the way all CRs are validated, right?

unmarshall, line 165: As discussed you will have to maintain 2 versions of the CRD - one for k8s versions < 1.30 and another which is for versions >=1.30.
How do we tackle the issue when now someone wishes to make a change to the API? How will you re-generate 2 versions of CRDs - one with CEL and one without?

unmarshall (changes requested):

api/core/crds/druid.gardener.cloud_etcds.yaml
shreyas-s-rao: @unmarshall I would prefer not increasing the scope of this PR now by introducing validations for `EtcdCopyBackupsTask` resources. We never ran validations for those, so we can continue live without validations for another release or two.

To clarify, gardener runs GRM webhook that uses the validation functions provided in [api/validation/etcdcopybackupstask.go](https://github.com/gardener/etcd-druid/blob/master/api/validation/etcdcopybackupstask.go), but for open-source users of etcd-druid, currently there's no validation run for either `Etcd` or `EtcdCopyBackupsTask` resources, sicne druid doesn't run a validating webhook of its own against these resources. So this PR is simply a bonus for open-source users of etcd-druid. And as for gardener, if we choose to retain the the GRM webhook validation for a couple more druid releases, then we are completely safe. And even if we remove GRM webhook validation for k8s versions v1.30 and above, since we know the `EtcdCopyBackupsTask` resource is well-formed in gardener, it should be ok to live without validating these resources for another release or two, and only for k8s clusters running v1.30 or above.

I would rather want focus on https://github.com/gardener/etcd-druid/pull/995#discussion_r1973394481 since that is more pressing right now.

test/it/crd-validation/etcd/specEtcd_test.go
shreyas-s-rao: @Shreyas-s14 additionally, please rename the directory `crdvalidation` instead. I think there's also [recommendation](https://go.dev/doc/effective_go#package-names) to not use hyphens in the directory/file names for go projects, and only use underscores or no separators at all.

api/core/crds/druid.gardener.cloud_etcds.yaml
unmarshall: @Shreyas-s14 can you confirm that you have tested with g/g locally setup that already has a GRM validating webhook enabled? Test it with the following combination:
- CRD with CEL expressions & GRM webhook
- CRD without CEL expressions & GRM webhook

>I would prefer not increasing the scope of this PR now by introducing validations for EtcdCopyBackupsTask resources.

Sure. Please create an issue to add validations for `EtcdCopyBackupsTask` so that this is not forgotten. This should be done before we remove the validating webhooks from GRM.

Shreyas-s14: addressed in https://github.com/gardener/etcd-druid/pull/995/commits/0e5e2797d9e2b47f1d8ec307df6685dc9ef38769

test/it/controller/etcd/reconciler_test.go
```diff
@@ -31,15 +32,28 @@ import (
...
 		err             error
 	)
-	sharedITTestEnv, itTestEnvCloser, err = setup.NewDruidTestEnvironment("etcd-reconciler", []string{assets.GetEtcdCrdPath()})
+	k8sVersion, err := assets.GetK8sVersionFromEnv()
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Failed to get the kubernetes version: %v\n", err)
+		os.Exit(1)
+	}
+
+	k8sVersionAbove129, err = crds.IsK8sVersionAbove129(k8sVersion)
```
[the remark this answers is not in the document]
Shreyas-s14: Decided to go forward with current implementation as discussed.

test/it/crdvalidation/etcd/specetcd_test.go
```diff
@@ -0,0 +1,186 @@
...
+// checks that if the values for etcd.spec.storageCapacity and etcd.spec.etcd.quota are valid, then if backups are enabled, the value of storageCapacity must be > 3x value of quota. If backups are not enabled, value of storageCapacity must be > quota
+func TestValidateSpecStorageCapacitySpecEtcdQuotaRelation(t *testing.T) {
+	skipCELTestsForOlderK8sVersions(t)
+	testNs, g := setupTestEnvironment(t)
+	tests := []struct {
+		name            string
+		etcdName        string
+		storageCapacity resource.Quantity
+		quota           resource.Quantity
+		backup          bool
```
shreyas-s-rao, line 126:
```suggestion
		backupEnabled          bool
```

test/it/crdvalidation/etcd/specsharedconfig_test.go
```diff
@@ -0,0 +1,61 @@
...
+		{
+			name:               "Valid autoCompactionMode #1: revision",
+			etcdName:           "etcd-valid-2",
+			autoCompactionMode: "revision",
+			expectErr:          false,
+		},
+		{
+			name:               "Invalid autoCompactionMode #1: invalid value",
+			etcdName:           "etcd-invalid-1",
+			autoCompactionMode: "frequent",
```
shreyas-s-rao, line 39:
```suggestion
			autoCompactionMode: "dummy",
```

`frequent` actually sounds like a possibly real compaction mode that could be supported any time in the future, so in testing we usually use explicitly garbage/invalid values to indicate that the value is invalid.

test/it/crdvalidation/etcd/specupdate_test.go
```diff
@@ -0,0 +1,216 @@
...
+			ctx := context.Background()
+			g.Expect(cl.Create(ctx, etcd)).To(Succeed())
+
+			etcd.Spec.StorageClass = &test.updatedStorageClassName
+			updateErr := cl.Update(ctx, etcd)
+			if test.expectErr {
+				g.Expect(updateErr).ToNot(BeNil())
+			} else {
+				g.Expect(updateErr).To(BeNil())
+			}
```
shreyas-s-rao, line 63: You can pull this into a separate helper function called `validateEtcdUpdation`, similar to `validateEtcdCreation`. Please add both to a `helper.go` file

test/it/crdvalidation/etcd/specupdate_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	. "github.com/onsi/gomega"
+)
+
+// tests for the Update Validations:
+
+// etcd.spec.storageClass is immutable
+func TestValidateSpecStorageClass(t *testing.T) {
```
shreyas-s-rao, line 21:
```suggestion
func TestValidateUpdateSpecStorageClass(t *testing.T) {
```

Use `TestValidateUpdate` prefix for all update tests, to make it easier to run only update validation tests via `go test` command.

shreyas-s-rao (changes requested):
@Shreyas-s14 thank you very much for the neat PR, and @unmarshall for your support in your help in getting it to work for k8s versions <1.29.

I have a few comments. PTAL, thanks.

docs/deployment/version-compatibility-matrix.md
```diff
@@ -9,6 +9,7 @@ The following is a list of kubernetes versions supported by the respective `etcd
 
 | etcd-druid version | Kubernetes version     |
 |--------------------|------------------------|
+| >=v0.28            | >=v1.29                |
```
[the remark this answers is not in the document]
Shreyas-s14: So what would the exact version be now?

test/it/assets/assets.go
```diff
@@ -0,0 +1,49 @@
...
+func GetK8sVersionFromEnv() (string, error) {
+	k8sVersion, isPresent := os.LookupEnv("ENVTEST_K8S_VERSION")
+	if isPresent {
+		return k8sVersion, nil
+	} else {
+		return "", errors.New("Error fetching k8s version from environment")
+	}
+}
+
+// Duplicating some code here that can be replaced with a call to the API module of etcd-druid.
```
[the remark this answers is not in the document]
Shreyas-s14: As discussed with @unmarshall , the function in the api module to return the CRDS, returns the CRD yaml files as a string. The test builds a test environment based on the `envtest` package's [`Environement struct`](https://github.com/kubernetes-sigs/controller-runtime/blob/52b17917caa97ec546423867d9637f1787830f3e/pkg/envtest/server.go#L111) which can either take in the path to the crds as a slice of `strings` or a slice of [`*apiExtensionsv1.CustomResourceDefinition`](https://github.com/kubernetes-sigs/controller-runtime/blob/52b17917caa97ec546423867d9637f1787830f3e/pkg/envtest/server.go#L143). If we were to use the function from the api module, we would have to then write additional helper functions to generate the `CustomResourceDefinition` objects from the string provided. Given that we plan to increase the minimum required kubernetes version to `1.29` in the near future, the function is duplicated here.

api/core/crds/druid.gardener.cloud_etcds.yaml
```diff
@@ -1822,22 +1840,38 @@ spec:
...
             - replicas
             - selector
             type: object
+            x-kubernetes-validations:
+            - message: If backups are enabled, etcd.spec.storageCapacity must be atleast
+                3x etcd.spec.etcd.quota. If not, it must be greater than etcd.spec.etcd.quota
+              rule: 'has(self.storageCapacity) && has(self.etcd.quota) ? (has(self.backup.store)
+                ? quantity(self.storageCapacity).compareTo(quantity(self.etcd.quota).add(quantity(self.etcd.quota)).add(quantity(self.etcd.quota)))
+                > 0 : quantity(self.storageCapacity).compareTo(quantity(self.etcd.quota))
+                > 0 ): true'
```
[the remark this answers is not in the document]
Shreyas-s14: These are the [currently supported functions by CEL for type `Quantity`](https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-quantity-library).

test/it/assets/assets.go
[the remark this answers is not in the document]
shreyas-s-rao: Understood, thanks for the explanation.

docs/deployment/version-compatibility-matrix.md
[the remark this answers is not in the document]
shreyas-s-rao: We don't really need this line entry here, because the next entry already says that all druid versions equal to or greater than v0.20 will support k8s clusters greater than or equal to v1.21. That is sufficient. You can remove the new entry that you have added. Once we truly get rid of the old CRD without CEL validations, then we udpate this table with the new compatibility.

shreyas-s-rao (approved):
@Shreyas-s14 thanks for making all the requested changes. 
/lgtm

seshachalam-yv:
Thank you, @Shreyas-s14, for addressing my review comments.
/lgtm

seshachalam-yv (approved):
Thank you, @Shreyas-s14, for addressing my review comments.
/lgtm

shreyas-s-rao (approved):
@unmarshall thanks for the fix.
/lgtm

unmarshall (approved):
/lgtm

hack/prepare-chart-resources.sh
```diff
@@ -12,31 +12,124 @@ source $(dirname $0)/openssl-utils.sh
...
-  declare -a crds=("druid.gardener.cloud_etcds.yaml" "druid.gardener.cloud_etcdcopybackupstasks.yaml")
   target_path="${PROJECT_ROOT}/charts/crds"
   echo "Creating ${target_path} to copy the CRDs if not present..."
   mkdir -p ${target_path}
-  for crd in "${crds[@]}"; do
-    local crd_path="${PROJECT_ROOT}/api/core/crds/${crd}"
-    if [ ! -f ${crd_path} ]; then
-      echo -e "CRD ${crd} not found in ${PROJECT_ROOT}/api/core/crds, run 'make generate' first"
+  echo "Cleaning up existing CRDs if any..."
+  rm -rf ${target_path}/*
```
seshachalam-yv, line 118:
```suggestion
  rm -rf "${target_path}"/*
```

seshachalam-yv (changes requested):

hack/prepare-chart-resources.sh
unmarshall: done

api/core/crds/druid.gardener.cloud_etcds.yaml
unmarshall: Raised https://github.com/gardener/etcd-druid/issues/1021

seshachalam-yv (approved):
/lgtm

PR #987, opened 2025-01-29
Introduce new Etcd status condition `AllMembersUpdated`

shreyas-s-rao:
**How to categorize this PR?**

/area control-plane monitoring ops-productivity usability
/kind enhancement api-change

**What this PR does / why we need it**:
This PR introduces a new Etcd status condition `AllMembersUpdated`, to indicate whether all the underlying etcd member pods have been updated with the latest Etcd spec changes after reconciliation. This allows consumers of the Etcd status to know when a rollout of their spec changes has completed.

Additionally, condition statuses `Progressing` and `ConditionCheckError` are now deprecated, and will soon be removed, since druid conditions will be replaced by `metav1.Condition` in the future.

Also, a small change has been made to the etcd components webhook handler, to output the `druid.gardener.cloud/disable-etcd-component-protection` annotation when disallowing an operation on an etcd component managed by druid. This helps users avoid searching through the documentation for the specific annotation to add in order to disable the webhook check.

**Which issue(s) this PR fixes**:
Fixes #985 

**Special notes for your reviewer**:
/invite @unmarshall 
/assign @unmarshall 
/cc @timuthy 

**Release note**:

```noteworthy user
Introduce new Etcd status condition `AllMembersUpdated`.
```
```noteworthy developer
Condition statuses `Progressing` and `ConditionCheckError` are now deprecated and will soon be removed, since druid conditions will be replaced by `metav1.Condition` in the future.
```

PR #996, opened 2025-02-04
Update `observedGeneration` in reconcileStatus flow instead of reconcileSpec flow

### LINKED ISSUE #985: Clients potentially miss `etcd` spec roll outs

/area control-plane
/kind bug

**What happened**:
When adjusting the spec of an `etcd` resource, Gardener waits for the change to be rolled out completely by consulting the `status` (see [CheckEtcdObject](https://github.com/gardener/gardener/blob/0d6120dbbb2d2d55212eed8233e291119224fb59/pkg/component/etcd/etcd/waiter.go#L49)).

Since Etcd-Druid changed to a multi-stage status update (probably in [v0.23.0](https://github.com/gardener/etcd-druid/releases/tag/v0.23.0)), those checks have become racy.

_Please consider the following example (only relevant steps are listed):_

1. The peer CA `Secret` reference is updated.

_Steps in [`reconcileSpec`](https://github.com/gardener/etcd-druid/blob/master/internal/controller/etcd/reconciler.go#L146):_

2. Druid rolls out the changes to the `StatefulSet`.
3. Druid updates `status.observedGeneration`.
4. Druid updates `status.lastOperation` and `status.lastErrors`
5. Druid removes `gardener.cloud/operation` annotation.

_Steps in [`reconcileStatus`](https://github.com/gardener/etcd-druid/blob/master/internal/controller/etcd/reconcile_status.go#L22):_

6. Druid updates `status.ready` and similar fields based on backing `Statefulset`.

Between step 5. and 6. is no hint in the resource that point towards an ongoing rollout of a spec change. Controllers/clients might accidentally continue their operations. This lately happened in the scope of credentials rotation for `gardener/gardener` where the new Peer CA Bundle was not completely rolled out to all replicas, but Gardener already continued with the rotation which led to a certificate mismatch in the etcd cluster.

Clients to know when a spec rollout is successfully finished.

/cc @shreyas-s-rao @LucaBernstein @dguendisch @hendrikKahl

**Environment**:

- etcd-druid version: v0.26.1
- Others:

shreyas-s-rao:

/area usability
/kind impediment

Follow-up to #987 

This PR enhances the reconciliation flow such that `observedGeneration` is updated in `reconcileStatus` flow instead of `reconcileSpec` flow.

Reason: Upon a spec reconciliation request (spec change or `gardener.cloud/operation: reconcile` annotation, depending on the reconcile strategy), the `observedGeneration` was updated in `reconcileSpec` flow, while the other status fields like `conditions`, sts-related fields, etc were updated later in `reconcileStatus` flow. After the user updates the spec, and waits for the spec changes to be reflected in the etcd cluster (which are accurately reflected by conditions), it can so happen that the `observedGeneration` is updated first, followed by a gap before the conditions are updated. This can lead to instances when the user watches the status and assumes that the spec has been fully reconciled and pods are updated with latest changes, because the previous conditions still applied while the observed generation got updated.

To avoid this, the `observedGeneration` will now be updated along with the rest of the status fields, in one single status patch call, so that at any given point of time, the user can check both `status.observedGeneration` and `status.conditions` to know whether the spec changes have been fully rolled out.

Additionally, the operation annotation is conditionally removed at the end of both spec and status reconciliation, to ensure that any failure in either spec or status reconciliation will result in a fresh spec reconciliation upon requeue, and will not get missed. This will ensure that the `observedGeneration` gets correctly updated, even if druid restarts in between reconciliation flow.
 
Fixes #985 

/invite @unmarshall 
/assign @unmarshall 
/cc @timuthy 

**Release note**:

```bugfix user
Updation of `status.observedGeneration` and optional removal of the `gardener.cloud/operation: reconcile` annotation on the Etcd resource are now executed after the reconciliation of the Etcd status, to depict accurate state of the cluster at any given point in time. Users must wait for the `status.observedGeneration` field to be updated (and additionally for the removal of the `gardener.cloud/operation: reconcile` annotation is CLI flag `enable-spec-auto-reconcile` is set to false) to confirm completion of reconciliation.
```

### DIFF
--- a/internal/controller/etcd/reconcile_status.go
+++ b/internal/controller/etcd/reconcile_status.go
@@ -17,9 +17,9 @@ import (
 )
 
 // mutateEtcdStatusFn is a function which mutates the status of the passed etcd object
-type mutateEtcdStatusFn func(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, logger logr.Logger, reconcileSpecResult ctrlutils.ReconcileStepResult) ctrlutils.ReconcileStepResult
+type mutateEtcdStatusFn func(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, logger logr.Logger) ctrlutils.ReconcileStepResult
 
-func (r *Reconciler) reconcileStatus(ctx component.OperatorContext, etcdObjectKey client.ObjectKey, reconcileSpecResult ctrlutils.ReconcileStepResult) ctrlutils.ReconcileStepResult {
+func (r *Reconciler) reconcileStatus(ctx component.OperatorContext, etcdObjectKey client.ObjectKey) ctrlutils.ReconcileStepResult {
 	etcd := &druidv1alpha1.Etcd{}
 	if result := ctrlutils.GetLatestEtcd(ctx, r.client, etcdObjectKey, etcd); ctrlutils.ShortCircuitReconcileFlow(result) {
 		return result
@@ -32,7 +32,7 @@ func (r *Reconciler) reconcileStatus(ctx component.OperatorContext, etcdObjectKe
 		r.mutateObservedGeneration,
 	}
 	for _, fn := range mutateETCDStatusStepFns {
-		if stepResult := fn(ctx, etcd, sLog, reconcileSpecResult); ctrlutils.ShortCircuitReconcileFlow(stepResult) {
+		if stepResult := fn(ctx, etcd, sLog); ctrlutils.ShortCircuitReconcileFlow(stepResult) {
 			return stepResult
 		}
 	}
@@ -43,7 +43,7 @@ func (r *Reconciler) reconcileStatus(ctx component.OperatorContext, etcdObjectKe
 	return ctrlutils.ContinueReconcile()
 }
 
-func (r *Reconciler) mutateETCDStatusWithMemberStatusAndConditions(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, logger logr.Logger, _ ctrlutils.ReconcileStepResult) ctrlutils.ReconcileStepResult {
+func (r *Reconciler) mutateETCDStatusWithMemberStatusAndConditions(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, logger logr.Logger) ctrlutils.ReconcileStepResult {
 	statusCheck := status.NewChecker(r.client, r.config.EtcdMember.NotReadyThreshold, r.config.EtcdMember.UnknownThreshold)
 	if err := statusCheck.Check(ctx, logger, etcd); err != nil {
 		logger.Error(err, "Error executing status checks to update member status and conditions")
@@ -52,7 +52,7 @@ func (r *Reconciler) mutateETCDStatusWithMemberStatusAndConditions(ctx component
 	return ctrlutils.ContinueReconcile()
 }
 
-func (r *Reconciler) inspectStatefulSetAndMutateETCDStatus(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, _ logr.Logger, _ ctrlutils.ReconcileStepResult) ctrlutils.ReconcileStepResult {
+func (r *Reconciler) inspectStatefulSetAndMutateETCDStatus(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, _ logr.Logger) ctrlutils.ReconcileStepResult {
 	sts, err := utils.GetStatefulSet(ctx, r.client, etcd)
 	if err != nil {
 		return ctrlutils.ReconcileWithError(err)
@@ -81,8 +81,11 @@ func (r *Reconciler) inspectStatefulSetAndMutateETCDStatus(ctx component.Operato
 	return ctrlutils.ContinueReconcile()
 }
 
-func (r *Reconciler) mutateObservedGeneration(_ component.OperatorContext, etcd *druidv1alpha1.Etcd, _ logr.Logger, reconcileSpecResult ctrlutils.ReconcileStepResult) ctrlutils.ReconcileStepResult {
-	if r.canReconcileSpec(etcd) && !ctrlutils.ShortCircuitReconcileFlow(reconcileSpecResult) {
+func (r *Reconciler) mutateObservedGeneration(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, _ logr.Logger) ctrlutils.ReconcileStepResult {
+	canReconcileSpec := utils.GetBoolValueOrDefault(ctx.Data, reconciliationContextDataKeyCanReconcileSpec, false)
+	shortCircuitSpecReconcile := utils.GetBoolValueOrDefault(ctx.Data, reconciliationContextDataKeyShortCircuitSpecReconcile, true)
+
+	if canReconcileSpec && !shortCircuitSpecReconcile {
 		etcd.Status.ObservedGeneration = &etcd.Generation
 	}
 	return ctrlutils.ContinueReconcile()
--- a/internal/controller/etcd/reconcile_status.go
+++ b/internal/controller/etcd/reconcile_status.go
@@ -5,12 +5,9 @@
 package etcd
 
 import (
-	"errors"
-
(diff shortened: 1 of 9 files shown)


shreyas-s-rao:
@seshachalam-yv thanks a lot for your inputs. As discussed, I have made the following changes to streamline the reconciliation flow. Changes made in https://github.com/gardener/etcd-druid/pull/996/commits/97282ed2a5fe3b1aab457b744b0203c5766fa8e0:
1. Call `canReconcileSpec` only once, to ensure that only one event is emitted when suspend-reconcile key is detected. Additionally, in the case where a new Etcd resource is created with operation-annotation set, computing `canReconcileSpec` only at the beginning ensures that it returns true, as opposed to recalculating it after updating observedGeneration, in which case it returns false and the operation annotation would never be removed.
2. Use operatorContext to pass information such as result of `canReconcileSpec` and `shortCircuitSpecReconcile` (both bool values), to avoid recalculating these within each of the other reconciliation steps like `reconcileStatus` and `removeOperationAnnotation`
3. Add helper functions with tests, along with some missing tests for other existing helper functions.

internal/controller/etcd/reconcile_status.go
```diff
@@ -79,3 +80,13 @@ func (r *Reconciler) inspectStatefulSetAndMutateETCDStatus(ctx component.Operato
 	}
 	return ctrlutils.ContinueReconcile()
 }
+
+func (r *Reconciler) mutateObservedGeneration(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, _ logr.Logger) ctrlutils.ReconcileStepResult {
+	canReconcileSpec := utils.GetBoolValueOrDefault(ctx.Data, reconciliationContextDataKeyCanReconcileSpec, false)
```
unmarshall, line 85: This looks like a bit of a hack to me and in my initial view is not a very clean design.
- First capturing reconcile spec status data as part of the context.
- Second having `canReconcileSpec` and `shortCircuitSpecReconcile` inside status update flow.

What if we do not update the observed generation as part of `reconcile_status`?

internal/controller/etcd/reconciler.go
```diff
@@ -143,15 +180,20 @@ func (r *Reconciler) reconcileEtcdDeletion(ctx component.OperatorContext, etcdOb
...
 		return result
 	}
-	if r.canReconcileSpec(etcd) {
-		rLog := r.logger.WithValues("etcd", etcdObjectKey, "operation", "reconcileSpec").WithValues("runID", ctx.RunID)
-		ctx.SetLogger(rLog)
-		return r.triggerReconcileSpecFlow(ctx, etcdObjectKey)
+
+	if metav1.HasAnnotation(etcdPartialObjMeta.ObjectMeta, v1beta1constants.GardenerOperation) {
+		ctx.Logger.Info("Removing operation annotation")
+		withOpAnnotation := etcdPartialObjMeta.DeepCopy()
```
unmarshall, line 191: do you wish to rename to `withoutOpAnnotation` because it is on this object you are removing the reconcile operation annotation?

internal/utils/miscellaneous.go
```diff
@@ -81,3 +82,33 @@ func ComputeScheduleInterval(cronSchedule string) (time.Duration, error) {
...
+	result, err := strconv.ParseBool(value)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse boolean value: %w", err)
+	}
+	return result, nil
+}
+
+// GetBoolValueOrDefault returns the boolean value for the given key from the data map,
+// and defaults to defaultValue if the key is not found or the value is not a valid boolean.
+func GetBoolValueOrDefault(data map[string]string, key string, defaultValue bool) bool {
```
unmarshall, line 104: if `parseBoolOrError` returns an error then is it right to use the default value? You are effectively hiding a programming bug.

internal/utils/miscellaneous.go
```diff
@@ -81,3 +82,33 @@ func ComputeScheduleInterval(cronSchedule string) (time.Duration, error) {
...
+	value, err := getValueOrError(data, key)
```
unmarshall, line 105: why should `getValueOrError` exist and why should it return an error if the key is not found?

internal/utils/miscellaneous_test.go
```diff
@@ -0,0 +1,255 @@
...
+import (
+	"testing"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestMergeMaps(t *testing.T) {
```
unmarshall, line 12: i already added unit tests for existing methods in the API module PR. Since you have added this in, i will perhaps remove what i have added then. Lets sync up before investing time so that duplicate efforts can be avoided in future.

unmarshall (changes requested):

internal/controller/etcd/reconciler.go
shreyas-s-rao: I just moved the `removeOperationAnnotation` method from one file to another, and haven't changed anything here. And the naming is still correct. `withOpAnnotation` is the original object, which has the operation annotation. We then modify the `etcdPartialObjMeta`, which will now be the without-op-annotation object. And then we run `r.client.Patch(ctx, etcdPartialObjMeta, client.MergeFrom(withOpAnnotation))`, so this is correct.

internal/utils/miscellaneous.go
shreyas-s-rao: We rely on `strconv.FormalBool()` and `strconv.ParseBool()` to convert between string and bool values, so I'm assuming that we should not be encountering any errors within `GetBoolValueOrDefault()` function. If you want, I can return the error and propagate it back up, which will result in a requeue of reconciliation. WDYT?

internal/utils/miscellaneous.go
shreyas-s-rao: Ok you're right. We don't really need the extra function unnecessarily. I'll remove this.

internal/utils/miscellaneous_test.go
shreyas-s-rao: Sure. I wasn't aware of that, and since I saw missing unit tests, I simply added them here.

internal/utils/miscellaneous.go
shreyas-s-rao: Fixed in https://github.com/gardener/etcd-druid/pull/996/commits/0f9cdd832d1be9e6cc36dcec8c17c1a97b0e12a9

internal/controller/etcd/reconcile_status.go
shreyas-s-rao: I have simplified the operatorContext.Data in https://github.com/gardener/etcd-druid/pull/996/commits/0f9cdd832d1be9e6cc36dcec8c17c1a97b0e12a9. Instead of passing `canReconcileSpec` and `shortCircuitSecReconcile`, I now simply pass `wasSpecReconciled`. If true, then `observedGeneration` must be updated. It's much simpler now.

internal/utils/miscellaneous.go
unmarshall: yes eating away errors is not good and points to a programming error which should surface and be fixed.

shreyas-s-rao: Addressed in https://github.com/gardener/etcd-druid/pull/996/commits/29e8a87f8edb87c35fe9144805e30b68d0cbda8e

internal/controller/etcd/reconcile_status.go
shreyas-s-rao: As discussed in an out-of-band conversation between @unmarshall and myself, we decided to introduce a new `r.completeReconcile()` flow, which executes the steps `updateObservedGeneration()` and `removeOperationAnnotation()`, since this is now semantically complete, and has no adverse effects on the overall flow of the reconciliation or the status fields updation. Any consumer of the Etcd resource will need to wait for `observedGeneration` to be updated (and the removal of operation annotation in case spec auto reconciliation is disabled) to confirm the completion of the Etcd resource reconciliation.

Changes are made in https://github.com/gardener/etcd-druid/pull/996/commits/4d873333e07b0d3ad0bb57101b2e65ccf945930b

unmarshall (approved):
/lgtm

PR #1024, opened 2025-03-10
Change GetAll CRD API to return a Map with keys as CRD name

### LINKED ISSUE #1023: Change CRD API to use CRD name as the key instead of Kind

/area control-plane
/kind enhancement

[GetAll](https://github.com/gardener/etcd-druid/blob/c083042ecbc7be41a1faad5bcce371f393b78236/api/core/crds/crd.go#L33) CRD API returns a `map[string]string`. They key of this map should be changed from `Kind` to `Name of the CRD` instead.

**Why is this needed**:
gardener/gardener programmatically deploys CRDs that are required for etcd-druid to function. For that they use the CRD string, convert it into a byte array and directly deploy the manifest. To delete these CRDs they will need the CRD name. What we pass today is the `Kind` of the CRD and that is not really required.

unmarshall:

/area control-plane
/kind enhancement

Changes the `GetAll` CRD API to return a map with keys being the CRD name instead of Kind.

Fixes #1023 

**Release note**:

```feature developer
Change the GetAll CRD API to return a map with keys having CRD names instead of Kind.
```

### DIFF
--- a/api/core/crds/crd.go
+++ b/api/core/crds/crd.go
@@ -20,16 +20,18 @@ var (
 )
 
 const (
-	// KindEtcd is the kind of the etcd CRD.
-	KindEtcd = "Etcd"
-	// KindEtcdCopyBackupsTask is the kind of the etcd-copy-backup-task CRD.
-	KindEtcdCopyBackupsTask = "EtcdCopyBackupsTask"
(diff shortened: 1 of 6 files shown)


shreyas-s-rao (approved):
@unmarshall thanks for the PR. 
/lgtm

PR #1027, opened 2025-03-11
Remove the direct depedency on `gopkg.in/yaml.v2`.

renormalize:

/area dev-productivity
/kind cleanup

* `gopkg.in/yaml.v2` was a direct dependency only because one file `internal/component/configmap/configmap.go` used it to marshal yaml. This direct depdendency is removed by replacing it with the package `sigs.k8s.io/yaml` used everywhere else in the repository to maintain consistency.

* Replace `yaml` with `json` struct tags since `sigs.k8s.io/yaml` first marshals the `struct` to a json, and then yaml. When `yaml` tags are specified, the specified tags are not used when marshalling to `json`, which causes the intermediate json values to have the keys as the `struct` field names, causing the final yaml also to have the struct field names as the keys.

* Make `omitempty` fields pointers. The general convention for fields which are `omitempty` is to make them pointers. If not, the fields will be their zero values which is not the intended marshalled output.

* Replace C-like for loops with range for loops for readability.

**Release note**:

```other dependency
Remove the direct depedency on `gopkg.in/yaml.v2`.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,6 @@ require (
 	go.uber.org/zap v1.27.0
 	golang.org/x/exp v0.0.0-20241204233417-43b7b7cde48d
 	golang.org/x/tools v0.29.0
-	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.32.1
 	k8s.io/apimachinery v0.32.1
 	k8s.io/client-go v0.32.1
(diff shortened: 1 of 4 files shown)


seshachalam-yv (approved):
Thanks, @renormalize, for replacing the direct dependency with `sigs.k8s.io/yaml`!

/lgtm
/approve

seshachalam-yv (approved):

PR #1032, opened 2025-03-14
Temporarily remove CEL validations for StorageCapacity

### LINKED ISSUE #1031: Remove CEL validations for StorageCapacity

/area control-plane
/kind enhancement

PR #995 introduced CEL validations for Etcd custom resource. All the validations introduced are correct. However validations w.r.t `StorageCapacity` needs to be temporarily removed. 

**Why is this needed**:

Validations on StorageCapacity (it is immutable and should be 3 times the etcd DB size) causes problems when running gardener/gardener e2e KIND tests.

See [test-failure](https://prow.gardener.cloud/view/gs/gardener-prow/pr-logs/pull/gardener_gardener/11545/pull-gardener-e2e-kind-ha-multi-zone-upgrade/1900227201136070656) as an example. The issue is that gardener incorrectly set the StorageCapacity as 10Gi when backup.Store is enabled while the DB size is set to 8Gi. The validation enforces that if the backup is enabled then the storage capacity should be 3 times the size of etcd DB size. This is a requirement from etcd-backup-restore and should be enforced. But all the gardener/gardener upgrade e2e tests fail and we do not want to change the StorageCapacity in the already released version. So for now we will remove these validations. We will make corrections to gardener allowing changing the StorageCapacity and once these are corrected then in the next version of etcd-druid we can re-introduce these validations.

If the user has chosen a storageSize which is less than 3 times the size of etcd DB and this configuration is used in a productive cluster then chances are that they might have already faced disk issues. We chose to include immutability rule for `StorageClass` initially because etcd-druid does not currently support zero-down time volume resizing for HA etcd clusters. Once that is introduced we will permanently remove this validation.

As explained above for short while we will be removing such validations to allow users to configure their `StorageCapacity` with an expected size. Once that is done we will re-introduce these validations and keep them till the time etcd-druid supports zero-down time scaling up of volumes.

unmarshall:

/area control-plane
/kind enhancement

Temporarily removes the CEL validations for `StorageCapacity`.

Fixes #1031 

**Release note**:

```other operator
Temporarily removes CEL validations for StorageCapacity to allow users to correctly configure volume size to be at least 3 times that of etcd DB size.
```

### DIFF
--- a/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
+++ b/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
@@ -1840,9 +1840,6 @@ spec:
                 description: StorageCapacity defines the size of persistent volume.
                 pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                 x-kubernetes-int-or-string: true
-                x-kubernetes-validations:
-                - message: etcd.spec.storageCapacity is an immutable field
-                  rule: self == oldSelf
               storageClass:
                 description: |-
                   StorageClass defines the name of the StorageClass required by the claim.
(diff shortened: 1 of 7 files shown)


shreyas-s-rao (approved):
@unmarshall thanks for addressing my requests. 
/lgtm

PR #1038, opened 2025-03-17
Forbid container privilege escalations for ETCD Druid component containers

georgibaltiev:

/area security
/area compliance
/kind enhancement

This PR sets `securityContext.allowPrivilegeEscalation` to false for every `Seed` and `Garden` cluster component container, which does not have `securityContext.Privileged` set to `true` or one of `CAP_SYS_ADMIN/SYS_ADMIN` capabilities added.

**Which issue(s) this PR 
Part of gardener/gardener#11139

cc @AleksandarSavchev @ialidzhikov 

**Release note**:

```other operator
Component containers, which do not require privilege escalations, now forbid privilege escalation explicitly.
```

### DIFF
--- a/internal/component/statefulset/builder.go
+++ b/internal/component/statefulset/builder.go
@@ -256,9 +256,10 @@ func (b *stsBuilder) getPodInitContainers() []corev1.Container {
 					Args:            []string{fmt.Sprintf("chown -R %d:%d /home/nonroot/%s", nonRootUser, nonRootUser, *b.etcd.Spec.Backup.Store.Container)},
 					VolumeMounts:    []corev1.VolumeMount{*etcdBackupVolumeMount},
 					SecurityContext: &corev1.SecurityContext{
-						RunAsGroup:   ptr.To[int64](0),
-						RunAsNonRoot: ptr.To(false),
-						RunAsUser:    ptr.To[int64](0),
+						AllowPrivilegeEscalation: ptr.To(false),
+						RunAsGroup:               ptr.To[int64](0),
+						RunAsNonRoot:             ptr.To(false),
+						RunAsUser:                ptr.To[int64](0),
 					},
 				})
 			}
--- a/internal/component/statefulset/builder.go
+++ b/internal/component/statefulset/builder.go
@@ -370,7 +370,10 @@ func (b *stsBuilder) getEtcdContainer() corev1.Container {
 				ContainerPort: b.clientPort,
 			},
 		},
-		Resources:    ptr.Deref(b.etcd.Spec.Etcd.Resources, defaultResourceRequirements),
+		Resources: ptr.Deref(b.etcd.Spec.Etcd.Resources, defaultResourceRequirements),
+		SecurityContext: &corev1.SecurityContext{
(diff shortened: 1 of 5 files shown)


anveshreddy18:
@georgibaltiev, thanks for making the change to set the `securityContext.allowPrivilegeEscalation` to `false`. Can you pls resolve the conflicts so we can review and merge the PR.

internal/component/statefulset/builder.go
```diff
@@ -256,9 +256,10 @@ func (b *stsBuilder) getPodInitContainers() []corev1.Container {
 					Args:            []string{fmt.Sprintf("chown -R %d:%d /home/nonroot/%s", nonRootUser, nonRootUser, *b.etcd.Spec.Backup.Store.Container)},
 					VolumeMounts:    []corev1.VolumeMount{*etcdBackupVolumeMount},
 					SecurityContext: &corev1.SecurityContext{
-						RunAsGroup:   ptr.To[int64](0),
-						RunAsNonRoot: ptr.To(false),
-						RunAsUser:    ptr.To[int64](0),
+						AllowPrivilegeEscalation: ptr.To(false),
```
renormalize, line 259: nit: this container is already running as root (is deployed only during tests), don't think setting this flag here would do anything. can this just be removed
```suggestion
```

internal/controller/etcdcopybackupstask/reconciler.go
```diff
@@ -362,9 +365,10 @@ func (r *Reconciler) createJobObject(ctx context.Context, task *druidv1alpha1.Et
 				Args:         []string{fmt.Sprintf("%s%s%s%s", "chown -R 65532:65532 /home/nonroot/", *targetStore.Container, " /home/nonroot/", *sourceStore.Container)},
 				VolumeMounts: volumeMounts,
 				SecurityContext: &corev1.SecurityContext{
-					RunAsGroup:   ptr.To[int64](0),
-					RunAsNonRoot: ptr.To(false),
-					RunAsUser:    ptr.To[int64](0),
+					AllowPrivilegeEscalation: ptr.To(false),
```
renormalize, line 368: https://github.com/gardener/etcd-druid/pull/1038#discussion_r2018154901
```suggestion
```

renormalize (changes requested):
sorry for the slow review @georgibaltiev!

internal/component/statefulset/builder.go
georgibaltiev: As far as I'm aware, when a container process is `privileged`, it indicates that the process in question can run as a `root` on the host. Starting a container process as `root` is not exactly the same - the process can run as root only in the context of the current container (it can't escalate privileges to run as root on the host). Therefore, forbidding container privilege escalations may still be mandatory.
Reference: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#securitycontext-v1-core

renormalize: I was not aware of the difference, thanks for pointing it out!

renormalize (approved):

shreyas-s-rao (approved):
@georgibaltiev thanks for the PR.
/lgtm

PR #1039, opened 2025-03-17
Enhance compaction metrics by segregating them into various categories

### LINKED ISSUE #1037: Enhance compaction Job metrics to expose the reason for Job Failures

/area monitoring
area control-plane
/kind enhancement

At present the metrics exposed by druid regarding compaction jobs are very minimal, i.e it categorises the jobs into just `succeeded` or `failed`, but in reality the `failed` category is vast, it can fail for multitude of reasons with the prominent reasons being 

1. Process Failure
2. Deadline Exceeded
3. Preemption
4. Eviction

There can be multiple sub reasons for `Eviction` like `DeletionByTaintManager`, `EvictionByEvictionAPI`, `TerminationByKubelet`, etc

The reason for why knowing about the failure reason matters is because currently we expose these compaction metrics from druid which gardener project scrapes through prometheus for monitoring and alerting, and we've been seeing a lot of jobs tagged under `failed` category but upon monitoring for some time we've seen that these are not because of the actual process failures, but due to external disruptions and job exceeding the deadline most often. So it's important that we expose the actual reason for the failure instead of just saying it's failed. This way, it becomes extremely clear as to what is causing the failure and the actions that can be taken to minimise such failures. 

This is important not just for gardener project, but for external independent consumers of `etcd-druid` as well as they can scrape the prometheus metrics exposed out of `etcd-druid` for the compactions job if they've enabled them to better monitor and understand the compaction job lifecycles.

**Why is this needed**:

To improve monitoring of the compaction jobs to better understand the reasons for job failures so that necessary targeted action can be taken to minimise such failures.

anveshreddy18:

/area monitoring
/kind enhancement

To improve monitoring of the compaction jobs to better understand the reasons for job failures so that necessary targeted action can be taken to minimise such failures. Pls see the issue #1037 for more details.

The metrics `etcddruid_compaction_jobs_total` and `etcddruid_compaction_job_duration_seconds` have been enhanced with a new label `failureReason` which tells the reason for why the job/pod failed.
The values for the `failureReason` label are the following:
* `preempted` indicates that the compaction pod has been preempted by the scheduler.
* `evicted` indicates that the compaction Pod has been evicted due to various eviction reasons outlined in #1037 
* `deadlineExceeded` indicates that compaction could not finish before the `activeDeadlineSeconds` of the job.
* `processFailure` indicates that the compaction process has failed.
* `unknown` indicates that the failure reason is not known.
* `none` this is used when combined with the {label, value} pair `succeeded`:`true`.

Fixes #1037 

To know the reason for the pod failure, we need to look at some of the pod statuses such as `DisruptionTarget` to see if it's subjected to any disruptions, `ContainerStatuses` to see if the pod is failed because of process failure, etc. For these, we need access to the pod for some sufficient time, so I've set `terminationGracePeriodSeconds` for the pod to `60s` to give druid enough time to fetch that before it is garbage collected. 

Connected to the above, there has been a change in how the status conditions for the job is populated in k8s `v1.31.0` with the change being that the `Failed` condition on the job will now be only added after the termination of all the pods instead of adding it as soon as the pod goes into `terminating` state. So, basically when the pod is in `terminating` state waiting for the `terminationGracePeriodSeconds`, only the `FailureTarget` condition is set initially, and later the `Failed` condition is also added once the pod is terminated. Due to this difference in behaviour between the recent k8s versions, I chose to consider both the `FailureTarget` and `Failed` conditions as indication of failure [here](https://github.com/gardener/etcd-druid/blob/bf2fa979123b83dc3f9fa6d1cb5f5b798ede56a3/internal/controller/compaction/reconciler.go#L430) so that it works for both versions smoothly. 

Changelog for the k8s `v1.32.0` detailing the above mentioned change. Read more [here](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.31.md)
```
Delay setting terminal Job conditions until all pods are terminal.
  
  Additionally, the FailureTarget condition is also added to the Job object in the first Job
  status update as soon as the failure conditions are met (backoffLimit is exceeded, maxFailedIndexes, 
  or activeDeadlineSeconds is exceeded).
  
  Similarly, the SuccessCriteriaMet condition is added in the first update as soon as the expected number
  of pod completions is reached.
  
  Also, introduce the following validation rules for Job status when JobManagedBy is enabled:
  1. the count of ready pods is less or equal than active
  2. when transitioning to terminal phase for Job, the number of terminating pods is 0
  3. terminal Job conditions (Failed and Complete) should be preceded by adding the corresponding interim conditions: FailureTarget and SuccessCriteriaMet ([#125510](https://github.com/kubernetes/kubernetes/pull/125510), [@mimowo](https://github.com/mimowo)) [SIG Apps and Testing]
```

**Release note**:

```noteworthy user
compaction job metrics are now enhanced with the new label `failureReason` to query the reason for failure job failures.
```


internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+
+func handleBackoffLimitExceeded(ctx context.Context, r *Reconciler, logger logr.Logger, job *batchv1.Job, etcd *druidv1alpha1.Etcd) error {
+	logger.Info("Job has been completed due to backoffLimitExceeded", "namespace", job.Namespace, "name", job.Name)
+	pod, err := getPodForJob(ctx, r.Client, logger, job)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			logger.Error(err, "Couldn't fetch pods for job", "namespace", job.Namespace, "name", job.Name)
+			return fmt.Errorf("error while fetching pod for job: %v", err)
+		}
+		logger.Info("No pods found for job", "namespace", job.Namespace, "name", job.Name)
```
unmarshall, line 386: if no pod is found why are you not returning and instead you are continuing to call `getPodFailureReasonAndLastTransitionTime` function?

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+	}
+	podFailureReason, lastTransitionTime := getPodFailureReasonAndLastTransitionTime(pod)
```
unmarshall, line 388: variable is shadowing the type alias defined above.

internal/controller/compaction/reconciler.go
```diff
@@ -34,13 +34,22 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
+type podFailureReason string
```
unmarshall, line 37: is there a reason to define a type alias instead of just defining string constants?
if the reason is code readability then you can name the constants properly.
For example:
```go
const (
  podFailureReasonDeletionByTaintManager = "DeletionByTaintManager"
)
```
Then just use the constants like `v1.PodReasonPreemptionByScheduler`, `v1.PodReasonTerminationByKubelet` that are defined in upstream k8s.

unmarshall: Also create a different constants section for these constants.

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+	pod, err := getPodForJob(ctx, r.Client, logger, job)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			logger.Error(err, "Couldn't fetch pods for job", "namespace", job.Namespace, "name", job.Name)
+			return fmt.Errorf("error while fetching pod for job: %v", err)
+		}
+		logger.Info("No pods found for job", "namespace", job.Namespace, "name", job.Name)
+	}
+	podFailureReason, lastTransitionTime := getPodFailureReasonAndLastTransitionTime(pod)
+	handlePodFailureReason(logger, pod, etcd, podFailureReason, lastTransitionTime, job.Status.StartTime.Time)
```
unmarshall, line 389: All this function does is record a metric. Why can it not have a more appropriate name?

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+		logger.Info("Pod has failed due to process failure", "namespace", pod.Namespace, "name", pod.Name)
+		labelValue = druidmetrics.ValueSucceededFalse
+	default:
+		logger.Info("Pod has failed due to unknown reason", "namespace", pod.Namespace, "name", pod.Name)
+		labelValue = druidmetrics.ValueSucceededUnknown
+	}
+
+	metricJobsTotal.With(prometheus.Labels{druidmetrics.LabelSucceeded: labelValue, druidmetrics.EtcdNamespace: etcd.Namespace}).Inc()
+	if !lastTransitionTime.IsZero() {
+		metricJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: labelValue, druidmetrics.EtcdNamespace: etcd.Namespace}).Observe(lastTransitionTime.Sub(jobStartTime).Seconds())
```
unmarshall, line 412: Again, not sure why `druidmetrics.LabelSucceeded` is used here to record failure reasons?

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
 }
 
+func handleFailedJob(ctx context.Context, r *Reconciler, logger logr.Logger, job *batchv1.Job, etcd *druidv1alpha1.Etcd, jobCompletionReason string) error {
+	switch jobCompletionReason {
+	case batchv1.JobReasonDeadlineExceeded:
+		logger.Info("Job has been completed due to deadline exceeded", "namespace", job.Namespace, "name", job.Name)
+		metricJobsTotal.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededDeadlineExceeded, druidmetrics.EtcdNamespace: etcd.Namespace}).Inc()
+		metricJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededDeadlineExceeded, druidmetrics.EtcdNamespace: etcd.Namespace}).Observe(float64(*job.Spec.ActiveDeadlineSeconds))
+	case batchv1.JobReasonBackoffLimitExceeded:
+		if err := handleBackoffLimitExceeded(ctx, r, logger, job, etcd); err != nil {
```
unmarshall, line 363: This is the parent method where except in one case of `batchv1.JobReasonBackoffLimitExceeded` in other cases the metrics are recorded in the same method. It becomes easy if you can compute the metrics values elsewhere but record it here instead of doing that finally in `handlePodFailureReason`

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+	}
+
+	metricJobsTotal.With(prometheus.Labels{druidmetrics.LabelSucceeded: labelValue, druidmetrics.EtcdNamespace: etcd.Namespace}).Inc()
+	if !lastTransitionTime.IsZero() {
+		metricJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: labelValue, druidmetrics.EtcdNamespace: etcd.Namespace}).Observe(lastTransitionTime.Sub(jobStartTime).Seconds())
+	}
+}
+
+// getJobCompletionStatusAndReason returns whether the job is successful or not and the reason for the completion.
+func getJobCompletionStatusAndReason(job *batchv1.Job) (bool, string) {
```
unmarshall, line 417: The name of the method indicates that you will return a Job completion status and a reason but what you return is  a bool and a reason. In the method you only handle successful and failed completion. If none of the conditions are set then you by default return false indicating its a failure but that is not true. It could also be suspended.

unmarshall (changes requested):

internal/metrics/metrics.go
```diff
@@ -14,6 +14,14 @@ const (
 	ValueSucceededTrue = "true"
 	// ValueSucceededFalse is value False for metric label failed.
 	ValueSucceededFalse = "false"
+	// ValueSucceededPreempted is value Preempted for metric label preempted.
+	ValueSucceededPreempted = "preempted"
```
unmarshall, line 18: These label values make no sense. How can you have succeeded with a failure reason?

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+	}
+	// If the DisruptionTarget condition is not present, then check the container status
+	if pod.Status.ContainerStatuses != nil {
+		for _, containerStatus := range pod.Status.ContainerStatuses {
+			if containerStatus.State.Terminated != nil {
+				return podReasonProcessFailure, containerStatus.State.Terminated.FinishedAt.Time
+			}
+		}
+	}
+	return podReasonUnknown, time.Time{}
```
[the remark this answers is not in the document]
anveshreddy18: Thanks, I've now modified it to report UTC time. Pls check

internal/controller/compaction/reconciler.go
```diff
@@ -131,36 +140,22 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
...
-			if job.Status.CompletionTime != nil {
-				metricJobDurationSeconds.With(prometheus.Labels{druidmetrics.LabelSucceeded: druidmetrics.ValueSucceededTrue, druidmetrics.EtcdNamespace: etcd.Namespace}).Observe(job.Status.CompletionTime.Time.Sub(job.Status.StartTime.Time).Seconds())
-			}
-			if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
-				logger.Error(err, "Couldn't delete the successful job", "namespace", etcd.Namespace, "name", compactionJobName)
-				return ctrl.Result{
-					RequeueAfter: 10 * time.Second,
-				}, fmt.Errorf("error while deleting successful compaction job: %v", err)
+		metricJobsCurrent.With(prometheus.Labels{druidmetrics.EtcdNamespace: etcd.Namespace}).Set(0)
+		isJobSuccessful, jobCompletionReason := getJobCompletionStatusAndReason(job)
```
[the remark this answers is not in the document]
anveshreddy18: Done

internal/controller/compaction/reconciler.go
```diff
@@ -131,36 +140,22 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
...
+		if !isJobSuccessful {
+			if err := handleFailedJob(ctx, r, logger, job, etcd, jobCompletionReason); err != nil {
```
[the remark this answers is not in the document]
anveshreddy18: added these suggested functions to record the metrics. Pls check

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
 	return job, nil
 }
 
+func handleFailedJob(ctx context.Context, r *Reconciler, logger logr.Logger, job *batchv1.Job, etcd *druidv1alpha1.Etcd, jobCompletionReason string) error {
```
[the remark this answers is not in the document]
anveshreddy18: I've now used `druidmetrics.LabelSucceeded` to record the status and `druidmetrics.LabelFailureReason` to record the reason for failure.

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +353,129 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+
+	podList := &v1.PodList{}
+	if err := cl.List(ctx, podList, &client.ListOptions{
+		Namespace:     job.Namespace,
+		LabelSelector: selector,
+	}); err != nil {
+		return nil, fmt.Errorf("error while fetching pods for job: %v", err)
+	}
+
+	if len(podList.Items) == 0 {
```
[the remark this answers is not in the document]
anveshreddy18: I removed the `NotFound` error now.

internal/controller/compaction/reconciler.go
anveshreddy18: Pls check now.

internal/controller/compaction/reconciler.go
anveshreddy18: Removed the type and placed the constants in a different const block.

internal/controller/compaction/reconciler.go
anveshreddy18: https://github.com/gardener/etcd-druid/pull/1039#discussion_r2013481429

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +374,144 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+		podFailureReasonLabelValue = druidmetrics.ValueFailureReasonProcessFailure
+	default:
+		logger.Info("Pod has failed due to unknown reason", "namespace", pod.Namespace, "name", pod.Name)
+		podFailureReasonLabelValue = druidmetrics.ValueFailureReasonUnknown
+	}
+	return podFailureReasonLabelValue, lastTransitionTime, nil
+}
+
+// isJobSuccessfulWithReason returns whether the job is successful or not and the reason for the completion.
+func isJobSuccessfulWithReason(job *batchv1.Job) (bool, string) {
```
unmarshall, line 464: As discussed introduce the following instead:

```go

const (
	JobSucceeded int = iota
	JobFailed
)

func getJobCompletionStateAndReason(job *batchv1.Job) (int, string) {
	jobConditions := job.Status.Conditions
	for _, condition := range jobConditions {
		if (condition.Type == batchv1.JobComplete || condition.Type == batchv1.JobSuccessCriteriaMet) && condition.Status == v1.ConditionTrue {
			return JobSucceeded, condition.Reason
		}
		if (condition.Type == batchv1.JobFailed || condition.Type == batchv1.JobFailureTarget) && condition.Status == v1.ConditionTrue {
			return JobFailed, condition.Reason
		}
	}
	return JobFailed, "" // the control will never reach here. But since the function signature requires a return value, this is added.
}

```

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +374,144 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+	if job.Status.CompletionTime != nil {
+		metricJobDurationSeconds.With(prometheus.Labels{
+			druidmetrics.LabelSucceeded:     druidmetrics.ValueSucceededTrue,
+			druidmetrics.LabelFailureReason: druidmetrics.ValueFailureReasonNone,
+			druidmetrics.LabelEtcdNamespace: job.Namespace,
+		}).Observe(job.Status.CompletionTime.Time.Sub(job.Status.StartTime.Time).Seconds())
+	}
+}
+
+func recordFailedJobMetrics(ctx context.Context, r *Reconciler, logger logr.Logger, job *batchv1.Job, etcd *druidv1alpha1.Etcd, jobCompletionReason string) error {
```
unmarshall, line 392: instead of passing reconciler as a function arg, please make this function as a method on `Reconciler`

internal/controller/compaction/reconciler.go
```diff
@@ -359,6 +374,144 @@ func (r *Reconciler) createCompactionJob(ctx context.Context, logger logr.Logger
...
+	var failureReasonLabelValue string
```
unmarshall, line 393: collect this into a single var block

unmarshall, line 393: rename this to `failureReason`

internal/controller/compaction/reconciler.go
```diff
@@ -126,44 +135,46 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
...
+		if err := r.Delete(ctx, job, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil {
+			logger.Error(err, "Couldn't delete the completed compaction job", "namespace", etcd.Namespace, "name", compactionJobName)
 			return ctrl.Result{
-				RequeueAfter: 10 * time.Second,
-			}, nil
+				RequeueAfter: 10 * time.Second}, fmt.Errorf("error while deleting the completed compaction job: %w", err)
 		}
 	}
 
+	jobMeta := &metav1.ObjectMeta{
```
unmarshall, line 160: can you add a comment stating the reason to loop up the pod. This will make things clear as it was never done before and is due to addition of graceful termination seconds defined in the PodSpec of the Job.

unmarshall (changes requested):

internal/controller/compaction/reconciler.go
anveshreddy18: Done.

internal/controller/compaction/register_test.go
```diff
@@ -216,6 +219,350 @@ func TestEtcdCompactionAnnotation(t *testing.T) {
 	g.Expect(compactionAnnotation).To(Equal(test1EtcdAnnotation))
 }
 
+func TestJobCompletionStatusAndReason(t *testing.T) {
```
unmarshall, line 222: why is this test here? Does it test anything in `register.go`?

internal/controller/compaction/register_test.go
```diff
@@ -216,6 +219,350 @@ func TestEtcdCompactionAnnotation(t *testing.T) {
...
+				},
+			}
+			status, reason := getJobCompletionStateAndReason(job)
+			g.Expect(status).To(Equal(test.expectedStatus))
+			g.Expect(reason).To(Equal(test.expectedReason))
+		})
+	}
+}
+
+func TestPodFailureReasonAndLastTransitionTime(t *testing.T) {
```
unmarshall, line 337: why is this test here? Does it test anything in register.go?

internal/controller/compaction/register_test.go
```diff
@@ -216,6 +219,350 @@ func TestEtcdCompactionAnnotation(t *testing.T) {
...
+				},
+			}
+			reason, lastTransitionTime := getPodFailureReasonAndLastTransitionTime(pod)
+			g.Expect(reason).To(Equal(test.expectedReason))
+			g.Expect(lastTransitionTime).To(BeTemporally("~", test.expectedTransitionTime, time.Second))
+		})
+	}
+}
+
+func TestPodForJob(t *testing.T) {
```
unmarshall, line 464: why is this test here? Does it test anything in register.go?

unmarshall (changes requested):

unmarshall (approved):
Looks good. Thanks for addressing the review comments.

PR #1041, opened 2025-03-18
Upgraded to golang 1.24 and now using first class support for tools

### LINKED ISSUE #1040: Upgrade golang version to 1.24

/area dev-productivity
/kind enhancement

* Upgrade golang version to 1.24
* Use support for tools in go.mod and replace empty imports in `hack/tools.go`

unmarshall:

/area dev-productivity
/kind enhancement

* Upgrades go.mod across all go modules to use version to 1.24
* Changes the pipeline definition file to use go 1.24.1
* Changes Dockerfile to use go 1.24 in the builder section
* Removed the need for `tools.go` from within `hack` directory in parent and API go module. Instead using `go get -tool <dependency-name>@<dependency-version` to add the dependency to go.mod under `tools` section.

Fixes #1040

**Release note**:

```other operator
Upgraded to golang 1.24
Removed the need for `hack/tools.go` and instead using first class support for tools in go.mod
```


unmarshall:
Since there is a change in `pipeline_definitions` the tests will fail. I have run - unit, integration and e2e tests and all pass. Please run them locally to double check.

renormalize (changes requested):
`make format` (and therefore, `make check`) fails locally since `goimports-reviser` is too old.
Changing the version to `v3.9.x` (`v3.9.1` being the latest) which adds support for `go1.24` will fix this error.

`make check` will also fail with `golangci-lint` errors. This can be solved by upgrading the `golangci-lint` version to `v1.64.8`.

Should we first upgrade the `.ci/pipeline_definitions` file in a separate PR so all tests run correctly in this PR?

seshachalam-yv (commented):
I encountered the following error while running make check after updating my environment. 

```bash
(⎈|N/A:N/A)➜  etcd-druid git:(go124) ✗ make check
> Format
go.mod:142: unknown block type: tool

Usage of goimports-reviser:
  -apply-to-generated-files
    	Apply imports sorting and formatting(if the option is set) to generated files. Generated file is a file with first comment which starts with comment '// Code generated'. Optional parameter.
  -company-prefixes string
    	Company package prefixes which will be placed after 3rd-party group by default(if defined). Values should be comma-separated. Optional parameters.
  -excludes string
    	Exclude files or dirs, example: '.git/,proto/*.go'.
  -file-path string
    	Deprecated. Put file name as an argument(last item) of command line.
  -format
    	Option will perform additional formatting. Optional parameter.
  -imports-order string
    	Your imports groups can be sorted in your way. 
    	std - std import group; 
    	general - libs for general purpose; 
    	company - inter-org or your company libs(if you set '-company-prefixes'-option, then 4th group will be split separately. In other case, it will be the part of general purpose libs); 
    	project - your local project dependencies;
    	blanked - imports with "_" alias;
    	dotted - imports with "." alias.
    	Optional parameter. (default "std,general,company,project")
  -list-diff
    	Option will list files whose formatting differs from goimports-reviser. Optional parameter.
  -local string
    	Deprecated
  -output string
    	Can be "file", "write" or "stdout". Whether to write the formatted content back to the file or to stdout. When "write" together with "-list-diff" will list the file name and write back to the file. Optional parameter. (default "file")
  -project-name string
    	Your project name(ex.: github.com/incu6us/goimports-reviser). Optional parameter.
  -recursive
    	Apply rules recursively if target is a directory. In case of ./... execution will be recursively applied by default. Optional parameter.
  -rm-unused
    	Remove unused imports. Optional parameter.
  -set-alias
    	Set alias for versioned package names, like 'github.com/go-pg/pg/v9'. In this case import will be set as 'pg "github.com/go-pg/pg/v9"'. Optional parameter.
  -set-exit-status
    	set the exit status to 1 if a change is needed/made. Optional parameter.
  -use-cache
    	Use cache to improve performance. Optional parameter.
make: *** [format] Error 1

```

unmarshall:
@seshachalam-yv  have you upgraded golang to 1.24?

seshachalam-yv:
> @seshachalam-yv have you upgraded golang to 1.24?

Yes

unmarshall:
@seshachalam-yv please check now. As @renormalize suggested bumping the versions for `golangci-lint` and `goimports-reviser` did the trick

renormalize (approved):

unmarshall:
> Should we first upgrade the .ci/pipeline_definitions file in a separate PR so all tests run correctly in this PR?

Some things like removal of `tools.go`, including tooling dependencies have to go together. What remains is not much. So i would recommend that we verify it locally

unmarshall:
I have also raised https://github.com/gardener/ci-infra/pull/3410 to update the images that used by etcd-druid prow jobs

seshachalam-yv (approved):
/lgtm

PR #1043, opened 2025-03-19
Add nodeSelector, tolerations, affinity & topologySpreadConstraints to charts

Hirostil:
**How to categorize this PR?**

/area robustness
/kind enhancement

**What this PR does / why we need it**:
This PR add to the `Etcd-druid` `Deployment` helm charts the possibility to add `nodeSelector`, `affinity`, `tolerations` & `topologySpreadConstraints`

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
feature operator
Add possibility to add `nodeSelector`, `affinity`, `tolerations` & `topologySpreadConstraints` on the helm charts
```

PR #1044, opened 2025-03-21
Fix charts by placing `volumeMounts` in the `containers` section of the pod spec.

renormalize:

/area delivery
/kind regression

#1043 caused a regression in the chart rendering. This PR fixes it.
`volumeMounts` must be present in the `containers` section of the pod spec. The charts currently render it under `nodeSelector`.

**Release note**:

```other operator
NONE
```


anveshreddy18 (approved):
Thanks for the fix @renormalize.
lgtm!

PR #1046, opened 2025-03-27
Do not specify a value for `nodeSelector` in `charts/values.yaml`.

renormalize:

/area delivery open-source
/kind cleanup

* `nodeSelector` was specified to be `kubernetes.io/os: linux` after #1043. However, this is not really general, and therefore is being removed `charts/values.yaml`. Consumers must specify their own `nodeSelector`, since `charts/values.yaml` is a very minimal example of configuration for etcd-druid.

cc @shreyas-s-rao

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -11,8 +11,7 @@ resources:
     cpu: 50m
     memory: 128Mi
 
-nodeSelector:
-  kubernetes.io/os: linux
+nodeSelector: {}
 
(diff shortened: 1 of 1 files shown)


renormalize:
@Hirostil Please specify `nodeSelector` in your `values.yaml` for your deployment of etcd-druid. Thanks.

PR #1053, opened 2025-04-04
CEL expression to invalidate setting/unsetting of immutable fields.

### LINKED ISSUE #1048: CEL transition rules do not apply to optional fields during set/unset

/area control-plane
/area robustness
/kind bug

**What happened**:

https://github.com/gardener/etcd-druid/blob/7818d2bc86be30ae9e812709b04340ae39d55b98/api/core/v1alpha1/etcd.go#L316-L327

has CEL validations like `self == oldSelf` defined. However, as can be seen in https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md#transition-rules, these transition rules do not apply for setting and unsetting optional fields.

Therefore, it is currently the case that we are able to unset a field which is considered immutable post creation, and then reset it to something different.

Such changes would obviously cause errors during reconciliation since the corresponding fields in the statefulset are immutable, but etcd-druid does not disallow them, as was the intent in #995.

Fields marked as immutable, must be so.

- Create an etcd cluster.
- Unset an "immutable" field like storage class, reconcile the etcd.
- Set it to something different, and then reconcile the etcd.

**Environment**:

- etcd-druid version: v0.28.0

Shreyas-s14:

/area control-plane
/kind bug

1) Added immutability rule to prevent setting/unsetting the immutable fields in `etcd.Spec` after creation. This was overlooked in the previous [PR for CEL Validation](https://github.com/gardener/etcd-druid/pull/995).
2) Added integration tests for the same. 

Fixes #1048 

**Release note**:

```other operator
Added CEL expression to prevent unsetting and setting of immutable fields in etcd.Spec after creation.
```

### DIFF
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -288,6 +288,8 @@ type SchedulingConstraints struct {
 }
 
 // EtcdSpec defines the desired state of Etcd
+// +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
 type EtcdSpec struct {
 	// selector is a label query over pods that should match the replica count.
 	// It must match the pod template's labels.
--- a/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
+++ b/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
@@ -1861,6 +1861,11 @@ spec:
             - labels
             - replicas
             type: object
+            x-kubernetes-validations:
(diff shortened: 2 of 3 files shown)


gardener-robot-ci-2:
Thank you @Shreyas-s14 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

shreyas-s-rao (approved):
@Shreyas-s14 thanks for the quick fix.
/lgtm 
/reviewed ok-to-test

api/core/v1alpha1/etcd.go
renormalize, line 1: The invalid CEL expressions in lines 321 and 328 were not removed. Please remove them.

renormalize (changes requested):

api/core/v1alpha1/etcd.go
Shreyas-s14: Those rules are still needed to prevent modifications to the fields if set initially. The new rule is just a check to make sure the fields aren't set/unset once the CRD is created.

renormalize (approved):

PR #1070, opened 2025-04-11
Re-introduce `/scale` subresource for Etcd CRD

### LINKED ISSUE #1069: Add scale subresource for Etcd custom resource

/area control-plane
/area open-source
/kind enhancement

We would like to add a `scale` subresource for `Etcd` custom resource. For this the following needs to be done:
* Add a kubebuilder tag on the `Etcd` type for scale subresource
```
// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=.status.selector
```
* Add a field for selector in `Etcd.Status`
```go
type EtcdStatus struct {
  Selector string
}
```

**Why is this needed**:

As of now in gardener clusters, VPA resource has been defined targeting StatefulSet that is created by the etcd reconciler for every `Etcd` resource.  We would like to change that and allow setting `vpa.Spec.TargetRef` directly to an`Etcd` resource itself. 

This has the following benefits:
* How an etcd cluster is realised is an implementation choice made by the etcd reconciler. Therefore having consumers depend upon STS exposes this fact and makes it harder to change the choice of using STS in future.
* Introducing `scale` subresource has a side-effect that it also now opens up possibility for an operator to execute [kubectl scale](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_scale/) to increase the replicas for an `Etcd` resource. If auto-reconciliation has been enabled then any change of replicas using `kubectl scale` command will get reflected immediately. If auto-reconciliation has been disabled then it will require an additional [reconcile annotation](https://github.com/gardener/etcd-druid/blob/48ecf9b109e906a29549147ff49cd4794888bf6a/api/core/v1alpha1/constants.go#L32) to be put on an `Etcd` resource. In case auto-reconciliation has been enabled then its quite convenient to use `kubectl scale` to change etcd-cluster replicas (We currently recommend to switch off auto-reconciliation for productive use cases to have better control of applying any changes (explicit or accidental) to `Etcd` resource during maintenance period as rolling update strategy of STS can potentially cause a transient downtime for quorum based applications.)

shreyas-s-rao:

/area scalability auto-scaling
/kind enhancement

Re-introduce `/scale` subresource for `Etcd` CRD, to enable usage of VPA to target Etcd resources instead of underlying statefulset resource.

Fixes #1069 

Can be tested using the following steps (pre-requisites: kubectl, helm): 
```console
# deploy etcd-druid
make kind-up
make deploy

# deploy Etcd resource, which should be reconciled immediately
kubectl apply -f examples/etcd/druid_v1alpha1_etcd.yaml

# check scale subresource 
kubectl get etcd etcd-test --subresource scale -oyaml

# -------- test vertical pod autoscaling --------

# install metrics server
kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml

# install VPA
helm repo add cowboysysop https://cowboysysop.github.io/charts/
helm install <release-name> cowboysysop/vertical-pod-autoscaler

# deploy VPA object for the Etcd resource
cat <<EOF | kubectl apply -f -
apiVersion: autoscaling.k8s.io/v1
kind: VerticalPodAutoscaler
metadata:
  name: etcd-test
spec:
  resourcePolicy:
    containerPolicies:
    - containerName: etcd
      controlledValues: RequestsOnly
      minAllowed:
        memory: 100M
      mode: Auto
    - containerName: backup-restore
      controlledValues: RequestsOnly
      mode: "Off"
  targetRef:
    kind: Etcd
    name: etcd-test
  updatePolicy:
    evictionRequirements:
    - changeRequirement: TargetHigherThanRequests
      resources:
      - memory
      - cpu
    updateMode: Auto
EOF

# check vpa-updater logs to confirm that VPA sees the etcd pods and updates their resources accordingly
# check for etcd pod restarts caused by VPA updates, where pod resources are set to the
# target recommendations from the VPA object status
```
/invite @unmarshall 

**Release note**:

```noteworthy user
Re-introduce `/scale` subresource for Etcd CRD.
```

### DIFF
--- a/docs/usage/managing-etcd-clusters.md
+++ b/docs/usage/managing-etcd-clusters.md
@@ -49,36 +49,36 @@ This will open up the linked editor where you can make the edits.
 
 ### Scale the Etcd cluster horizontally
 
-To scale the etcd cluster horizontally, you can update the `spec.replicas` field in the `Etcd` custom resource. For example, to scale the etcd cluster to 5 replicas, you can run:
+To scale an etcd cluster horizontally, you can update the `spec.replicas` field in the `Etcd` custom resource. For example, to scale an etcd cluster to 5 replicas, you can run:
 
 ```bash
 kubectl patch etcd <etcd-name> -n <namespace> --type merge -p '{"spec":{"replicas":5}}'
 ```
 
-Additionally, you can use the `/scale` subresource to scale the etcd cluster horizontally. For example, to scale the etcd cluster to 5 replicas, you can run:
+Alternatively, you can use the `/scale` subresource to scale an etcd cluster horizontally. For example, to scale an etcd cluster to 5 replicas, you can run:
 
 ```bash
 kubectl scale etcd <etcd-name> -n <namespace> --replicas=5
 ```
 
 !!! note
-    While the Etcd cluster can be scaled out, it cannot be scaled in, ie, the replicas cannot be decreased to a non-zero value. The Etcd cluster can still be scaled to 0 replicas, indicating that the cluster is to be "hibernated". This is beneficial for use-cases where the etcd cluster is not needed for a certain period of time, and the user does not want to pay for the compute resources. The hibernated etcd cluster can be resumed later by scaling it back to a non-zero value. Please note that the data volumes backing the Etcd cluster will be retained during hibernation, and will still be charged for.
+    While an Etcd cluster can be scaled out, it cannot be scaled in, i.e., the replicas cannot be decreased to a non-zero value. An Etcd cluster can still be scaled to 0 replicas, indicating that the cluster is to be "hibernated". This is beneficial for use-cases where an etcd cluster is not needed for a certain period of time, and the user does not want to pay for the compute resources. A hibernated etcd cluster can be resumed later by scaling it back to a non-zero value. Please note that the data volumes backing an Etcd cluster will be retained during hibernation, and will still be charged for.
 
 ### Scale the Etcd cluster vertically
 
-To scale the etcd cluster vertically, you can update the `spec.etcd.resources` and `spec.backup.resources` fields in the `Etcd` custom resource. For example, to scale the etcd container to 2 CPU and 4Gi memory, you can run:
+To scale an Etcd cluster vertically, you can update the `spec.etcd.resources` and `spec.backup.resources` fields in the `Etcd` custom resource. For example, to scale an Etcd cluster's `etcd` container to 2 CPU and 4Gi memory, you can run:
 
 ```bash
 kubectl patch etcd <etcd-name> -n <namespace> --type merge -p '{"spec":{"etcd":{"resources":{"requests":{"cpu":"2","memory":"4Gi"}}}}}'
 ```
 
-Additionally, you can use an external pod autoscaler such as [vertical pod autoscaler](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) to scale the etcd cluster vertically. This is made possible by the `/scale` subresource on the Etcd custom resource.
+Additionally, you can use an external pod autoscaler such as [vertical pod autoscaler](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) to scale an Etcd cluster vertically. This is made possible by the `/scale` subresource on the Etcd custom resource.
 
 !!! note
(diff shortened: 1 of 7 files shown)


unmarshall (approved):

unmarshall:
@shreyas-s-rao can you also add the following to the docs:

An operator can use `kubectl scale` to horizontally scale an etcd cluster. However this will get automatically reflected if auto-reconciliation is in progress. If its not enabled then one needs to put the reconcile annotation on an Etcd resource.

shreyas-s-rao:
@unmarshall thanks for the suggestion. I've added documentation for both types of scaling in https://github.com/gardener/etcd-druid/pull/1070/commits/2a427ebda5e684be1b6b10f76a0c23660e19e407

docs/usage/managing-etcd-clusters.md
```diff
@@ -47,6 +47,39 @@ kubectl edit etcd <etcd-name> -n <namespace>
...
 
+### Scale the Etcd cluster horizontally
+
+To scale the etcd cluster horizontally, you can update the `spec.replicas` field in the `Etcd` custom resource. For example, to scale the etcd cluster to 5 replicas, you can run:
+
+```bash
+kubectl patch etcd <etcd-name> -n <namespace> --type merge -p '{"spec":{"replicas":5}}'
+```
+
+Additionally, you can use the `/scale` subresource to scale the etcd cluster horizontally. For example, to scale the etcd cluster to 5 replicas, you can run:
```
unmarshall, line 58:
```suggestion
Alternatively, you can use the `/scale` subresource to scale an etcd cluster horizontally. For example, to scale an etcd cluster to 5 replicas, you can run:
```

docs/usage/managing-etcd-clusters.md
```diff
@@ -47,6 +47,39 @@ kubectl edit etcd <etcd-name> -n <namespace>
 
 This will open up the linked editor where you can make the edits.
 
+### Scale the Etcd cluster horizontally
+
+To scale the etcd cluster horizontally, you can update the `spec.replicas` field in the `Etcd` custom resource. For example, to scale the etcd cluster to 5 replicas, you can run:
```
unmarshall, line 52:
```suggestion
To scale an etcd cluster horizontally, you can update the `spec.replicas` field in the `Etcd` custom resource. For example, to scale an etcd cluster to 5 replicas, you can run:
```

docs/usage/managing-etcd-clusters.md
```diff
@@ -47,6 +47,39 @@ kubectl edit etcd <etcd-name> -n <namespace>
...
+```
+
+Additionally, you can use the `/scale` subresource to scale the etcd cluster horizontally. For example, to scale the etcd cluster to 5 replicas, you can run:
+
+```bash
+kubectl scale etcd <etcd-name> -n <namespace> --replicas=5
+```
+
+!!! note
+    While the Etcd cluster can be scaled out, it cannot be scaled in, ie, the replicas cannot be decreased to a non-zero value. The Etcd cluster can still be scaled to 0 replicas, indicating that the cluster is to be "hibernated". This is beneficial for use-cases where the etcd cluster is not needed for a certain period of time, and the user does not want to pay for the compute resources. The hibernated etcd cluster can be resumed later by scaling it back to a non-zero value. Please note that the data volumes backing the Etcd cluster will be retained during hibernation, and will still be charged for.
```
unmarshall, line 65: `ie` should be `i.e.`

docs/usage/managing-etcd-clusters.md
```diff
@@ -47,6 +47,39 @@ kubectl edit etcd <etcd-name> -n <namespace>
...
+```
+
+Additionally, you can use an external pod autoscaler such as [vertical pod autoscaler](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) to scale the etcd cluster vertically. This is made possible by the `/scale` subresource on the Etcd custom resource.
+
+!!! note
+    Scaling the Etcd cluster vertically can potentially result in downtime, based on its failure tolerance. This is due to the fact that the etcd cluster is actuated by pods, and the pods need to be restarted in order to apply the new resource values. A multi-member etcd cluster can tolerate a failure of one member, but if the number of replicas is set to 1, then the etcd cluster will not be able to tolerate any failures. Therefore, it is recommended to scale the etcd cluster vertically only when it is not under heavy load, and to monitor the cluster closely during the scaling process. A multi-member Etcd cluster is also accompanied by a pod disruption budget to ensure that the cluster can tolerate a certain number of failures. The pod disruption budget is automatically created and managed by etcd-druid, and it is recommended to not modify it manually.
```
unmarshall, line 78: You mention:
```
Therefore, it is recommended to scale the etcd cluster vertically only when it is not under heavy load, and to monitor the cluster closely during the scaling process
```
Well, should that not be precisely the case when it should be scaled up?

docs/usage/managing-etcd-clusters.md
```diff
@@ -47,6 +47,39 @@ kubectl edit etcd <etcd-name> -n <namespace>
...
+
+!!! note
+    While the replicas and resources in the Etcd resource spec can be modified, please ensure to read the next section to understand when and how these changes are actuated by etcd-druid.
```
unmarshall, line 81:
```suggestion
    While the replicas and resources in the Etcd resource spec can be modified, please ensure to read the next section to understand when and how these changes are reconciled by etcd-druid.
```

unmarshall (commented):
Minor comments to docs.

Issue #1071, opened 2025-04-15
Support self-hosted shoot cluster use-cases


shreyas-s-rao:
**How to categorize this issue?**

/area control-plane high-availability
/kind enhancement

**What would you like to be added**:
Gardener is on its way to support running [self-hosted shoot clusters](https://github.com/gardener/gardener/blob/master/docs/proposals/28-self-hosted-shoot-clusters.md), and would like etcd-druid to support this use-case. Currently, the etcd pods of an self-hosted shoot cluster's control plane are planned to be deployed as static pods, in order to tolerate node restarts/failures. While the etcd static pod will initially be deployed by gardenadm, its management will be taken over by etcd-druid once the control plane is stable.

For druid to manage etcd cluster as static pods, we require a few changes in the way the etcd pods are created and managed. Following are a list of changes already discussed:
- [x] https://github.com/gardener/etcd-druid/pull/1086
- [x] https://github.com/gardener/etcd-druid/pull/1088
- [x] https://github.com/gardener/etcd-druid/pull/1100
- [x] https://github.com/gardener/etcd-druid/pull/1106
- [ ] De-couple etcd-backup-restore from Kubernetes API. Currently, both etcd-backup-restore bootstrap logic as well as scale-up from 1->3 replicas, have a dependency on k8s API to fetch statefulset to check its annotations, or to update peer-TLS info in member leases, etc. This will not work for static pods that do not use statefulsets or leases to indicate intent/info to etcd-druid.
  - [x] https://github.com/gardener/etcd-backup-restore/pull/764
- [ ] Decouple from using pod name for etcd member name and URLs. This is because, pod names for static pods will be suffixed by the node name, and druid cannot know this information beforehand in order to correctly set the etcd configuration in the configmap.
- [x] ~~Allow specifying volume type in the Etcd resource, so that we can specify either PVC (default) or a hostPath which points to the data directory already created for the existing etcd pod which druid aims to "adopt" and manage, without having to recreate the etcd data.~~
  - not needed since this will be taken over by the [static pod translator](https://github.com/gardener/gardener/pull/11349)

Some considerations:
[...]

PR #1087, opened 2025-05-06
Augment `Etcd` API with `enable{Member,Snapshot}LeaseRenewal` fields

rfranzke:

/area usability
/kind enhancement

After https://github.com/gardener/etcd-druid/pull/1075, it is possible to configure `etcd-backup-restore` sidecar with the new `--enable-{member,snapshot}-lease-renewal` flags. However, they are not exposed via the `Etcd` API, i.e., it was not possible to configure them. This PR addresses this aspect.

Part of https://github.com/gardener/etcd-druid/issues/1071
Follow-up of https://github.com/gardener/etcd-druid/pull/1075

/cc @shreyas-s-rao @ishan16696 

**Release note**:

```feature developer
It is now possible to disable member and snapshot `Lease` renewals by setting `.spec.backup.enable{Member,Snapshot}LeaseRenewal` in the `Etcd` API to `false`.
```


api/core/v1alpha1/etcd.go
```diff
@@ -188,6 +188,12 @@ type BackupSpec struct {
 	// EnableProfiling defines if profiling should be enabled for the etcd-backup-restore-sidecar
 	// +optional
 	EnableProfiling *bool `json:"enableProfiling,omitempty"`
+	// EnableMemberLeaseRenewal defines if member lease renewal should be enabled for the etcd-backup-restore-sidecar
+	// +optional
+	EnableMemberLeaseRenewal *bool `json:"enableMemberLeaseRenewal,omitempty"`
```
unmarshall, line 193: Today as part of the Etcd API we do not expose in any way a list of components that are created to realise an etcd-cluster since that is an implementation detail and can change. For instance we would get rid of snapshot lease as it is a big time hack and replace it with `EtcdMember` custom resource in near future. So we consciously try not to leak implementation details into the API.

From this perspective these 2 configuration properties now leak implementation details into the API.

Secondly, should etcd-druid even create snapshot/member lease resources if they are never going to be renewed by backup-restore? Probably not, right?

While i do understand the need to have a handle to configure etcd-backup-restore, i am not entirely sure if this is the right change to the etcd-druid API.

Let us also understand the consequences of disabling renewal of leases:
- Disabling renewal of snapshot lease will lead to inability to run compaction jobs. We need to evaluate if compaction jobs are relevant for autonomous clusters.
- Disabling renewal of member lease will lead to inability to switch from non-TLS to TLS for peer URLs. This is the case for g/g clusters which move from non-HA to HA. Perhaps for autonomous you can always enable peer TLS and circumvent this issue.
- Member status updates is now incomplete. What will not be available: 
  - `Etcd.Status.Members` can no longer be computed because it uses member leases.
  -  As a consequence of the above, `AllMembersReady`can no longer be computed.
  - `BackupReady` condition is no longer valid as it depends on the information from snapshot leases.

What you really need is to disable k8s client communication from etcd-members to a kube-apiserver. If this is disabled then there is no longer a need to create the following additional resources for an etcd-cluster (apart from member and snapshot leases) :
- role
- rolebinding
- serviceaccount

We would propose that instead of changing the API you annotate the `Etcd` resource with `druid.gardener.cloud/disable-member-status-reporting`. Once this is set, then the following will be the consequences (which perhaps works for single node etcd clusters):
- No member status is available to etcd-druid.
- Scale from single to HA etcd cluster will also no longer work.

In medium to long term we would like to brainstorm on a holistic approach to allow etcd-druid to completely manage the lifecycle and monitoring of etcd clusters even for on-prem installations (i.e. autonomous clusters use case).

ishan16696, line 193: Currently, we have not exposed the `--enable-{member,snapshot}-lease-renewal`  flags via API because we assume that `etcd-druid` will always operate within Kubernetes environments. From an open-source perspective and consumption of etcd-druid outside gardener, it is not advisable to allow users to set these flags to `false`, as doing so would make `etcd-druid` unable to scale up the etcd cluster. 
These flags should only be set to `false` in non-HA conditions where there is no intention to transition to HA in the future.

api/core/v1alpha1/etcd.go
```diff
@@ -188,6 +188,12 @@ type BackupSpec struct {
...
+	// EnableSnapshotLeaseRenewal defines if snapshot lease renewal should be enabled for the etcd-backup-restore-sidecar
+	// +optional
+	EnableSnapshotLeaseRenewal *bool `json:"enableSnapshotLeaseRenewal,omitempty"`
```
ishan16696, line 196: same as above

api/core/v1alpha1/etcd.go
rfranzke: `gardenadm` (as bring-up tool for autonomous shoot clusters) is also just a user of `etcd-druid` and its `Etcd` API. It requires these flags to be set to `false` on `etcd-backup-restore`. Hence, we need to make this configurable somehow, be it via `spec` or via annotation as proposed by https://github.com/gardener/etcd-druid/pull/1087#pullrequestreview-2827028689

rfranzke: Done, PTAL :)

PR #1095, opened 2025-05-13
Generate CRD for Etcd without CEL expressions

anveshreddy18:

/area usability
/kind bug

This PR re-adds the Etcd CRD for when CEL expressions are not used, which was existing before [this PR](https://github.com/gardener/etcd-druid/pull/1084) seems to have removed it accidentally. Hence adding it now after manually invoking the `generate_crd_without_cel_expressions` function in the bash script `api/hack/generate.sh`

Fixes #

**Release note**:

```other operator

```


renormalize:
A patch with this commit in would be necessary for the current release, considering both CEL and without-CEL are to be supported currently if I am not mistaken. Am I correct @anveshreddy18?

anveshreddy18:
Thanks for taking a look @renormalize. The `etcd-druid` release `v0.30.0` has not been made yet, will probably happen today which will include these changes too, so I think we're good in that way.

renormalize (approved):

renormalize:
@anveshreddy18 thanks!

Shreyas-s14 (approved):
Thanks for the PR @anveshreddy18 !!
LGTM

seshachalam-yv (approved):
/lgtm

PR #1088, opened 2025-05-06
Augment `Etcd` API with `runAsRoot`

rfranzke:
**How to categorize this PR?**

/area usability
/kind enhancement

**What this PR does / why we need it**:
In the autonomous shoot cluster scenario, the etcd pods have to run as `root` in order to be able to use the host path volumes for reading config files and writing data.

**Which issue(s) this PR fixes**:
Part of https://github.com/gardener/etcd-druid/issues/1071

**Special notes for your reviewer**:
/cc @shreyas-s-rao @ishan16696 

**Release note**:

```feature developer
It is now possible to run the `etcd` pod as `root` by setting `.spec.runAsRoot=true` in the `Etcd` API.
```

PR #1100, opened 2025-05-15
Use `/root` as home dir instead of `/home/nonroot` when running as `root`

rfranzke:

/area usability
/kind bug

In https://github.com/gardener/etcd-druid/pull/1088, we introduced `spec.runAsRoot` in the `Etcd` API. However, when set to `true` and when using the `local` provider for backups, the bucket path is mounted to `/home/nonroot`. This does not play well together with `etcd-backup-restore` which uses the home directory for looking up the bucket: https://github.com/gardener/etcd-backup-restore/blob/beafd20c79dd41e266c303590e3fe8cc316a46bb/pkg/snapstore/utils.go#L72-L76

This PR changes the mount path to `/root` when `.spec.runAsRoot` is set to `true`. The init container is not needed in this case.

Follow-up of https://github.com/gardener/etcd-druid/pull/1088

/cc @unmarshall @anveshreddy18 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/internal/utils/kubernetes/store.go
+++ b/internal/utils/kubernetes/store.go
@@ -1,0 +1,23 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package kubernetes
+
+import (
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
+	druidstore "github.com/gardener/etcd-druid/internal/store"
+)
+
+// GetBackupStoreProvider returns the provider name for the backup store. If the provider is not known, an error is
+// returned.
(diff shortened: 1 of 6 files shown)


internal/utils/kubernetes/store.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	provider, err := druidstore.StorageProviderFromInfraProvider(etcd.Spec.Backup.Store.Provider)
+	if err != nil {
+		return nil, err
+	}
+	return &provider, nil
+}
+
+// MountPathLocalStore returns the mount path for the local store if the provider matches.
+func MountPathLocalStore(etcd *druidv1alpha1.Etcd, provider *string) string {
```
ishan16696, line 28: would you like to add a basic couple of unit tests for this func ?

anveshreddy18 (approved):
Thanks for the PR. LGTM!

Issue #1010, opened 2025-02-27
Fix local provider setup for running etcd clusters & e2e tests


anveshreddy18:
**How to categorize this issue?**

/area testing
/area usability
/kind bug

**What happened**:

When running etcd-druid locally with `local` provider, I've seen the functionality is broken with a volume mount error for the pods as shown below: 

```
MountVolume.SetUp failed for volume "local-backup" : hostPath type check failed: /etc/gardener/local-backupbuckets/etcd-bucket is not a directory
```

This is due to the `HostPath` type being set to `Directory` instead of `DirectoryOrCreate` [here](https://github.com/gardener/etcd-druid/blob/c5d8f6ed8f4cbc04718ab8783250157da5aa2bac/internal/component/statefulset/builder.go#L810). 

Also all the e2e tests for the local provider fail with the following error: 

```
[FAILED] Unexpected error:
      <*errors.StatusError | 0x140005ff5e0>: 
      jobs.batch "purge-local-snapstore" already exists
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "jobs.batch \"purge-local-snapstore\" already exists",
              Reason: "AlreadyExists",
              Details: {
                  Name: "purge-local-snapstore",
                  Group: "batch",
                  Kind: "jobs",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 409,
          },
      }
  occurred
  In [It] at: /Users/i586337/work/etcd-druid/test/e2e/etcd_multi_node_test.go:264
```

This needs to be looked into and fix it. 

This issue acts as a tracker to fix the problems with `local` provider.

**What you expected to happen**:

The etcd-druid setup with `local` provider should be as robust as it is with other providers/emulators.

**How to reproduce it (as minimally and precisely as possible)**:

In the `examples/etcd/druid_v1alpha1_etcd.yaml` file, change the `spec.backup.store` to the following and try deploying the Etcd CR and observe the pod events.

```
store:
  provider: local # options: aws,azure,gcp,openstack,alicloud,dell,openshift,local
  prefix: etcd-prefix
  container: etcd-bucket
```

For e2e tests, just run them locally by doing below and observe the logs:
```
make kind-up
[...]

PR #1102, opened 2025-05-20
Fix local provider setup for running Etcd clusters

anveshreddy18:

/area usability
/kind bug

When etcd-druid is run with `local` provider with the `spec.backup.store.container` set, the functionality is broken with the following volume mount error: 

```
MountVolume.SetUp failed for volume "local-backup" : hostPath type check failed: /etc/gardener/local-backupbuckets/etcd-bucket is not a directory
```

This is due to the `HostPath` type being set to `Directory` instead of `DirectoryOrCreate` at multiple places, because of which the directory has to exist first before being used as a volume type for storing backups. But when started fresh, the node need not have the directory created beforehand, hence the volume error we see above. 

To fix this, the change is to basically convert all instances of hostPath type from `Directory` to `DirectoryOrCreate`, this configuration will ensure the directory is created if it doesn't exist, so that the CSI can create it into a volume to be mounted onto the `backup-restore` container. 

Fixes part of #1010

**Release note**:

```bugfix user
fix a volume creation issue with using `local` provider for etcd backups
```

### DIFF
--- a/internal/component/statefulset/builder.go
+++ b/internal/component/statefulset/builder.go
@@ -833,7 +833,7 @@ func (b *stsBuilder) getBackupVolume(ctx component.OperatorContext) (*corev1.Vol
 			return nil, fmt.Errorf("error getting host mount path for etcd: %v Err: %w", druidv1alpha1.GetNamespaceName(b.etcd.ObjectMeta), err)
 		}
 
-		hpt := corev1.HostPathDirectory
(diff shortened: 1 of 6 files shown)


ishan16696 (approved):
Thanks @anveshreddy18 for the fix, can you update the release note with something like this:
```
Fix a volume mount issue of provider :`Local` for etcd backups
```

PR #1014, opened 2025-03-04
Cleanup older configmaps with the naming format `etcd-bootstrap-UID[:6]`

anveshreddy18:
**How to categorize this PR?**

/area quality
/kind cleanup

**What this PR does / why we need it**:

This PR cleans up the unused configmaps with the naming format `etcd-bootstrap-UID[:6]` that were created by `etcd-druid` until `v0.24.0`. The name of the configmap was changed to `{etcd.Name}-config` from `v0.25.0` onwards but the older configmaps were not deleted, this PR removes them. Refer to the issue #931 for more details. 

**Which issue(s) this PR fixes**:
Fixes #931 

**Special notes for your reviewer**:

**Release note**:

```noteworthy operator
cleanup unused older configmaps with the naming format `etcd-bootstrap-UID[:6]` from the cluster.
```

PR #1103, opened 2025-05-21
Remove functionality that deletes old ConfigMaps

### LINKED ISSUE #1015: Remove the functionality in druid that deletes old configmaps

/area quality
/kind cleanup

The PR #1014 adds functionality to cleanup the old configmaps with naming format `etcd-bootstrap-UID[:6]`. However, as it's only a cleanup logic this needs to be removed from druid code after 3 releases i.e in `v0.31.0` ensuring sufficient time for users to upgrade to versions that has this cleanup logic. After that, it is unnecessary to maintain this piece of code. 

**Why is this needed**:

Remove cleanup logic for older configmaps in `v0.31.0`

anveshreddy18:

/area quality
/kind cleanup

This PR removes the code functionality that was introduced in #1014 to remove older ConfigMaps with the naming format `etcd-bootstrap-UID[:6]`. The functionality was only intended to be there for 3 releases starting from `v0.28.0` till `v0.30.0`. Now it's no longer needed to be present, hence removing the functionality.

Fixes #1015 

**Release note**:

```other developer
Remove the functionality that deletes old ConfigMaps from the cluster
```

### DIFF
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -29,12 +29,6 @@ func GetServiceAccountName(etcdObjMeta metav1.ObjectMeta) string {
 	return etcdObjMeta.Name
 }
 
-// GetOldConfigMapName returns the name of the old configmap for the Etcd.
-// TODO: @anveshreddy18: Remove this function after 3 releases, i.e in v0.31.0
-func GetOldConfigMapName(etcdObjMeta metav1.ObjectMeta) string {
-	return fmt.Sprintf("etcd-bootstrap-%s", string(etcdObjMeta.UID[:6]))
-}
-
 // GetConfigMapName returns the name of the configmap for the Etcd.
 func GetConfigMapName(etcdObjMeta metav1.ObjectMeta) string {
 	return fmt.Sprintf("%s-config", etcdObjMeta.Name)
--- a/internal/component/types.go
+++ b/internal/component/types.go
@@ -17,8 +17,6 @@ import (
 // However, it can also be used where ever operation context is required to be specified.
 // Each component can in turn define their own fine-grained operation labels as well.
 const (
-	// OperationPreSync is the PreSync operation of the Operator.
-	OperationPreSync = "PreSync"
 	// OperationSync is the Sync operation of the Operator.
 	OperationSync = "Sync"
 	// OperationTriggerDelete is the TriggerDelete operation of the Operator.
--- a/internal/component/types.go
+++ b/internal/component/types.go
@@ -17,6 +17,8 @@ import (
 // However, it can also be used where ever operation context is required to be specified.
 // Each component can in turn define their own fine-grained operation labels as well.
 const (
+	// OperationPreSync is the PreSync operation of the Operator.
+	OperationPreSync = "PreSync"
 	// OperationSync is the Sync operation of the Operator.
 	OperationSync = "Sync"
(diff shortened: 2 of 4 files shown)


api/core/v1alpha1/helper.go
```diff
@@ -32,7 +32,11 @@ func GetServiceAccountName(etcdObjMeta metav1.ObjectMeta) string {
 // GetOldConfigMapName returns the name of the old configmap for the Etcd.
 // TODO: @anveshreddy18: Remove this function after 3 releases, i.e in v0.31.0
 func GetOldConfigMapName(etcdObjMeta metav1.ObjectMeta) string {
-	return fmt.Sprintf("etcd-bootstrap-%s", string(etcdObjMeta.UID[:6]))
+	etcdRole := etcdObjMeta.Labels["role"]
+	if etcdRole == "" {
+		etcdRole = "unknown"
+	}
+	return fmt.Sprintf("etcd-bootstrap-%s", etcdRole)
```
seshachalam-yv, line 39: Should we wait for three more releases after `v0.34.0` before removing this function?
Also, do we need to update the comment accordingly in this PR?

seshachalam-yv (commented):
Overall looks good to me. 
Just a small nitpick: please address the minor comment I’ve left.

api/core/v1alpha1/helper.go
anveshreddy18: Thanks for the review @seshachalam-yv, as I informed you offline, I've repurposed this PR to remove the cleanup logic itself. PTAL at the new version. Thanks

internal/component/types.go
```diff
@@ -17,8 +17,6 @@ import (
 // However, it can also be used where ever operation context is required to be specified.
 // Each component can in turn define their own fine-grained operation labels as well.
 const (
-	// OperationPreSync is the PreSync operation of the Operator.
-	OperationPreSync = "PreSync"
```
shreyas-s-rao, line 21: We can retain this, to allow any new components that want to implement the `PreSync` method, to re-use this constant.

shreyas-s-rao (changes requested):
@anveshreddy18 thanks a lot for the PR! Just one small nit, but overall looks good. PTAL.

internal/component/types.go
anveshreddy18: I thought to, but then it may be unused for a long time so linters may issue a warning, so I decided to remove it. What is the best practice here? I have no hard opinion, just trying to understand when to keep vs when to remove these consts.

shreyas-s-rao: That's a fair point. But this constant is to be used by other components when they implement `PreSync` method, so that makes it part of the developer API (not to be confused with the user API), since we want developers (including ourselves) to re-use this constant for return `druidErrors` from the `PreSync` method, to retain consistency.

Unused constants are generally not flagged by linters, since APIs can always define constants that may not be used within the codebase, but can be used by external users. In our case, this is part of the developer API which is needed to retain consistency when `PreSync` is used once again, and not re-define it (where mistakes can be made and we may end up changing the constant's value).

anveshreddy18: Agreed!. Thanks for explaining. 

I've addressed, PTAL. Thanks

shreyas-s-rao (approved):
/lgtm

seshachalam-yv (approved):
/lgtm

PR #1117, opened 2025-06-02
Support disabling creation of etcd runtime components

### LINKED ISSUE #1106: Support disabling creation of runtime components for etcd cluster

/area usability
/kind enhancement

As part of #1071 , we need to allow specifying to druid to provision an etcd cluster with the correct expected configuration, but without running any of the etcd pods. Additionally, the etcd-backup-restore sidecar should be configured to disable lease renewal, since the pods do not have access to the same kube-apiserver that druid talks to to create the member leases and snapshot leases. Since lease renewal is disabled, druid need not create components such as member leases, snapshot leases, RBAC (serviceaccount, role, rolebinding) resources required by the backup sidecar to renew these leases. Since etcd-druid derives many of the Etcd status fields and conditions from these leases, these checks will also need to be disabled.

**Why is this needed**:
This is a replacement of https://github.com/gardener/etcd-druid/pull/1087, which introduced the ability to disable renewal of member leases and snapshot leases by the etcd-backup-restore sidecar, but requires the additional changes mentioned above.

shreyas-s-rao:

/area control-plane
/kind enhancement task

Introduce annotation `druid.gardener.cloud/disable-etcd-runtime-component-creation` for the Etcd resource to instruct druid to not create runtime components for the etcd cluster, such as pods, volumes, member leases and snapshot leases used by the pods to communicate status to druid, and RBAC resources (serviceaccount, role, rolebinding) needed by the etcd member pods to update said leases. This is useful when the operator wants to manage the runtime resources manually or when the runtime resources are not needed, but still requires the correct etcd cluster configuration to be made available.

This also disables status checks, since these rely on the existence of said leases and pods to derive etcd cluster status. In the future, we will work on ways for the running members to directly communicate status with druid without the need for a k8s cluster and leases to do so.

Fixes #1106 

/cc @rfranzke 
/invite @ishan16696 

**Release note**:

```noteworthy user
NONE
```

### DIFF
--- a/api/core/v1alpha1/constants.go
+++ b/api/core/v1alpha1/constants.go
@@ -36,7 +36,5 @@ const (
 	DruidOperationReconcile = "reconcile"
 	// DisableEtcdRuntimeComponentCreationAnnotation is an annotation set by an operator to disable the creation and management of
 	// runtime components of the etcd cluster such as pods, PVCs, leases, RBAC resources, etc.
-	// This is useful when the operator wants to manage the runtime resources manually or when the runtime resources are not needed,
-	// but still requires the correct etcd cluster configuration to be made available.
 	DisableEtcdRuntimeComponentCreationAnnotation = "druid.gardener.cloud/disable-etcd-runtime-component-creation"
 )
--- a/api/core/v1alpha1/constants.go
+++ b/api/core/v1alpha1/constants.go
@@ -35,6 +35,6 @@ const (
 	// --enable-etcd-spec-auto-reconcile CLI flag.
 	DruidOperationReconcile = "reconcile"
 	// DisableEtcdRuntimeComponentCreationAnnotation is an annotation set by an operator to disable the creation and management of
-	// runtime components of the etcd cluster such as pods, PVCs, leases, RBAC resources, etc.
+	// runtime components of the etcd cluster such as pods, PVCs, leases, RBAC resources, PDBs, services, etc.
 	DisableEtcdRuntimeComponentCreationAnnotation = "druid.gardener.cloud/disable-etcd-runtime-component-creation"
 )
--- a/api/core/v1alpha1/constants.go
+++ b/api/core/v1alpha1/constants.go
@@ -34,4 +34,9 @@ const (
 	// This value will only be effective if etcd-druid is not configured with auto-reconciliation of Etcd resource specification via
 	// --enable-etcd-spec-auto-reconcile CLI flag.
 	DruidOperationReconcile = "reconcile"
+	// DisableEtcdRuntimeComponentCreationAnnotation is an annotation set by an operator to disable the creation and management of
+	// runtime components of the etcd cluster such as pods, PVCs, leases, RBAC resources, etc.
+	// This is useful when the operator wants to manage the runtime resources manually or when the runtime resources are not needed,
+	// but still requires the correct etcd cluster configuration to be made available.
+	DisableEtcdRuntimeComponentCreationAnnotation = "druid.gardener.cloud/disable-etcd-runtime-component-creation"
 )
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -165,3 +165,8 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
 	delete(etcdObjMeta.Annotations, DruidOperationAnnotation)
 	delete(etcdObjMeta.Annotations, GardenerOperationAnnotation)
 }
+
+// IsEtcdRuntimeComponentCreationEnabled checks if the creation of runtime components is not disabled for the Etcd resource.
+func IsEtcdRuntimeComponentCreationEnabled(etcdObjMeta metav1.ObjectMeta) bool {
+	return !metav1.HasAnnotation(etcdObjMeta, DisableEtcdRuntimeComponentCreationAnnotation)
+}
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -166,7 +166,7 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
 	delete(etcdObjMeta.Annotations, GardenerOperationAnnotation)
 }
 
-// IsEtcdRuntimeComponentCreationEnabled checks if the creation of runtime components is not disabled for the Etcd resource.
+// IsEtcdRuntimeComponentCreationEnabled checks if the creation of runtime components is enabled for an Etcd resource.
 func IsEtcdRuntimeComponentCreationEnabled(etcdObjMeta metav1.ObjectMeta) bool {
 	return !metav1.HasAnnotation(etcdObjMeta, DisableEtcdRuntimeComponentCreationAnnotation)
 }
--- a/docs/usage/managing-etcd-clusters.md
+++ b/docs/usage/managing-etcd-clusters.md
@@ -32,23 +32,6 @@ kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=instance=
 kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=app.kubernetes.io/managed-by=etcd-druid,app.kubernetes.io/part-of=<etcd-name>
 ```
 
-### Disable creation of runtime components
-
-If you do not want to create runtime components in order to manage them externally, you can set the annotation `druid.gardener.cloud/disable-etcd-runtime-component-creation` on the Etcd custom resource. This is useful in scenarios where the etcd pods need to be run as static pods on a Kubernetes node, where druid provides the configuration for the etcd cluster, and an external actor creates the corresponding static pod manifests and related resources on the node.
-
-Setting this annotation will disable the creation of the following resources/components:
-- `pods` - while the Etcd `spec.replicas` still decides the number of etcd members in the etcd cluster configuration, no pods will be created for these.
-- `volumes` - no persistent volumes will be created for the etcd cluster, since no pods are created.
-- `member leases`
-- `snapshot leases`
-- `serviceaccount`
-- `role`
-- `rolebinding`
-
-The statefulset will still be created, with `spec.replicas` set to 0, but the lease renewal flags `--enable-member-lease-renewal` and `--enable-snapshot-lease-renewal` on the `etcd-backup-restore` config will be disabled.
-
-Additionally, certain checks on the `Etcd` resource status will be disabled, since according to etcd-druid, the etcd cluster is not running.
-
 ## Update & Reconcile an Etcd Cluster
 
 ### Edit the Etcd custom resource
--- a/docs/usage/managing-etcd-clusters.md
+++ b/docs/usage/managing-etcd-clusters.md
@@ -32,7 +32,22 @@ kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=instance=
 kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=app.kubernetes.io/managed-by=etcd-druid,app.kubernetes.io/part-of=<etcd-name>
 ```
 
-  
+### Disable creation of runtime components
+
+If you do not want to create runtime components in order to manage them externally, you can set the annotation `druid.gardener.cloud/disable-etcd-runtime-component-creation` on the Etcd custom resource. This is useful in scenarios where the etcd pods need to be run as static pods on a Kubernetes node, where druid provides the configuration for the etcd cluster, and an external actor creates the corresponding static pod manifests and related resources on the node.
+
+Setting this annotation will disable the creation of the following resources/components:
+- `pods` - while the Etcd `spec.replicas` still decides the number of etcd members in the etcd cluster configuration, no pods will be created for these.
+- `volumes` - no persistent volumes will be created for the etcd cluster, since no pods are created.
+- `member leases`
+- `snapshot leases`
+- `serviceaccount`
+- `role`
+- `rolebinding`
+
+The statefulset will still be created, with `spec.replicas` set to 0, but the lease renewal flags on the `etcd-backup-restore` sidecar container will remain disabled.
+
+Additionally, certain checks on the `Etcd` resource status will be disabled, since according to etcd-druid, the etcd cluster is not running.
 
 ## Update & Reconcile an Etcd Cluster
 
--- a/docs/usage/managing-etcd-clusters.md
+++ b/docs/usage/managing-etcd-clusters.md
@@ -45,7 +45,7 @@ Setting this annotation will disable the creation of the following resources/com
 - `role`
 - `rolebinding`
 
-The statefulset will still be created, with `spec.replicas` set to 0, but the lease renewal flags on the `etcd-backup-restore` sidecar container will remain disabled.
+The statefulset will still be created, with `spec.replicas` set to 0, but the lease renewal flags `--enable-member-lease-renewal` and `--enable-snapshot-lease-renewal` on the `etcd-backup-restore` config will be disabled.
 
 Additionally, certain checks on the `Etcd` resource status will be disabled, since according to etcd-druid, the etcd cluster is not running.
 
--- a/internal/component/statefulset/builder.go
+++ b/internal/component/statefulset/builder.go
@@ -136,7 +136,11 @@ func (b *stsBuilder) getStatefulSetLabels() map[string]string {
 
 func (b *stsBuilder) createStatefulSetSpec(ctx component.OperatorContext) error {
 	err := b.createPodTemplateSpec(ctx)
-	b.sts.Spec.Replicas = ptr.To(b.replicas)
+	b.sts.Spec.Replicas = ptr.To[int32](0)
+	if druidv1alpha1.IsEtcdRuntimeComponentCreationEnabled(b.etcd.ObjectMeta) {
+		b.sts.Spec.Replicas = ptr.To(b.replicas)
+	}
+	b.logger.Info("Creating StatefulSet spec", "replicas", b.sts.Spec.Replicas, "name", b.sts.Name, "namespace", b.sts.Namespace)
 	b.sts.Spec.UpdateStrategy = defaultUpdateStrategy
 	if err != nil {
 		return err
@@ -164,7 +168,6 @@ func (b *stsBuilder) createPodTemplateSpec(ctx component.OperatorContext) error 
 	podTemplateSpec := corev1.PodTemplateSpec{
 		Spec: corev1.PodSpec{
 			HostAliases:           b.getHostAliases(),
-			ServiceAccountName:    druidv1alpha1.GetServiceAccountName(b.etcd.ObjectMeta),
 			ShareProcessNamespace: ptr.To(true),
 			InitContainers:        b.getPodInitContainers(),
 			Containers: []corev1.Container{
@@ -177,6 +180,9 @@ func (b *stsBuilder) createPodTemplateSpec(ctx component.OperatorContext) error 
 			Volumes:                   podVolumes,
 			PriorityClassName:         ptr.Deref(b.etcd.Spec.PriorityClassName, ""),
 		},
+	}
+	if druidv1alpha1.IsEtcdRuntimeComponentCreationEnabled(b.etcd.ObjectMeta) {
+		podTemplateSpec.Spec.ServiceAccountName = druidv1alpha1.GetServiceAccountName(b.etcd.ObjectMeta)
 	}
 	podTemplateLabels := b.getStatefulSetPodLabels()
 	selectorMatchesLabels, err := kubernetes.DoesLabelSelectorMatchLabels(b.sts.Spec.Selector, podTemplateLabels)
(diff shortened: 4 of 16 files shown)


docs/usage/managing-etcd-clusters.md
```diff
@@ -32,7 +32,22 @@ kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=instance=
...
+Setting this annotation will disable the creation of the following resources/components:
+- `pods` - while the Etcd `spec.replicas` still decides the number of etcd members in the etcd cluster configuration, no pods will be created for these.
+- `volumes` - no persistent volumes will be created for the etcd cluster, since no pods are created.
+- `member leases`
+- `snapshot leases`
+- `serviceaccount`
+- `role`
+- `rolebinding`
+
+The statefulset will still be created, with `spec.replicas` set to 0, but the lease renewal flags on the `etcd-backup-restore` sidecar container will remain disabled.
```
ishan16696:
```suggestion
The statefulset will still be created, with `spec.replicas` set to 0, but the lease renewal flags: `--enable-member-lease-renewal` and `--enable-snapshot-lease-renewal` of the `etcd-backup-restore` config will be disabled.
```
I suggest to remove the word `sidecar container` as it will be no longer deployed by sts, so I want it to keep doc flexible as we don't know how user will use it in general and sts just provide the backup-restore config, wdyt ?

internal/controller/etcd/reconcile_status.go
```diff
@@ -27,11 +27,20 @@ func (r *Reconciler) reconcileStatus(ctx component.OperatorContext, etcdObjectKe
...
-		r.setSelector,
+
+	var mutateETCDStatusStepFns []mutateEtcdStatusFn
+	if druidv1alpha1.IsEtcdRuntimeComponentCreationEnabled(etcd.ObjectMeta) {
+		mutateETCDStatusStepFns = []mutateEtcdStatusFn{
+			r.mutateETCDStatusWithMemberStatusAndConditions,
+			r.inspectStatefulSetAndMutateETCDStatus,
+		}
+	} else {
+		sLog.Info("Skipping status checks since etcd runtime component creation is disabled")
```
ishan16696: There is one problem I see @shreyas-s-rao ,
suppose a user deployed the Etcd CR without this annotation with `.spec.replicas: 3`, after that when all resources get created successfully then user adds this annotation `druid.gardener.cloud/disable-etcd-runtime-component-creation` (may be by mistake), now the problem what I see is druid controller will always skip the update of Etcd status but resources are already got created, IMO, this is not the correct behaviour. wdyt ?

ishan16696 (changes requested):

internal/controller/etcd/reconcile_status.go
shreyas-s-rao: Agreed, thanks for pointing this out!

shreyas-s-rao: @ishan16696 I've added the cleanup logic. PTAL.

api/core/v1alpha1/constants.go
```diff
@@ -34,4 +34,9 @@ const (
 	// This value will only be effective if etcd-druid is not configured with auto-reconciliation of Etcd resource specification via
 	// --enable-etcd-spec-auto-reconcile CLI flag.
 	DruidOperationReconcile = "reconcile"
+	// DisableEtcdRuntimeComponentCreationAnnotation is an annotation set by an operator to disable the creation and management of
+	// runtime components of the etcd cluster such as pods, PVCs, leases, RBAC resources, etc.
+	// This is useful when the operator wants to manage the runtime resources manually or when the runtime resources are not needed,
```
unmarshall: >This is useful when the operator wants to manage the runtime resources manually

This statement has a few issues:
- There has not been any known use case for managing resource manually.
- Secondly to manage the resources manually etcd-druid as an operator needs to advertise what are the resources that needs to be created for a functioning etcd cluster.
- The set of resources required for autonomous use case might differ as compared to the ones required for hosted control planes.
Therefore i would really not add this to the doc string.

api/core/v1alpha1/constants.go
```diff
@@ -34,4 +34,9 @@ const (
...
+	// but still requires the correct etcd cluster configuration to be made available.
+	DisableEtcdRuntimeComponentCreationAnnotation = "druid.gardener.cloud/disable-etcd-runtime-component-creation"
```
unmarshall, line 39: What are the runtime components are not publicly known. This is completely internally managed by etcd-druid. This is on purpose so that it then has the flexibility to change the set of resources that it deploys without impacting consumers as no API is then changed. Therefore when you add this to the API and mention `disable runtime component` then the context is unclear as now it leaks implementation details.

Instead you choose a semantic name

docs/usage/managing-etcd-clusters.md
```diff
@@ -32,7 +32,22 @@ kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=instance=
 kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=app.kubernetes.io/managed-by=etcd-druid,app.kubernetes.io/part-of=<etcd-name>
 ```
 
-  
+### Disable creation of runtime components
+
+If you do not want to create runtime components in order to manage them externally, you can set the annotation `druid.gardener.cloud/disable-etcd-runtime-component-creation` on the Etcd custom resource. This is useful in scenarios where the etcd pods need to be run as static pods on a Kubernetes node, where druid provides the configuration for the etcd cluster, and an external actor creates the corresponding static pod manifests and related resources on the node.
```
unmarshall: as commented earlier this ideally should not be exposed.

internal/controller/etcd/reconcile_spec.go
```diff
@@ -172,17 +200,40 @@ func (r *Reconciler) getOrderedOperatorsForPreSync() []component.Kind {
...
-		component.ConfigMapKind,
 		component.PodDisruptionBudgetKind,
+		component.ConfigMapKind,
+		component.StatefulSetKind,
+	)
+
+	return operators
+}
+
+func (r *Reconciler) getOperatorsForCleanup(etcdObjMeta metav1.ObjectMeta) []component.Kind {
```
unmarshall, line 230: what happens when we have a fully managed etcd cluster and now this annotation is added on it. It is not prevented via any webhook. This means that now there will be leaked resources.

ishan16696: I also pointed this out in this comment: https://github.com/gardener/etcd-druid/pull/1117#discussion_r2126218726 then to address that @shreyas-s-rao added a cleanup logic which means when you have managed etcd cluster and then user add this annotation then druid will cleanup the respective resources, .... I hope my understanding is correct @shreyas-s-rao ?
I'm kind of not very comfortable with this cleanup logic as people by mistake can also adds this annotation, I'm also more inclining towards prevent this adding of annotation via webhook.

api/core/v1alpha1/constants.go
shreyas-s-rao: > There has not been any known use case for managing resource manually

The main usecase is https://github.com/gardener/etcd-druid/issues/1071, but I didn't want to bake that into the code/docstring since the feature from druid should be as generic as possible.

> Secondly to manage the resources manually etcd-druid as an operator needs to advertise what are the resources that needs to be created for a functioning etcd cluster

Well, the docs do provide that. What would you suggest is a better way to really advertise the created components? Etcd status?

> The set of resources required for autonomous use case might differ as compared to the ones required for hosted control planes.

This annotation isn't required for hosted control planes to begin with. It's only used for autonomous control planes for the moment, so I don't see a problem.

api/core/v1alpha1/constants.go
shreyas-s-rao: Then the only way is to name it something like `druid.gardener.cloud/for-autonomous-cluster-control-plane`, which makes it very specific to the usecase, but does not leak implementation detail. WDYT?

internal/controller/etcd/reconcile_spec.go
shreyas-s-rao: > what happens when we have a fully managed etcd cluster and now this annotation is added on it. It is not prevented via any webhook. This means that now there will be leaked resources

As @ishan16696 mentioned, I have added cleanup logic that deletes the unnecessary components after the annotation is set, and recreates them once the annotation is removed. If we want a webhook for the Etcd resource, then we add a latency overhead of the webhook, which is the reason we added CEL validations for the Etcd resource in the first place.

I don't see value in preventing the annotation from being added/removed, unless we want to force the policy that once an Etcd is decided to be used for hosted or autonomous control planes, it cannot be changed later. Is that something we want to enforce, given that we can easily handle changing between the two modes? If yes, and we want to achieve the enforcement without a webhook, then we'll need to convert this from an annotation into a well-defined Spec field, which will make it part of the API and cannot be easily changed later if we want to.

unmarshall: >I don't see value in preventing the annotation from being added/removed, unless we want to force the policy that once an Etcd is decided to be used for hosted or autonomous control planes, it cannot be changed later.

If a cluster is changed from being completely managed by etcd-druid to (in your words) manually managed then we cannot just clean up the resources assuming that these resources are no longer required. Your assumption comes from the fact that in autonomous cluster use case it is not required. But manually managing an etcd cluster does not always mean that these resources are no longer required, right? 

If you are making that assumption then you are already tying your implementation to autonomous cluster use case requirements, is it not?

api/core/v1alpha1/constants.go
unmarshall: >The main usecase is https://github.com/gardener/etcd-druid/issues/1071, but I didn't want to bake that into the code/docstring since the feature from druid should be as generic as possible.

This is not completely true. #1071 is for supporting autonomous cluster and cannot be clubbed with manual management of etcd cluster resources which is a more generic requirement as opposed to autonomous cluster use case which is a very specific requirement due to the implementation choices that gardener has currently made where it only wishes to use etcd-druid as a STS spec generator and nothing more than that (at least in the first iteration).

unmarshall: >Well, the docs do provide that. What would you suggest is a better way to really advertise the created components? Etcd status?

Sure the docs do provide information for operators and developers. However it is clear that they do not need to worry and should not worry about these resources as it is intrinsically managed by etcd-druid. Which is also the reason that we prevent any change to these resources via a validating webhook (etcd-component-protection webhook).

It has been documented only for transparency purpose so that an operator can check if there are issues in reconciliation due to which some resources are not created or updated successfully.

api/core/v1alpha1/constants.go
unmarshall: Actually your implementation assumes autonomous cluster use case only and based on that you make a LOT of assumptions but label it as `manual control for etcd cluster resources` which as i have explained elsewhere is not entirely true and correct.

internal/controller/etcd/reconcile_spec.go
ishan16696: I would like to make one more point, currently etcd-druid doesn't allow any user to delete any etcd-druid managed resources unless user explicitly disable the protection by annotating the Etcd resource with annotation `druid.gardener.cloud/disable-etcd-component-protection` but here isn't what we're doing is completely opposite and semantically incorrect ? we are allowing druid to itself cleaup the resources by providing special annotation which completely voilates and bypass the protection rule.
I will also vote for not allowing the migration from completely managed to manually managed for now by blindly deleting the resources.

shreyas-s-rao: Ok, if that's the case, how do we disallow changing it? Since it's an annotation and not a well-defined spec field, I cannot utilise CEL validations to check whether the annotation is added or removed (this is true for most metadata fields, expect for `name` and `generateName`). So I will now be forced to write a webhook for the Etcd resource, just to validate this annotation. Either that, or I will need to make it part of the spec for better control over immutability. What would be the better option here? @unmarshall @ishan16696

ishan16696: other option I would suggest let the user add the annotation but druid shouldn't take any action may be just print the warning that "currently migration from completely managed to manually managed is not supported".

shreyas-s-rao: The etcd reconciler is stateless, so it doesn't necessarily know whether the annotation was previously present or not in order to print such warnings. And it won't be right to rely on the resources themselves, because the user could have deployed those too.

api/core/v1alpha1/constants.go
shreyas-s-rao: Ok, `manual management of resources` may not the right term, but I think `external management of resources` still makes sense. ASC is a case of external management of runtime resources, since it's the `static pod translator` that crates the pod spec and gets it deployed it as a static pod on the node.

> It has been documented only for transparency purpose so that an operator can check if there are issues in reconciliation due to which some resources are not created or updated successfully.

Yes, that's correct. So I think we should keep this in the docs as is.

I can re-word "manual" to "external" management of runtime resources, if that sounds more apt.

api/core/v1alpha1/constants.go
shreyas-s-rao: As mentioned above, I can re-word it to `external` management of runtime resources, since that seems correct. WDYT?

api/core/v1alpha1/helper.go
```diff
@@ -165,3 +165,8 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
 	delete(etcdObjMeta.Annotations, DruidOperationAnnotation)
 	delete(etcdObjMeta.Annotations, GardenerOperationAnnotation)
 }
+
+// IsEtcdRuntimeComponentCreationEnabled checks if the creation of runtime components is not disabled for the Etcd resource.
```
unmarshall:
```suggestion
// IsEtcdRuntimeComponentCreationEnabled checks if the creation of runtime components is enabled for an Etcd resource.
```

internal/component/statefulset/builder.go
```diff
@@ -136,7 +136,11 @@ func (b *stsBuilder) getStatefulSetLabels() map[string]string {
 
 func (b *stsBuilder) createStatefulSetSpec(ctx component.OperatorContext) error {
 	err := b.createPodTemplateSpec(ctx)
-	b.sts.Spec.Replicas = ptr.To(b.replicas)
+	b.sts.Spec.Replicas = ptr.To[int32](0)
+	if druidv1alpha1.IsEtcdRuntimeComponentCreationEnabled(b.etcd.ObjectMeta) {
```
unmarshall: It is generally a good practice to have the uncommon usage within condition. So ideally i would have written it as 
```go
b.sts.Spec.Replicas = ptr.To(b.replicas)
if druidv1alpha1.IsEtcdRuntimeComponentCreationDisabled(b.etcd.ObjectMeta) {
   b.sts.Spec.Replicas = 0
}
```
Alternatively you can also do this.
```go
replicas = utils.IfConditionOr(druidv1alpha1.IsEtcdRuntimeComponentCreationDisabled(b.etcd.ObjectMeta), 0, etcd.Spec.Replicas)
}
```

internal/controller/etcd/reconcile_spec.go
```diff
@@ -99,6 +102,31 @@ func (r *Reconciler) syncEtcdResources(ctx component.OperatorContext, etcdObjKey
 	return ctrlutils.ContinueReconcile()
 }
 
+// cleanupEtcdResources cleans up the resources that are no longer required for the Etcd cluster.
```
unmarshall, line 105: can you add a little more context in doc string? Now we can understand the context but a few months later this will be lost. Why are these resources no longer required needs to be explicitly mentioned.

internal/controller/etcd/reconcile_spec.go
```diff
@@ -172,17 +200,40 @@ func (r *Reconciler) getOrderedOperatorsForPreSync() []component.Kind {
...
+			component.ServiceAccountKind,
+			component.RoleKind,
+			component.RoleBindingKind,
+			component.MemberLeaseKind,
+			component.SnapshotLeaseKind,
+		}
+	}
+
+	// add the rest of the operators that are always needed for the etcd cluster
+	operators = append(operators,
```
unmarshall, line 222: is client and peer service really required in case of autonomous cluster use case? Similarly is PDB component required as these are static pods where does PDB fit?

internal/controller/etcd/reconcile_status.go
unmarshall, line 31: this log is not clear.

internal/controller/etcd/reconcile_status.go
```diff
@@ -27,11 +27,20 @@ func (r *Reconciler) reconcileStatus(ctx component.OperatorContext, etcdObjectKe
...
 	}
+	mutateETCDStatusStepFns = append(mutateETCDStatusStepFns,
```
unmarshall: why is this always required? Do we have a VPA used in autonomous use case and will this be able to scale a static pod?

internal/webhook/etcdcomponentprotection/handler.go
```diff
@@ -164,6 +173,13 @@ func isObjManagedByDruid(objMeta metav1.ObjectMeta) bool {
 	return hasLabel && managedBy == druidv1alpha1.LabelManagedByValue
 }
 
+func isRuntimeComponent(requestGK schema.GroupKind) bool {
```
unmarshall, line 178: what does `isRuntimeComponent` mean actually?

unmarshall (changes requested):

internal/component/statefulset/builder.go
shreyas-s-rao: The reason I chose `IsEtcdRuntimeComponentCreationEnabled` was because it's easier to read since enablement of runtime components results in creation of additional components by the etcd controller, rather than reading `IsEtcdRuntimeComponentCreationDisabled` and negating it our head all the time. WDYT?

internal/controller/etcd/reconcile_spec.go
shreyas-s-rao: Done

internal/controller/etcd/reconcile_spec.go
shreyas-s-rao: Valid points. PDB is definitely not required. The service could still be used by the translator to expose specific ports from the etcd static pods. 
@rfranzke are client and peer service resources created by druid used in the ASC setup?

internal/controller/etcd/reconcile_status.go
shreyas-s-rao: It simply states that etcd status checks are skipped since runtime components have not been created, meaning there are no running pods or leases to fetch any status information about the etcd cluster.

internal/controller/etcd/reconcile_status.go
shreyas-s-rao: We don't know. Maybe it could be used. Since druid doesn't know how exactly the pod template is realised as actual running pods, we cannot make any assumptions about the selector. And if a VPA is added for the Etcd resource without `status.selector` set, it throws an error and cannot scale the etcd pod resources.

internal/webhook/etcdcomponentprotection/handler.go
shreyas-s-rao: Whether the current resource for which the webhook has been triggered is a runtime component for the etcd cluster. I'll add a docstring to make this clear.

internal/controller/etcd/reconcile_spec.go
rfranzke: No, at the moment they are not used. Hence, you could remove them for now. When we come to HA, we have to figure out how the communication shall work.

internal/controller/etcd/reconcile_status.go
rfranzke: We don't have a VPA for now (even if we have, I don't see how it could scale the static pods 👀)

internal/controller/etcd/reconcile_spec.go
shreyas-s-rao: Sure, thanks Rafael.

shreyas-s-rao: @rfranzke I've made the change in https://github.com/gardener/etcd-druid/pull/1117/commits/7519be5fb7f86ba0f1ce4da354b2bf2af605754a, so PDB, client service and peer service are not created for the Etcd when runtime components are disabled. Additionally this also removed `sts.Spec.ServiceName` field, as well as `--service-endpoints` CLI flag from the backup sidecar. Please check whether this works well with the translated static pods in the ASC setup.

internal/controller/etcd/reconcile_status.go
shreyas-s-rao: I still somehow feel that pod selector should be part of the status, for any possible usage of `/scale` subresource, unless @unmarshall you feel otherwise.

internal/controller/etcd/reconcile_spec.go
rfranzke: I don't think this would cause problems, but if you want to check it (I don't have the bandwidth until mid of next week), feel free to use https://github.com/gardener/gardener/pull/12391. You could run the e2e test ([ref](https://github.com/gardener/gardener/blob/master/docs/deployment/getting_started_locally_with_gardenadm.md#running-e2e-tests-for-gardenadm)). Generally, the static pod translator doesn't use the `Service`s or PDBs, and communication happens directly via the host network. Hence, I don't anticipate any issues given your description.

internal/controller/etcd/reconcile_status.go
unmarshall: Lets keep the reconciliation steps to what is required today. As and when more resources/configuration is required then we incrementally add/enable them.

shreyas-s-rao: Sure, makes sense.

I have made the fix in https://github.com/gardener/etcd-druid/pull/1117/commits/b3f0e481946fe40b96c7d54e79f2776d23cb9578. PTAL

PR #1119, opened 2025-06-05
Add support to create and migrate etcd clusters to use `emptyDir` volumes for etcd data directories.

### LINKED ISSUE #887: Support Ephemeral Storage
Please support the operation of ETCD with ephemeral [persistent volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes) (sounds like a contradiction), e.g. [hostpath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) or better/safer yet [local](https://kubernetes.io/docs/concepts/storage/volumes/#local), so that network attached persistent volumes can be avoided that are often a scarce machine resource (e.g. AWS can only attach 26 resp. 32 volumes for most machine types; Alicloud and Azure even less).

**Why is this needed**:
We observe that machines can rarely be fully utilised because of the high ratio of pods-with-volumes to pods-without-volumes in a Gardener managed shoot cluster control plane. If the ETCD for events could be configured to avoid network attached persistent volumes, we could improve the machine utilisation considerably (at the expense of only limited additional network costs to "catch up" when a pod is moved to another node).

**Considerations**:
- Losing 1 of 1 pods (non-HA) or 2 of 3 pods (HA) will result in an unrecoverable permanent quorum loss. Because without  network attached persistent volumes this could happen more frequently, ETCD druid should detect that and in the case of ephemeral persistent volumes, discard the statefulset and recreate it from scratch (in the context of events, this seems acceptable in many cases as the default events TTL is anyway only 1h and events are no critical/essential resource for the operation of a cluster).
- While backup and restore can be added (later), it doesn't have to be added right from the start. Whoever uses ETCD druid should have the liberty to decide for ephemeral persistent volumes.
- In order to stick to stateful sets (we don't have to, but it would make things easier), we need to find a PV(C) type that would work for us, e.g. [local](https://kubernetes.io/docs/concepts/storage/volumes/#local). So we need to experiment with it and see whether it works as expected, can be dynamically configured (now multiple ETCD pods would need different local paths on the node), and also the cleanup works (data is deleted once the pod is descheduled from the node).

renormalize:

/area cost
/area storage
/kind api-change
/kind enhancement

Adds support to create, and migrate existing etcd clusters to use node storage by the means of `emptyDir` volumes, for etcd data directories. The inverse is explicitly not supported, and disallowed by etcd-druid.

This feature orphan deletes the `StatefulSet`, and recreates it with a different template, which causes a rolling update to occur, creating new etcd pods which use `emptyDir` volumes.

However, this strategy does not work when etcd clusters are to be migrated from `emptyDir` volumes to CSI volumes. This is a consequence of the behavior of `StatefulSet`s, which errors when the pod template using `emptyDir` volumes is changed to use `volumeClaimTemplates`.

**This feature is considered experimental, and is marked an alpha feature.**

Fixes #887 

**Release note**:

```feature operator
[ALPHA] etcd clusters can now be created with `emptyDir` volumes as the etcd data directories.
```
```feature operator
[ALPHA] Existing etcd clusters can now be migrated to use `emptyDir` volumes as the etcd data directories. The leftover PVs are not affected by etcd-druid - manual intervention is need to clean them up.
```
```feature operator
[ALPHA] etcd-druid *does not* support migrating etcd clusters back to CSI volumes, from `emptyDir` volumes; and explicitly disallows it.
```

### DIFF
--- a/test/e2e/etcd_multi_node_test.go
+++ b/test/e2e/etcd_multi_node_test.go
@@ -83,6 +83,11 @@ var _ = Describe("Etcd", func() {
 			checkForUnreadyEtcdMembers(ctx, cl, objLogger, etcd, 30*time.Second)
 			checkEtcdReady(ctx, cl, objLogger, etcd, multiNodeEtcdTimeout)
 
+			By("Performing emptyDir migration")
+			Expect(cl.Get(ctx, client.ObjectKeyFromObject(etcd), etcd)).To(Succeed())
+			setEmptyDirVolumeSource(etcd)
+			updateAndCheckEtcdWithOrphanDeletedSts(ctx, cl, objLogger, etcd, multiNodeEtcdTimeout)
+
 			By("Deleting debug pod")
 			Expect(client.IgnoreNotFound(cl.Delete(ctx, debugPod))).ToNot(HaveOccurred())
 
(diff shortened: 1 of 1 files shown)


modzilla99:
@renormalize any news on this? Would you mind rebasing? We would be very interested in this feature.

renormalize:
@modzilla99 thanks for showing interest in this feature. I will try to resume work on this PR as soon as possible. Hopefully by the start of next week. 

There's a tab bit more left to this, other than the rebase (and the documentation, of course)

renormalize:
@modzilla99 feel free to play around with this branch now. Merging this is subject to review capacity, but the functionality will be identical to this. As mentioned in the PR description, migration from CSI to `emptyDir` is supported, but the inverse is explicitly disallowed.

You would need to switch the `AllowEmptyDir` feature on, through the `OperatorConfiguration`.

Issue #708, opened 2023-10-12
Enhance Etcd spec to include snapshot compaction spec


shreyas-s-rao:

unmarshall:
@shreyas-s-rao  we already have [CompactionResources](https://github.com/gardener/etcd-druid/blob/794de68b2f41f8484de5e94b91ac7ec18c895acd/api/v1alpha1/etcd.go#L151) as part of the `Backup` spec. This is also used in the compaction reconciler. See [here](https://github.com/gardener/etcd-druid/blob/794de68b2f41f8484de5e94b91ac7ec18c895acd/internal/controller/compaction/reconciler.go#L350-L352)

shreyas-s-rao:
@unmarshall yes it should. It's up to the consumer to utilise this spec field to set resource requests/limits for the compaction job pods. @anveshreddy18 can you confirm if you added this to g/g when gardenlet creates the `Etcd` resources?

shreyas-s-rao:
@anveshreddy18 But it makes sense to utilise this spec field for g/g, rather than hardcoding the values in the reconciler. Can you please make this change?

ashwani2k:
[like]  Kumar, Ashwani reacted to your message:
________________________________

PR #1130, opened 2025-07-03
Trigger full snapshot when spinning up a compaction job is unviable

### LINKED ISSUE #502: [Feature] Support configurable `etcd-events-threshold` per etcd object
`etcd-druid` is designed to start a compaction job if the difference between the latest revision stored in the latest delta snapshot and the latest revision stored in the latest full snapshot is greater than a predefined threshold. This threshold has a default value of `1000000` and can be adjusted by a flag `--etcd-events-threshold` to `etcd-druid`.

The drawback with this approach is that `--etcd-events-threshold` is now a value at `etcd-druid` level and will be applied to all etcd objects in it's control. This is not ideal given that each `etcd` can be used in a different use case and some `etcd's` might desire a different `--etcd-events-threshold` value given it's use case.

This issue proposes to move `etcd-events-threshold` to be an etcd level parameter so that each etcd object can configure it's own compaction threshold value and have etcd-druid honour this value

This is needed given that etcd's can have various use cases and the compaction threshold value should be decided per individual etcd and not one blanket value for all `etcd's`

1. Add a new field to the etcd CR where a compaction threshold value can be defined
2. `compaction_lease_controller` would then read this value from the etcd object and use this value to decide whether or not to run a compaction job
3. In the event that a compaction threshold is not specified in the etcd object, `etcd-druid` can fall back to it's default threshold value `1000000`
4. The `etcd-druid` flag `--etcd-events-threshold` will still be present to give the option of modifying the default threshold value if at all desired

anveshreddy18:

/area disaster-recovery
/area backup
/area monitoring
/kind api-change
/kind enhancement

This PR enhances the compaction controller logic of creating a compaction job upon delta revisions reaching a certain threshold by introducing an additional logic of triggering a full snapshot of the etcd db when it is decided that creating a compaction job is unviable to compact the backups, so it falls back on triggering a full snapshot instead. 

To support the above process, a new field `triggerFullSnapshotThreshold` has been introduced under the `compactionSpec` in Etcd API. The field acts as the upper threshold for the number of delta revisions beyond which the compaction controller should stop creating a compaction job and instead trigger a full snapshot. 

It also triggers the full snapshot when the previous compaction job has gone into `DeadlineExceeded` state as identified by the compaction controller. Again, this is to ensure that we don't waste resources trying to compact when the previous compaction itself timed out due to large number of deltas. 

So taking a full snapshot is the easiest fallback mechanism that we can employ to make sure any successive restoration efforts by the `backup-restore` container are performed fast so that there is minimal downtime to the operating `etcd` cluster. 

The enhancement follows the technical discussion whose meeting minutes are located at https://github.com/gardener/etcd-druid/issues/708#issuecomment-2863310023

Edit: This PR also extends the EtcdStatus API with a new condition `ConditionTypeLastSnapshotCompactionSucceeded` that reports the status of last compaction activity. Compaction activity can either mean a compaction job, or a fullSnapshot in the case of unviability of triggering a compaction job.

---

> Note: There is an existing problem with metrics getting reported incorrectly because of the staleness in the informer cache, which the compaction controller relies on. However, the fix to that will come in a separate PR, which uses expectations to ensure that we don't operate on an event / object more than once due to staleness.

---

Fixes #502

**Release note**:

```feature operator
Trigger full snapshot when spinning up a compaction job is unviable, in order to expedite the successive restoration efforts by `etcd-backup-restore`.
```
```breaking operator
Snapshot compaction resources API has been changed in incompatible way. Refer the `snapshotCompactionSpec` section in the [document](https://github.com/gardener/etcd-druid/blob/master/docs/api-reference/etcd-druid-api.md#snapshotcompactionspec) to learn more on how to configure snapshot compaction controller.
```
```noteworthy operator
EtcdStatus.Conditions slice is extended with a new condition type `ConditionTypeLastSnapshotCompactionSucceeded` that reports the status of last compaction activity ( compaction job / fullSnapshot ). See https://github.com/gardener/etcd-druid/pull/1130 for more details.
```

### DIFF
--- a/internal/controller/compaction/snapshot.go
+++ b/internal/controller/compaction/snapshot.go
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	etcdbrFullSnapshotReqTimeout = 60 * time.Second
+	defaultEtcdFullSnapshotReqTimeout = 600 * time.Second
 )
 
 type httpClientInterface interface {
@@ -102,8 +102,13 @@ func newHTTPClient(ctx context.Context, cl client.Client, etcd *druidv1alpha1.Et
 		}
 	}
 
+	etcdFullSnapshotReqTimeout := defaultEtcdFullSnapshotReqTimeout
+	if etcd.Spec.Backup.EtcdSnapshotTimeout != nil {
+		etcdFullSnapshotReqTimeout = etcd.Spec.Backup.EtcdSnapshotTimeout.Duration
+	}
+
 	httpClient := &http.Client{
-		Timeout:   etcdbrFullSnapshotReqTimeout,
+		Timeout:   etcdFullSnapshotReqTimeout,
 		Transport: httpTransport,
 	}
 	return httpClient, httpScheme, nil


api/config/v1alpha1/defaults.go
```diff
@@ -129,6 +129,8 @@ const (
 	DefaultCompactionConcurrentSyncs = 3
 	// DefaultCompactionEventsThreshold is the default number of events that triggers compaction.
 	DefaultCompactionEventsThreshold = 1000000
+	// DefaultCompactionTriggerFullSnapshotThreshold is the default upper threshold for the number of etcd revisions before giving up on compaction job and triggering a full snapshot.
+	DefaultCompactionTriggerFullSnapshotThreshold = 3000000
```
unmarshall, line 133: If this is going to be a multiple of `EventThreshold` then i would recommend that you compute it as a multiple of `DefaultCompactionEventsThreshold`. This allows automatic change of the value if `DefaultCompactionEventsThreshold` undergoes a change.

api/config/v1alpha1/defaults.go
```diff
@@ -129,6 +129,8 @@ const (
 	DefaultCompactionConcurrentSyncs = 3
 	// DefaultCompactionEventsThreshold is the default number of events that triggers compaction.
 	DefaultCompactionEventsThreshold = 1000000
+	// DefaultCompactionTriggerFullSnapshotThreshold is the default upper threshold for the number of etcd revisions before giving up on compaction job and triggering a full snapshot.
```
unmarshall:
```suggestion
	// DefaultCompactionTriggerFullSnapshotThreshold is the default upper threshold for the number of etcd events before giving up on compaction job and triggering a full snapshot.
```

api/config/v1alpha1/defaults.go
unmarshall: Can you also provide some context on how this number was arrived? At some later point of time when we look at the code we would have lost the context on why this threshold value was chosen. Maybe we create a markdown for compaction in `docs` directory and give a detailed rational on why this value was chosen. Bring in aspects of resource consumption, time taken to complete compaction job and probability of preemption that is seen when trying to compact snapshots (delta + full) beyond a certain event threshold.

api/config/v1alpha1/types.go
```diff
@@ -166,6 +166,8 @@ type CompactionControllerConfiguration struct {
 	ConcurrentSyncs *int `json:"concurrentSyncs,omitempty"`
 	// EventsThreshold denotes total number of etcd events to be reached upon which a backup compaction job is triggered.
 	EventsThreshold int64 `json:"eventsThreshold"`
+	// TriggerFullSnapshotThreshold denotes the upper threshold for the number of etcd revisions before giving up on compaction job and triggering a full snapshot.
+	TriggerFullSnapshotThreshold int64 `json:"triggerFullSnapshotThreshold"`
```
unmarshall, line 170: we seem to have mixed conventions. While `ConcurrentSyncs` is optional and also a pointer, `TriggerFullSnapshotThreshold` and `EventsThreshold` is also optional but not a pointer. We have defined this as `*int64` for `CompactionSpec.TriggerFullSnapshotThreshold`

api/config/v1alpha1/validation/validation_test.go
```diff
@@ -355,12 +356,18 @@ func TestValidateCompactionControllerConfiguration(t *testing.T) {
...
-			expectedErrors: 1,
-
-			matcher: ConsistOf(PointTo(MatchFields(IgnoreExtras, Fields{"Type": Equal(field.ErrorTypeInvalid), "Field": Equal("controllers.compaction.eventsThreshold")}))),
+			name:            "should forbid event threshold less than zero",
+			enabled:         true,
+			eventsThreshold: ptr.To(int64(-1)),
+			expectedErrors:  1,
+			matcher:         ConsistOf(PointTo(MatchFields(IgnoreExtras, Fields{"Type": Equal(field.ErrorTypeInvalid), "Field": Equal("controllers.compaction.eventsThreshold")}))),
+		},
+		{
```
unmarshall, line 372: The validation returns an error when `TriggerFullSnapshotThreshold` is <= 0 but there is no test if it is zero. Can you add that test as well?

internal/controller/compaction/reconciler.go
```diff
@@ -59,12 +62,25 @@ const (
...
+	// defaultCompactionJobCpuRequests defines the default cpu requests for the compaction job
+	defaultCompactionJobCpuRequests = resource.MustParse("600m")
+	// defaultCompactionJobMemoryRequests defines the default memory requests for the compaction job
+	defaultCompactionJobMemoryRequests = resource.MustParse("3Gi")
+	// lastJobCompletionReason is used to store the reason of the last completed job.
+	lastJobCompletionReason *string
+)
+
+// FullSnapshotTriggerFunc defines a function type for triggering a full snapshot.
+type FullSnapshotTriggerFunc func(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd) error
```
unmarshall: why is this required?

internal/controller/compaction/reconciler.go
```diff
@@ -59,12 +62,25 @@ const (
...
 // Reconciler reconciles compaction jobs for Etcd resources.
 type Reconciler struct {
 	client.Client
-	config      druidconfigv1alpha1.CompactionControllerConfiguration
-	imageVector imagevector.ImageVector
-	logger      logr.Logger
+	config              druidconfigv1alpha1.CompactionControllerConfiguration
+	imageVector         imagevector.ImageVector
+	logger              logr.Logger
+	FullSnapshotTrigger FullSnapshotTriggerFunc
```
unmarshall: why is this required? It is a locally defined function and that should be sufficient, right?

internal/controller/compaction/reconciler.go
```diff
@@ -274,6 +277,102 @@ func (r *Reconciler) delete(ctx context.Context, logger logr.Logger, etcd *druid
...
+		if compactionSpec.TriggerFullSnapshotThreshold != nil {
+			triggerFullsnapshotThreshold = *compactionSpec.TriggerFullSnapshotThreshold
+		}
+	}
+
+	// Trigger full snapshot if the delta revisions over the last full snapshot are more than the configured upper threshold
+	// or if the last job completion reason is DeadlineExceeded.
+	// This is to ensure that we avoid spinning up compaction jobs even when we know that the probability of it getting succeeded is very low due to the large number of revisions.
+	// This avoids unnecessary resource consumption and delays in the compaction process
+	if (lastJobCompletionReason != nil && *lastJobCompletionReason == batchv1.JobReasonDeadlineExceeded) || accumulatedEtcdRevisions >= triggerFullsnapshotThreshold {
```
unmarshall: It would be beneficial if we introduce a condition on the `Etcd` resource which records the last compaction job state and rely on that instead as it will always be persisted. It also provides better observability as one can see the compaction status for a specific etcd resource directly as a condition on the same resource.

internal/controller/compaction/reconciler.go
```diff
@@ -274,6 +277,102 @@ func (r *Reconciler) delete(ctx context.Context, logger logr.Logger, etcd *druid
 	}, nil
 }
 
+func (r *Reconciler) createCompactionJobOrTriggerFullSnapshot(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd, accumulatedEtcdRevisions int64, compactionJobName string, job *batchv1.Job) (ctrl.Result, error) {
+	var err error
+	eventsThreshold := r.config.EventsThreshold
+	triggerFullsnapshotThreshold := r.config.TriggerFullSnapshotThreshold
```
unmarshall: rename the variable to `triggerFullSnapshotThreshold`

internal/controller/compaction/reconciler.go
unmarshall: if the last job failed due to deadline exceeded and you have not yet crossed the `fullSnapshotEventThreshold` and now you trigger a full snapshot then you are effectively not honoring the full snapshot threshold. Job failing due to deadline exceeded is not certainly an indication that the compaction job will now never succeed unless a full snapshot is taken. This is why you introduced `fullSnapshotEventThreshold` in the first place.

internal/controller/compaction/reconciler.go
```diff
@@ -274,6 +277,102 @@ func (r *Reconciler) delete(ctx context.Context, logger logr.Logger, etcd *druid
...
+		logger.Info("Triggering full snapshot",
+			"namespace", etcd.Namespace,
+			"name", compactionJobName,
+			"accumulatedRevisions", accumulatedEtcdRevisions,
+			"triggerFullSnapshotThreshold", triggerFullsnapshotThreshold,
+			"reason", reason)
+		// Trigger full snapshot
+		if err = r.FullSnapshotTrigger(ctx, logger, etcd); err != nil {
+			recordFullSnapshotsTriggered(druidmetrics.ValueSucceededFalse, etcd.Namespace)
+			return ctrl.Result{
```
unmarshall: why would you requeue after 10s and not use the exponential backoff that the controller-runtime already provides when an error is returned as part of ctrl.Result?

internal/controller/compaction/reconciler.go
```diff
@@ -274,6 +277,102 @@ func (r *Reconciler) delete(ctx context.Context, logger logr.Logger, etcd *druid
 	}, nil
 }
 
+func (r *Reconciler) createCompactionJobOrTriggerFullSnapshot(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd, accumulatedEtcdRevisions int64, compactionJobName string, job *batchv1.Job) (ctrl.Result, error) {
```
unmarshall: break this function into 2 - one which takes a full snapshot and another that creates a compaction Job.

internal/controller/compaction/reconciler.go
```diff
@@ -274,6 +277,102 @@ func (r *Reconciler) delete(ctx context.Context, logger logr.Logger, etcd *druid
...
+		if err = r.FullSnapshotTrigger(ctx, logger, etcd); err != nil {
+			recordFullSnapshotsTriggered(druidmetrics.ValueSucceededFalse, etcd.Namespace)
+			return ctrl.Result{
+				RequeueAfter: 10 * time.Second,
+			}, fmt.Errorf("error while triggering full snapshot: %v", err)
+		}
+		recordFullSnapshotsTriggered(druidmetrics.ValueSucceededTrue, etcd.Namespace)
+		// Reset the last job completion reason after triggering full snapshot
+		lastJobCompletionReason = nil
+		return ctrl.Result{Requeue: false}, nil
```
unmarshall: you do not need to set `Requeue` to false explicitly. Also this is now a deprecated field.

api/config/v1alpha1/defaults.go
unmarshall: be consistent and used events or revisions. For `DefaultCompactionEventsThreshold` we use events. If we are going to also measure thresholds via number of events then we should adjust the doc string and also rename this variable to `DefaultCompactionTriggerFullSnapshotEventsThreshold`

internal/controller/compaction/reconciler.go
```diff
@@ -274,6 +277,102 @@ func (r *Reconciler) delete(ctx context.Context, logger logr.Logger, etcd *druid
...
+
+	return ctrl.Result{Requeue: false}, nil
+}
+
+// triggerFullSnapshot triggers a full snapshot for the given Etcd resource's associated etcd cluster.
+func (r *Reconciler) triggerFullSnapshot(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd) error {
+	httpScheme := "http"
+	httpTransport := &http.Transport{}
+
+	if tlsConfig := etcd.Spec.Backup.TLS; tlsConfig != nil {
```
unmarshall: move the construction of HTTP client to its own function.

internal/controller/compaction/utils.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type httpClientInterface interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+func fullSnapshot(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd, httpClient httpClientInterface, httpScheme string) error {
```
unmarshall: why is this function in utils.go? Is it used anywhere other than compaction?
instead create a new file and name it `snapshot.go` and move all functions related to taking a snapshot there.

unmarshall (changes requested):

api/config/v1alpha1/types.go
anveshreddy18: Yeah but can we leave it like that because if these fields are made pointers, then we would have to make another level of defaulting to some value in the compaction reconciler code, like `eventsThreshold := ptr.Deref(r.config.EventsThreshold, <default_value>)`

There are already too many levels of defaults we have set at different layers. These fields are set at `Etcd` level, druid configuration level, and a defaulting at druid level, now this.. I feel this makes it more complicated to follow and know which level is being set. And this is not at the `Etcd` API level, but at druid configuration level. Considering these, I feel there's no harm in having them as values. Wdyt?

internal/controller/compaction/reconciler.go
anveshreddy18: Right now we have a default value of `3M` events for the `triggerFullSnapshotThreshold`, which is configured at the druid level, although we provide an option to configure this from the `Etcd` level, we do not provide any means to calculate and arrive at this number, so users may just continue using the default that we have set, but this may not cater to the independent needs of all users, also because we currently look at just no.of events, but the size of the event also matters, the size of the <key , value> pair can range from kbs to `1.5MB`. If that's the case, then having the `DeadlineExceeded` can be a fairly good indication of compaction job never succeeding. Since each cluster is different, the default threshold cannot cater to the independent needs, so this deadline exceeded acts as a nice indicator to make the decision of taking full snapshot.

internal/controller/compaction/reconciler.go
anveshreddy18: The only reason was that if a transient network issue was present for some time, then I thought a periodic retry would make sense, which if left to exponential backoff may sometimes take long after some initial retries fail due to mentioned issue. But no hard opinion honestly. Anything looks fine to me.

Edit: I've changed it to use exponential backoff

api/config/v1alpha1/defaults.go
anveshreddy18: Addressed in https://github.com/gardener/etcd-druid/pull/1130/commits/d8ab94ed872334d3777c4cfab87f14da5478761d

api/config/v1alpha1/validation/validation_test.go
anveshreddy18: Added now

internal/controller/compaction/reconciler.go
anveshreddy18: Done

internal/controller/compaction/reconciler.go
anveshreddy18: Done

internal/controller/compaction/reconciler.go
anveshreddy18: Have separated out the full snapshot process into it's own function, but left the compaction job creation in it because now it's already very small.

internal/controller/compaction/reconciler.go
anveshreddy18: Done

internal/controller/compaction/utils.go
anveshreddy18: Thanks. I've now moved it to `snapshot.go` file and corresponding tests to `snapshot_test.go` file.

Also moved metrics related functionality that's present in `reconciler.go` to `metrics.go` file.

api/config/v1alpha1/defaults.go
anveshreddy18: > If this is going to be a multiple of EventThreshold then i would recommend that you compute it as a multiple of DefaultCompactionEventsThreshold

I'm not sure if we want to couple them in that way, we arrived at this number from our observations of busy clusters so far, but I can't say if it's always going to be a multiple of `eventsThreshold`. 


>Can you also provide some context on how this number was arrived?

In our discussion to finalise the approach for compaction, we saw some busy clusters and made this rough approximation, so I can mention that in the doc. But I don't have any precise calculation, nor I think it can be made because each cluster and their usage is independent.

internal/controller/compaction/reconciler.go
anveshreddy18: This is just passing the full snapshot triggering function as a value in the reconciler struct so that the controller can be tested in integration tests by injecting the testing full snapshot function while constructing the reconciler [here](https://github.com/gardener/etcd-druid/blob/d8ab94ed872334d3777c4cfab87f14da5478761d/test/integration/controllers/compaction/compaction_suite_test.go#L62). The tests are located [here](https://github.com/gardener/etcd-druid/blob/d8ab94ed872334d3777c4cfab87f14da5478761d/test/integration/controllers/compaction/reconciler_test.go#L291)

internal/controller/compaction/reconciler.go
anveshreddy18: Addressed in https://github.com/gardener/etcd-druid/pull/1130#discussion_r2221576484

internal/controller/compaction/reconciler.go
anveshreddy18: @shreyas-s-rao @ishan16696 would you like to add any points here. It makes sense to me, if you don't have any objections as well, then I can go ahead and make these changes.

api/core/v1alpha1/conditions.go
```diff
@@ -32,6 +32,13 @@ const (
 	ConditionCheckError ConditionStatus = "ConditionCheckError"
 )
 
+const (
+	// ConditionReasonFullSnapshotTakenSuccessfully is the reason for a successful full snapshot.
+	ConditionReasonFullSnapshotTakenSuccessfully string = "FullSnapshotTakenSuccessfully"
+	// ConditionReasonFullSnapshotError is the reason for a failed full snapshot.
+	ConditionReasonFullSnapshotError string = "ErrorTriggeringFullSnapshot"
+)
```
shreyas-s-rao: I would expect we would have more reasons like `CompactionJobSucceeded`, `CompactionJobFailed`, etc and publish those as well from the compaction controller, not just when full snapshot is triggered.

internal/controller/compaction/reconciler.go
```diff
@@ -59,12 +59,22 @@ const (
 	jobFailed
 )
 
+var (
+	// defaultCompactionJobCpuRequests defines the default cpu requests for the compaction job
+	defaultCompactionJobCpuRequests = resource.MustParse("600m")
```
shreyas-s-rao:
```suggestion
	defaultCompactionJobCPURequests = resource.MustParse("600m")
```

internal/controller/compaction/reconciler.go
```diff
@@ -79,12 +89,13 @@ func NewReconciler(mgr manager.Manager, config druidconfigv1alpha1.CompactionCon
 // NewReconcilerWithImageVector creates a new reconciler for Compaction with an ImageVector.
 // This constructor will mostly be used by tests.
 func NewReconcilerWithImageVector(mgr manager.Manager, config druidconfigv1alpha1.CompactionControllerConfiguration, imageVector imagevector.ImageVector) *Reconciler {
-	return &Reconciler{
+	reconciler := &Reconciler{
```
shreyas-s-rao: You can return the Reconciler object as is

internal/controller/compaction/reconciler.go
```diff
@@ -213,65 +243,84 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
...
+		if compactionSpec.TriggerFullSnapshotThreshold != nil {
+			triggerFullSnapshotThreshold = *compactionSpec.TriggerFullSnapshotThreshold
+		}
+	}
+
+	// Trigger full snapshot if the delta revisions over the last full snapshot are more than the configured upper threshold
+	// or if the last job completion reason is DeadlineExceeded.
+	// This is to ensure that we avoid spinning up compaction jobs even when we know that the probability of it getting succeeded is very low due to the large number of revisions.
+	// This avoids unnecessary resource consumption and delays in the compaction process
+	if (lastJobCompletionReason != nil && *lastJobCompletionReason == batchv1.JobReasonDeadlineExceeded) || accumulatedEtcdRevisions >= triggerFullSnapshotThreshold {
```
shreyas-s-rao: What happens if for some reason druid was down, and when it came up the threshold for full-snapshot trigger had already been reached? At that time, there is no `lastJobCompletionReason` to check, so would it go ahead and run a compaction job now and only after that fails it attempts to trigger full snapshot?

internal/controller/compaction/snapshot.go
shreyas-s-rao, line 1: Add licence header

internal/controller/compaction/snapshot_test.go
shreyas-s-rao, line 1: Add licence header

internal/controller/compaction/status_compaction.go
shreyas-s-rao, line 1: Add licence header

internal/controller/compaction/status_compaction.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func (r *Reconciler) triggerFullSnapshotAndUpdateStatus(ctx context.Context, logger logr.Logger, etcd *druidv1alpha1.Etcd, accumulatedEtcdRevisions, triggerFullSnapshotThreshold int64) (ctrl.Result, error) {
+	latestSnapshotCompactionCondition := druidv1alpha1.Condition{
```
shreyas-s-rao: You can use a simpler variable name like `condition` since that's the only condition updated by this controller, and makes the code easier to read.

internal/controller/compaction/status_compaction.go
```diff
@@ -0,0 +1,97 @@
...
+	if fullSnapErr != nil {
+		latestSnapshotCompactionCondition.Status = druidv1alpha1.ConditionFalse
+		latestSnapshotCompactionCondition.Reason = druidv1alpha1.ConditionReasonFullSnapshotError
+		latestSnapshotCompactionCondition.Message = fmt.Sprintf("Error while triggering full snapshot for etcd %s/%s: %v", etcd.Namespace, etcd.Name, fullSnapErr)
+	} else {
+		latestSnapshotCompactionCondition.Status = druidv1alpha1.ConditionTrue
+		latestSnapshotCompactionCondition.Reason = druidv1alpha1.ConditionReasonFullSnapshotTakenSuccessfully
+		latestSnapshotCompactionCondition.Message = fmt.Sprintf("Full snapshot taken successfully for etcd %s/%s", etcd.Namespace, etcd.Name)
+	}
+	etcdStatusUPDErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
shreyas-s-rao:
```suggestion
	etcdStatusUpdateErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```

internal/controller/compaction/status_compaction.go
```diff
@@ -0,0 +1,97 @@
...
+		if err := r.Get(ctx, types.NamespacedName{Namespace: etcd.Namespace, Name: etcd.Name}, latestEtcd); err != nil {
+			return err
+		}
+		return r.updateCompactionJobEtcdStatusCondition(ctx, latestEtcd, latestSnapshotCompactionCondition)
+	})
+
+	if fullSnapErr != nil || etcdStatusUPDErr != nil {
+		var requeueErrReason error
+		if fullSnapErr != nil && etcdStatusUPDErr != nil {
+			requeueErrReason = fmt.Errorf("error while triggering full snapshot and updating compaction-fullSnapshot etcd status condition: %w", fullSnapErr)
```
shreyas-s-rao: Can we break this into two separate cases: one where there was an error in full snapshot, and a generic one where the etcd status update itself failed? It doesn't make a difference for the requeue, but makes the flow cleaner.

shreyas-s-rao (changes requested):
@anveshreddy18 thanks a lot for the well-written PR and for incorporating changes discussed offline. I have a few comments/suggestions. PTAL, thanks.

internal/controller/compaction/reconciler.go
anveshreddy18: Yes, that's what happens with the current code. 

I left it like this for now as I wasn't very sure about getting the condition from the `Etcd.Status`. We can do that instead of depending on the in-memory status, wdyt?

api/core/v1alpha1/conditions.go
anveshreddy18: >I would expect we would have more reasons like CompactionJobSucceeded, CompactionJobFailed, etc

They are present and being published in the Etcd.Status, just that the constants are defined elsewhere at :

https://github.com/anveshreddy18/etcd-druid/blob/5d8e075f2ccb71c2424da0b56ae2f09b7d8c8156/internal/metrics/metrics.go#L18-L31

internal/controller/compaction/status_compaction.go
anveshreddy18: Agree. I retained the prefix and suffix, just removed the middle `SnapshotCompaction` term. Hope it's fine.

internal/controller/compaction/snapshot.go
anveshreddy18: Added

internal/controller/compaction/reconciler.go
anveshreddy18: Even bigger issue is that this `lastJobCompletionReason` is currently not resource scoped, but since compaction controller acts on multiple Etcd clusters, it is incorrect to have a single variable to track the status, fetching it via `Etcd.Status` seems like the correct way to go.

api/config/v1alpha1/defaults.go
unmarshall: These are magic numbers which might make sense for some clusters but for others they do not make sense. Therefore does it not make sense to have this configuration as part of an `Etcd` custom resource?

api/config/v1alpha1/types.go
unmarshall: I do not follow your arguments.
There are just 2 levels:
- Per Etcd resource: This is done as part of `SnapshotCompactionSpec.TriggerFullSnapshotThreshold`
- At etcd-druid operator level as `CompactionControllerConfiguration.TriggerFullSnapshotThreshold`

The usage should be:
* If it is defined at `Etcd` resource level use it.
* If it not defined at `Etcd` resource level then check if there is something defined at etcd-druid operator level. If it is defined use it.
* If not defined anywhere, assume a default.

This needs to be consistent for all fields that are defined at 2 levels. In addition if there is going to be just one default value for all fields defined at 2 levels then they should be `optional`and therefore pointers.

Please also refer to recommendation by k8s upstream:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

api/core/v1alpha1/conditions.go
unmarshall: Please move all constants that are used as part of API status conditions to the API packages only.

internal/controller/compaction/reconciler.go
```diff
@@ -59,12 +59,22 @@ const (
...
 // Reconciler reconciles compaction jobs for Etcd resources.
 type Reconciler struct {
 	client.Client
-	config      druidconfigv1alpha1.CompactionControllerConfiguration
-	imageVector imagevector.ImageVector
-	logger      logr.Logger
+	config           druidconfigv1alpha1.CompactionControllerConfiguration
+	imageVector      imagevector.ImageVector
+	logger           logr.Logger
+	EtcdbrHTTPClient httpClientInterface
```
unmarshall, line 74: why is this exported and the type unexported?
I would recommend that you expose the type and have an unexported field instead. Create another constructor function to accept an exported HTTPClient interface that you defined.

internal/controller/compaction/snapshot.go
```diff
@@ -0,0 +1,137 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	etcdbrFullSnapshotReqTimeout = 30 * time.Second
+)
+
+type httpClientInterface interface {
```
unmarshall, line 30: Please export this type and rename this to `EtcdBackupRestoreHTTPClient` as this is only introduced for that purpose and has no other purpose currently.
Move this type to `reconciler.go` just after `Reconciler` struct type or before it - you choose.

internal/controller/compaction/reconciler.go
```diff
@@ -142,29 +148,58 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
 
 		// Check if there is one active job or not
 		if job.Status.Active > 0 {
+			logger.Info("Compaction job is currently running", "namespace", job.Namespace, "name", job.Name)
 			metricJobsCurrent.With(prometheus.Labels{druidmetrics.LabelEtcdNamespace: etcd.Namespace}).Set(1)
 			// Don't need to requeue if the job is currently running
 			return ctrl.Result{}, nil
```
unmarshall, line 148: Consider a case where a job is already running. If the number of deltas and full snapshot are large in number, chances are that the compaction job can take a couple of hours (default is DefaultCompactionActiveDeadlineDuration which is 3hrs). If the cluster is very busy and if the events accumulated have significantly increased and waiting any longer would anyways result in taking a full snapshot then should you wait for the current job to finish?

internal/controller/compaction/reconciler.go
```diff
@@ -142,29 +148,58 @@ func (r *Reconciler) reconcileJob(ctx context.Context, logger logr.Logger, etcd
...
 		}
+
 		// Set the current job count to 0 if the job is not active
 		metricJobsCurrent.With(prometheus.Labels{druidmetrics.LabelEtcdNamespace: etcd.Namespace}).Set(0)
+
+		// Check the job completion state and reason to capture the metrics and update the etcd status condition
+		var (
```
unmarshall: Move all this logic to a function which for an existing Job which has completed:
* inspects its status
* collects failure reason, job duration seconds
* Records job metrics
* Updates `Etcd` condition

Of course you should delegate each of the above to a dedicated function, some of these you already have defined.

internal/controller/compaction/reconciler.go
anveshreddy18: This is now addressed by fetching the condition from the Etcd.Status.

internal/controller/compaction/status_compaction.go
anveshreddy18: Done.

internal/controller/compaction/metrics.go
```diff
@@ -143,46 +144,52 @@ func recordSuccessfulJobMetrics(job *batchv1.Job) {
...
 			druidmetrics.LabelSucceeded:     druidmetrics.ValueSucceededFalse,
-			druidmetrics.LabelFailureReason: failureReason,
+			druidmetrics.LabelFailureReason: jobFailureReasonMetricLabelValue,
 			druidmetrics.LabelEtcdNamespace: job.Namespace,
 		}).Observe(durationSeconds)
 	}
 }
 
-func (r *Reconciler) fetchFailedJobMetrics(ctx context.Context, logger logr.Logger, job *batchv1.Job, jobCompletionReason string) (string, float64, error) {
+func (r *Reconciler) fetchFailedJobMetrics(ctx context.Context, logger logr.Logger, job *batchv1.Job, jobCompletionReason string) (string, string, float64, error) {
```
shreyas-s-rao, line 164: Add docstring about return parameters

internal/controller/compaction/reconciler.go
```diff
@@ -293,7 +284,7 @@ func (r *Reconciler) createCompactionJobOrTriggerFullSnapshot(ctx context.Contex
 	// or if the last job completion reason is DeadlineExceeded.
 	// This is to ensure that we avoid spinning up compaction jobs even when we know that the probability of it getting succeeded is very low due to the large number of revisions.
 	// This avoids unnecessary resource consumption and delays in the compaction process
-	if (lastJobCompletionReason != nil && *lastJobCompletionReason == batchv1.JobReasonDeadlineExceeded) || accumulatedEtcdRevisions >= triggerFullSnapshotThreshold {
+	if isLastCompactionConditionDeadlineExceeded(etcd) || accumulatedEtcdRevisions >= triggerFullSnapshotThreshold {
```
shreyas-s-rao: I would suggest re-fetching the etcd resource here before checking previous compaction condition state.

shreyas-s-rao (changes requested):
@anveshreddy18 thanks for making the requested changes. Couple of comments, but overall LGTM. PTAL

internal/controller/compaction/metrics.go
anveshreddy18: Added now, pls check.

internal/controller/compaction/reconciler.go
anveshreddy18: As discussed offline, this won't be necessary with the current flow.

internal/controller/compaction/reconciler.go
```diff
@@ -187,10 +206,106 @@ func isJobPresent(job *batchv1.Job) bool {
...
+	}
+
+	return nil
+}
+
+// waitForEtcdStatusConditionUpdate waits for the etcd status condition to reflect the expected state.
+// This ensures that the next reconciliation will not see a stale etcd status due to informer cache staleness.
+func (r *Reconciler) waitForEtcdStatusConditionUpdate(ctx context.Context, logger logr.Logger, etcdNamespace, etcdName string, expectedCondition druidv1alpha1.Condition) error {
+	logger.Info("Waiting for etcd status condition to be updated",
+		"etcdName", etcdName, "namespace", etcdNamespace,
```
shreyas-s-rao: The `logger` object should ideally be created with `etcdName` and `namespace` key-values at the very beginning itself, so we can avoid re-specifying these throughout the reconciler code, similar to https://github.com/gardener/etcd-druid/blob/34d685263a3d89e6cc0802dddc836356b6b1d8fe/internal/controller/etcd/reconcile_spec.go#L29

internal/controller/compaction/reconciler.go
```diff
@@ -278,26 +395,36 @@ func (r *Reconciler) triggerFullSnapshotAndUpdateStatus(ctx context.Context, log
...
 		latestEtcd := &druidv1alpha1.Etcd{}
 		if err := r.Get(ctx, types.NamespacedName{Namespace: etcd.Namespace, Name: etcd.Name}, latestEtcd); err != nil {
 			return fmt.Errorf("error while fetching etcd %s/%s: %w", etcd.Namespace, etcd.Name, err)
 		}
 		return r.updateCompactionJobEtcdStatusCondition(ctx, latestEtcd, latestCondition)
 	})
+	// If the etcd status update was successful, we will wait for the condition to be reflected in the cache.
+	if etcdStatusUpdateErr == nil {
+		if err := r.waitForEtcdStatusConditionUpdate(ctx, logger, etcd.Namespace, etcd.Name, latestCondition); err != nil {
+			// Don't fail the reconciliation - this is just a cache consistency check
```
shreyas-s-rao, line 405: So if the check fails to confirm the cache is updated within the given timeout period, then we simply continue with the rest of the flow?

internal/controller/compaction/reconciler.go
```diff
@@ -278,26 +395,36 @@ func (r *Reconciler) triggerFullSnapshotAndUpdateStatus(ctx context.Context, log
...
+		// Fetch the latest etcd resource to avoid conflict errors
 		latestEtcd := &druidv1alpha1.Etcd{}
 		if err := r.Get(ctx, types.NamespacedName{Namespace: etcd.Namespace, Name: etcd.Name}, latestEtcd); err != nil {
 			return fmt.Errorf("error while fetching etcd %s/%s: %w", etcd.Namespace, etcd.Name, err)
 		}
 		return r.updateCompactionJobEtcdStatusCondition(ctx, latestEtcd, latestCondition)
 	})
+	// If the etcd status update was successful, we will wait for the condition to be reflected in the cache.
+	if etcdStatusUpdateErr == nil {
+		if err := r.waitForEtcdStatusConditionUpdate(ctx, logger, etcd.Namespace, etcd.Name, latestCondition); err != nil {
```
shreyas-s-rao, line 404: What happens if druid restarts while waiting for the cache to be updated?

shreyas-s-rao (changes requested):
@anveshreddy18 thanks for making the requested changes. Most of the changes seem fine, but I do have a few doubts about the behavior of the reconciler upon restarts, as well as a suggestion for simpler logging throughout the reconciler. PTAL, thanks.

internal/controller/compaction/reconciler.go
anveshreddy18: A timeout of 30s is generally sufficient time for the cache to get updated with the latest events, so in most cases this won't be a problem. However, if it still times out, then we simply continue with the rest of the logic, which anyways requeue's because of the next steps in the code: Ref https://github.com/anveshreddy18/etcd-druid/blob/b2aa35cde50a281c72eaa94fbdb2a7898e3c7256/internal/controller/compaction/reconciler.go#L422-L434. 

Essentially we tried to reduce the issues caused by cache inconsistency which usually gets updated in seconds, but we still gave 30s time ( in the worst case, usually the check completes much faster) since compaction is not triggered very frequently, so as such it might not affect the controller a lot. 

This is a temporary measure since we ruled out the idea of hitting API Server directly by uncaching, the next best solution we thought out was to have some consistency check with a timeout, but in the worst case, if even that fails to get the cache updated, we might have some inconsistencies in the metrics, which was anyway happening previously. 

Once a proper solution is implemented in the form of expectations, then this temporary code can be removed.

internal/controller/compaction/reconciler.go
anveshreddy18: If a restart happens while waiting on this step, it behaves similar to what would've happened without this check, which is, when it starts up, the cache gets updated freshly, so the next reconciliation will operate on latest & correct data, which is what we want. I do not see any issue here 🤔. Do you have a specific concern in mind?

internal/controller/compaction/reconciler.go
anveshreddy18: I agree, this was a miss from my side. Thanks for the catch.

anveshreddy18: Addressed.

internal/controller/compaction/reconciler.go
shreyas-s-rao: Got it, thanks

internal/controller/compaction/reconciler.go
shreyas-s-rao: Thinking out loud here. Since this is just a status update, it doesn't really affect metrics. Upon restart, the worst thing the controller might do is re-trigger a full snapshot, if the snapshot lease has not been updated and it doesn't yet know that it already took a full snapshot in the previous reconciliation. But again, we rely on the compactor job to update the snapshot lease as soon as the snapshot succeeds, so even if the controller restarts without status update, we wouldn't actually take another full snapshot.

Now, the only downside is that the status condition remains stale until the next compaction job is triggered, which could be misleading to human operators (and also to gardenlet if it chooses to use this condition in the control plane health checks). There's no way for the controller to know what exactly was done in the previous reconciliation run, so there's no way to deterministically update the condition in the next reconciliation run. The only thing for human operators to do is to check the compaction controller logs and derive whether the condition is stale in this case. I think we should be ok with this. WDYT?

shreyas-s-rao (approved):
@anveshreddy18 thanks a lot for al the changes and incorporating suggestions. 
/lgtm

internal/controller/compaction/reconciler.go
anveshreddy18: You're right @shreyas-s-rao. this isn't a very big deal as you've pointed out. The logs are very descriptive of what is happening, so they can act as the backup in the case of a suspicion that the Etcd status is not upto date. Today, any external entity does not rely on this status information, but we should soon be taking a relook once the expectations solution is ready across the druid for all controllers, to make it more deterministic.

api/config/v1alpha1/defaults.go
anveshreddy18: These are made available to configure in the API, ref: 
https://github.com/anveshreddy18/etcd-druid/blob/1ee944a85175dd36203dd8fd8e62516cdaa339bc/api/core/v1alpha1/etcd.go#L201-L212

Just the defaults are mentioned in the `defaults.go` file that's all.

internal/controller/compaction/reconciler.go
anveshreddy18: This is being done now. I've extracted out all the noisy code from the main reconcile method into their own functions. PTAL.

internal/controller/compaction/reconciler.go
anveshreddy18: IMO we cannot assume that a long running compaction job would eventually fail, as there are a lot of compaction jobs that run for long and succeed in landscapes.

Lets take the example of same busy cluster, let say the accumulated events was increasing rapidly but the first time it reaches `eventsThreshold` < `diff` < `triggerFullSnapshotThreshold`, then it triggers a compaction job (because at this point, we still feel a compaction job would be sufficient) , the compaction process will then first fetch all the snapshot information from the backup bucket at this point [ref](https://github.com/gardener/etcd-backup-restore/blob/e945b4493d6cebce35914ce7228ccfe2a61b1801/cmd/miscellaneous.go#L50-L54) and start restoration of just these snapshots ( it doesn't keep considering new deltas as they come ) & applying them to embedded etcd. What we've decided to do before still holds true as the compaction process is only processing those snapshots which were there at the time of start of this job. So it can indeed succeed.

Let say while this compaction process is running, the `etcd-main` has taken few more deltas, so now let's assume it became `diff` > `triggerFullSnapshotThreshold`, if I understand your point right, you want to stop the running compaction process & instead trigger the full snapshot. But I have a different argument, if we let the compaction to get completed, and if it succeeds, then we wouldn't always need to even trigger full snapshot or even a compaction job because the `diff` can become very less due to the previous compaction. So we can wait for the compaction to finish.

Overall, I think making it more dynamic may get very complex very quickly.

ishan16696 (approved):

PR #1132, opened 2025-07-04
Fixes helm chart service template to correctly use .Values.enabledOperatorConfig

### LINKED ISSUE #1131: Fix the operator service template yaml to use .Values.enabledOperatorConfig

/area control-plane
/kind bug

**What happened**:
Operator Service template YAML file defined in `charts/templates/service.yaml` currently only uses deprecated configuration values defined in `values.yaml`. Specifically it uses `.Values.controllerManager`.

We should first check if `.Values.enabledOperatorConfig` is set to true then it should pick the ports configuration from `.Values.operatorConfig.server` else it should pick it up from `.Values.controllerManager.server` 

Currently it will not cause any issues because of 2 reasons:
* The values in old and new configuration is the same.
* Old configuration is only deprecated but not removed.

However, this needs to be fixed anyhow as its incorrect and if someone does change the ports when using operator config then the new ports will not be reflected correctly in the service.

unmarshall:

/area control-plane
/kind bug

Fixes the etcd-druid operator service template yaml to first check `.Values.enabledOperatorConfig` and based on the value pick the ports configuration from the relevant section in `values.yaml`.

Fixes #1131 

**Release note**:

```other operator
Fixes etcd-druid service template yaml to properly support operator configuration values defined in values.yaml
```

### DIFF
--- a/charts/templates/_helpers.tpl
+++ b/charts/templates/_helpers.tpl
@@ -87,3 +87,33 @@ etcd-druid-operator-configmap-{{ include "operator.config.data" . | sha256sum | 
 {{- define "webhook.etcdcomponentprotection.reconcilerServiceAccountFQDN" -}}
 {{- printf "system:serviceaccount:%s:%s" .Release.Namespace .Values.serviceAccount.name }}
 {{- end -}}
+
+{{ define "operator.service.ports" -}}
+{{- if .Values.enabledOperatorConfig }}
+{{- if .Values.operatorConfig.server.metrics.port }}
+- name: metrics
+  port: {{ .Values.operatorConfig.server.metrics.port }}
+  protocol: TCP
+  targetPort: {{ .Values.operatorConfig.server.metrics.port }}
+{{- end }}
+{{- if .Values.operatorConfig.server.webhooks.port }}
+- name: webhooks
+  port: {{ .Values.operatorConfig.server.webhooks.port }}
+  protocol: TCP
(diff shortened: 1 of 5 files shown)


charts/templates/_helpers.tpl
```diff
@@ -87,3 +87,27 @@ etcd-druid-operator-configmap-{{ include "operator.config.data" . | sha256sum |
...
+{{- end }}
+{{- if $metricsPort }}
+- name: metrics
+  port: {{ $metricsPort }}
+  protocol: TCP
+  targetPort: {{ $metricsPort }}
+{{- end }}
+{{- if $webhooksPort }}
+- name: webhooks
+  port: {{ .Values.operatorConfig.server.webhooks.port }}
```
Shreyas-s14: Thanks for the quick PR @unmarshall.
Just a small change:

Shreyas-s14: Here, the values are being taken from the operatorConfig instead of the variable defined above.
```suggestion
  port: {{ $webhooksPort }}
```

charts/templates/_helpers.tpl
```diff
@@ -87,3 +87,27 @@ etcd-druid-operator-configmap-{{ include "operator.config.data" . | sha256sum |
...
+  protocol: TCP
+  targetPort: {{ .Values.operatorConfig.server.webhooks.port }}
```
Shreyas-s14: Same as https://github.com/gardener/etcd-druid/pull/1132#discussion_r2185150945.

PR #1133, opened 2025-07-08
Fix etcd-wrapper related tls config flags

anveshreddy18:

/area user-management
/kind bug

The tls configuration related flags in the etcd container args in statefulset builder are incorrectly populated, where the `backup-restore` tls specific flags were dependent on etcd clientURL tls, but it should not depend on that and should instead depend on whether `etcd-backup-restore` server is TLS enabled or not. Have made the required changes in this PR.

Try configuring druid setup by just enabling backup-restore server TLS, it throws error as the client in the etcd container makes `http` request instead of `https` as the container's cli flags are incorrectly populated in druid. With these changes, it should work fine.

**Release note**:

```bugfix operator
Fix backup-restore TLS related CLI flags to etcd-wrapper.
```


shreyas-s-rao (approved):
@anveshreddy18 thanks a lot for the fix!
/lgtm

seshachalam-yv (approved):
/lgtm

PR #1134, opened 2025-07-09
Fix compaction e2e test setup in skaffold profile

anveshreddy18:

/area testing
/kind bug

The PR #1090 had introduced OperatorConfiguration but the skaffold profiles for the e2e test setup had an issue with an incorrect config field name, which the e2e tests depend upon. The field name was supposed to be `eventsThreshold` but was incorrectly named `etcdEventsThreshold` due to which the compaction e2e tests were failing consistently in PR pipelines. 

Code ref: https://github.com/gardener/etcd-druid/blob/f2552bdd84afa7f9b3ec1ffbc70c451865fff66c/skaffold.yaml#L69

**Release note**:

```bugfix developer
Fix incorrectly named config field in skaffold profile for e2e tests
```

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -66,7 +66,7 @@ profiles:
         path: /deploy/helm/releases/0/setValues/operatorConfig
         value:
           controllers.etcd.etcdStatusSyncPeriod: 5s
-          controllers.compaction.etcdEventsThreshold: "15"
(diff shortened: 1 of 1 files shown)


Shreyas-s14 (approved):
Thank you for the fix @anveshreddy18 . LGTM

PR #1135, opened 2025-07-15
migrate CICD-Pipeline to GitHub-Actions

ccwienk:
see: https://gardener.github.io/cc-utils/github_actions.html

/area delivery
/kind task
/kind technical-debt

Fixes #

**Release note**:

```other developer
migrate CICD-Pipeline to GitHub-Actions
```


.ci/component_descriptor
shreyas-s-rao, line 1: If this is no longer required, I think you can also remove https://github.com/gardener/etcd-druid/blob/master/.ci/hack/component_descriptor, since its not used anywhere else in the repo

.github/workflows/build.yaml
```diff
@@ -0,0 +1,108 @@
...
+    steps:
+      - uses: gardener/cc-utils/.github/actions/trusted-checkout@master
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.24'
+      - name: run-integration-tests
+        shell: bash
+        run: |
+          set -euo pipefail
+          .ci/test_integration
```
shreyas-s-rao: We need to run both [test](https://github.com/gardener/etcd-druid/blob/76eb742a53659290ea1a8d97939c3c71d3db4d20/.ci/pipeline_definitions#L14) and [integration-test](https://github.com/gardener/etcd-druid/blob/76eb742a53659290ea1a8d97939c3c71d3db4d20/.ci/pipeline_definitions#L16). Please include the `test` step here as well.

Also, we don't want to run these two steps for PRs (since that's already handled by prow), but only for head-update and release. Can you please split `non-release.yaml` into `head-update` and `pull-request`, so that we can call `test` and `integration-test` only for `head-update` and then also for `release`?

ccwienk: I added the `test`-script as well.

As for your second ask, I added a new input (run-tests), which can be used to either run, or not run tests + used a condition to skip tests for pullrequests. This approach has the benefit to introduce less boilerplate. If you still would prefer separate triggering workflows, we can of course also change it as you suggested (please check)

.github/workflows/build.yaml
```diff
@@ -0,0 +1,118 @@
...
+    steps:
+      - uses: gardener/cc-utils/.github/actions/trusted-checkout@master
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.24'
+      - name: run-integration-tests
+        shell: bash
+        run: |
+          set -euo pipefail
+          ${{ matrix.args.run }}
```
shreyas-s-rao, line 109: Since both `test` and `test_integration` are part of the same step, would they be run as separate steps or would they show up as one single line item in the github PR checks section?

.ci/component_descriptor
shreyas-s-rao: @ccwienk Can you also please remove this if it's no longer necessary for CI jobs (we don't use it anywhere else in the repo)?

.github/workflows/build.yaml
shreyas-s-rao: Makes sense, thanks. The new changes are easy to read and avoid duplication, so it's good.

.ci/component_descriptor
ccwienk: @shreyas-s-rao : sure (I just saw your response today - sorry for taking thus long).

.github/workflows/build.yaml
ccwienk: each instance of a "matrix"-job will show up separately (as you can, btw, even see for _this_ pullrequest in the checks-section :-))

shreyas-s-rao:
@ccwienk [concourse job](https://concourse.ci.gardener.cloud/teams/gardener/pipelines/etcd-druid-master/jobs/master-pull-request-job/builds/1438) is failing at `build` step. Should we be worried about this?

shreyas-s-rao:
Also, integration test fails in github actions, but passes in concourse and locally as well. Any idea why?

shreyas-s-rao:
@ccwienk thanks for the clarification.

PR #1155, opened 2025-08-13
Add validation for EtcdCopyBackupsTask

### LINKED ISSUE #1021: CRD validation for EtcdCopyBackupsTask

/area control-plane
/area open-source
/kind enhancement

Introduce CEL validations for `EtcdCopyBackupsTask` CRD. In #989 CEL validations were introduced for `Etcd` CRD. There is a need to also do the same for `EtcdCopyBackupsTask` CRD.

**Why is this needed**:
To ensure that invalid `EtcdCopyBackupsTask` CR's are checked even before an event is raised in the respective reconciler in etcd-druid.

CaptainIRS:

/area control-plane
/area open-source
/kind enhancement

* Introduces CR validations for EtcdCopyBackupsTask.
* Validation for fields with metav1.Duration are made uniform across the repo following the upstream gardener/gardener regex (https://github.com/gardener/gardener/blob/master/pkg/apis/operator/v1alpha1/types_garden.go#L393)

Fixes #1021 

* Fields with unsigned integer and duration types are validated.
* Other validations (present in [./api/validation/etcdcopybackupstask.go](https://github.com/gardener/etcd-druid/blob/e88eafba4b049abff0526cf723e38a36c373b258/api/validation/etcdcopybackupstask.go)) which could not be implemented using markers/CEL and the reasoning behind it:
  * Validating that spec is immutable when deletionTimestamp is set: This is not possible as metadata.deletionTimestamp is not exposed for validation purposes:
    > The apiVersion, kind, metadata.name and metadata.generateName are always accessible from the root of the object and from any x-kubernetes-embedded-resource annotated objects. No other metadata properties are accessible.
    
    Source: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/
  * Validating provider prefix: it is a Gardener-specific validation
* The test helpers are a bit complicated because the Go client doesn't allow for invalid data to be sent to the apiserver to trigger the validation failures. So, the objects are marshaled to a map and patched with invalid data before being sent through the client.
* The duration field validation needs to be tested thoroughly because there are discrepancies between how the duration format is handled in the apiserver and elsewhere (Refer to kubernetes/apimachinery#131)
  * There are cases like "1d" which are allowed by the apiserver and stored in etcd, but throws a runtime error when a controller tries to access it ("d" for days is not allowed by the unmarshaling logic)

**Release note**:

```feature operator
Validations are added to `EtcdCopyBackupsTask`.
```


api/core/v1alpha1/etcdcopybackupstask.go
```diff
@@ -63,6 +65,8 @@ type WaitForFinalSnapshotSpec struct {
 	// Timeout is the timeout for waiting for a final full snapshot. When this timeout expires, the copying of backups
 	// will be performed anyway. No timeout or 0 means wait forever.
 	// +optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^(0|([0-9]+(\\.[0-9]+)?(s|m|h))+)$"
```
Shreyas-s14: Are we allowing values such as `10h2h3h4s3m1h` to be passed as valid duration? The `time` package in go does allow these values, but will the controller throw an error for these?

test/it/crdvalidation/etcdcopybackupstask/helper.go
```diff
@@ -0,0 +1,96 @@
...
+		currentObject = currentObject[key].(map[string]interface{})
+	}
+	currentObject[path[len(path)-1]] = value
+
+	return &unstructured.Unstructured{
+		Object: unstructuredObject,
+	}
+}
+
+// initial setup for setting up namespace and test environment
```
Shreyas-s14: Could you rephrase the function comment to something like:
`// setupTestEnvironment performs the initial setup for namespace and test environment`

test/it/crdvalidation/etcdcopybackupstask/helper.go
```diff
@@ -0,0 +1,96 @@
...
+	}, "5s", "1s").Should(Succeed())
+}
+
+// Helper to patch a key within the given Kubernetes object. This is needed to simulate invalid values
+// passed in through YAML to the api-server.
+func patchObject[T client.Object](object T, path []string, value interface{}) *unstructured.Unstructured {
+	// Convert the object to a vanilla Go map
+	unstructuredObject, err := runtime.DefaultUnstructuredConverter.ToUnstructured(object)
+	if err != nil {
+		panic(fmt.Sprintf("failed to convert EtcdCopyBackupsTask to unstructured object: %v", err))
```
renormalize: Any reason to panic here? Why don't we just return an error and check in the caller of `patchObject`? I understand it is a test but we don't typically use panics in the codebase, even with test utilities like this.

test/it/crdvalidation/etcdcopybackupstask/helper.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		typedObject := new(T)
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredObject.Object, typedObject); err != nil {
+			return fmt.Errorf("failed to convert unstructured object to %T: %w", typedObject, err)
+		}
+		return nil
+	}, "5s", "1s").Should(Succeed())
+}
+
+// Helper to patch a key within the given Kubernetes object. This is needed to simulate invalid values
```
Shreyas-s14: Could you change the function comment to start with the function name as well?

test/it/crdvalidation/etcdcopybackupstask/helper.go
```diff
@@ -0,0 +1,96 @@
...
+		Object: unstructuredObject,
+	}
+}
+
+// initial setup for setting up namespace and test environment
+func setupTestEnvironment(t *testing.T) (string, *WithT) {
+	g := NewWithT(t)
+	testNs := utils.GenerateTestNamespaceName(t, testNamespacePrefix)
+
+	t.Logf("successfully create namespace: %s to run test => '%s'", testNs, t.Name())
```
renormalize:
```suggestion
	t.Logf("Successfully created namespace: %s to run test => '%s'", testNs, t.Name())
```

test/it/crdvalidation/etcdcopybackupstask/validation_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/gardener/etcd-druid/test/it/assets"
+	"github.com/gardener/etcd-druid/test/it/setup"
+)
+
+func TestMain(m *testing.M) {
+	var (
+		itTestEnvCloser setup.DruidTestEnvCloser
+		err             error
+	)
+	itTestEnv, itTestEnvCloser, err = setup.NewDruidTestEnvironment("etcdcopybackuptast-validation", []string{assets.GetEtcdCopyBackupsTaskCrdPath()})
```
renormalize:
```suggestion
	itTestEnv, itTestEnvCloser, err = setup.NewDruidTestEnvironment("etcdcopybackuptask-validation", []string{assets.GetEtcdCopyBackupsTaskCrdPath()})
```

test/it/crdvalidation/etcdcopybackupstask/spec_test.go
```diff
@@ -0,0 +1,157 @@
...
+package etcdcopybackupstask
+
+import (
+	"testing"
+
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
+	"github.com/gardener/etcd-druid/test/utils"
+)
+
+func TestValidateSpecWaitForFinalSnapshotTimeout(t *testing.T) {
```
Shreyas-s14, line 17: Could you add a function comment here? Something like
`TestValidateSpecWaitForFinalSnapshotTimeout tests the spec.waitForFinalSnapshotTimeout field of the EtcdCopyBackup CR.`

test/it/crdvalidation/etcdcopybackupstask/spec_test.go
```diff
@@ -0,0 +1,157 @@
...
+			copyBackupTask := utils.CreateEtcdCopyBackupsTask(test.taskName, testNs, "aws", true)
+			patchedCopyBackupTask := patchObject(copyBackupTask, []string{
+				"spec", "waitForFinalSnapshot", "timeout",
+			}, test.value)
+			validatePatchedObjectCreation[*druidv1alpha1.EtcdCopyBackupsTask](g, patchedCopyBackupTask, test.expectErr)
+		})
+	}
+}
+
+func TestValidateSpecMaxBackupAge(t *testing.T) {
```
Shreyas-s14, line 88: Same as https://github.com/gardener/etcd-druid/pull/1155#discussion_r2276057321

test/it/crdvalidation/etcdcopybackupstask/spec_test.go
```diff
@@ -0,0 +1,157 @@
...
+			copyBackupTask := utils.CreateEtcdCopyBackupsTask(test.taskName, testNs, "aws", true)
+			patchedCopyBackupTask := patchObject(copyBackupTask, []string{
+				"spec", "maxBackupAge",
+			}, test.value)
+			validatePatchedObjectCreation[*druidv1alpha1.EtcdCopyBackupsTask](g, patchedCopyBackupTask, test.expectErr)
+		})
+	}
+}
+
+func TestValidateSpecMaxBackups(t *testing.T) {
```
Shreyas-s14, line 128: https://github.com/gardener/etcd-druid/pull/1155#discussion_r2276057321

test/it/crdvalidation/etcdcopybackupstask/helper.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Navigate to the specified path and set the value
+	currentObject := unstructuredObject
+	for _, key := range path[:len(path)-1] {
+		if _, exists := currentObject[key]; !exists {
+			currentObject[key] = make(map[string]interface{})
+		}
+		currentObject = currentObject[key].(map[string]interface{})
+	}
+	currentObject[path[len(path)-1]] = value
```
renormalize: Can you add a comment which explains that you're traversing the tree and initializing each field in 72, and then moving to the child in line 74. From first glance it wasn't obvious to me, at least.

Shreyas-s14 (commented):
Had a few small nits. PTAL!

test/it/crdvalidation/etcdcopybackupstask/helper.go
```diff
@@ -0,0 +1,96 @@
...
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredObject.Object, typedObject); err != nil {
+			return fmt.Errorf("failed to convert unstructured object to %T: %w", typedObject, err)
+		}
+		return nil
+	}, "5s", "1s").Should(Succeed())
+}
+
+// Helper to patch a key within the given Kubernetes object. This is needed to simulate invalid values
+// passed in through YAML to the api-server.
+func patchObject[T client.Object](object T, path []string, value interface{}) *unstructured.Unstructured {
```
renormalize: Can you rename `path` to `fieldPath` so it is makes it easier for the reader to understand what the parameter is for?
```suggestion
// Helper to patch a key within the given Kubernetes object. This is needed to simulate invalid values
// passed in through YAML to the api-server.
func patchObject[T client.Object](object T, fieldPath []string, value interface{}) *unstructured.Unstructured {
```

renormalize (changes requested):
Thanks for the PR @CaptainIRS. Great job with your first PR!

Regarding your comments to reviewers:

> Validating that spec is immutable when deletionTimestamp is set: This is not possible as metadata.deletionTimestamp is not exposed for validation purposes:

If the deletion timestamp is set, we do not need to check immutability and establish spec == status. The deletion timestamp being set indicates that the resource is to be deleted, and all its children are to be cleaned up. This isn't really a limitation in our case, and even if we could implement it, there is no reason to.

> Validating provider prefix: it is a Gardener-specific validation

I'm not so sure about this. The prefix is at the end of the day is part of the filepath of an object in remote object storage. Therefore, it would be correct to ensure that the prefix that the user specifies actually is allowed by the remote object store providers, and does not have any special characters which will not be allowed in filepath names.

For example, AWS S3 has a document on how to name objects: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html. Similarly for GCS https://cloud.google.com/storage/docs/objects#naming.

Consumers like gardener/gardener will obviously have their own syntax they enforce onto the prefix, but the prefix must work with any consumer, for any remote object storage provider that is supported by etcd-backup-restore.

In essence the prefix must be a valid filepath. We can add a validation for this.

api/core/v1alpha1/etcdcopybackupstask.go
renormalize: Interesting find @Shreyas-s14!

Shreyas-s14: Thanks!
The main issue is with the inconsistency with how the values are parsed, as @CaptainIRS pointed out in the PR description.
As of now, for the Etcd api, the fields strictly ensure `h,m,s` format. We'll have to decide a consistent format for all fields.

CaptainIRS: In the sync today, it was decided that validations for fields with metav1.Duration are to be made uniform across etcd-druid, following the upstream gardener/gardener regex (https://github.com/gardener/gardener/blob/master/pkg/apis/operator/v1alpha1/types_garden.go#L393)

test/it/crdvalidation/etcdcopybackupstask/helper.go
CaptainIRS: Thanks for catching this! Fixed.

CaptainIRS:
Since there are no major implications when there is no strict prefix validation, it is decided that we can skip out on adding the validations in the interest of avoiding breaking changes. The duration validation would not be a breaking change as the regex used in upstream gardener (the new validation) is more relaxed than the existing validation in druid.

renormalize (approved):

Shreyas-s14 (approved):
LGTM!

Issue #1146, opened 2025-08-07
Disallow compaction reconciliation of unnecessary events by setting tighter predicates


anveshreddy18:
**How to categorize this issue?**

/area quality
/area performance
/kind enhancement

**What would you like to be added**:

Currently the compaction controller registration logic allows unnecessary events on the owned resources ( Leases & Jobs in this case ) to be reconciled by the controller. However, these can be made more stricter by having tighter predicate conditions which only allows the necessary events to trigger reconciliation. 

To give more details, currently the controller watches for Jobs and Leases owned by Etcd but has loose restrictions with the help of predicates. 

- [x] The predicate for job allows events if the job status has changed, doesn't take into consideration which kind of job it is, see below: https://github.com/gardener/etcd-druid/blob/c7d1e687148ec2bdfcab54cd2e35e6dd6deb271a/internal/controller/compaction/register.go#L83 All that compaction controller needs to care about is compaction jobs, which are named in a specific format of `<etcdName>-compactor`. Ref: https://github.com/gardener/etcd-druid/blob/c7d1e687148ec2bdfcab54cd2e35e6dd6deb271a/api/core/v1alpha1/helper.go#L37-L40 So having a validation on the name is necessary. Although today the only kind of job Etcd owns is compaction job, but that need not be the case always. Being more specific about the kind of job ensures we don't run into problems when new kind of jobs are introduced.

- [ ]  Similarly, the predicate for the lease currently allows reconciliations for both full snapshot and delta snapshot lease's `Spec.HolderIdentity`. While it is required to allow events to reconcile when the delta snapshot lease holderIdentity updates, it is not required to be done for full snapshot lease updates in the current implementation, however this may change when compaction job is made more smarter to cancel compaction jobs whenever it sees a parallel full snapshot is taken by the scheduled snapshotting in the `etcd-backup-restore` server. See the reference below for the code location where this is allowed currently: https://github.com/gardener/etcd-druid/blob/c7d1e687148ec2bdfcab54cd2e35e6dd6deb271a/internal/controller/compaction/register.go#L43

**Why is this needed**:

The reason for proposing stricter rules is to reduce unnecessary load it adds on etcd-druid, and extra logs that it produces which can be cumbersome to navigate & debug.

PR #1158, opened 2025-08-18
Set Stricter Predicates For Compaction Controller

anveshreddy18:

/area performance
/area quality
/kind enhancement

This PR fixes the issue of unnecessary compaction controller reconciliations by setting stricter predicates to allow only those events that are required for the functioning of the controller. This is done to reduce the unnecessary load on druid & to make the logs more leaner.

Fixes point 1 of the issue #1146 

Try running compaction locally and you'll see a lot of reconciliations happen even when no important changes happens on the job that the controller should be aware of. Can add temporary debug logs in the predicates to see it in action. 

**Release note**:

```other operator
set stricter predicates for compaction controller : reduce unnecessary reconciliations triggered by job object.
```

### DIFF
--- a/internal/controller/compaction/register.go
+++ b/internal/controller/compaction/register.go
@@ -12,7 +12,6 @@ import (
 
 	batchv1 "k8s.io/api/batch/v1"
 	coordinationv1 "k8s.io/api/coordination/v1"
-	apiequality "k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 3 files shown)


shreyas-s-rao (approved):
@anveshreddy18 thanks a lot for the optimisation! 
/lgtm

ishan16696:
@anveshreddy18 this PR only address the point 1 of the issue, can you update PR description.

PR #1164, opened 2025-09-02
export testresults as inlined ocm-resource

heldkat:

/area compliance
/kind enhancement

Fixes #

**Release note**:

```other operator
export testresults as inlined ocm-resource
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -96,7 +96,7 @@ jobs:
       matrix:
         args:
           - run: .ci/test_integration
-          - run: .ci/test
+
     steps:
       - uses: gardener/cc-utils/.github/actions/trusted-checkout@master
       - uses: actions/setup-go@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/base.yaml
```diff
@@ -134,8 +134,23 @@ jobs:
...
-
+          mkdir tmp/blobs.d
+          ${{ matrix.args.run }} |& tee tmp/blobs.d/test-log.txt
+          tar czf tmp/blobs.d/test-log.tar.gz -C tmp/blobs.d test-log.txt
+      - name: add-test-results-to-component-descriptor
+        uses: gardener/cc-utils/.github/actions/export-ocm-fragments@master
+        with:
+          blobs-directory: /tmp/blobs.d
+          ocm-resources: |
+            - name: test-results
```
8R0WNI3, line 145: This won't work as both matrix runs will try to create the same OCM resource. Instead, you could use the matrix run name for the OCM resource name as well:
```yaml
- name: ${{ matrix.args.name }}-results
```

heldkat: Following your advice, as the `ocm-resources name` is used as a filtering option in subsequent pipelines, the `extraIdentity` was introduced to distinguish the two tests.

Issue #1022, opened 2025-03-10
Enhance detection, validation and potential recovery from etcd split brain


unmarshall:
**How to categorize this issue?**

/area control-plane
/area robustness
/area monitoring
/kind enhancement

**What would you like to be added**:
Recently when running gardener/gardener e2e tests etcd-cluster split brain issues were observed.
See [gardener-etcd channel](https://sap-ti.slack.com/archives/C0177NLL8V9/p1741081613999379?thread_ts=1739880572.704909&cid=C0177NLL8V9) for more information on the test failures.

**Test flow observations**
----------------------------------------------------------------------------------------------------
* HA shoot cluster gets created with hibernation enabled. This will create an `Etcd` resource with replicas=0.
* During the restoring phase of `Control-Plane-Migration` process, etcd-cluster started with replicas=1 in the target seed cluster. Restoration phase succeeds.
* We now expected etcd-clusters to be scaled to replicas=3 (since its a HA shoot cluster), instead we observed that they were scaled down to replicas=0 followed by scaling of the replicas=3. This results in a split brain issue due to the following reasons:
  * etcd-member-0 has an existing data directory so it disregards any change in the etcd-config (as a result of scaling it to 3 replicas) and loads the etcd-config from the existing db even though its initial-cluster-state=new. As a result etcd-member-0 creates its own cluster with replicas=1.
  * etcd-member-1 and etcd-member-2 are brand new members and they have initial-cluster-state=new and they do not have any data directory (a classic bootstrap case). These 2 members do not see Memer-0 yet as all 3 pods were started simultaneously. Thus etcd-member-1 and etcd-member-2 create their own cluster.
* Split brain problem is not visible via etcd-druid at the moment since ClusterID information is not available to it.
* In the tests we also see that VPA evicted etcd-events pods one at a time starting with etcd-events-0 and followed by etcd-events-2 and etcd-events-1.
* This caused quorum loss in one of these clusters resulting in AllMembersReady to become false and Ready to also become false.

**Issue with the the test in question** 
-----------------------------------------------------------------------------------------------------
It was changing the replicas of an etcd-cluster from 1->0->3. Typically we today only support the following cluster-size transitions:
  * 0->3 (bootstrap of an etcd-cluster)
  * 0->1 and 1->3 (scale-up from a single node to a multi-node etcd cluster)
  * 1->0 OR 3->0 (scale down of either a single or multi-node etcd cluster) 

_**Additional Note:**_
[...]

PR #1165, opened 2025-09-05
Add ability to detect split-brain/split-quorum scenarios

CaptainIRS:
/area control-plane
/area robustness
/area monitoring

Adds the ability to detect split-brain/split-quorum scenarios. This allows for operators to look at the member leases of etcds deployed by druid and determine if it is stuck in a split-brain or split-quorum scenario.

Related to #1022 

To ensure that this update doesn't break existing clusters, both an update and revert scenario was tested using https://gardener.github.io/etcd-druid/usage/managing-etcd-clusters.html#overwrite-container-oci-images. No issues were found in the test.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/internal/health/condition/check_ready.go
+++ b/internal/health/condition/check_ready.go
@@ -6,15 +6,52 @@ package condition
 
 import (
 	"context"
+	"strings"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
 
+	coordinationv1 "k8s.io/api/coordination/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type readyCheck struct{}
+type readyCheck struct {
+	client client.Client
+}
 
-func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
+const holderIdentitySeparator = ":"
+
+func extractClusterIdAndRole(holderIdentity *string) (*string, *druidv1alpha1.EtcdRole) {
+	if holderIdentity == nil {
+		return nil, nil
+	}
+	_, remainder, hasSeparator := strings.Cut(*holderIdentity, holderIdentitySeparator)
+	if !hasSeparator {
+		return nil, nil
+	}
+
+	before, after, hasSeparator := strings.Cut(remainder, holderIdentitySeparator)
+	var clusterIdString, roleString string
+	if hasSeparator {
+		clusterIdString = before
+		roleString = after
+	} else {
+		clusterIdString = ""
+		roleString = before
+	}
+
+	switch druidv1alpha1.EtcdRole(roleString) {
+	case druidv1alpha1.EtcdRoleLeader:
+		role := druidv1alpha1.EtcdRoleLeader
+		return &clusterIdString, &role
+	case druidv1alpha1.EtcdRoleMember:
+		role := druidv1alpha1.EtcdRoleMember
+		return &clusterIdString, &role
+	default:
+		return &clusterIdString, nil
+	}
+}
+
+func (r *readyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
 
 	// TODO: remove this case as soon as leases are completely supported by etcd-backup-restore
 	if len(etcd.Status.Members) == 0 {
@@ -48,6 +85,61 @@ func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
 		}
 	}
 
+	// Look for split-quorum/split-brain scenario.
+	leaseNames := druidv1alpha1.GetMemberLeaseNames(etcd.ObjectMeta, etcd.Spec.Replicas)
+	leases := make([]*coordinationv1.Lease, 0, len(leaseNames))
+	for _, leaseName := range leaseNames {
+		lease := &coordinationv1.Lease{}
+		if err := r.client.Get(ctx, client.ObjectKey{Namespace: etcd.Namespace, Name: leaseName}, lease); err != nil {
+			// Lease has not been created yet, so we pass on the split-quorum/split-brain check for now.
+			continue
+		}
+		leases = append(leases, lease)
+	}
+	if len(leases) == size {
+		// All leases are present, so we can check for split-brain/split-quorum scenario.
+		clusterIDs := make([]string, 0)
+		roles := make([]druidv1alpha1.EtcdRole, 0)
+		for _, lease := range leases {
+			clusterID, role := extractClusterIdAndRole(lease.Spec.HolderIdentity)
+			if clusterID != nil {
+				clusterIDs = append(clusterIDs, *clusterID)
+			}
+			if role != nil {
+				roles = append(roles, *role)
+			}
+		}
+		// ClusterIDs are not populated in old druid/backup-restore versions which do
+		// not support checking for split-brain/split-quorum, so we ignore such cases.
+		if len(clusterIDs) == size && len(roles) == size {
+			leaderCount := 0
+			for _, role := range roles {
+				if role == druidv1alpha1.EtcdRoleLeader {
+					leaderCount++
+				}
+			}
+			if leaderCount > 1 {
+				return &result{
+					conType: druidv1alpha1.ConditionTypeReady,
+					status:  druidv1alpha1.ConditionFalse,
+					reason:  "SplitBrainDetected",
+					message: "Split-brain detected",
+				}
+			}
+			firstClusterID := clusterIDs[0]
+			for _, clusterID := range clusterIDs[1:] {
+				if clusterID != firstClusterID {
+					return &result{
+						conType: druidv1alpha1.ConditionTypeReady,
+						status:  druidv1alpha1.ConditionFalse,
+						reason:  "SplitQuorumDetected",
+						message: "Split-quorum detected",
+					}
+				}
+			}
+		}
+	}
+
 	return &result{
 		conType: druidv1alpha1.ConditionTypeReady,
 		status:  druidv1alpha1.ConditionTrue,
@@ -57,6 +149,8 @@ func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
 }
 
 // ReadyCheck returns a check for the "Ready" condition.
-func ReadyCheck(_ client.Client) Checker {
-	return &readyCheck{}
+func ReadyCheck(client client.Client) Checker {
+	return &readyCheck{
+		client: client,
+	}
 }
--- a/internal/health/condition/check_ready_test.go
+++ b/internal/health/condition/check_ready_test.go
@@ -6,8 +6,15 @@ package condition_test
 
 import (
 	"context"
+	"fmt"
 
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
+	testutils "github.com/gardener/etcd-druid/test/utils"
+
+	coordinationv1 "k8s.io/api/coordination/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	. "github.com/gardener/etcd-druid/internal/health/condition"
 	. "github.com/onsi/ginkgo/v2"
@@ -16,7 +23,21 @@ import (
 
 var _ = Describe("ReadyCheck", func() {
 	Describe("#Check", func() {
-		var readyMember, notReadyMember, unknownMember druidv1alpha1.EtcdMemberStatus
+		const (
+			etcdName      = "etcd"
+			etcdNamespace = "etcd-test"
+			clusterID     = "0"
+			newClusterID  = "1"
+			member1ID     = "1"
+			member2ID     = "2"
+			member3ID     = "3"
+		)
+		var (
+			member1Name                                = fmt.Sprintf("%s-%d", etcdName, 0)
+			member2Name                                = fmt.Sprintf("%s-%d", etcdName, 1)
+			member3Name                                = fmt.Sprintf("%s-%d", etcdName, 2)
+			readyMember, notReadyMember, unknownMember druidv1alpha1.EtcdMemberStatus
+		)
 
 		BeforeEach(func() {
 			readyMember = druidv1alpha1.EtcdMemberStatus{
@@ -33,6 +54,10 @@ var _ = Describe("ReadyCheck", func() {
 		Context("when members in status", func() {
 			It("should return that the cluster has a quorum (all members ready)", func() {
 				etcd := druidv1alpha1.Etcd{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      etcdName,
+						Namespace: etcdNamespace,
+					},
 					Spec: druidv1alpha1.EtcdSpec{Replicas: 3},
 					Status: druidv1alpha1.EtcdStatus{
 						Members: []druidv1alpha1.EtcdMemberStatus{
@@ -42,7 +67,12 @@ var _ = Describe("ReadyCheck", func() {
 						},
 					},
 				}
-				check := ReadyCheck(nil)
+				member1Lease := createMemberLease(member1Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s:%s", member1ID, clusterID, druidv1alpha1.EtcdRoleLeader)))
+				member2Lease := createMemberLease(member2Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s:%s", member2ID, clusterID, druidv1alpha1.EtcdRoleMember)))
+				member3Lease := createMemberLease(member3Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s:%s", member3ID, clusterID, druidv1alpha1.EtcdRoleMember)))
+				existingObjects := mapToClientObjects([]*coordinationv1.Lease{member1Lease, member2Lease, member3Lease})
+				cl := testutils.CreateTestFakeClientForObjects(nil, nil, nil, nil, existingObjects)
+				check := ReadyCheck(cl)
 
 				result := check.Check(context.TODO(), etcd)
 
@@ -52,6 +82,10 @@ var _ = Describe("ReadyCheck", func() {
 
 			It("should return that the cluster has a quorum (members are partly unknown)", func() {
 				etcd := druidv1alpha1.Etcd{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      etcdName,
+						Namespace: etcdNamespace,
+					},
 					Spec: druidv1alpha1.EtcdSpec{Replicas: 3},
 					Status: druidv1alpha1.EtcdStatus{
 						Members: []druidv1alpha1.EtcdMemberStatus{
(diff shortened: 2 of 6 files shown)


gardener-robot-ci-3:
Thank you @CaptainIRS for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

internal/health/condition/check_ready.go
```diff
@@ -6,15 +6,52 @@ package condition
...
+	}
+
+	before, after, hasSeparator := strings.Cut(remainder, holderIdentitySeparator)
+	var clusterIdString, roleString string
+	if hasSeparator {
+		clusterIdString = before
+		roleString = after
+	} else {
+		clusterIdString = ""
+		roleString = before
```
ishan16696: just above this, if lease's holderIdentity doesn't have clusterID then you are returning nil, but here you are setting clusterID with empty string, why two behaviour ?

internal/health/condition/check_ready.go
```diff
@@ -48,6 +85,61 @@ func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
...
+			clusterID, role := extractClusterIdAndRole(lease.Spec.HolderIdentity)
+			if clusterID != nil {
+				clusterIDs = append(clusterIDs, *clusterID)
+			}
+			if role != nil {
+				roles = append(roles, *role)
+			}
+		}
+		// ClusterIDs are not populated in old druid/backup-restore versions which do
+		// not support checking for split-brain/split-quorum, so we ignore such cases.
```
ishan16696:
```suggestion
		// ClusterIDs are not populated by `etcd-backup-restore` version less than `v0.39.0`.
		// so, for which druid not support checking for split-brain/split-quorum, so we ignore such cases.
```

internal/health/condition/check_ready.go
```diff
@@ -48,6 +85,61 @@ func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
...
+					leaderCount++
+				}
+			}
+			if leaderCount > 1 {
+				return &result{
+					conType: druidv1alpha1.ConditionTypeReady,
+					status:  druidv1alpha1.ConditionFalse,
+					reason:  "SplitBrainDetected",
+					message: "Split-brain detected",
+				}
```
ishan16696: actually this condition is also not correct, take this example:
```
etcd-main-0                  111:xxxx:Member                                                           376d
etcd-main-1                  222:xxxx:Leader                                                           376d
etcd-main-2                  333:xxxx:Leader                                                           376d
```
In this case as `etcd-main-2` and `etcd-main-1` both are shown as leader but actually `etcd-main-2` which was previous leader and now it is down, so it's backup-restore unable to update it's lease, hence it's contain stale information.
But clearly it's not split brain.

internal/health/condition/check_ready_test.go
```diff
@@ -61,16 +95,112 @@ var _ = Describe("ReadyCheck", func() {
...
+				}
+				// Note that the cluster ID is not part of the HolderIdentity here, which is the old format used by backup-restore
+				member1Lease := createMemberLease(member1Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s", member1ID, druidv1alpha1.EtcdRoleLeader)))
+				member2Lease := createMemberLease(member2Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s", member2ID, druidv1alpha1.EtcdRoleMember)))
+				member3Lease := createMemberLease(member3Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s", member3ID, druidv1alpha1.EtcdRoleMember)))
+				existingObjects := mapToClientObjects([]*coordinationv1.Lease{member1Lease, member2Lease, member3Lease})
+				cl := testutils.CreateTestFakeClientForObjects(nil, nil, nil, nil, existingObjects)
+				check := ReadyCheck(cl)
+
+				result := check.Check(context.TODO(), etcd)
```
ishan16696:
```suggestion
				result := check.Check(context.TODO(), etcd)
```
may be declare one context and use it everywhere. Moreover, use `context.Background()` instead of `context.TODO()`

internal/health/condition/check_ready_test.go
```diff
@@ -61,16 +95,112 @@ var _ = Describe("ReadyCheck", func() {
...
+					},
+				}
+				member1Lease := createMemberLease(member1Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s:%s", member1ID, clusterID, druidv1alpha1.EtcdRoleLeader)))
+				member2Lease := createMemberLease(member2Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s:%s", member2ID, clusterID, druidv1alpha1.EtcdRoleMember)))
+				member3Lease := createMemberLease(member3Name, etcdNamespace, ptr.To(fmt.Sprintf("%s:%s:%s", member3ID, clusterID, druidv1alpha1.EtcdRoleLeader)))
+				existingObjects := mapToClientObjects([]*coordinationv1.Lease{member1Lease, member2Lease, member3Lease})
+				cl := testutils.CreateTestFakeClientForObjects(nil, nil, nil, nil, existingObjects)
+				check := ReadyCheck(cl)
+
+				result := check.Check(context.TODO(), etcd)
```
ishan16696: same as above and where ever it used.

internal/health/condition/check_ready_test.go
```diff
@@ -61,16 +95,112 @@ var _ = Describe("ReadyCheck", func() {
...
+
+				Expect(result.ConditionType()).To(Equal(druidv1alpha1.ConditionTypeReady))
+				Expect(result.Status()).To(Equal(druidv1alpha1.ConditionTrue))
+			})
+
+			It("should return that the cluster has a split-brain (there are 2 leaders simultaneously)", func() {
```
ishan16696: I think this tests need to be removed as this condition may not be the correct way to detect split-brain with surety.

internal/health/condition/check_ready.go
CaptainIRS: In this case should we report "Quorate"?

ishan16696: I would suggest, let's first define the condition for split-brain.
IMO, clusterID mismatch can be combined with 2 leader detection . wdyt ?
If, if we found both of the condition to be true then only declare it as split-brain.

internal/health/condition/check_ready.go
```diff
@@ -6,15 +6,52 @@ package condition
...
 )
 
-type readyCheck struct{}
+type readyCheck struct {
+	client client.Client
+}
+
+const holderIdentitySeparator = ":"
+
+func extractClusterIdAndRole(holderIdentity *string) (*string, *druidv1alpha1.EtcdRole) {
```
unmarshall: Please add a docstring clearly mentioning the revised format for the `holderIdentity`. As per the implementation you are opting to use `memberID:clusterID:role` right?

internal/health/condition/check_ready.go
```diff
@@ -6,15 +6,52 @@ package condition
...
+	if holderIdentity == nil {
+		return nil, nil
+	}
+	_, remainder, hasSeparator := strings.Cut(*holderIdentity, holderIdentitySeparator)
```
unmarshall: the correct order should have been `<clusterID>:<memberID>:<role>`
And the implementation can be simplified:

```go
func extractParts(s string) (string, string, string, err error) {
	splits := strings.Split(s, ":")
	if len(splits) < 2 || len(splits) > 3 {
		return "", "", "", fmt.Errorf("invalid format of holderIdentity :%s, expected at least two parts separated by ':'", s)
	}
	if len(splits) == 2 {
		return "", splits[0], asEtcdRole(splits[1]), nil
	} else {
		return splits[0], splits[1], asEtcdRole(splits[2]), nil
	}
}
```

internal/health/condition/check_ready.go
```diff
@@ -6,15 +6,52 @@ package condition
...
+	var clusterIdString, roleString string
+	if hasSeparator {
+		clusterIdString = before
+		roleString = after
+	} else {
+		clusterIdString = ""
+		roleString = before
+	}
+
+	switch druidv1alpha1.EtcdRole(roleString) {
```
unmarshall: move this to a separate function:
```go
func asEtcdRole(string) druidv1alpha1.EtcdRole
```

internal/health/condition/check_ready.go
```diff
@@ -48,6 +85,61 @@ func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
 		}
 	}
 
+	// Look for split-quorum/split-brain scenario.
+	leaseNames := druidv1alpha1.GetMemberLeaseNames(etcd.ObjectMeta, etcd.Spec.Replicas)
+	leases := make([]*coordinationv1.Lease, 0, len(leaseNames))
+	for _, leaseName := range leaseNames {
+		lease := &coordinationv1.Lease{}
+		if err := r.client.Get(ctx, client.ObjectKey{Namespace: etcd.Namespace, Name: leaseName}, lease); err != nil {
+			// Lease has not been created yet, so we pass on the split-quorum/split-brain check for now.
```
unmarshall:
```suggestion
			// Lease has not been created yet, so we skip the split-quorum/split-brain check for now.
```

internal/health/condition/check_ready.go
```diff
@@ -48,6 +85,61 @@ func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
 		}
 	}
 
+	// Look for split-quorum/split-brain scenario.
+	leaseNames := druidv1alpha1.GetMemberLeaseNames(etcd.ObjectMeta, etcd.Spec.Replicas)
+	leases := make([]*coordinationv1.Lease, 0, len(leaseNames))
+	for _, leaseName := range leaseNames {
+		lease := &coordinationv1.Lease{}
+		if err := r.client.Get(ctx, client.ObjectKey{Namespace: etcd.Namespace, Name: leaseName}, lease); err != nil {
```
unmarshall: if `err != nil` does not mean that it is not found. You have to explicitly check for `NotFoundError` If it is any other error then you should indicate that at this time you are unable to determine the condition.

internal/health/condition/check_ready.go
unmarshall: Have exported functions/methods first and then unexported functions.

internal/health/condition/check_ready.go
```diff
@@ -48,6 +85,61 @@ func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
...
+			// Lease has not been created yet, so we pass on the split-quorum/split-brain check for now.
+			continue
+		}
+		leases = append(leases, lease)
+	}
+	if len(leases) == size {
```
unmarshall: if this condition is false then the function returns:
```go
return &result{
		conType: druidv1alpha1.ConditionTypeReady,
		status:  druidv1alpha1.ConditionTrue,
		reason:  "Quorate",
		message: "The majority of ETCD members is ready",
	}
```
This is incorrect.

internal/health/condition/check_ready.go
```diff
@@ -6,15 +6,52 @@ package condition
 
 import (
 	"context"
+	"strings"
```
unmarshall: was there a rational to link split-brain with `ready` check? You can have a split brain but still be ready in a 3 member-cluster, right? Was a separate condition considered?

unmarshall (changes requested):

internal/health/condition/check_ready.go
ishan16696: > You can have a split brain but still be ready in a 3 member-cluster, right?

yes, but that's why druid controller is detecting this and capturing in etcd ready status, otherwise split-brain will go undetected as all members of cluster are in ready state.

ishan16696: > Was a separate condition considered?

why should have separate condition ? from druid point of etcd cluster is one complete entity and Etcd 's ready status should reflect that only.

unmarshall: It depends on how we define `Ready` condition at `Etcd` resource level which represents a single etcd-cluster. This ready state is not going to influence registration of endpoints from individual pods to endpoint slice for the client service. It is only meant for monitoring with no other impact. 

If you see the current implementation (in master) you see the following:
```go
	if readyMembers < quorum {
		return &result{
			conType: druidv1alpha1.ConditionTypeReady,
			status:  druidv1alpha1.ConditionFalse,
			reason:  "QuorumLost",
			message: "The majority of ETCD members is not ready",
		}
	}
```
So ready == quorate at this point of time. We also do not have any documentation as part of the API explaining the semantics on how we should interpret `ready` condition.

ishan16696: yes, I also checked the etcd status:
```yaml
 - lastTransitionTime: "2025-09-29T06:00:40Z"
    lastUpdateTime: "2025-09-29T11:33:56Z"
    message: The majority of ETCD members is ready
    reason: Quorate
    status: "True"
    type: Ready
```
It says "majority of ETCD members is ready" which is also true in split-brain, now I got your point ... should we create a separate of may be `type:  ClusterReady` which reflects/capture all members readiness ?

internal/health/condition/check_ready.go
CaptainIRS: I have updated the condition as we had discussed offline. Now, we just use the combination of member readiness ClusterID mismatch as the criteria for detecting a split brain situation.

internal/health/condition/check_ready.go
CaptainIRS: I have simplified the implementation. But, I wonder why this ordering is preferred over the current one. Changing the ordering would require another PR in the backup-restore as well, so if there is a strong reason to prefer the suggested ordering, I can raise the PR for that as well.

internal/health/condition/check_ready.go
CaptainIRS: The split-brain check is done only if certain conditions are satisfied, one of which being that all the member leases are available. If any of the leases aren't populated it just defaults to the existing behaviour of the ready check (where there is no split-brain detection).

internal/health/condition/check_ready.go
```diff
@@ -6,15 +6,22 @@ package condition
...
+	coordinationv1 "k8s.io/api/coordination/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type readyCheck struct{}
+type readyCheck struct {
+	client client.Client
+}
+
+const holderIdentitySeparator = ":"
```
ishan16696:
```suggestion
const memberLeaseHolderIdentitySeparator = ":"
```
just to avoid confusion with other lease like snapshot leases.

internal/health/condition/check_ready.go
```diff
@@ -48,6 +55,67 @@ func (r *readyCheck) Check(_ context.Context, etcd druidv1alpha1.Etcd) Result {
...
+		// So, druid does not support checking for split-brain in such cases.
+		if size > 0 && len(clusterIDs) == size && len(roles) == size {
+			firstClusterID := clusterIDs[0]
+			for _, clusterID := range clusterIDs[1:] {
+				if clusterID != firstClusterID {
+					return &result{
+						conType: druidv1alpha1.ConditionTypeReady,
+						status:  druidv1alpha1.ConditionFalse,
+						reason:  "SplitBrainDetected",
+						message: "Split-brain detected",
```
ishan16696:
```suggestion
						message: "ClusterID mismatch has been detected",
```
may be different message can be more informative.

internal/health/condition/check_ready.go
shreyas-s-rao: I agree with @unmarshall that ideally a separate condition should be used for displaying network partition or split brain for the etcd cluster. Since readiness is defined on the entire cluster and not individual members, a split brain can result in the cluster still being ready (2 out of 3 members are healthy and part of the original cluster). So readiness is not the right way to show the end user whether a split brain has occurred.

I would suggest a new condition type like `NetworkPartitioned` or `ClusterPartitioned` with reason `MultipleClusterIDsDetected`. This makes the status very clear and removes ambiguity around the ready check.

internal/health/condition/check_ready.go
unmarshall: A member belongs to an etcd cluster and plays a role in that cluster. Therefore the semantically correct hierarchy is first the <clusterID> that establishes the context well. This is then followed by the memberID which is part of that cluster. What role does this member play in context to a cluster is then the last thing.

internal/health/condition/check_ready.go
CaptainIRS: Update from the discussions with the team: We're going with a `ClusterIDMismatch` condition with reason `MultipleClusterIDsDetected`. I've updated the PR with the changes for this.

internal/health/condition/check_cluster_id_mismatch.go
```diff
@@ -0,0 +1,134 @@
...
+
+type clusterIDMismatchCheck struct {
+	client client.Client
+}
+
+const memberLeaseHolderIdentitySeparator = ":"
+
+func (r *clusterIDMismatchCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) Result {
+	memberNames := make([]string, 0)
+	memberConditionStatuses := make(map[string]druidv1alpha1.EtcdMemberConditionStatus)
```
shreyas-s-rao:
```suggestion
	memberStatuses := make(map[string]druidv1alpha1.EtcdMemberConditionStatus)
```

I just realised that `druidv1alpha1.EtcdMemberConditionStatus` is a confusing name used to denote `etcd.status.members[i].status`, since `EtcdMemberStatus` is already used up by `etcd.status.members[i]` field. We should change this to a better name to avoid the confusion here.

But in the meantime, it would make it easier to read the code if you changed the variable name to `memberStatuses` rather than `memberConditionStatuses`, especially because this piece of code deals with the status conditions, and there's an overlap of the use of the word `condition`.

internal/health/condition/check_cluster_id_mismatch.go
```diff
@@ -0,0 +1,134 @@
...
+
+	switch len(uniqueClusterIDs) {
+	case 0:
+		return &result{
+			conType: druidv1alpha1.ConditionTypeClusterIDMismatch,
+			status:  druidv1alpha1.ConditionUnknown,
+			reason:  "NoClusterIDDetected",
+			message: "No cluster ID could be detected from known ETCD members",
+		}
+	case 1:
```
shreyas-s-rao, line 88: What happens if we have a cluster of size 5, and 2 members get a new cluster ID? Can we set the conditions here in such a way that we handle any cluster size? Essentially, if length of uniqueClusterIDs is between 1 to n-1, then we can saefly assume that there's a mismatch.

internal/health/condition/check_cluster_id_mismatch.go
```diff
@@ -0,0 +1,134 @@
...
+		return &result{
+			conType: druidv1alpha1.ConditionTypeClusterIDMismatch,
+			status:  druidv1alpha1.ConditionFalse,
+			reason:  "NoClusterIDMismatchDetected",
+			message: "No cluster ID mismatch detected among ETCD members",
```
shreyas-s-rao, line 93: It would be nice to know which member(s) have a mismatched clusterID, since the Etcd resource doesn't publish this information in the Etcd status (yet). Speaking of which, wouldn't it be nice to publish the clusterID in `etcd.status.members` for each member? That way, users can simply look at the Etcd status and also visually inspect which member(s) has/have mismatched clusterID.

If an API change will take long, I will vote for retaining that information in the condition reason itself for now, and re-looking at enhancing the `members` section later.

@unmarshall WDYT?

shreyas-s-rao (changes requested):
@CaptainIRS thanks a lot for the well-written PR! I have a few questions/suggestions. PTAL, thanks.

internal/health/condition/check_cluster_id_mismatch.go
CaptainIRS: This is already handled in the code - the default case in the switch statement handles the case for 1 to n-1 unique cluster IDs, in which case it reports "MultipleClusterIDsDetected"

internal/health/condition/check_cluster_id_mismatch.go
CaptainIRS: The API change would not take long from my side. But if we were to add that information in the condition message, I think it would make it cluttered, especially when you consider the case where there are more than 2 unique cluster IDs. Currently, the easiest way for the operator to visually inspect the cluster IDs is to look at the leases.

internal/health/condition/check_cluster_id_mismatch.go
CaptainIRS: I've changed the variable to `memberStatuses` for now.

internal/health/condition/check_cluster_id_mismatch.go
shreyas-s-rao: It would be nice to atleast point the user to the member leases for them to check, although I would prefer this information being available in the Etcd resource status itself. If a user created the Etcd resource, it's only fair for them to check everything about the cluster frm the resource's status section. So ideally clusterID should be published in the `status.members` section, so let's take that up in another PR, because I don't want to block this one.

shreyas-s-rao (approved):
@CaptainIRS thanks for making the requested changes.
/lgtm

PR #1166, opened 2025-09-05
Reduce the default value of snapshot-count from 75k to 10k.

### LINKED ISSUE #1161: Set the snapshot-count from `100k` to `10k`
/area high-availability
/kind enhancement

In etcd v3.2, upstream etcd increased the default snapshot count from `10k` to `100k` https://github.com/etcd-io/etcd/pull/7160 and we followed the same behaviour, but in this issue https://github.com/etcd-io/etcd/issues/13889,  they came to the consensus to reverted this snapshot count back to `10k`.

**Why is this needed**:

The benefits of making snapshot-count to 10k:

- etcd will consume less memory and less likely to oom killed
- etcd will comsume less disk space because the snapshot will release the WAL file locks and purgeFile can delete those old files before throwing error "wal: no space is left"

ishan16696:
/area high-availability
/area auto-scaling
/kind enhancement

This PR reduces the default value of `--snapshot-count` from 75k to 10k.
- `--snapshot-count` is a tradeoff between higher memory usage and better availabilities of slow followers but in production usually clusters don't have slow followers, hence it better to reduce the snapshot-count to reduce the memory and disk usuage of etcd.
Please check doc: https://etcd.io/docs/v3.4/op-guide/maintenance/#raft-log-retention

Fixes #1161 

**Release note**:

```other operator
Reduces the default value of etcd's `--snapshot-count` from 75,000 to 10,000 set by `spec.etcd.snapshotCount`. This will reduce memory and disk usage of etcd. 
If you rely on the previous default, then set `spec.etcd.snapshotCount: 75000` to keep the old behavior.
```

### DIFF
--- a/api/core/v1alpha1/etcd_test.go
+++ b/api/core/v1alpha1/etcd_test.go
@@ -123,7 +123,7 @@ func createEtcd(name, namespace string) *Etcd {
 		backupPort    int32 = 8080
 		wrapperPort   int32 = 9095
 		metricLevel         = Basic
-		snapshotCount int64 = 75000
+		snapshotCount int64 = 10000
 	)
 
 	garbageCollectionPeriod := metav1.Duration{
--- a/examples/client/main.go
+++ b/examples/client/main.go
@@ -57,7 +57,7 @@ func main() {
 			},
 			Etcd: druidv1alpha1.EtcdConfig{
 				Quota:                   ptr.To(resource.MustParse("8Gi")),
-				SnapshotCount:           ptr.To(int64(75000)),
+				SnapshotCount:           ptr.To(int64(10000)),
 				DefragmentationSchedule: ptr.To("0 */24 * * *"),
 				ServerPort:              ptr.To[int32](2380),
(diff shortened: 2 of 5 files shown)


seshachalam-yv (approved):
/lgtm — thanks for adding this! Approving.

Nit: please add a release note advising operators who rely on the previous default to set the field explicitly:

```
Default for `spec.etcd.snapshotCount` reduced from 75000 to 10000 when unset. This changes snapshot/compaction cadence and reduces retained Raft log size. If you rely on the previous default, set `spec.etcd.snapshotCount: 75000` to
  keep the old behavior.
```

ishan16696:
> Nit: please add a release note advising operators who rely on the previous default to set the field explicitly:

done

anveshreddy18 (approved):
Thanks for PR @ishan16696. We went with 75k as an experiment and I feel it might have been the reason for some VPA evictions that we see so frequently in recent times, due to higher memory consumption. 

I hope this fix improves that.

PR #1170, opened 2025-09-18
Add make target `make check-license-headers` and include it in `.ci/check`

shreyas-s-rao:

/area testing
/kind enhancement

This PR adds a new make target `make check-license-headers` to check missing license headers, and also includes it in check step in CI pipeline.

**Release note**:

```other developer
Add make target `make check-license-headers` to check missing license headers.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ check: $(GOLANGCI_LINT) $(GOIMPORTS) format
 # Check license headers
 .PHONY: check-license-headers
 check-license-headers: $(GO_ADD_LICENSE)
-	@$(HACK_DIR)/check-license-headers.sh "$(REPO_ROOT)"
+	@$(HACK_DIR)/check-license-headers.sh
 
 .PHONY: sast
 sast: $(GOSEC)


hack/check-license-headers.sh
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env bash
+#
+# SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and Gardener contributors
```
unmarshall, line 3: This is very similar to `api/hack/check-generate.sh` - can we extract the common parts as reusable functions which can then be included wherever required?

unmarshall (changes requested):

hack/check-license-headers.sh
```diff
@@ -0,0 +1,115 @@
...
+  if [[ "$(git rev-parse --abbrev-ref HEAD)" == "$check_branch" ]]; then
+    echo "Already on check branch, aborting"
+    exit 1
+  fi
+  delete-check-branch
+
+  if [[ "$(git status -s)" != "" ]]; then
+    stashed=true
+    git stash --include-untracked -q
+    git stash apply -q &>/dev/null
```
renormalize: The working directory is stashed, and then un-stashed again immediately. What's the intended side effect of these two commands?

From what I can tell, the state of the git repository just returns to its original state right before running these two commands.

shreyas-s-rao: @renormalize I have taken @CaptainIRS 's suggestion and simplified the check script. `check generate` still uses this "complex" way of stashing current changes, checking out a new git branch, running `make generate`, detecting uncommited files, stashing them, checking out original branch and then unstashing the previously stashed changes from the dev. This is done so that devs can run `make check-generate` without disrupting their current local changes. I can try and simplify this, but in a separate PR, in order to keep the scope of this PR limited.

hack/check-license-headers.sh
shreyas-s-rao: @unmarshall I have taken @CaptainIRS 's suggestion and simplified the check script so there's no commonality between the two check scripts anymore.

PR #1184, opened 2025-10-03
Add OWNERS files

dimityrmirchev:
This PR prepares the repository for migration to prow.

Part of https://github.com/gardener/org/issues/3

@aaronfern @anveshreddy18 @CaptainIRS @ishan16696 @renormalize @seshachalam-yv @shreyas-s-rao @Shreyas-s14 @unmarshall please review and adjust assignments if necessary.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,23 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+aliases:
+  etcd-druid-reviewers:
+  - aaronfern
+  - anveshreddy18
+  - CaptainIRS
+  - ishan16696
+  - renormalize
+  - seshachalam-yv
+  - shreyas-s-rao
+  - Shreyas-s14
+  - unmarshall
+  etcd-druid-approvers:
+  - aaronfern
+  - anveshreddy18
+  - CaptainIRS
+  - ishan16696
+  - renormalize
+  - seshachalam-yv
+  - shreyas-s-rao
+  - Shreyas-s14
+  - unmarshall
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,6 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+- etcd-druid-reviewers
+approvers:
+- etcd-druid-approvers


OWNERS_ALIASES
```diff
@@ -0,0 +1,23 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+aliases:
+  etcd-druid-reviewers:
+  - aaronfern
```
ishan16696, line 5: @ashwani2k @unmarshall @shreyas-s-rao 
Should we now remove aaron as he's now MCM maintainer and no longer actively work on etcd-druid ?

OWNERS_ALIASES
```diff
@@ -0,0 +1,23 @@
...
+  - anveshreddy18
+  - CaptainIRS
+  - ishan16696
+  - renormalize
+  - seshachalam-yv
+  - shreyas-s-rao
+  - Shreyas-s14
+  - unmarshall
+  etcd-druid-approvers:
+  - aaronfern
```
ishan16696: same as above

unmarshall: you can remove as an approver.
In general @ishan16696 and @shreyas-s-rao  we need to decide who can approve. It cannot be the same list as the reviewers. Can you short list some names who can approve?

OWNERS_ALIASES
unmarshall: Well a reviewer can leave review comments but not approve. So having him as a reviewer for some more time is fine. In case we see that there is zero reviews from him then we can remove him.

OWNERS_ALIASES
ashwani2k: I agree you should remove as an approver as he may not be aware of context of other things that are part of a release and might approve something which should have been on hold as you don't wish to include it in the release.

ashwani2k: I'll also propose to put approver list to a limited few:
- unmarshall
- ishan
- shreyas-s-rao 

Reason being:
- lets start with a small list and expand it as the need arise
- we anyways needs lgtm before an approve so having limited eyes on approve ensures that we have an oversight if everything is covered from a review point of view.
- having everyone as revieweres and approver also dilutes the purpose of approvers
- when an external member joins he feels part of reviewer list and we have a path for people to become approver

shreyas-s-rao: @dimityrmirchev hi, any update on this?

shreyas-s-rao (approved):
@dimityrmirchev thanks for making the requested changes. 
/lgtm

ashwani2k (approved):

Issue #1185, opened 2025-10-03
Bundle CRDs with the Helm Chart for Standard Installation


mjudeikis:
**How to categorize this issue?**

/area usability
/area dev-productivity
/kind enhancement

**What would you like to be added**:

The Custom Resource Definitions (CRDs) should be included directly in the Helm chart (e.g., in the crds/ directory) to allow for a standard, one-step installation using `helm install`.

**Why is this needed**:

Currently, CRDs are not shipped with the Helm chart or committed to the repository. Users are required to clone the source and run a project-specific `make` command to generate them. This is a non-standard workflow that complicates automated deployments and creates significant challenges for GitOps environments.

Bundling the CRDs with the Helm chart would enable:
- Standard, one-step installation using `helm install`
- Better support for GitOps tools like Argo CD and Flux
- Lower barrier to entry for new users
- Simplified automated deployments

PR #1195, opened 2025-10-14
Add CRDs into git & helm

mjudeikis:
/area control-plane
/kind enhancement

This PR adds Custom Resource Definitions (CRDs) directly into the git repository and includes them with helm charts. Previously, CRDs were generated at build time and not tracked in version control.

**Key changes:**
- CRDs are now committed directly to the repository under `charts/crds/`
- `.gitignore` updated to remove the exclusion of CRD files
- GitHub workflow enhanced to prepare helm charts and verify git status is clean
- New Makefile targets: `check-git-status` to make sure we commit changed and dont have dirty repo.
- Added test validation for CRD specifications (EtcdCopyBackupsTask)

Fixes https://github.com/gardener/etcd-druid/issues/1185

The workflow now ensures that generated CRDs are committed and that the repository state remains clean after generation steps.

**Release note**:
```noteworthy operator
CRDs are now included directly in the git repository and helm charts, improving deployment reliability and version control tracking.
```

### DIFF
--- a/.github/workflows/base.yaml
+++ b/.github/workflows/base.yaml
@@ -114,6 +114,8 @@ jobs:
           make --directory=api check-generate
       - name: prepare helm charts
         run: make prepare-helm-charts
+        env:
+          K8S_VERSION: 1.34
       - name: check-git-status
         run: make check-git-status
   tests:


.github/workflows/base.yaml
```diff
@@ -112,7 +112,10 @@ jobs:
           git config --global user.email "gardener@sap.com"
           git config --global user.name "Gardener CI/CD"
           make --directory=api check-generate
-
+      - name: prepare helm charts
+        run: make prepare-helm-charts
```
shreyas-s-rao, line 116: Running this will not work in the github actions workflow since the `K8S_VERSION` variable is not set. This is necessary to run `make prepare-helm-charts`, since the k8s version determines which CRDs are copied to the helm chart - ones with CEL expressions or ones without (since CEL-based validation became stable only in k8s v1.29). 

Please specify the k8s version in the action. Since the latest k8s version is 1.34, I would recommend using that here.

shreyas-s-rao (changes requested):
@mjudeikis on second look, I realised I missed one point. Please check and address this, thanks.

PR #1202, opened 2025-10-21
Set the default snapshot compression to true.

ishan16696:
/area backup
/kind technical-debt

Snapshot compression feature was introduced long back by this PR https://github.com/gardener/etcd-backup-restore/pull/293 but we decided not to enable it by default in etcd-druid. Now I think the feature has been matured enough and it can be enabled by default. Moreover, I have seen that while doing perf testing for snapshot timeout sometimes dev forgot to enable the snapshot compression which result in failure while uploading/downloading the big snapshots as those snapshots aren't compressed.

Fixes #

**Release note**:

```other user
Snapshot compression is now enabled by default.
If you wish to disable the snapshot compression then please set the etcd resource: `.spec.backup.compression.enabled` to false.
```

### DIFF
--- a/examples/etcd/druid_v1alpha1_etcd.yaml
+++ b/examples/etcd/druid_v1alpha1_etcd.yaml
@@ -54,7 +54,7 @@ spec:
     #   provider: local # options: aws,azure,gcp,openstack,alicloud,dell,openshift,local
     #   prefix: etcd-test
     compression:
-      enabled: false
+      enabled: true # false (to disable compression)
       policy: "gzip"
     leaderElection:
       reelectionPeriod: 5s
--- a/internal/component/statefulset/builder.go
+++ b/internal/component/statefulset/builder.go
@@ -565,14 +565,14 @@ func (b *stsBuilder) getBackupStoreCommandArgs() []string {
 
 	// Snapshot compression and timeout command line args
 	// -----------------------------------------------------------------------------------------------------------------
+	compressionEnabled, compressionPolicy := druidv1alpha1.DefaultCompressionEnabled, druidv1alpha1.DefaultCompression
 	if b.etcd.Spec.Backup.SnapshotCompression != nil {
-		if ptr.Deref(b.etcd.Spec.Backup.SnapshotCompression.Enabled, false) {
-			commandArgs = append(commandArgs, fmt.Sprintf("--compress-snapshots=%t", *b.etcd.Spec.Backup.SnapshotCompression.Enabled))
-		}
-		if b.etcd.Spec.Backup.SnapshotCompression.Policy != nil {
-			commandArgs = append(commandArgs, fmt.Sprintf("--compression-policy=%s", string(*b.etcd.Spec.Backup.SnapshotCompression.Policy)))
-		}
+		compressionEnabled = ptr.Deref(b.etcd.Spec.Backup.SnapshotCompression.Enabled, druidv1alpha1.DefaultCompressionEnabled)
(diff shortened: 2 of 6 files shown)


examples/etcd/druid_v1alpha1_etcd.yaml
```diff
@@ -54,7 +54,7 @@ spec:
     #   provider: local # options: aws,azure,gcp,openstack,alicloud,dell,openshift,local
     #   prefix: etcd-test
     compression:
-      enabled: false
+      enabled: true
```
renormalize: Since omitting the `compression` block would mean compression is enabled, perhaps we can show in the example how to disable it instead?
The current proposed changes set it to `true`, which doesn't really change the behavior from the default.

```suggestion
      enabled: false # compression is enabled by default, and can be disabled like so.
```

internal/component/statefulset/builder.go
```diff
@@ -572,6 +572,9 @@ func (b *stsBuilder) getBackupStoreCommandArgs() []string {
 		if b.etcd.Spec.Backup.SnapshotCompression.Policy != nil {
 			commandArgs = append(commandArgs, fmt.Sprintf("--compression-policy=%s", string(*b.etcd.Spec.Backup.SnapshotCompression.Policy)))
 		}
+	} else {
+		commandArgs = append(commandArgs, fmt.Sprintf("--compress-snapshots=%t", druidv1alpha1.DefaultCompressionEnabled))
+		commandArgs = append(commandArgs, fmt.Sprintf("--compression-policy=%s", druidv1alpha1.DefaultCompression))
```
renormalize: I think the enclosing block, from line 568 to 578 can be simplified like so:

```go
	compressionEnabled, compressionPolicy := druidv1alpha1.DefaultCompressionEnabled, druidv1alpha1.DefaultCompression
	if b.etcd.Spec.Backup.SnapshotCompression != nil {
		compressionEnabled = ptr.Deref(b.etcd.Spec.Backup.SnapshotCompression.Enabled, druidv1alpha1.DefaultCompressionEnabled)
		compressionPolicy = ptr.Deref(b.etcd.Spec.Backup.SnapshotCompression.Policy, druidv1alpha1.DefaultCompression)
	}
	commandArgs = append(commandArgs, fmt.Sprintf("--compress-snapshots=%t", compressionEnabled))
	commandArgs = append(commandArgs, fmt.Sprintf("--compression-policy=%s", compressionPolicy))
```

****

According to the already existing code, if we only specify `SnapshotCompression.Enabled,` but do not specify `SnapshotCompression.Policy`, we never set this flag for the etcd-backup-restore container.

What ends up happening is that the default specified in etcd-backup-restore is used for compression.

Ideally, if a default for this field is specified in the API for etcd-druid, we must always use that specific value. This will ensure that changes in the default behavior of etcd-backup-restore do not cause changes in the behavior of etcd-druid.

If a consumer only sets the enabled flag, and the behavior of etcd-backup-restore is modified, then the compression they achieve will change.

The code I've suggested above ensures that this does not occur.

renormalize (changes requested):

examples/etcd/druid_v1alpha1_etcd.yaml
ishan16696: As discussed, we would like to have enabled set to `true`, so that user by mistake should disable the compression while do testing.

internal/component/statefulset/builder.go
ishan16696: I have made the changes, PTAL

Issue #1171, opened 2025-09-19
Druidctl (CLI) to Manage And Operate Etcd-druid managed etcd clusters.


anveshreddy18:
**How to categorize this issue?**

/area dev-productivity
/area usability
/kind enhancement

**What would you like to be added**:

To develop a CLI for druid which makes it easier for operators/developers to consume etcd-druid, take actions on any running etcd cluster, do certain kind of tasks as outlined in the [proposal](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/05-etcd-operator-tasks.md), etc. This can significantly help with the day-to-day operations of the operators and developers in viewing the state, making informed decisions to act upon the etcd cluster, etc.

Initially we plan to develop few general CLI commands and later extend it to include other commands such as operator task commands and any others as need arises.

The implementation will be broken down into the following phases:

### Phase 1: General Management Commands

This phase focuses on establishing the plugin framework and implementing the essential commands for day-to-day cluster management.

-   [x] **Setup Cobra CLI framework for the `kubectl druid` plugin.**
-   [x] **Implement General Management Commands:**
    -   [x] kubectl druid reconciliation [command] [resources] [flags]
        - [x] `kubectl druid reconciliation trigger [etcd-resources]`    # Trigger a reconciliation for selected `Etcd` resources.
        - [x] `kubectl druid reconciliation suspend [etcd-resources]`    # Suspend reconciliation capability
        - [x] `kubectl druid reconciliation resume [etcd-resources]`     # Resume (unsuspend) reconciliation capability
    - [x] kubectl druid component-protection [command] [resources] [flags]
        - [x] `kubectl druid component-protection add [etcd-resources]` # Add resource protection to all managed components for the selected Etcd resources
        - [x] `kubectl druid component-protection remove [etcd-resources]` # Remove resource protection from all managed components for the selected Etcd resources.
    -   [x] `kubectl druid version` # Prints out the build version of the CLI tool.
    -   [x] `kubectl druid list-resources [etcd-resources]`: List all managed and referenced resources for an etcd cluster.
    -   [ ] `kubectl druid debug <etcd-resource-name>`: Attach an ephemeral debug container to an etcd member pod.

---

### Phase 2: Monitoring and Diagnostics

This phase adds commands focused on observing the state and health of etcd clusters.

-   [ ] **Implement Monitoring Commands:**
    -   [ ] `kubectl druid member-status <etcd-resource-name>`: Provide a detailed status for etcd members.
[...]

PR #1212, opened 2025-11-06
Druidctl :  Phase one

anveshreddy18:

/area usability
/kind enhancement

This PR contains the ground work for the druidctl CLI along with implementation for few commands from the Phase 1 of the #1171 except for the `debug` subcommand. Since the PR keeps getting bigger with every subcommand, I wanted to push that to the next phase to make it easy for the reviewers. 

Fixes Phase#1 of #1171 

**Release note**:

```feature user
Druidctl phase one is out now and can be tested by getting into `druidctl/` folder and do `make install` or `make plugin-install` to install as standalone binary or as a kubectl plugin respectively.
```


CaptainIRS (commented):
Thanks for the PR @anveshreddy18! Here are the general points that we had discussed out-of-band:
* The shell completion command (offered by Cobra) seems to be broken as the logic in the codebase seems to interfere with its operation.
* Every command (except for `version`) invokes `cmdutils.GetEtcdList` as part of its `Execute` method, so it might be good to just have an `EtcdList` as part of the global context and populate it using the global `Complete` method.
* Most of the tests have a flow of getting the output from the command as a buffer and performing string matching to determine the outcome, but this feels more like integration tests rather than unit tests. We could consider extracting the logic into smaller functions that return (output, error) and have unit tests for such functions instead.
* The `rootCmd` is a global variable (which is similar to how the boilerplate for a new Cobra project looks like), but not sure if the approach is scalable (as discussed in https://github.com/spf13/cobra-cli/issues/12)
* We have a lot of logic for resolution of the resource GVR from short-hand resource names like `po`, `no`, `sts`, etc. for the list-resources command, but this could be simplified with a function like this using the `RestMapper`:
  ```go
  func (l *listResourcesCmdCtx) parseFilter() ([]schema.GroupVersionResource, error) {
	mapper := l.GenericClient.RESTMapper()
	parts := strings.Split(l.Filter, ",")
	out := make([]schema.GroupVersionResource, 0, len(parts))
	for _, part := range parts {
		gvr, err := mapper.ResourceFor(schema.GroupVersionResource{Resource: part})
		if err != nil {
			return nil, fmt.Errorf("failed to resolve resource type '%s': %w", part, err)
		}
		out = append(out, gvr)
	}
	for _, o := range out {
		fmt.Printf("GVR: %+v\n", o)
	}
	return out, nil
  }
  ```
* The list-resources command seems to take a long time to execute when testing with the new e2e test setup, retaining the test artifacts, so this might be a problem in prod clusters as well. One thing we could try to evaluate to improve the performance is [label selector set-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement).

Some opinionated comments:
* I feel like a kubectl-like syntax is more familiar to users of this command (typically k8s operators). This would also simplify the implementation using features like `ResourceBuilderFlags` from `cli-runtime`. The changes in this case would look something like https://gist.github.com/CaptainIRS/a6243cbb311ec95f7523a6884ce98fc4, allowing for command like `kubectl druid list-resources etcd-main etcd-events`, `kubectl druid list-resources -n shoot--test`, `kubectl druid list-resources -A --label-selector=role=main`, etc.
* The structure of commands currently in a way that we have `(suspend|resume)-reconcile`, `(add|remove)-component-protection)` as separate commands, but I feel like this could be structured as sub-commands, like `reconcile (suspend|resume)`, `component-protection (add|remove)`, etc. similar to `etcdctl` commands like `member add`, `member remove`, etc.

druidctl/cmd/listresources/command.go
```diff
@@ -0,0 +1,88 @@
...
+	cmdutils "github.com/gardener/etcd-druid/druidctl/cmd/utils"
+	"github.com/gardener/etcd-druid/druidctl/internal/printer"
+
+	"github.com/spf13/cobra"
+)
+
+const defaultFilter = "all"
+
+var (
+	example = `
```
CaptainIRS, line 17: This could be wrapped with https://pkg.go.dev/k8s.io/kubectl/pkg/util/templates#Examples to get rid of the indentation when printed in the terminal

druidctl/cmd/listresources/command.go
```diff
@@ -0,0 +1,88 @@
...
+		Short:   "List managed resources for an etcd cluster filtered by the specified types",
+		Long:    `List managed resources for an etcd cluster filtered by the specified types. If no types are specified, all managed resources will be listed.`,
+		Args:    cobra.ArbitraryArgs,
+		Example: example,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			listResourcesCmdCtx := &listResourcesCmdCtx{
+				listResourcesOptions: listResourcesOptions,
+			}
+			if err := listResourcesCmdCtx.validate(); err != nil {
+				if herr := cmd.Help(); herr != nil {
```
CaptainIRS, line 63: I'm curious if there is a reason we set `SlienceUsage` to true and handle this ourselves.

druidctl/cmd/listresources/command.go
```diff
@@ -0,0 +1,88 @@
...
+		Use:     "list-resources <etcd-resource-name> --filter=<comma separated types> (optional flag) --output=<output-format> (optional flag)",
+		Short:   "List managed resources for an etcd cluster filtered by the specified types",
+		Long:    `List managed resources for an etcd cluster filtered by the specified types. If no types are specified, all managed resources will be listed.`,
+		Args:    cobra.ArbitraryArgs,
+		Example: example,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			listResourcesCmdCtx := &listResourcesCmdCtx{
+				listResourcesOptions: listResourcesOptions,
+			}
+			if err := listResourcesCmdCtx.validate(); err != nil {
```
CaptainIRS, line 62: Could this be moved to `PreRunE`, since that semantically seems to be the right place for validation?

druidctl/cmd/listresources/command.go
```diff
@@ -0,0 +1,88 @@
...
+					options.Logger.Warning(options.IOStreams.ErrOut, "Failed to show help: ", herr.Error())
+				}
+				return err
+			}
+
+			if err := listResourcesCmdCtx.complete(options); err != nil {
+				return err
+			}
+
+			if err := listResourcesCmdCtx.execute(context.TODO()); err != nil {
```
CaptainIRS: Would passing Cobra's context here work?
```suggestion
			if err := listResourcesCmdCtx.execute(cmd.Context()); err != nil {
```

druidctl/cmd/listresources/command.go
```diff
@@ -0,0 +1,88 @@
...
+		Long:    `List managed resources for an etcd cluster filtered by the specified types. If no types are specified, all managed resources will be listed.`,
+		Args:    cobra.ArbitraryArgs,
+		Example: example,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			listResourcesCmdCtx := &listResourcesCmdCtx{
+				listResourcesOptions: listResourcesOptions,
+			}
+			if err := listResourcesCmdCtx.validate(); err != nil {
+				if herr := cmd.Help(); herr != nil {
+					options.Logger.Warning(options.IOStreams.ErrOut, "Failed to show help: ", herr.Error())
```
CaptainIRS: In general, having the first parameter be `io.Writer` adds too much noise to the repo. I think this could be set in the global `Complete` function.

druidctl/cmd/listresources/listresources.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// EtcdRef identifies an Etcd custom resource by name and namespace.
+type EtcdRef struct {
+	Name      string `json:"name" yaml:"name"`
+	Namespace string `json:"namespace" yaml:"namespace"`
+}
+
+// ResourceListPerKey groups resources under a single ResourceKey.
+type ResourceListPerKey struct {
```
CaptainIRS: Could the types in this file be moved to a `types.go` file as the types defined here seem to be used in `output.go` as well?

druidctl/cmd/listresources/listresources.go
```diff
@@ -0,0 +1,258 @@
...
+// Result is the top-level aggregation for managed resources across Etcds.
+type Result struct {
+	Etcds []EtcdResourceResult `json:"etcds" yaml:"etcds"`
+	Kind  string               `json:"kind" yaml:"kind"`
+}
+
+func (l *listResourcesCmdCtx) complete(options *cmdutils.GlobalOptions) error {
+	etcdClient, err := options.Clients.EtcdClient()
+	if err != nil {
+		options.Logger.Error(l.IOStreams.ErrOut, "Unable to create etcd client: ", err)
```
CaptainIRS: Instead of explicit log statements in this function, `fmt.Errorf` could be returned and the logging can be done in the caller (since the return type is `error` anyways).

druidctl/cmd/listresources/command.go
```diff
@@ -0,0 +1,88 @@
...
+			if err := listResourcesCmdCtx.complete(options); err != nil {
+				return err
+			}
+
+			if err := listResourcesCmdCtx.execute(context.TODO()); err != nil {
+				options.Logger.Error(options.IOStreams.ErrOut, "Listing Managed resources for Etcds failed", err)
+				return err
+			}
+
+			options.Logger.Success(options.IOStreams.Out, "Listing Managed resources for Etcds completed successfully")
```
CaptainIRS: I think this kind of statements should be logged only if the user explicitly requests for verbose logs. Also, if the command exits with a status code of 0, it is implicitly assumed to have executed successfully.

druidctl/cmd/listresources/listresources.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (l *listResourcesCmdCtx) validate() error {
+	return l.GlobalOptions.ValidateResourceSelection()
+}
+
+// execute lists the managed resources for the selected etcd resources based on the filter.
+func (l *listResourcesCmdCtx) execute(ctx context.Context) error {
+	if l.AllNamespaces {
+		l.Logger.Info(l.IOStreams.Out, "Listing all Managed resources for Etcds across all namespaces")
```
CaptainIRS: This also should be logged only if verbose log is explicitly requested by the user.

druidctl/cmd/listresources/listresources.go
```diff
@@ -0,0 +1,258 @@
...
+
+	etcdClient := l.EtcdClient
+	genClient := l.GenericClient
+
+	tokens := parseFilter(l.Filter)
+	if len(tokens) == 0 || (len(tokens) == 1 && tokens[0] == "all") {
+		tokens = defaultResourceTokens()
+	}
+	resolver := newRESTMapperResolver(genClient.RESTMapper())
+	metas, err := resolver.resolve(tokens)
```
CaptainIRS: I think we can completely get rid of the `resourceresolver.go` file. Functions like `RESTMapper.ResourceFor` or `RESTMapper.KindFor` could be used for resolution, without hardcoding all possible variations like short-form, plural, etc. which are implicitly handled by the `RESTMapper`. This would reduce a lot of complexity that is still present here.

druidctl/internal/log/utils.go
```diff
@@ -0,0 +1,24 @@
...
+func DefaultLogger() Logger {
+	return charm.NewCharmLogger()
+}
+
+// NewLogger constructs a logger based on the provided LogType.
+func NewLogger(logType LogType) Logger {
+	switch logType {
+	case LogTypeCharm:
+		return charm.NewCharmLogger()
+	default:
```
CaptainIRS, line 21: Both cases seem to return the same logger, is this intended?

druidctl/cmd/utils/clientbundle.go
```diff
@@ -0,0 +1,47 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
```
CaptainIRS, line 1: File could be named `clientbundle.go` for consistency.

CaptainIRS (changes requested):
@anveshreddy18 thanks for making the changes. I've added some review comments, PTAL, thanks. Note that some of the comments are applicable for all commands though I've mentioned it for the `list-resources` command to avoid repetition.

Also, please run `make check` within `druidctl` as it seems to point to some issues.

druidctl/cmd/listresources/command.go
anveshreddy18: Nice find, but I'm afraid that would bring in the entire kubectl dependency into the module for a simple non functional requirement, It would've been a good option if we already had a dependency.

druidctl/cmd/listresources/command.go
anveshreddy18: The only reason was that I felt it was too much spam to just show usage everytime and I instead had some descriptive error logs that described why the error happened, and there was a conflict where we could have only one of those two options and I went with the latter. But I'm open to rethinking that if there's an opinion on that.

druidctl/cmd/listresources/command.go
anveshreddy18: semantically yes but it would complicate the code as the `listResourcesCmdCtx` has to be established there as well or share it across these funcs which I feel makes it not so easy to follow.

druidctl/cmd/listresources/command.go
anveshreddy18: Nice catch, fixed it now.

druidctl/cmd/listresources/command.go
anveshreddy18: These writers were added because in the tests we test each command's functionality independently and we use buffers there to collect the output to check against in the tests, there a different writer will be passed, to inject dependencies basically.

as per your other suggestions, I've removed most of the success logs that felt spammy, now there are only a few important logs like this so you might not feel it that spammy, also because it's essential for the tests.

druidctl/cmd/listresources/listresources.go
anveshreddy18: yeah you're right, I've now moved these.

druidctl/cmd/listresources/listresources.go
anveshreddy18: Thanks for the catch, I've now fixed them.

druidctl/cmd/listresources/command.go
anveshreddy18: Yes, kubectl also doesn't have these success logs, I've now removed as you've made a good point about the 0 status code.

druidctl/cmd/listresources/listresources.go
anveshreddy18: sure, modified now.

druidctl/internal/log/utils.go
anveshreddy18: Yeah it's intended as we currently support just `charm` logs so that'll be the default as well, just as a safety measure, that's the only reason.

druidctl/cmd/listresources/listresources.go
anveshreddy18: Yeah my bad I missed pushing the changes last time, now I've pushed. PTAL.

druidctl/cmd/listresources/command.go
CaptainIRS: That's a fair point, I agree it might not be good to introduce a kubectl dependency. But, I think we should get rid of the indentation before the lines when printing. Could we just remove the spaces before `druidctl` in this multi-line string instead?

druidctl/cmd/listresources/command.go
CaptainIRS: Sure, let's keep it as it is then.

druidctl/cmd/listresources/command.go
CaptainIRS: Sure, then I'm okay with the current approach. Thanks for the explanation!

druidctl/cmd/listresources/command.go
CaptainIRS: That makes sense. Thanks for the clarification, let's keep it as it is.

druidctl/cmd/listresources/listresources.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// parseFilter splits and normalizes the filter string
+func parseFilter(filter string) []string {
+	if strings.TrimSpace(filter) == "" {
+		return nil
+	}
+	parts := strings.Split(filter, ",")
+	out := make([]string, 0, len(parts))
+	seen := map[string]struct{}{}
```
CaptainIRS: I think we can get rid of the duplicate checking here as we're doing it within `execute` as well.

druidctl/cmd/listresources/listresources.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	// Identify etcds to operate on
+	etcdList, err := cmdutils.GetEtcdList(ctx, etcdClient, l.etcdRefList, l.AllNamespaces, l.GetNamespace(), l.LabelSelector)
+	if err != nil {
+		return err
+	}
+	if len(etcdList.Items) == 0 {
+		if !l.AllNamespaces {
+			return fmt.Errorf("no Etcd resources found in namespace %q", l.GetNamespace())
```
CaptainIRS: I'm not sure about this since this will lead to a non-zero exit code. In kubectl's case at least, if no resources are found, it is not treated as an error (we just get an informational text, but with exit code 0).

druidctl/cmd/reconcile/command.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+const (
+	defaultTimeout = 5 * time.Minute
+)
+
+var (
+	reconcileExample = `
+		# Reconcile an Etcd resource named "my-etcd" in the test namespace
+		kubectl druid reconcile test/my-etcd
```
CaptainIRS: Nit: We have `druid` and `kubectl druid` in some commands, so I think we could pick one and make it consistent across the codebase.

druidctl/cmd/reconcile/reconcile.go
```diff
@@ -0,0 +1,249 @@
...
+	if r.waitTillReady || r.watch {
+		printReconcileStatus(statusMgr)
+	}
+
+	if len(failedResults) > 0 {
+		r.Logger.Info(r.IOStreams.Out, "Reconciliation failed for the following etcd resources:")
+		for _, result := range failedResults {
+			r.Logger.Error(r.IOStreams.ErrOut, "Reconciliation failed", result.Error, result.Etcd.Name, result.Etcd.Namespace)
+		}
+		return fmt.Errorf("one or more reconciliations failed")
```
CaptainIRS: Could we aggregate the errors and send one error instead of logging here? (Using something like https://pkg.go.dev/k8s.io/apimachinery/pkg/util/errors#NewAggregate)

druidctl/cmd/reconcile/resumeetcdreconcile.go
```diff
@@ -0,0 +1,109 @@
...
+			r.Logger.Error(r.IOStreams.ErrOut, "Failed to resume reconciliation for etcd", result.Error, result.Etcd.Name, result.Etcd.Namespace)
+			failedEtcds = append(failedEtcds, fmt.Sprintf("%s/%s", result.Etcd.Namespace, result.Etcd.Name))
+		}
+	}
+	if len(failedEtcds) > 0 {
+		r.Logger.Warning(r.IOStreams.Out, "Failed to resume reconciliation for etcd resources", failedEtcds...)
+		return fmt.Errorf("failed to resume reconciliation for etcd resources: %s", strings.Join(failedEtcds, ", "))
+	}
+	r.Logger.Success(r.IOStreams.Out, "Resumed reconciliation for all etcd resources")
+	return nil
```
CaptainIRS: In this case too the logging could be done only with verbose logs enabled. Also, the failure logs can be aggregated and sent to the caller instead of being logged here.

druidctl/cmd/reconcile/suspendetcdreconcile.go
```diff
@@ -0,0 +1,109 @@
...
+			s.Logger.Error(s.IOStreams.ErrOut, "Failed to suspend reconciliation for etcd", result.Error, result.Etcd.Name, result.Etcd.Namespace)
+			failedEtcds = append(failedEtcds, fmt.Sprintf("%s/%s", result.Etcd.Namespace, result.Etcd.Name))
+		}
+	}
+	if len(failedEtcds) > 0 {
+		s.Logger.Warning(s.IOStreams.Out, "Failed to suspend reconciliation for etcd resources", failedEtcds...)
+		return fmt.Errorf("suspending reconciliation failed for etcd resources: %v", failedEtcds)
+	}
+	s.Logger.Success(s.IOStreams.Out, "Suspended reconciliation for all etcd resources")
+	return nil
```
CaptainIRS: Same as https://github.com/gardener/etcd-druid/pull/1212#discussion_r2715673322

druidctl/cmd/resourceprotection/etcdprotection.go
```diff
@@ -0,0 +1,127 @@
...
+		confirmed, err := cmdutils.ConfirmAllNamespaces(r.IOStreams.Out, r.IOStreams.In, "remove protection from")
+		if err != nil {
+			return fmt.Errorf("confirmation failed: %w", err)
+		}
+		if !confirmed {
+			return cmdutils.ErrConfirmationDeclined
+		}
+		r.Logger.Info(r.IOStreams.Out, "Removing component protection from Etcds across all namespaces")
+	} else {
+		r.Logger.Info(r.IOStreams.Out, "Removing component protection from selected Etcds")
```
CaptainIRS: In this case too the logging could be done only with verbose logs enabled.

druidctl/cmd/resourceprotection/etcdprotection.go
```diff
@@ -0,0 +1,127 @@
...
+		confirmed, err := cmdutils.ConfirmAllNamespaces(r.IOStreams.Out, r.IOStreams.In, "add protection to")
+		if err != nil {
+			return fmt.Errorf("confirmation failed: %w", err)
+		}
+		if !confirmed {
+			return cmdutils.ErrConfirmationDeclined
+		}
+		r.Logger.Info(r.IOStreams.Out, "Adding component protection to all namespaces")
+	} else {
+		r.Logger.Info(r.IOStreams.Out, "Adding component protection to selected Etcds")
```
CaptainIRS: In this case too the logging could be done only with verbose logs enabled

druidctl/cmd/resourceprotection/etcdprotection.go
```diff
@@ -0,0 +1,127 @@
...
+			if e.Annotations != nil {
+				delete(e.Annotations, druidv1alpha1.DisableEtcdComponentProtectionAnnotation)
+			}
+		}
+		if err := r.etcdClient.UpdateEtcd(ctx, &etcd, etcdModifier); err != nil {
+			r.Logger.Error(r.IOStreams.ErrOut, "Failed to update etcd object", err, etcd.Name, etcd.Namespace)
+			errList = append(errList, fmt.Errorf("unable to update etcd object %s/%s: %w", etcd.Namespace, etcd.Name, err))
+			continue
+		}
+		r.Logger.Success(r.IOStreams.Out, "Removed protection annotation from etcd", etcd.Name, etcd.Namespace)
```
CaptainIRS: In this case too the logging could be done only with verbose logs enabled

druidctl/cmd/resourceprotection/etcdprotection.go
```diff
@@ -0,0 +1,127 @@
...
+		etcdModifier := func(e *druidv1alpha1.Etcd) {
+			if e.Annotations != nil {
+				delete(e.Annotations, druidv1alpha1.DisableEtcdComponentProtectionAnnotation)
+			}
+		}
+		if err := r.etcdClient.UpdateEtcd(ctx, &etcd, etcdModifier); err != nil {
+			r.Logger.Error(r.IOStreams.ErrOut, "Failed to update etcd object", err, etcd.Name, etcd.Namespace)
+			errList = append(errList, fmt.Errorf("unable to update etcd object %s/%s: %w", etcd.Namespace, etcd.Name, err))
+			continue
+		}
```
CaptainIRS, line 129: The failure logs can be aggregated and sent to the caller instead of being logged here.

druidctl/cmd/utils/clientbundle.go
CaptainIRS: This seems to be `clientBundle.go` now too, could you please check this?

druidctl/cmd/utils/options.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+	}
+	return false
+}
+
+// BuildEtcdRefList builds a list of NamespacedName from ResourceArgs respecting the -n flag.
+// kubectl-compatible behavior:
+//   - "name" : uses namespace from -n flag or default
+//   - "ns/name" : uses explicit namespace (cross-namespace selection)
+//   - Empty args : returns nil (caller should list all in namespace)
```
CaptainIRS, line 105: I think we could add unit tests for this for clarity, since there are many permutations and combinations.

druidctl/internal/log/utils.go
CaptainIRS: Do we intend to support other loggers in the future? If not I think we can remove a layer of abstraction for simplicity.

druidctl/cmd/listresources/command.go
anveshreddy18: I've removed that extra indentation now, PTAL. Thanks.

druidctl/internal/log/utils.go
anveshreddy18: Yes, we do intend to support other logger, Madhav raised this point previously about being able to do something like that and this was added then.

druidctl/cmd/utils/clientbundle.go
anveshreddy18: My bad, file name changes from camelcase to all small is somehow not being picked up by git, so I had to first delete the file, commit it, then create a new file with that name and then amend it to previous commit to make it work. Didn't check that previously.

druidctl/cmd/listresources/listresources.go
anveshreddy18: Thanks for the catch, I've now removed it.

druidctl/cmd/listresources/listresources.go
anveshreddy18: Agreed, changed it to just return nil.

druidctl/cmd/reconcile/command.go
anveshreddy18: thanks for identifying that, I've now changed everything to `kubectl druid`.

druidctl/cmd/reconcile/reconcile.go
anveshreddy18: I've used the inbuilt errors pkg which provides the `errors.Join()` method to join errors, and returning that now. PTAL.

druidctl/cmd/reconcile/resumeetcdreconcile.go
anveshreddy18: Yep, pls check now.

druidctl/cmd/reconcile/suspendetcdreconcile.go
anveshreddy18: Same as above.

druidctl/cmd/resourceprotection/etcdprotection.go
anveshreddy18: Done.

druidctl/cmd/resourceprotection/etcdprotection.go
anveshreddy18: Done.

druidctl/cmd/resourceprotection/etcdprotection.go
anveshreddy18: sure, changed it now. PTAL.

druidctl/cmd/utils/options.go
anveshreddy18: I've now added unit tests, PTAL.

druidctl/cmd/utils/clientbundle.go
CaptainIRS: I've run into the same issue in the past too 😆
Thanks for fixing!

druidctl/cmd/utils/options.go
CaptainIRS: The tests looks pretty thorough. Thanks for adding them!

druidctl/cmd/reconcile/reconcile.go
CaptainIRS: That's neat. Thanks.

PR #1214, opened 2025-11-14
Add support for externally managed members

CaptainIRS:
/area control-plane
/area high-availability
/kind enhancement
/kind api-change

This pull request introduces a new mode of operation in `etcd-druid` to support etcd clusters whose members are managed by an external actor, rather than by `etcd-druid` via the `StatefulSet` controller. This is primarily to enable use cases like Gardener's [self-hosted shoot clusters (GEP-28)](https://github.com/gardener/gardener/blob/master/docs/proposals/28-self-hosted-shoot-clusters.md), where a tool like `gardenadm` is responsible for deploying and managing etcd members as static pods on the control plane nodes.

To enable this, a new field, `spec.externallyManagedMemberAddresses`, is introduced in the `Etcd` API. When this field is populated with a list of member IP addresses, `etcd-druid`'s behavior changes as follows:

*   **Decoupled Pod Management**: `etcd-druid` no longer manages the lifecycle of etcd pods. It creates the `StatefulSet` with `replicas=0` to serve as a template but does not create any pods itself.
    * The template now sets the `POD_NAME` to be `<etcd-name>-<ip-address>` for the external tool to spin up `etcd` pods with the `etcd-backup-restore` sidecar using the template.
*   **IP-based Communication**: All etcd configuration is generated to use the provided IP addresses for peer communication. This removes the dependency on Kubernetes DNS and headless services.
    *   The `initial-cluster` and `advertised-<peer|client>-urls` configuration in the `ConfigMap` is populated using the IPs from `spec.externallyManagedMemberAddresses`.
    *   Member identities are based on their IP addresses (e.g., `etcd-main-192.168.0.1`) instead of StatefulSet ordinals.
*   **Disabled Component Creation**: The creation of several components that are tied to pod lifecycle management is disabled:
    *   Client and Peer `Service`s.
    *   `PodDisruptionBudget`.
*   **Continued Management Role**: While pod management is delegated, `etcd-druid` continues to provide value by:
    *   Generating and reconciling the etcd `ConfigMap`.
    *   Providing status reporting and facilitating maintenance operations for the externally managed cluster (enabling all functionality of `etcd-backup-restore` full/delta snapshots, compaction, de-fragmentation, etc.).

This approach provides a generic mechanism to integrate `etcd-druid` with diverse deployment strategies beyond the default `StatefulSet`-based model, making it more flexible.

This work supersedes the previous approach in PR #1117, opting for a more explicit API field (`spec.externallyManagedMemberAddresses`) over an annotation, along with design changes that overcome the limitations of the previous approach.

Part of #1071
Supersedes #1117

The core logic is triggered by the presence of the `spec.externallyManagedMemberAddresses` field. Please pay close attention to the validation rules for this new field, as they are crucial for preventing unsupported transitions:

*   The field must contain a list of valid IPv4 address strings.
*   If specified, its length must be equal to `spec.replicas`.
*   The field can be specified during the creation of a new `Etcd` resource.
*   **Mode Conversion is Not Supported**:
    *   The field **cannot** be added to an existing, running cluster or a hibernating cluster that is currently managed by `etcd-druid`
    *   The field **cannot** be removed from a running, externally managed cluster to convert it back to a `etcd-druid`-managed one.
    *   Externally managed clusters cannot be hibernated, i.e it cannot be scaled-in to zero replicas.

When `externallyManagedMemberAddresses` is set, `etcd-druid` effectively transitions from a pod lifecycle manager to a configuration and maintenance provider for an externally managed cluster.

**Release note**:
```other operator
NONE
```

### DIFF
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -358,6 +358,9 @@ type EtcdSpec struct {
 	// run as root. By default, they run as non-root with user 'nobody'.
 	// +optional
 	RunAsRoot *bool `json:"runAsRoot,omitempty"`
+	// ExternallyManagedMemberAddresses defines the list of addresses of externally managed etcd members. Specifying this
+	// will disable components that are involved in management of etcd members like Pods, Services and PDBs.
+	ExternallyManagedMemberAddresses []string `json:"externallyManagedMemberAddresses,omitempty"`
 }
 
 // CrossVersionObjectReference contains enough information to let you identify the referred resource.
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -312,6 +312,9 @@ type SchedulingConstraints struct {
 // EtcdSpec defines the desired state of Etcd
 // +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be empty when specified.",rule="has(self.externallyManagedMemberAddresses) ? self.externallyManagedMemberAddresses.size() > 0 : true"
 type EtcdSpec struct {
 	// selector is a label query over pods that should match the replica count.
 	// It must match the pod template's labels.
@@ -360,6 +363,9 @@ type EtcdSpec struct {
 	RunAsRoot *bool `json:"runAsRoot,omitempty"`
 	// ExternallyManagedMemberAddresses defines the list of addresses of externally managed etcd members. Specifying this
 	// will disable components that are involved in management of etcd members like Pods, Services and PDBs.
+	// +optional
+	// +kubebuilder:validation:items:Format=ipv4
+	// +listType=set
 	ExternallyManagedMemberAddresses []string `json:"externallyManagedMemberAddresses,omitempty"`
 }
 
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -310,10 +310,10 @@ type SchedulingConstraints struct {
 }
 
 // EtcdSpec defines the desired state of Etcd
-// +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
-// +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.storageClass field cannot be added or removed dynamically.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate field cannot be added or removed dynamically.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
-// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses is an immutable field.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
 type EtcdSpec struct {
 	// selector is a label query over pods that should match the replica count.
 	// It must match the pod template's labels.
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -313,8 +313,7 @@ type SchedulingConstraints struct {
 // +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
-// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
-// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be empty when specified.",rule="has(self.externallyManagedMemberAddresses) ? self.externallyManagedMemberAddresses.size() > 0 : true"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses is an immutable field.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
 type EtcdSpec struct {
 	// selector is a label query over pods that should match the replica count.
 	// It must match the pod template's labels.
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -220,7 +220,7 @@ type EtcdConfig struct {
 	// +optional
 	Quota *resource.Quantity `json:"quota,omitempty"`
 	// SnapshotCount defines the number of applied Raft entries to hold in-memory before compaction.
-	// More info: https://etcd.io/docs/v3.4/op-guide/maintenance/#raft-log-retention
+	// More info: https://etcd.io/docs/v3.5/op-guide/maintenance/#raft-log-retention
 	// +optional
 	SnapshotCount *int64 `json:"snapshotCount,omitempty"`
 	// EnableGRPCGateway enables the gRPC-Gateway proxy for etcd.
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -362,8 +362,8 @@ type EtcdSpec struct {
 	RunAsRoot *bool `json:"runAsRoot,omitempty"`
 	// ExternallyManagedMemberAddresses defines the list of addresses of externally managed etcd members. Specifying this
 	// will disable components that are involved in management of etcd members like Pods, Services and PDBs.
+	// Allowed values include: IPv4/IPv6 addresses and hostnames. Protocol or port shall not be specified.
 	// +optional
-	// +kubebuilder:validation:items:Format=ipv4
 	// +listType=set
 	ExternallyManagedMemberAddresses []string `json:"externallyManagedMemberAddresses,omitempty"`
 }
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -5,7 +5,9 @@
 package v1alpha1
 
 import (
+	"crypto/rand"
 	"fmt"
+	"math/big"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -22,6 +24,22 @@ func GetPeerServiceName(etcdObjMeta metav1.ObjectMeta) string {
 // GetClientServiceName returns the client service name for the Etcd cluster reachable by external clients.
 func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
 	return fmt.Sprintf("%s-client", etcdObjMeta.Name)
+}
+
+// GetClientHostname returns the hostname of the client endpoint for the Etcd cluster. This is the client service hostname when the Etcd members
+// are managed by etcd-druid, else it is a random member address from the externally managed member addresses.
+func GetClientHostname(etcd *Etcd) string {
+	if IsPodManagementEnabled(etcd) {
+		return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.ObjectMeta.Namespace)
+	} else {
+		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(etcd.Spec.ExternallyManagedMemberAddresses))))
+		if err != nil {
+			// Fallback to first member address in case of an error
+			return etcd.Spec.ExternallyManagedMemberAddresses[0]
+		}
+		randomIndex := int(n.Int64())
+		return etcd.Spec.ExternallyManagedMemberAddresses[randomIndex]
+	}
 }
 
 // GetServiceAccountName returns the service account name for the Etcd.
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -44,6 +44,11 @@ func GetOrdinalPodName(etcdObjMeta metav1.ObjectMeta, ordinal int) string {
 	return fmt.Sprintf("%s-%d", etcdObjMeta.Name, ordinal)
 }
 
+// GetMemberNameFromAddress returns the name of the etcd member based on the address.
+func GetMemberNameFromAddress(etcdObjMeta metav1.ObjectMeta, memberAddress string) string {
+	return fmt.Sprintf("%s-%s", etcdObjMeta.Name, memberAddress)
+}
+
 // GetAllPodNames returns the names of all pods for the Etcd.
 func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
 	podNames := make([]string, replicas)
@@ -54,12 +59,22 @@ func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
 }
 
 // GetMemberLeaseNames returns the name of member leases for the Etcd.
-func GetMemberLeaseNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
-	leaseNames := make([]string, replicas)
-	for i := range int(replicas) {
-		leaseNames[i] = fmt.Sprintf("%s-%d", etcdObjMeta.Name, i)
+func GetMemberLeaseNames(etcd *Etcd) []string {
+	if IsPodManagementEnabled(etcd) {
+		replicas := etcd.Spec.Replicas
+		podNames := make([]string, replicas)
+		for i := range int(replicas) {
+			podNames[i] = GetOrdinalPodName(etcd.ObjectMeta, i)
+		}
+		return podNames
+	} else {
+		memberAddresses := etcd.Spec.ExternallyManagedMemberAddresses
+		memberNames := make([]string, len(memberAddresses))
+		for i, memberAddress := range memberAddresses {
+			memberNames[i] = GetMemberNameFromAddress(etcd.ObjectMeta, memberAddress)
+		}
+		return memberNames
 	}
-	return leaseNames
 }
 
 // GetPodDisruptionBudgetName returns the name of the pod disruption budget for the Etcd.
@@ -166,7 +181,7 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
 	delete(etcdObjMeta.Annotations, GardenerOperationAnnotation)
 }
 
-// IsEtcdRuntimeComponentCreationEnabled checks if the creation of runtime components is enabled for an Etcd resource.
-func IsEtcdRuntimeComponentCreationEnabled(etcdObjMeta metav1.ObjectMeta) bool {
-	return !metav1.HasAnnotation(etcdObjMeta, DisableEtcdRuntimeComponentCreationAnnotation)
+// IsPodManagementEnabled checks if the management of pods is handled by etcd-druid for an Etcd resource.
+func IsPodManagementEnabled(etcd *Etcd) bool {
+	return len(etcd.Spec.ExternallyManagedMemberAddresses) == 0
 }
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -27,9 +27,9 @@ func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
 }
 
 // GetClientHostname returns the hostname of the client endpoint for the Etcd cluster. This is the client service hostname when the Etcd members
-// are managed by etcd-druid, else it is a random member address from the externally managed member addresses.
+// are managed by etcd-druid, else it is a randomly selected member address out of the externally managed member addresses.
 func GetClientHostname(etcd *Etcd) string {
-	if IsPodManagementEnabled(etcd) {
+	if ArePodsManagedByEtcdDruid(etcd) {
 		return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.Namespace)
 	} else {
 		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(etcd.Spec.ExternallyManagedMemberAddresses))))
@@ -78,7 +78,7 @@ func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
 
 // GetMemberLeaseNames returns the name of member leases for the Etcd.
 func GetMemberLeaseNames(etcd *Etcd) []string {
-	if IsPodManagementEnabled(etcd) {
+	if ArePodsManagedByEtcdDruid(etcd) {
 		return GetAllPodNames(etcd.ObjectMeta, etcd.Spec.Replicas)
 	} else {
 		memberAddresses := etcd.Spec.ExternallyManagedMemberAddresses
@@ -194,7 +194,7 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
 	delete(etcdObjMeta.Annotations, GardenerOperationAnnotation)
 }
 
-// IsPodManagementEnabled checks if the management of pods is handled by etcd-druid for an Etcd resource.
-func IsPodManagementEnabled(etcd *Etcd) bool {
+// ArePodsManagedByEtcdDruid checks if the management of pods is handled by etcd-druid for an Etcd resource.
+func ArePodsManagedByEtcdDruid(etcd *Etcd) bool {
 	return len(etcd.Spec.ExternallyManagedMemberAddresses) == 0
 }
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -79,12 +79,7 @@ func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
 // GetMemberLeaseNames returns the name of member leases for the Etcd.
 func GetMemberLeaseNames(etcd *Etcd) []string {
 	if IsPodManagementEnabled(etcd) {
-		replicas := etcd.Spec.Replicas
-		podNames := make([]string, replicas)
-		for i := range int(replicas) {
-			podNames[i] = GetOrdinalPodName(etcd.ObjectMeta, i)
-		}
-		return podNames
+		return GetAllPodNames(etcd.ObjectMeta, etcd.Spec.Replicas)
 	} else {
 		memberAddresses := etcd.Spec.ExternallyManagedMemberAddresses
 		memberNames := make([]string, len(memberAddresses))
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -30,7 +30,7 @@ func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
 // are managed by etcd-druid, else it is a random member address from the externally managed member addresses.
 func GetClientHostname(etcd *Etcd) string {
 	if IsPodManagementEnabled(etcd) {
-		return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.ObjectMeta.Namespace)
+		return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.Namespace)
 	} else {
 		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(etcd.Spec.ExternallyManagedMemberAddresses))))
 		if err != nil {
--- a/api/core/v1alpha1/helper_test.go
+++ b/api/core/v1alpha1/helper_test.go
@@ -86,11 +86,42 @@ func TestGetFullSnapshotLeaseName(t *testing.T) {
 	g.Expect(fullSnapshotLeaseName).To(Equal("etcd-test-full-snap"))
 }
 
-func TestGetMemberLeaseNames(t *testing.T) {
+func TestGetMemberNameFromAddress(t *testing.T) {
 	g := NewWithT(t)
 	etcdObjMeta := createEtcdObjectMetadata(uuid.NewUUID(), nil, nil, false)
-	leaseNames := GetMemberLeaseNames(etcdObjMeta, 3)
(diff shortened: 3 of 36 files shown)


docs/concepts/externally-managed-members.md
```diff
@@ -0,0 +1,42 @@
+# Externally Managed Members
+
+In certain scenarios, such as [self-hosted shoot clusters](https://github.com/gardener/gardener/blob/master/docs/proposals/28-self-hosted-shoot-clusters.md) in Gardener, the etcd members are managed by an external actor (such as static pods managed by the kubelet on the control plane nodes). In these cases, etcd-druid cannot manage the lifecycle of the etcd members directly through Kubernetes constructs like StatefulSets. To accommodate this use case, support for externally managed etcd members is introduced in etcd-druid.
```
Shreyas-s14: This url points to the older location of the doc. Could you update it with : https://github.com/gardener/enhancements/tree/main/geps/0028-self-hosted-shoot-clusters

docs/concepts/externally-managed-members.md
```diff
@@ -0,0 +1,42 @@
...
+* Removal of the field is not allowed once set.
+* The list of member IPs cannot contain duplicates.
+* The list of member IPs can be updated, but the updated list must still satisfy the above constraints.
+
+In the etcd-backup-restore sidecar,
+* If the "service-endpoints" field is not supplied, the endpoints used for creating the etcd client will be derived from the member IPs obtained from the config file as fallback (since the client service is not created in this scenario).
+
+Using externally managed members allows etcd-druid to support scenarios where etcd members are managed outside of its direct control, while still providing essential functionalities like backup/restore, maintenance, and monitoring. This allows for etcd-druid to be used for managing etcd clusters in use-cases like [self-hosted shoot clusters in Gardener](https://github.com/gardener/gardener/blob/master/docs/proposals/28-self-hosted-shoot-clusters.md), [kubeadm](https://kubernetes.io/docs/reference/setup-tools/kubeadm/), [k3s](https://k3s.io/), etc.
```
Shreyas-s14: Same as https://github.com/gardener/etcd-druid/pull/1214#discussion_r2992487554

api/core/v1alpha1/etcd.go
```diff
@@ -312,6 +312,9 @@ type SchedulingConstraints struct {
 // EtcdSpec defines the desired state of Etcd
 // +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
```
Shreyas-s14, line 316: Could the message for this be renamed to indicate the field is immutable as done in other immutable fields? This essentially means that right?

api/core/v1alpha1/etcd.go
```diff
@@ -312,6 +312,9 @@ type SchedulingConstraints struct {
...
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be empty when specified.",rule="has(self.externallyManagedMemberAddresses) ? self.externallyManagedMemberAddresses.size() > 0 : true"
```
Shreyas-s14: Just a question, isn't this rule inherently enforced by the rule in line 315?
If the replicas is set to 0, then it is passed anyways
Same goes for it being >0.

Correct me if I'm wrong.

api/core/v1alpha1/helper.go
```diff
@@ -24,6 +26,22 @@ func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
...
 }
 
+// GetClientHostname returns the hostname of the client endpoint for the Etcd cluster. This is the client service hostname when the Etcd members
+// are managed by etcd-druid, else it is a random member address from the externally managed member addresses.
+func GetClientHostname(etcd *Etcd) string {
+	if IsPodManagementEnabled(etcd) {
+		return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.ObjectMeta.Namespace)
+	} else {
+		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(etcd.Spec.ExternallyManagedMemberAddresses))))
+		if err != nil {
```
Shreyas-s14, line 36: Just a nit here:
This function wont be called/executed until the resource has been validated by the API server with the CEL expressions.
In that case, would there be an error for fetching the value from the index?
```suggestion
		if err != nil {
```
    if IsPodManagementEnabled(etcd) {
          return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.ObjectMeta.Namespace)
    }
    return etcd.Spec.ExternallyManagedMemberAddresses[rand.IntN(len(etcd.Spec.ExternallyManagedMemberAddresses))]
```

File: api/core/v1alpha1/helper.go [hunk 6]
```diff
@@ -54,12 +77,22 @@ func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
...
-	for i := range int(replicas) {
-		leaseNames[i] = fmt.Sprintf("%s-%d", etcdObjMeta.Name, i)
+func GetMemberLeaseNames(etcd *Etcd) []string {
+	if IsPodManagementEnabled(etcd) {
+		replicas := etcd.Spec.Replicas
+		podNames := make([]string, replicas)
+		for i := range int(replicas) {
+			podNames[i] = GetOrdinalPodName(etcd.ObjectMeta, i)
+		}
+		return podNames
```
Shreyas-s14: The block inside 
```go
if IsPodManagementEnabled(etcd) {}
```
is basically defined in the function `GetAllPodNames`. Could that be used instead?

Shreyas-s14:
```suggestion
	if IsPodManagementEnabled(etcd) {
        return GetAllPodNames(etcd.ObjectMeta, etcd.Spec.Replicas)
    }
```

test/it/crdvalidation/etcd/spec_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestSpecExternallyManagedMemberAddresses(t *testing.T) {
+	tests := []struct {
+		name                           string
+		etcdName                       string
+		replicas                       int32
+		externallyManagedMemberAddress []string
+		expectErr                      bool
+	}{
+		{
+			name:                           "Valid externallyManagedMemberAddresses #1: no addresses with non-zero replicas",
```
Shreyas-s14: Could you rename this test case to indicate that this is a scenario for druid managed setup?

test/it/crdvalidation/etcd/spec_test.go
```diff
@@ -0,0 +1,71 @@
...
+			expectErr:                      true,
+		},
+	}
+
+	testNs, g := setupTestEnvironment(t)
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			etcd := utils.EtcdBuilderWithoutDefaults(test.etcdName, testNs).WithReplicas(test.replicas).Build()
+			etcd.Spec.ExternallyManagedMemberAddresses = test.externallyManagedMemberAddress
```
Shreyas-s14: Can the new builder method be used here? The one introduced in `test/utils/etcd.go`

internal/component/configmap/etcdconfig.go
```diff
@@ -118,12 +118,20 @@ func getSchemeAndSecurityConfig(tlsConfig *druidv1alpha1.TLSConfig, caPath, serv
 }
 
 func prepareInitialCluster(etcd *druidv1alpha1.Etcd, peerScheme string) string {
-	domainName := fmt.Sprintf("%s.%s.%s", druidv1alpha1.GetPeerServiceName(etcd.ObjectMeta), etcd.Namespace, "svc")
 	serverPort := strconv.Itoa(int(ptr.Deref(etcd.Spec.Etcd.ServerPort, common.DefaultPortEtcdPeer)))
 	builder := strings.Builder{}
-	for i := range int(etcd.Spec.Replicas) {
-		podName := druidv1alpha1.GetOrdinalPodName(etcd.ObjectMeta, i)
-		builder.WriteString(fmt.Sprintf("%s=%s://%s.%s:%s,", podName, peerScheme, podName, domainName, serverPort))
+	if druidv1alpha1.IsPodManagementEnabled(etcd) {
```
Shreyas-s14: Do you think we should extracted into a separate function since it it used here and in line 150-161? If you feel thats unnecessary, then you can go ahead and resolve this. Just a thought

internal/component/statefulset/builder.go
```diff
@@ -522,11 +519,9 @@ func (b *stsBuilder) getBackupRestoreContainerCommandArgs() []string {
 func (b *stsBuilder) getBackupStoreCommandArgs() []string {
 	var commandArgs []string
 
-	if druidv1alpha1.IsEtcdRuntimeComponentCreationEnabled(b.etcd.ObjectMeta) {
-		commandArgs = append(commandArgs, "--enable-snapshot-lease-renewal=true")
-		commandArgs = append(commandArgs, fmt.Sprintf("--full-snapshot-lease-name=%s", druidv1alpha1.GetFullSnapshotLeaseName(b.etcd.ObjectMeta)))
-		commandArgs = append(commandArgs, fmt.Sprintf("--delta-snapshot-lease-name=%s", druidv1alpha1.GetDeltaSnapshotLeaseName(b.etcd.ObjectMeta)))
-	}
+	commandArgs = append(commandArgs, "--enable-snapshot-lease-renewal=true")
```
Shreyas-s14, line 522: Not a comment, but a question. Is the check no longer needed here ebcause the annotation has been removed altogether? Same for other removed checks in the file.

internal/controller/etcd/reconcile_spec.go
```diff
@@ -90,29 +87,6 @@ func (r *Reconciler) syncEtcdResources(ctx component.OperatorContext, etcd *drui
 	return ctrlutils.ContinueReconcile()
 }
```
Shreyas-s14, line 88: Do we no longer need the Cleanup logic? Maybe I'm missing the context here...

Shreyas-s14 (changes requested):
Thanks for the PR @CaptainIRS. Have a few comments. PTAL!

api/core/v1alpha1/helper.go
CaptainIRS: > In that case, would there be an error for fetching the value from the index?

Since the validation would fail in that case, the CR will never get created/updated with an invalid value. So, out-of-bounds access is not possible. Though, if you feel it is required, I can add out-of-bounds check and handle the error.

> return etcd.Spec.ExternallyManagedMemberAddresses[rand.IntN(len(etcd.Spec.ExternallyManagedMemberAddresses))]

Initially I had a simplified implementation like this, but the SAST lint complains with this:
```
G404 (CWE-338): Use of weak random number generator (math/rand or math/rand/v2 instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
```

internal/component/configmap/etcdconfig.go
CaptainIRS: I'm not sure there is much in common because one writes a string in a particular format while the other populates entries in a map. If you have in mind an elegant way of extracting the common parts, please let me know.

internal/component/statefulset/builder.go
CaptainIRS: In the previous implementation, leases had to be disabled. But in the current approach, member and snapshot leases continue to work like the hosted control-plane setup.

api/core/v1alpha1/helper.go
CaptainIRS: Yes, thanks for pointing this out!
Addressed in 3b8bbd9b

internal/controller/etcd/reconcile_spec.go
CaptainIRS: The cleanup logic was added in the previous implementation to handle cases where runtime components were disabled at runtime. But in the current implementation we don't support the transition from druid-managed to externally-managed modes and vice-versa, so we would no longer need the cleanup logic.

test/it/crdvalidation/etcd/spec_test.go
CaptainIRS: Addressed in 3b8bbd9b

test/it/crdvalidation/etcd/spec_test.go
CaptainIRS: Addressed in 3b8bbd9b

docs/concepts/externally-managed-members.md
CaptainIRS: Addressed in 3b8bbd9b

docs/concepts/externally-managed-members.md
CaptainIRS: Addressed in 3b8bbd9b

api/core/v1alpha1/etcd.go
CaptainIRS: You're right. Addressed in 3b8bbd9b.

api/core/v1alpha1/etcd.go
CaptainIRS: Yes, thanks for catching this! Addressed in 3b8bbd9b.

CaptainIRS (commented):
@Shreyas-s14 thanks for the review. I've addressed the the review comments. PTAL, thanks!

internal/component/configmap/etcdconfig.go
Shreyas-s14: Yeah fair enough. The extraction would look messier than the current.

api/core/v1alpha1/helper.go
Shreyas-s14: >Initially I had a simplified implementation like this, but the SAST lint complains with this:

Oh then it's fine as is.

internal/component/statefulset/builder.go
Shreyas-s14: Thanks for the clarification!

Shreyas-s14 (approved):
Thanks for the changes. Looks good

docs/concepts/externally-managed-members.md
```diff
@@ -0,0 +1,42 @@
...
+  * PodDisruptionBudget
+* Populate the etcd ConfigMap with peer/client URLs that use the member IPs from `spec.externallyManagedMemberAddresses` instead of DNS names.
+* Populate the Lease object used for sharing state with the sidecar with the member IPs from `spec.externallyManagedMemberAddresses` instead of DNS names.
+* Create and assume that the member identities will no longer rely on StatefulSet semantics (like etcd-main-0, etcd-main-1, etc.), instead the identity will be based on the member IPs from `spec.externallyManagedMemberAddresses`. (like etcd-main-192.168.0.1, etcd-main-192.168.0.2, etc.)
+* The `Status.Ready` field in the `Etcd` CR will be populated with the value of the `AllMembersReady` condition instead of the current logic that depends on the StatefulSet Ready status when members are being managed externally.
+* When the list of member IPs is changed, etcd-druid will update the ConfigMap and Lease objects accordingly (i.e getting rid of old leases), but it will be the external actor's responsibility to ensure that the old members are removed from the cluster and the new members are added correctly.
+
+The field is subject to the following validations:
```
anveshreddy18, line 29: do we also want to ensure that these members count is an odd number or do you feel it's upto whoever is the consumer? Don't have any strong opinion here but just bringing this up for discussion.

api/core/v1alpha1/etcd.go
```diff
@@ -358,6 +360,12 @@ type EtcdSpec struct {
 	// run as root. By default, they run as non-root with user 'nobody'.
 	// +optional
 	RunAsRoot *bool `json:"runAsRoot,omitempty"`
+	// ExternallyManagedMemberAddresses defines the list of addresses of externally managed etcd members. Specifying this
+	// will disable components that are involved in management of etcd members like Pods, Services and PDBs.
```
anveshreddy18, line 364:
```suggestion
	// will disable components that are involved in etcd-druid managed etcd members like Pods, Services and PDBs.
```

api/core/v1alpha1/helper.go
```diff
@@ -166,7 +194,7 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
 	delete(etcdObjMeta.Annotations, GardenerOperationAnnotation)
 }
 
-// IsEtcdRuntimeComponentCreationEnabled checks if the creation of runtime components is enabled for an Etcd resource.
-func IsEtcdRuntimeComponentCreationEnabled(etcdObjMeta metav1.ObjectMeta) bool {
-	return !metav1.HasAnnotation(etcdObjMeta, DisableEtcdRuntimeComponentCreationAnnotation)
+// IsPodManagementEnabled checks if the management of pods is handled by etcd-druid for an Etcd resource.
+func IsPodManagementEnabled(etcd *Etcd) bool {
```
anveshreddy18:
```suggestion
func ArePodsManagedByEtcdDruid(etcd *Etcd) bool {
```

imo having the name imply what it is actually doing will improve the readability, reducing the need for checking what's inside the function. Wdyt?

api/core/v1alpha1/helper.go
```diff
@@ -166,7 +194,7 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
...
+	return len(etcd.Spec.ExternallyManagedMemberAddresses) == 0
```
anveshreddy18, line 199: This condition would lead to a false state when the ExternallyManagedMemberAddresses is an empty list, it says it is managed by druid when it should've been by external actor because from the CEL expression & documentation, I see that the entire decision making between who manages depends entirely on the "presence" of this field and not based on the contents inside it. If someone starts with empty list and later adds more member ips to this list, then this toggle would switch from (true: druidmanaged ) to (false: external managed). 

We may not hit this practically but semantically I feel it carries meaning and we might want to check for the presence of the field in this function by making the `ExternallyManagedMemberAddresses` field a pointer to slice and here we just check if it's nil or not.

Pls correct me if my understanding is incorrect.

api/core/v1alpha1/helper.go
```diff
@@ -24,6 +26,22 @@ func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
 	return fmt.Sprintf("%s-client", etcdObjMeta.Name)
 }
 
+// GetClientHostname returns the hostname of the client endpoint for the Etcd cluster. This is the client service hostname when the Etcd members
+// are managed by etcd-druid, else it is a random member address from the externally managed member addresses.
```
anveshreddy18:
```suggestion
// are managed by etcd-druid, else it is a randomly selected member address out of the externally managed member addresses.
```
NIT: just that when I read it first, it felt like "random" ip instead of randomly selected ip from the list.

api/core/v1alpha1/helper.go
```diff
@@ -24,6 +26,22 @@ func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
...
+func GetClientHostname(etcd *Etcd) string {
```
anveshreddy18: I don't see this function being used anywhere, can you pls lmk why we need this?

api/core/v1alpha1/etcd.go
```diff
@@ -358,6 +360,12 @@ type EtcdSpec struct {
...
+	// +optional
+	// +kubebuilder:validation:items:Format=ipv4
+	// +listType=set
+	ExternallyManagedMemberAddresses []string `json:"externallyManagedMemberAddresses,omitempty"`
```
anveshreddy18, line 368:
```suggestion
	ExternallyManagedMemberIPs []string `json:"ExternallyManagedMemberIPs,omitempty"`
```
If this just has list of IPs, then can we instead name it to reflect that? because the "address" can mean different things as there's no standard, it can be just ip, or ip:port, protocol://ip:port, protocol://ip:port/path, etc. We can avoid some potential confusion by explicitly naming it based on what we accept here. Wdyt?

api/core/v1alpha1/helper.go
```diff
@@ -24,6 +26,22 @@ func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
...
+	if IsPodManagementEnabled(etcd) {
+		return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.Namespace)
+	} else {
+		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(etcd.Spec.ExternallyManagedMemberAddresses))))
```
anveshreddy18, line 35: Is there any strong reason for randomising this, can we not always pick the first member from the list which would make it much simpler and deterministic. If you think there's a reasonable problem with it, pls lmk.

api/core/v1alpha1/helper_test.go
```diff
@@ -43,6 +43,37 @@ func TestGetClientServiceName(t *testing.T) {
...
+	g := NewWithT(t)
+	etcdObjMeta := createEtcdObjectMetadata(uuid.NewUUID(), nil, nil, false)
+	etcd := &Etcd{
+		ObjectMeta: etcdObjMeta,
+		Spec: EtcdSpec{
+			Replicas: 3,
+		},
+	}
+	clientHostname := GetClientHostname(etcd)
+	g.Expect(clientHostname).To(Equal("etcd-test-client.etcd-test-namespace.svc"))
```
anveshreddy18:
```suggestion
	g.Expect(clientHostname).To(Equal(fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcdNamespace)))
```
The reason being the names of the `etcdName` and `etcdNamespace` constants can change anytime and the unit tests which are testing other functionality shouldn't break because of etcd name change, it should be resilient to that.

api/core/v1alpha1/helper_test.go
```diff
@@ -86,13 +117,44 @@ func TestGetFullSnapshotLeaseName(t *testing.T) {
 	g.Expect(fullSnapshotLeaseName).To(Equal("etcd-test-full-snap"))
 }
 
-func TestGetMemberLeaseNames(t *testing.T) {
+func TestGetMemberNameFromAddress(t *testing.T) {
 	g := NewWithT(t)
 	etcdObjMeta := createEtcdObjectMetadata(uuid.NewUUID(), nil, nil, false)
-	leaseNames := GetMemberLeaseNames(etcdObjMeta, 3)
+	memberName := GetMemberNameFromAddress(etcdObjMeta, "1.1.1.1")
+	g.Expect(memberName).To(Equal("etcd-test-1.1.1.1"))
```
anveshreddy18: Similar to https://github.com/gardener/etcd-druid/pull/1214#discussion_r3027492549

api/core/v1alpha1/helper_test.go
```diff
@@ -86,13 +117,44 @@ func TestGetFullSnapshotLeaseName(t *testing.T) {
...
+			Replicas: 3,
+			ExternallyManagedMemberAddresses: []string{
+				"1.1.1.1",
+				"1.1.1.2",
+				"1.1.1.3",
+			},
+		},
+	}
+	leaseNames := GetMemberLeaseNames(etcd)
+	g.Expect(leaseNames).To(Equal([]string{"etcd-test-1.1.1.1", "etcd-test-1.1.1.2", "etcd-test-1.1.1.3"}))
```
anveshreddy18: Similar to https://github.com/gardener/etcd-druid/pull/1214#discussion_r3027492549

internal/component/configmap/configmap_test.go
```diff
@@ -287,14 +304,17 @@ func TestTriggerDelete(t *testing.T) {
...
-func buildEtcd(replicas int32, clientTLSEnabled, peerTLSEnabled bool) *druidv1alpha1.Etcd {
+func buildEtcd(replicas int32, clientTLSEnabled, peerTLSEnabled bool, externallyManagedMemberAddresses []string) *druidv1alpha1.Etcd {
 	etcdBuilder := testutils.EtcdBuilderWithDefaults(testutils.TestEtcdName, testutils.TestNamespace).WithReplicas(replicas)
 	if clientTLSEnabled {
 		etcdBuilder.WithClientTLS()
 	}
 	if peerTLSEnabled {
 		etcdBuilder.WithPeerTLS()
 	}
+	if len(externallyManagedMemberAddresses) > 0 {
```
anveshreddy18, line 315: Similar to https://github.com/gardener/etcd-druid/pull/1214#discussion_r3027350156. We might want to check for the presence of the field itself instead of no.of elements in it.

internal/component/memberlease/memberlease_test.go
```diff
@@ -158,35 +188,40 @@ func TestSync(t *testing.T) {
 			t.Parallel()
 			// *************** set up existing environment *****************
 			etcdBuilder := testutils.EtcdBuilderWithDefaults(testutils.TestEtcdName, testutils.TestNamespace)
-			etcd := etcdBuilder.WithReplicas(tc.etcdReplicas).Build()
+			if len(tc.existingExternalMembers) > 0 {
```
anveshreddy18, line 191: Similar to https://github.com/gardener/etcd-druid/pull/1214#discussion_r3027350156, here and few other places.

internal/component/memberlease/memberlease.go
```diff
@@ -106,6 +134,18 @@ func (r _resource) doCreateOrUpdate(ctx component.OperatorContext, etcd *druidv1
 	return nil
 }
 
+func (r _resource) doDelete(ctx component.OperatorContext, objKey client.ObjectKey) error {
+	lease := emptyMemberLease(objKey)
+	if err := r.client.Delete(ctx, lease); err != nil {
```
anveshreddy18: Can we handle the case where it returns a NotFound error as the deletion can happen concurrently or asynchronously so when this call might fail because the corresponding lease might've already been deleted, in which case we don't need to requeue again with a failure.

internal/component/statefulset/builder.go
```diff
@@ -137,7 +137,7 @@ func (b *stsBuilder) getStatefulSetLabels() map[string]string {
 
 func (b *stsBuilder) createStatefulSetSpec(ctx component.OperatorContext) error {
 	err := b.createPodTemplateSpec(ctx)
-	b.sts.Spec.Replicas = ptr.To(utils.IfConditionOr(druidv1alpha1.IsEtcdRuntimeComponentCreationEnabled(b.etcd.ObjectMeta), b.replicas, 0))
+	b.sts.Spec.Replicas = ptr.To(utils.IfConditionOr(druidv1alpha1.IsPodManagementEnabled(b.etcd), b.replicas, 0))
 	b.logger.Info("Creating StatefulSet spec", "replicas", b.sts.Spec.Replicas, "name", b.sts.Name, "namespace", b.sts.Namespace)
 	b.sts.Spec.UpdateStrategy = defaultUpdateStrategy
 	if err != nil {
```
anveshreddy18, line 140: The error check happens at a later stage which needs to be moved up to fail fast upon an non-nil error is returned in the call to `createPodTemplateSpec`.

api/core/v1alpha1/helper.go
```diff
@@ -24,6 +26,22 @@ func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
 	return fmt.Sprintf("%s-client", etcdObjMeta.Name)
 }
 
+// GetClientHostname returns the hostname of the client endpoint for the Etcd cluster. This is the client service hostname when the Etcd members
+// are managed by etcd-druid, else it is a random member address from the externally managed member addresses.
+func GetClientHostname(etcd *Etcd) string {
+	if IsPodManagementEnabled(etcd) {
+		return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.Namespace)
+	} else {
```
anveshreddy18: Since the `if` returns always, we can get rid of the else block to make it more readable.

api/core/v1alpha1/helper.go
```diff
@@ -54,12 +77,17 @@ func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
...
 // GetMemberLeaseNames returns the name of member leases for the Etcd.
-func GetMemberLeaseNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
-	leaseNames := make([]string, replicas)
-	for i := range int(replicas) {
-		leaseNames[i] = fmt.Sprintf("%s-%d", etcdObjMeta.Name, i)
+func GetMemberLeaseNames(etcd *Etcd) []string {
+	if IsPodManagementEnabled(etcd) {
+		return GetAllPodNames(etcd.ObjectMeta, etcd.Spec.Replicas)
+	} else {
+		memberAddresses := etcd.Spec.ExternallyManagedMemberAddresses
```
anveshreddy18: Same as https://github.com/gardener/etcd-druid/pull/1214#discussion_r3027881841

anveshreddy18 (changes requested):
Thank you @CaptainIRS for the PR, the approach looks good, I have few comments, PTAL and lmk what you think of those, also I observed one thing 👇 

I think the `Status.Ready` is still taken from STS and will always be `false` because it checks the below condition:
```
if statefulSet.Status.ReadyReplicas < etcdReplicas {
		return false, fmt.Sprintf("not enough ready replicas (%d/%d)", statefulSet.Status.ReadyReplicas, etcdReplicas)
	}
```
Which will always be `false` since sts replicas is `0` and `etcdReplicas` is usually >0.

Also, the other conditions like `Status.CurrentReplicas`, `Status.ReadyReplicas`, and `Status.Replicas` are all derived from STS status and will be 0,0,0.

api/core/v1alpha1/etcd.go
CaptainIRS: I think `involved in management of etcd members like Pods, Services and PDBs.` would be the correct way. Could you reconsider this suggestion?

api/core/v1alpha1/etcd.go
CaptainIRS: After a discussion with @unmarshall we decided to keep it open for future extensibility by retaining the name of `ExternallyManagedMemberAddresses`. I've also updated the validation to not consider just IPv4 addresses. Since there is no way to perform the validation in a way compatible with Kubernetes v1.29, I've just updated the docstring with the valid values for now. After #1211, we could add the validations for this.

api/core/v1alpha1/helper.go
CaptainIRS: It is just implemented this way to mimic the behaviour of the client service as much as possible, so unless there is a strong reason to move to a static endpoint like the first member, we can keep it this way.

api/core/v1alpha1/helper.go
CaptainIRS: After discussions with @unmarshall we decided to keep the existing field as a slice as in effect the `len` check is the idiomatic way that handles all cases including 0 and nil.

docs/concepts/externally-managed-members.md
CaptainIRS: We would have to support sequential scale up from 1->2->3, so the intermediate even-numbered state would also be a valid state.

internal/component/configmap/configmap_test.go
CaptainIRS: I think in tests, this is the simpler way to do it. Unless you envision a better way, WDYT?

internal/component/memberlease/memberlease_test.go
CaptainIRS: Same as discussion in https://github.com/gardener/etcd-druid/pull/1214/files#r3027563795

internal/component/statefulset/builder.go
CaptainIRS: Addressed in [de3145e](https://github.com/gardener/etcd-druid/pull/1214/commits/de3145ebfb366369daac19cb8e246cb3f963b3e0)

CaptainIRS (commented):
@anveshreddy18 thanks for the review. I've addressed your comments, PTAL, thanks. The ready condition now works as expected (thanks for catching this!):

Image

@Shreyas-s14, I have made a change in [c9559a3](https://github.com/gardener/etcd-druid/pull/1214/commits/c9559a3278b08d76408b80d8302497fbc9ed7a8d) where I have changed the message emitted for some validations. I think the correct message should point to the fact that the dynamic addition/removal of the field is not allowed rather than the immutability since the condition itself is not checking for immutability in these cases. PTAL, thanks.

internal/component/memberlease/memberlease.go
CaptainIRS: Addressed in [de3145e](https://github.com/gardener/etcd-druid/pull/1214/commits/de3145ebfb366369daac19cb8e246cb3f963b3e0)

api/core/v1alpha1/helper.go
CaptainIRS: I had written it this way to explicitly convey the fact that only in case of externally managed members the following logic is executed. If you feel strongly about this structure, I can change this to an implicit return too, please do let me know your thoughts.

api/core/v1alpha1/helper_test.go
CaptainIRS: I've made the change and propagated this change to the whole file in [de3145e](https://github.com/gardener/etcd-druid/pull/1214/commits/de3145ebfb366369daac19cb8e246cb3f963b3e0). PTAL, thanks.

api/core/v1alpha1/helper_test.go
CaptainIRS: Same as https://github.com/gardener/etcd-druid/pull/1214#discussion_r3063865007

api/core/v1alpha1/helper_test.go
CaptainIRS: Same as https://github.com/gardener/etcd-druid/pull/1214#discussion_r3063865007

api/core/v1alpha1/helper.go
CaptainIRS: Same as https://github.com/gardener/etcd-druid/pull/1214#discussion_r3063859206

docs/concepts/externally-managed-members.md
anveshreddy18: Understood, thanks for the clarification.

api/core/v1alpha1/etcd.go
anveshreddy18: It's fine, I just wanted to insert the `etcd-druid` word there to differentiate two types of "management" as we'll be having, i.e `etcd-druid` vs `externally managed`. So it's upto you, no strong opinion.

internal/component/configmap/configmap_test.go
anveshreddy18: Now that we've agreed to keeping it as a slice directly, my above comment stands invalid. We can continue the current way.

api/core/v1alpha1/helper.go
anveshreddy18: Both are fine, we can keep it this way.

anveshreddy18 (approved):
Thanks for addressing my comments @CaptainIRS. LGTM!

PR #1226, opened 2025-12-08
Add cli flags for `etcdopstask` controller configurations in etcd-druid deployment when `OperatorConfig` is not used.

Shreyas-s14:

/area usability

Current setup for etcdopstask when not using the operator config based setup does not allow configurability in the controller config since there were no `CLI-flags` introduced in the `etcd-druid` deployment. This would then be set by the defaults.
This PR provides the cli flags to do so.

Fixes #

**Release note**:

```other operator
Added cli flags for etcdopstask controller configuration for non `Operator-Config` based `etcd-druid` setup.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ clean-mod-cache:
 
 .PHONY: update-dependencies
 update-dependencies:
-	@env GO111MODULE=on go get -u
+	@env GO111MODULE=on go get -u ./...
 	@make tidy
 
 .PHONY: add-license-headers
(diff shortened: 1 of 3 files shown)


charts/values.yaml
```diff
@@ -29,7 +29,7 @@ webhookPKI:
 # This configuration property is a short-lived property. It will be used to toggle between the now deprecated
 # CLI flags and the new OperatorConfiguration (which is the recommended way to configure etcd-druid operator).
 # To preserve backward compatibility the default value for this property is set to false.
```
anveshreddy18:
```suggestion
```
I think you can remove this line.

anveshreddy18 (approved):

PR #1229, opened 2025-12-10
Enable GHA test runs across all builds AND add api unit tests to it

anveshreddy18:

/area dev-productivity
/area testing
/kind task

This PR enables running unit & integration tests with Github Actions as these steps are pre-requisites for publishing OCI images on PR builds that can come in handy in PR & release testing.

It also adds api unit tests to run with GHA.

**Release note**:

```other developer
api module unit tests are now part of GHA tests and enabled GHA tests across all builds
```

### DIFF
--- a/.github/workflows/base.yaml
+++ b/.github/workflows/base.yaml
@@ -95,12 +95,12 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version: '1.25'
+      - name: tidy
+        run: make tidy
       - name: check
         run: make check
       - name: check-license-headers
         run: make check-license-headers
-      - name: tidy
-        run: make tidy
       - name: api-tidy
         run: make --directory=api tidy
       - name: api-check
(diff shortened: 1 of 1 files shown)


anveshreddy18:
Hold till we confirm whether to remove unit test, api unit test, integration tests from prow and instead enable all of them with GHA through this PR.
/hold

.github/workflows/base.yaml
```diff
@@ -103,6 +99,8 @@ jobs:
         run: make check
       - name: check-license-headers
         run: make check-license-headers
+      - name: tidy
+        run: make tidy
```
shreyas-s-rao: We'll need to run `make tidy` before `make check`, so that the check is performed after dependencies are downloaded, giving an accurate picture about static code vulnerabilities.

anveshreddy18: Thanks for pointing this out. I missed it.

PR #1236, opened 2025-12-18
Introduce `BackupSpec.StoreSpec.EndpointOverride` to override the object store's endpoint used by etcd-backup-restore

### LINKED ISSUE #1234: Enhance the API to support overriding backup bucket's endpoint

/area control-plane backup
/kind enhancement

The API that we expose through the etcd resource must be enhanced to allow users to configure the endpoint of their remote object store. This would be necessary for "special" regions like ME in GCP, China in Azure and so on, where the endpoint URL constructed for buckets does not follow the same syntax as the buckets in other regions like APJ, EU, NA, etc.

This would also be beneficial for consumers who use S3 compatible providers, by specifying the endpoint directly in the resource, instead of the credential secret as is currently being done. The following need to be done:

- [ ] etcd-backup-restore: https://github.com/gardener/etcd-backup-restore/issues/943
- [ ] etcd-druid: Enhance the API, and pass configuration to etcd-backup-restore
- [ ] provider-extensions: Enhance all relevant provider extensions with a mutating webhook to modify the `spec.backup` field of the etcd resource with the endpoint override.

**Why is this needed**:

Currently, configuration to etcd-backup-restore to override the bucket's endpoint is passed through credential secret files, which conflates credentials and configuration. Configuration must be explicit, and must be passed through the spec of the resource. The secret must only contain the credentials. https://github.com/gardener/etcd-backup-restore/issues/943 was raised for this.

renormalize:

/area backup compliance robustness
/kind enhancement

This PR enhances the API by introducing an optional `BackupSpec.StoreSpec.EndpointOverride` field to pass the object store's endpoint override information to etcd-backup-restore through its `--store-endpoint-override` flag, as introduced in https://github.com/gardener/etcd-backup-restore/pull/952.

Fixes #1234

The current code works with both S3, GCS and ABS emulators.

This PR requires https://github.com/gardener/etcd-backup-restore/pull/952 to be consumed by etcd-druid for this feature to work as intended.

**Release note**:

```feature operator
An optional field `BackupSpec.StoreSpec.EndpointOverride` is introduced in the etcd API, which enables operators to specify the object store's endpoint, overriding the cloud provider's default.
```

### DIFF
--- a/examples/etcd/druid_v1alpha1_etcd_azurite.yaml
+++ b/examples/etcd/druid_v1alpha1_etcd_azurite.yaml
@@ -51,7 +51,7 @@ spec:
       container: etcd-bucket
       prefix: etcd-test
       provider: ABS
-      endpointOverride: http://azurite-service.default:10000/devstoreaccount1/etcd-bucket
+      endpointOverride: http://azurite-service.default:10000/devstoreaccount1/
       secretRef:
         name: etcd-backup-azurite
     compression:
--- a/hack/ci-e2e-kind-azure.sh
+++ b/hack/ci-e2e-kind-azure.sh
@@ -30,8 +30,8 @@ rm -rf "${AZURE_APPLICATION_CREDENTIALS}"
 mkdir -p "${AZURE_APPLICATION_CREDENTIALS}"
 echo -n "${STORAGE_ACCOUNT}" > "${AZURE_APPLICATION_CREDENTIALS}/storageAccount"
 echo -n "${STORAGE_KEY}" > "${AZURE_APPLICATION_CREDENTIALS}/storageKey"
-echo -n "true" > "${AZURE_APPLICATION_CREDENTIALS}/emulatorEnabled"
-echo -n "${AZURITE_DOMAIN_LOCAL}" > "${AZURE_APPLICATION_CREDENTIALS}/domain"
+# echo -n "true" > "${AZURE_APPLICATION_CREDENTIALS}/emulatorEnabled"
+# echo -n "${AZURITE_DOMAIN_LOCAL}" > "${AZURE_APPLICATION_CREDENTIALS}/domain"
 
 # Deploy Azurite and run end-to-end tests
 make deploy-azurite
--- a/hack/ci-e2e-kind-gcs.sh
+++ b/hack/ci-e2e-kind-gcs.sh
@@ -17,7 +17,7 @@ trap '{
 
 kubectl wait --for=condition=ready node --all
 
-echo "{ \"serviceaccount.json\": { \"type\": \"service_account\", \"project_id\": \"theworld\" } }" >/tmp/svc_acc.json
+echo "{ \"type\": \"service_account\", \"project_id\": \"theworld\" }" >/tmp/svc_acc.json
 
 # Deploy fake-gcs and run e2e tests
 make deploy-fakegcs


api/core/v1alpha1/store.go
```diff
@@ -14,6 +14,10 @@ type StoreSpec struct {
 	// Container is the name of the container the backup is stored at.
 	// +optional
 	Container *string `json:"container,omitempty"`
+	// Endpoint denotes the storage endpoint that will be used to override the provider's default.
+	// +optional
+	// +kubebuilder:validation:XValidation:message="endpoint override must be a valid URL.",rule="isURL(self)"
+	Endpoint *string `json:"endpoint,omitempty"`
```
CaptainIRS: (feel free to ignore) I think this could be called `EndpointOverride` as having a field like `Endpoint` might lead to the assumption that this must be provided, whereas an `-Override` suffix implicitly conveys that it is an optional field.

internal/store/store.go
```diff
@@ -128,6 +128,11 @@ func GetProviderEnvVars(store *druidv1alpha1.StoreSpec) ([]corev1.EnvVar, error)
 
 	case GCS:
 		envVars = append(envVars, utils.GetEnvVarFromValue(common.EnvGoogleApplicationCredentials, fmt.Sprintf("%sserviceaccount.json", common.VolumeMountPathGCSBackupSecret)))
+		// `http` communication is disallowed by GCS APIs, so we assume an emulator is being used when "http://fake-gcs" is present in the endpoint.
+		// The GCS SDK requires EnvGoogleStorageEmulatorHost to be set to the emulator's endpoint.
+		if store.Endpoint != nil && strings.Contains(*store.Endpoint, "http://fake-gcs") {
+			envVars = append(envVars, utils.GetEnvVarFromValue(common.EnvGoogleStorageEmulatorHost, *store.Endpoint))
+		}
```
CaptainIRS: (not a strong opinion, but could be considered): Assuming the Google Cloud SDK's suboptimal handling of this, the workaround we're going to have is going to be hacky. But, I feel like the hack being in `etcd-backup-restore` is better suited than exposing this through `etcd-druid`. Since the SDK has a quirk that the URL has to be `https` unless it is specifically an emulator as you rightly mentioned in the comments, we can have a special case for `http` endpoints in the `etcd-backup-restore` codebase to set the environment variable through Go code (as suggested in the official documentation, which itself is not optimal).

renormalize: Thanks for the feedback! I agree that the handling existing in etcd-backup-restore might be the better way to go about this. Will push the corresponding change.

api/core/v1alpha1/store.go
renormalize: That's a fair suggestion. I don't have any objections to it. Will push the corresponding change.

internal/store/store.go
renormalize: Addressed in [c95a80b](https://github.com/gardener/etcd-druid/pull/1236/commits/c95a80b84e84d85ac083bf069a5b88ebeb6957df).

Made the corresponding change in etcd-backup-restore in [2a3b891](https://github.com/gardener/etcd-backup-restore/pull/952/commits/2a3b8912a3556cbeaf7de88108d21290091094e3).

api/core/v1alpha1/store.go
renormalize: Addressed in [61ea557](https://github.com/gardener/etcd-druid/pull/1236/commits/61ea557fa7304f82969a5fbf4acd7444c9969e11).

examples/etcd/druid_v1alpha1_etcd_azurite.yaml
```diff
@@ -51,6 +51,7 @@ spec:
       container: etcd-bucket
       prefix: etcd-test
       provider: ABS
+      endpoint: http://azurite-service.default:10000/devstoreaccount1/etcd-bucket
```
CaptainIRS:
```suggestion
      endpointOverride: http://azurite-service.default:10000/devstoreaccount1/etcd-bucket
```

examples/etcd/druid_v1alpha1_etcd_fakegcs.yaml
```diff
@@ -48,6 +48,7 @@ spec:
     deltaSnapshotPeriod: 300s
     deltaSnapshotMemoryLimit: 1Gi
     store:
+      endpoint: http://fake-gcs.default:8000/storage/v1/b
```
CaptainIRS:
```suggestion
      endpointOverride: http://fake-gcs.default:8000/storage/v1/b
```

examples/etcd/druid_v1alpha1_etcd_localstack.yaml
```diff
@@ -51,6 +51,7 @@ spec:
       container: etcd-bucket
       prefix: etcd-test
       provider: S3
+      endpoint: http://localstack.default:4566
```
CaptainIRS:
```suggestion
      endpointOverride: http://localstack.default:4566
```

renormalize: Addressed in [75a03f3](https://github.com/gardener/etcd-druid/pull/1236/commits/75a03f377260cc7301c0808d70f24f712cba8339).

examples/etcd/druid_v1alpha1_etcd_azurite.yaml
renormalize: Addressed in [75a03f3](https://github.com/gardener/etcd-druid/pull/1236/commits/75a03f377260cc7301c0808d70f24f712cba8339).

examples/etcd/druid_v1alpha1_etcd_fakegcs.yaml
renormalize: Addressed in [75a03f3](https://github.com/gardener/etcd-druid/pull/1236/commits/75a03f377260cc7301c0808d70f24f712cba8339).

api/core/v1alpha1/store.go
```diff
@@ -14,6 +14,10 @@ type StoreSpec struct {
 	// Container is the name of the container the backup is stored at.
 	// +optional
 	Container *string `json:"container,omitempty"`
+	// EndpointOverride denotes the storage endpoint that will be used to override the provider's default.
```
ishan16696:
```suggestion
	// EndpointOverride denotes the storage endpoint that will be used to override the storage provider's default endpoint.
```

ishan16696 (approved):
LGTM, just one nit:

examples/objstore-emulator/etcd-secret-azurite.yaml
```diff
@@ -8,6 +8,4 @@ type: Opaque
 data: 
   storageAccount: ZGV2c3RvcmVhY2NvdW50MQ==
   storageKey: RWJ5OHZkTTAyeE5PY3FGbHFVd0pQTGxtRXRsQ0RYSjFPVXpGVDUwdVNSWjZJRnN1RnEyVVZFckN6NEk2dHEvSzFTWkZQVE90ci9LQkhCZWtzb0dNR3c9PQ==
-  emulatorEnabled: dHJ1ZQ== # true
-  # TODO: going further, support [production-style URLs](https://github.com/Azure/Azurite?tab=readme-ov-file#production-style-url) for Azurite
-  domain: YXp1cml0ZS1zZXJ2aWNlLmRlZmF1bHQ6MTAwMDA= # azurite-service.default:10000 (assuming azurite service runs in `default` namespace), emulatorEnabled has to be true
+  # domain: YXp1cml0ZS1zZXJ2aWNlLmRlZmF1bHQ6MTAwMDA= # azurite-service.default:10000 (assuming azurite service runs in `default` namespace) # passing domain through credentials is deprecated
```
shreyas-s-rao, line 11: Please re-add the TODO for supporting production-style URLs, because that would still be a good-to-have feature for this

shreyas-s-rao (changes requested):
@renormalize thanks for the PR. Just one small nit, but overall LGTM.

examples/objstore-emulator/etcd-secret-azurite.yaml
renormalize: @shreyas-s-rao thanks for the review!

With the new changes being brought in etcd-backup-restore in https://github.com/gardener/etcd-backup-restore/pull/952, we need no additional changes to support production style URLs from etcd-backup-restore.

While running Azurite in the kind cluster, the URL for the bucket becomes `http://devstoreaccount1.blob.azurite-service.default:10000/container`, since the account and the azure service type needs to be a part of the hostname. This becomes a problem in a Kubernetes context since a simple service `azurite-service.default` can not have subdomains like this, and will fail to route requests correctly to the emulator.

The correct way to approach this is to deploy an ingress resource and controller that would be able to route the requests correctly to the emulator.

However, I feel that would not be worth the effort. Thus I removed the TODO. If you still think there is indeed some value in supporting production style URLs, I will reintroduce this TODO.

shreyas-s-rao: Sure, thanks for the clarification. For the emulator usecase, what we have is sufficient. If at all at a later point someone has a need to use a custom Azure endpoint, then we can bring up this topic again.

shreyas-s-rao: Sure, thanks for the clarification. For the emulator usecase, what we have is sufficient. If at all at a later point someone has a need to use a custom Azure endpoint, then we can bring up this topic again.

PR #1255, opened 2026-01-12
Code improvements using modernize

shafeeqes:

/area quality
/kind enhancement

The fixes in this PR are applied using 
```
$ go run golang.org/x/tools/go/analysis/passes/modernize/cmd/modernize@latest -omitzero=false -fix ./...
```
Refer https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize for details on each fix.

Fixes #

**Release note**:

```other operator
NONE
```


anveshreddy18:
Thanks for "modernising" the code @shafeeqes 😉 

As for changing from empty interface `interface{}` to `any`, since it's only a stylistic change I don't have any opinion but are we doing the same across all repos to be consistent? If already done or in progress, then looks good.

renormalize (approved):
thanks~

PR #1257, opened 2026-01-12
Enhance `OperatorConfiguration` to allow it to be incorporated as part of other CRs

### LINKED ISSUE #1256: Enhance `OperatorConfiguration` to allow it to be incorporated as part of other CRs

/area open-source usability
/kind api-change enhancement

* Fields in `druidconfigv1alpha1.OperatorConfiguration` that have the `bool` datatype should be changed to `*bool` to be able to distinguish between an unset field and a field set to `false`.
* Validations for subfields of `OperatorConfiguration` should also be exported for consumers of the API.

**Why is this needed**:

As a consumer of `druidconfigv1alpha1.OperatorConfiguration`, I would like to expose fields in the `OperatorConfiguration` to my users (as subfields in my own CR), but with my own defaults/fallbacks instead of ones provided by etcd-druid. For example, I would like to expose `.controllers.etcd.concurrentSyncs` with my default of `50` instead of `3` (etcd-druid default). While trying to accomplish this, I run into 2 issues:
* For `bool` fields that I would like to default to `true`, there is no way to distinguish between a field that is not set by my user vs. a field that is set to `false`.
* Only validations for the whole `OperatorConfiguration` is exported to be used by consumers, but if I expose just a subset of fields within `OperatorConfiguration`, I'd like to have the granular validations exposed to be consumed as well.

CaptainIRS:

/area open-source usability
/kind api-change enhancement

* Fields in `druidconfigv1alpha1.OperatorConfiguration` that have the `bool` datatype are changed to `*bool` to be able to distinguish between an unset field and a field set to `false`.
* Validations for subfields of `OperatorConfiguration` are exported for consumers of the API to utilize.

Fixes #1256 

**Checklist**:
    - [x] Unit tests

**Release note**:

```breaking user
Users who programmatically use `OperatorConfiguration` to setup etcd-druid should adapt their logic to use `bool` pointers in place of `bool` fields.
```
```noteworthy user
Validations for sub-fields of `OperatorConfiguration` are now exported.
```

### DIFF
--- a/api/config/v1alpha1/validation/validation.go
+++ b/api/config/v1alpha1/validation/validation.go
@@ -40,7 +40,7 @@ func ValidateClientConnectionConfiguration(clientConnConfig druidconfigv1alpha1.
 // ValidateLeaderElectionConfiguration validates the leader election configuration.
 func ValidateLeaderElectionConfiguration(leaderElectionConfig druidconfigv1alpha1.LeaderElectionConfiguration, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
-	if leaderElectionConfig.Enabled == nil || !*leaderElectionConfig.Enabled {
+	if !ptr.Deref(leaderElectionConfig.Enabled, false) {
 		return allErrs
 	}
 	allErrs = append(allErrs, mustBeGreaterThanZeroDuration(leaderElectionConfig.LeaseDuration, fldPath.Child("leaseDuration"))...)
@@ -93,7 +93,7 @@ func validateEtcdControllerConfiguration(etcdControllerConfig druidconfigv1alpha
 
 func validateCompactionControllerConfiguration(compactionControllerConfig druidconfigv1alpha1.CompactionControllerConfiguration, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
-	if compactionControllerConfig.Enabled == nil || !*compactionControllerConfig.Enabled {
+	if !ptr.Deref(compactionControllerConfig.Enabled, false) {
 		return allErrs
 	}
 	allErrs = append(allErrs, validateConcurrentSyncs(compactionControllerConfig.ConcurrentSyncs, fldPath.Child("concurrentSyncs"))...)
@@ -110,7 +110,7 @@ func validateCompactionControllerConfiguration(compactionControllerConfig druidc
 
 func validateEtcdCopyBackupsTaskControllerConfiguration(etcdCopyBackupsTaskControllerConfig druidconfigv1alpha1.EtcdCopyBackupsTaskControllerConfiguration, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
-	if etcdCopyBackupsTaskControllerConfig.Enabled == nil || !*etcdCopyBackupsTaskControllerConfig.Enabled {
+	if !ptr.Deref(etcdCopyBackupsTaskControllerConfig.Enabled, false) {
 		return allErrs
 	}
 	allErrs = append(allErrs, validateConcurrentSyncs(etcdCopyBackupsTaskControllerConfig.ConcurrentSyncs, fldPath.Child("concurrentSyncs"))...)
--- a/api/config/v1alpha1/validation/validation_test.go
+++ b/api/config/v1alpha1/validation/validation_test.go
@@ -116,7 +116,7 @@ func TestValidateLeaderElectionConfiguration(t *testing.T) {
 				Enabled: test.enabled,
 			}
 			druidconfigv1alpha1.SetDefaults_LeaderElectionConfiguration(leaderElectionConfig)
-			if test.enabled != nil && *test.enabled {
+			if ptr.Deref(test.enabled, false) {
 				updateLeaderElectionConfig(leaderElectionConfig, test.overrideLeaseDuration, test.overrideRenewDeadline, test.overrideRetryPeriod, test.overrideResourceLock, test.overrideResourceName)
 			}
 			actualErrList := ValidateLeaderElectionConfiguration(*leaderElectionConfig, fldPath)


internal/webhook/register.go
```diff
@@ -15,7 +15,7 @@ import (
 // Register registers all etcd-druid webhooks with the controller manager.
 func Register(mgr ctrl.Manager, config druidconfigv1alpha1.WebhookConfiguration) error {
 	// Add Etcd Components webhook to the manager
-	if config.EtcdComponentProtection.Enabled {
+	if config.EtcdComponentProtection.Enabled != nil && *config.EtcdComponentProtection.Enabled {
```
renormalize: Nit: just a stylistic suggestion since this reads better. This could be applied to other similar changes in this PR.

```suggestion
	if ptr.Deref(config.EtcdComponentProtection.Enabled, false) {
```

internal/webhook/utils/config.go
```diff
@@ -9,5 +9,5 @@ import druidconfigv1alpha1 "github.com/gardener/etcd-druid/api/config/v1alpha1"
 // AtLeaseOneEnabled returns true if at least one webhook is enabled.
 // NOTE for contributors: For every new webhook, add a disjunction condition with the webhook's Enabled field.
 func AtLeaseOneEnabled(config druidconfigv1alpha1.WebhookConfiguration) bool {
-	return config.EtcdComponentProtection.Enabled
+	return config.EtcdComponentProtection.Enabled != nil && *config.EtcdComponentProtection.Enabled
```
renormalize: https://github.com/gardener/etcd-druid/pull/1257#discussion_r2703890777

internal/manager/manager.go
```diff
@@ -65,7 +65,7 @@ func createManager(operatorConfig *druidconfigv1alpha1.OperatorConfiguration) (c
 		&eventsv1.Event{},
 	}
 
-	if operatorConfig.Controllers.DisableLeaseCache {
+	if operatorConfig.Controllers.DisableLeaseCache != nil && *operatorConfig.Controllers.DisableLeaseCache {
```
renormalize: https://github.com/gardener/etcd-druid/pull/1257#discussion_r2703890777

internal/controller/etcd/reconciler.go
```diff
@@ -149,7 +149,7 @@ func (r *Reconciler) GetOperatorRegistry() component.Registry {
 
 func createAndInitializeOperatorRegistry(client client.Client, config druidconfigv1alpha1.EtcdControllerConfiguration, imageVector imagevector.ImageVector) component.Registry {
 	reg := component.NewRegistry()
-	reg.Register(component.ServiceAccountKind, serviceaccount.New(client, config.DisableEtcdServiceAccountAutomount))
+	reg.Register(component.ServiceAccountKind, serviceaccount.New(client, config.DisableEtcdServiceAccountAutomount != nil && *config.DisableEtcdServiceAccountAutomount))
```
renormalize: https://github.com/gardener/etcd-druid/pull/1257#discussion_r2703890777

api/config/v1alpha1/validation/validation.go
```diff
@@ -19,26 +19,28 @@ import (
...
+	allErrs = append(allErrs, ValidateControllerConfiguration(config.Controllers, field.NewPath("controllers"))...)
+	allErrs = append(allErrs, ValidateLogConfiguration(config.Logging, field.NewPath("log"))...)
+	allErrs = append(allErrs, ValidateWebhookConfiguration(config.Webhooks, field.NewPath("webhooks"))...)
 
 	return allErrs
 }
 
-func validateClientConnectionConfiguration(clientConnConfig druidconfigv1alpha1.ClientConnectionConfiguration, fldPath *field.Path) field.ErrorList {
+// ValidateClientConnectionConfiguration validates the client connection configuration.
+func ValidateClientConnectionConfiguration(clientConnConfig druidconfigv1alpha1.ClientConnectionConfiguration, fldPath *field.Path) field.ErrorList {
```
renormalize, line 32: Can you help me understand why these helper functions are also now exposed? Do we not expect consumers to use `ValidateOperatorConfiguration` directly?

cmd/opts/deprecatedoptions.go
```diff
@@ -123,27 +124,27 @@ func (d *deprecatedOperatorConfiguration) ToOperatorConfiguration() *druidconfig
 	config.Server.Webhooks.BindAddress = d.webhookServerBindAddress
 	config.Server.Webhooks.Port = d.webhookServerPort
 	config.Server.Webhooks.ServerCertDir = d.webhookServerTLSCertDir
-	config.LeaderElection.Enabled = d.leaderElectionEnabled
+	config.LeaderElection.Enabled = ptr.To(d.leaderElectionEnabled)
```
renormalize: Nit: `ptr.To` is helpful with values more than variables - since we already have variables we can just use `&`? It also keeps consistent with the other lines in this function using `&`. Objections?

```suggestion
	config.LeaderElection.Enabled = &d.leaderElectionEnabled
```

CaptainIRS: Thanks for the suggestion, fixed in 77a900ccee13ef51d7b776952deee9e764bc2523

internal/webhook/register.go
CaptainIRS: That is very elegant, thanks for the suggestion! Addressed in 77aa9e7b8abef6544961012f0c1778a25dcc2858

internal/webhook/utils/config.go
CaptainIRS: Addressed in 77aa9e7b8abef6544961012f0c1778a25dcc2858

internal/controller/etcd/reconciler.go
CaptainIRS: Addressed in 77aa9e7b8abef6544961012f0c1778a25dcc2858

internal/manager/manager.go
CaptainIRS: Addressed in 77aa9e7b8abef6544961012f0c1778a25dcc2858

api/config/v1alpha1/validation/validation.go
CaptainIRS: Usually, that is true. But in some cases (like gardener/gardener), we intend to expose just a subset of fields from `OperatorConfiguration` to consumers (just `ControllerConfiguration` in case of gardener/gardener). In such cases, the whole configuration is not available when the validation is done, and hence it is beneficial to have validations for fields within `OperatorConfiguration` to also be exposed to druid API consumers.

renormalize: Understood, thanks for the explanation!

api/config/v1alpha1/types.go
```diff
@@ -51,7 +51,7 @@ type ClientConnectionConfiguration struct {
 type LeaderElectionConfiguration struct {
 	// Enabled specifies whether leader election is enabled. Set this
 	// to true when running replicated instances of the operator for high availability.
-	Enabled bool `json:"enabled"`
+	Enabled *bool `json:"enabled,omitempty"`
```
unmarshall, line 54: ideally it does not make sense to have this as a pointer to boolean. The only valid values are `false|true` with default being as false which is the same as if its not set.

api/config/v1alpha1/types.go
```diff
@@ -110,7 +110,7 @@ type ControllerConfiguration struct {
 	// Deprecated: This field will be eventually removed. It is recommended to not use this.
 	// It has only been introduced to allow for backward compatibility with the old CLI flags.
 	// +optional
-	DisableLeaseCache bool `json:"disableLeaseCache"`
+	DisableLeaseCache *bool `json:"disableLeaseCache,omitempty"`
```
unmarshall, line 113: This field has been marked as deprecated. Is it still being configured? We should not be changing this field as it is going to be removed.

api/config/v1alpha1/types.go
```diff
@@ -135,9 +135,9 @@ type EtcdControllerConfiguration struct {
 	// NOTE: Decision to enable it should be carefully taken as spec updates could potentially result in rolling update
 	// of the StatefulSet which will cause a minor downtime for a single node etcd cluster and can potentially cause a
 	// downtime for a multi-node etcd cluster.
-	EnableEtcdSpecAutoReconcile bool `json:"enableEtcdSpecAutoReconcile"`
+	EnableEtcdSpecAutoReconcile *bool `json:"enableEtcdSpecAutoReconcile,omitempty"`
```
unmarshall, line 138: For this field as well, the bool default value of false is correct. There is no need to distinguish if this was set or not.

api/config/v1alpha1/types.go
```diff
@@ -162,7 +162,7 @@ type SecretControllerConfiguration struct {
 // CompactionControllerConfiguration defines the configuration for the compaction controller.
 type CompactionControllerConfiguration struct {
 	// Enabled specifies whether backup compaction should be enabled.
-	Enabled bool `json:"enabled"`
+	Enabled *bool `json:"enabled,omitempty"`
```
unmarshall, line 165: These are added capabilities that are provided by etcd-druid. By default its value should be false and thus it does not make sense to keep it as a pointer to bool.
Same for `EtcdCopyBackupsTaskControllerConfiguration` and `EtcdComponentProtectionWebhookConfiguration`

PR #1266, opened 2026-02-02
Use `StoreSpec.EndpointOverride` in `EtcdCopyBackupsTask`.

### LINKED ISSUE #1234: Enhance the API to support overriding backup bucket's endpoint

renormalize:

/area backup
/kind enhancement

* Append etcdcopybackupstask args with `--store-endpoint-override` when required.

Fixes #1234 

This was missed in #1236. I seem to have not committed this to the right branch during development. This miss was caught during release validation testing.

**Release note**:

```other operator
`EtcdCopyBackupsTask` also supports bucket override specified in `StoreSpec.EndpointOverride`.
```

### DIFF
--- a/internal/controller/etcdcopybackupstask/reconciler_test.go
+++ b/internal/controller/etcdcopybackupstask/reconciler_test.go
@@ -274,22 +274,33 @@ var _ = Describe("EtcdCopyBackupsTaskController", func() {
 					Container: ptr.To("store_container"),
 				}
 				provider = "storage_provider"
-				prefix = "prefix"
+				prefix = "prefix-"
 			})
 
 			It("returns a argument slice with provider, prefix, and container information", func() {
 				expected := []string{
-					"--prefixstorage-provider=storage_provider",
-					"--prefixstore-prefix=store_prefix",
-					"--prefixstore-container=store_container",
+					"--prefix-storage-provider=storage_provider",
+					"--prefix-store-prefix=store_prefix",
+					"--prefix-store-container=store_container",
+				}
+				Expect(createJobArgumentFromStore(&store, provider, prefix)).To(Equal(expected))
+			})
+
+			It("returns a argument slice with provider, prefix, container, and endpoint override information", func() {
+				store.EndpointOverride = ptr.To("http://override-information")
+				expected := []string{
+					"--prefix-storage-provider=storage_provider",
+					"--prefix-store-prefix=store_prefix",
+					"--prefix-store-container=store_container",
+					"--prefix-store-endpoint-override=http://override-information",
 				}
 				Expect(createJobArgumentFromStore(&store, provider, prefix)).To(Equal(expected))
 			})
 
 			It("should return a argument slice with provider and prefix information only when StoreSpec.Container is nil", func() {
 				expected := []string{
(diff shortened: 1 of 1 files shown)


internal/controller/etcdcopybackupstask/reconciler.go
```diff
@@ -553,5 +553,9 @@ func createJobArgumentFromStore(store *druidv1alpha1.StoreSpec, provider, prefix
 	if store.Container != nil && len(*store.Container) > 0 {
 		arguments = append(arguments, argPrefix+"store-container="+*store.Container)
 	}
+
+	if store.EndpointOverride != nil && len(*store.EndpointOverride) > 0 {
+		arguments = append(arguments, argPrefix+"store-endpoint-override="+*store.EndpointOverride)
```
CaptainIRS, line 558: Could we add a unit test for consistency? (since other flags seem to have unit tests too)
Something like:
```go
It("should return an argument slice with endpoint override information", func() {
	expected := []string{
		"--store-endpoint-override=custom-endpoint",
	}
	store.EndpointOverride = &endpointOverride
	Expect(createJobArgumentFromStore(&store, provider, prefix)).To(Equal(expected))
})
```

CaptainIRS (approved):
@renormalize thanks for fixing this. Just one comment about unit tests, but other than that, LGTM!

internal/controller/etcdcopybackupstask/reconciler.go
renormalize: Of course, I forgot marking this as draft while raising this since I wanted to test a bit more. Thanks for the review!

renormalize: Adapted an existing unit test in [f51180f](https://github.com/gardener/etcd-druid/pull/1266/commits/f51180ff94518b19082b4d03efcf8afc392936ac). Don't want to add tests to test different permutations, such a stub would be sufficient in my opinion.

CaptainIRS: Sounds good, no issues.

PR #1277, opened 2026-02-06
Add GHA workflow `PR Review Timeline`

shreyas-s-rao:

/area dev-productivity
/kind enhancement

Introduce new GHA workflow `PR Review Timeline`, which uses https://github.com/marketplace/actions/pr-review-timeline to publish the review timeline for a PR in the Checks tab. This makes it easier to track PR timelines, especially useful for large or long-running PRs with multiple review cycles.

**Checklist**:
NA

This is what the output will look like:
Screenshot 2026-02-06 at 11 14 09 AM

Current summary for this action is already available for this PR here: https://github.com/gardener/etcd-druid/actions/runs/21740272050?pr=1277

Please note, this will only be available in the action execution page and not the PR description itself, since the action input `publish-to-pr` is set to `false` by default.

**Release note**:

```other developer
Add GHA workflow `PR Review Timeline` for easy tracking of reviewer timelines for PRs.
```

### DIFF
--- a/hack/tools.mk
+++ b/hack/tools.mk
@@ -91,19 +91,19 @@ $(MOCKGEN):
 	GOBIN=$(abspath $(TOOLS_BIN_DIR)) go install go.uber.org/mock/mockgen@$(MOCKGEN_VERSION)
 
 $(SETUP_ENVTEST):
-	curl -Lo $(SETUP_ENVTEST) https://github.com/kubernetes-sigs/controller-runtime/releases/download/$(CONTROLLER_RUNTIME_VERSION)/setup-envtest-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
+	curl -fLo $(SETUP_ENVTEST) --retry 5 --retry-delay 5 https://github.com/kubernetes-sigs/controller-runtime/releases/download/$(CONTROLLER_RUNTIME_VERSION)/setup-envtest-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
 	chmod +x $(SETUP_ENVTEST)
 
 $(SKAFFOLD):
-	curl -Lo $(SKAFFOLD) https://storage.googleapis.com/skaffold/releases/$(SKAFFOLD_VERSION)/skaffold-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
+	curl -fLo $(SKAFFOLD) --retry 5 --retry-delay 5 https://storage.googleapis.com/skaffold/releases/$(SKAFFOLD_VERSION)/skaffold-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
 	chmod +x $(SKAFFOLD)
 
 $(KIND):
-	curl -Lo $(KIND) https://kind.sigs.k8s.io/dl/$(KIND_VERSION)/kind-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-review-timeline.yaml
```diff
@@ -0,0 +1,26 @@
+name: PR Review Timeline
+
+on:
+  pull_request:
+    types: [opened, closed, reopened, ready_for_review, review_requested, assigned, synchronize]
```
ishan16696, line 5: what is `synchronize` here ? It's not mentioned here: https://github.com/marketplace/actions/pr-review-timeline, that's why I'm asking

shreyas-s-rao: The link is only an example. We can choose which triggers to use for this action. I've added `synchronize` because whenever a new commit is added (or the commit history is changed by a force push / rebase), it can invalidate/re-assign reviewers to the PR, so the PR review timeline needs to be recomputed at that point.

ishan16696: ok, thanks for explaining.

hack/tools.mk
```diff
@@ -91,19 +91,19 @@ $(MOCKGEN):
 	GOBIN=$(abspath $(TOOLS_BIN_DIR)) go install go.uber.org/mock/mockgen@$(MOCKGEN_VERSION)
 
 $(SETUP_ENVTEST):
-	curl -Lo $(SETUP_ENVTEST) https://github.com/kubernetes-sigs/controller-runtime/releases/download/$(CONTROLLER_RUNTIME_VERSION)/setup-envtest-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
+	curl -fLo $(SETUP_ENVTEST) --retry 5 --retry-delay 5 https://github.com/kubernetes-sigs/controller-runtime/releases/download/$(CONTROLLER_RUNTIME_VERSION)/setup-envtest-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
```
ishan16696, line 94: do we need 5 retry ? Isn't too much ... I'm thinking of having 2 or 3 , wdyt ?

hack/tools.mk
```diff
@@ -91,19 +91,19 @@ $(MOCKGEN):
...
 	chmod +x $(SETUP_ENVTEST)
 
 $(SKAFFOLD):
-	curl -Lo $(SKAFFOLD) https://storage.googleapis.com/skaffold/releases/$(SKAFFOLD_VERSION)/skaffold-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
+	curl -fLo $(SKAFFOLD) --retry 5 --retry-delay 5 https://storage.googleapis.com/skaffold/releases/$(SKAFFOLD_VERSION)/skaffold-$(SYSTEM_NAME)-$(SYSTEM_ARCH)
```
ishan16696, line 98: same as above and all places where you've added this

hack/tools.mk
shreyas-s-rao: I think 5 is standard for HTTP calls, although 3 is also ok, I have no strong preferences. It's more probable for HTTP calls to temporarily fail, so allowing 5 attempts should be fine. For CI pipeline, more retries would be better (to increase the chances of success), and when running locally 5 attempts is generally ok as compared to 3, since the end goal is for the HTTP call to finally succeed. WDYT?

hack/tools.mk
shreyas-s-rao: Same as above

hack/tools.mk
ishan16696: ok, make sense ... I was thinking `5` seems more but I also don't have strong opinion.

ishan16696 (approved):

Issue #1237, opened 2025-12-23
☂️ Support for LiveCPM


acumino:
**How to categorize this issue?**

/area open-source
/kind enhancement

**What would you like to be added**:
This is an umbrella issue for all the changes required to support [LiveCPM](https://github.com/gardener/gardener/issues/10686) in Gardener.

**Tasks**

PR #1309, opened 2026-03-12
Introduce `MemberNamePrefix`

acumino:

/area control-plane-migration delivery open-source
/kind api-change enhancement epic

Adds an optional, immutable `spec.memberNamePrefix` field to the Etcd API. When set, member names become `<prefix>-<pod-name>` instead of defaulting to the pod name. Immutability is enforced via CEL validation since renaming live etcd members requires a full re-bootstrap.

Part of https://github.com/gardener/etcd-backup-restore/issues/896 https://github.com/gardener/etcd-druid/issues/1237 https://github.com/gardener/gardener/issues/10686

**Checklist**:
    - [x] Unit tests
    - [ ] Integration tests
    - [ ] E2E tests

- ~~WIP till TSC is approved.~~
- A PR will be opened in https://github.com/gardener/etcd-backup-restore to adapt these changes. - https://github.com/gardener/etcd-backup-restore/pull/1030 ✅

**Release note**:

```feature operator
The `Etcd` API now supports an optional `spec.memberNamePrefix` field. When set, each etcd member name becomes `<prefix>-<pod-name>` instead of defaulting to the pod name. The field is immutable once set.
```


acumino:
PR in backup-restore is also there - https://github.com/gardener/etcd-backup-restore/pull/1030
This PR can be reviewed now.
cc @ishan16696 @anveshreddy18

api/core/v1alpha1/etcd.go
```diff
@@ -425,7 +429,8 @@ const (
 
 // EtcdMemberStatus holds information about etcd cluster membership.
 type EtcdMemberStatus struct {
-	// Name is the name of the etcd member. It is the name of the backing `Pod`.
+	// Name is the name of the etcd member. It matches the member lease name.
+	// When MemberNamePrefix is set, it is "<prefix>-<pod-name>" otherwise it is the name of the backing `Pod`.
```
anveshreddy18:
```suggestion
	// When EtcdSpec.MemberNamePrefix is set, it is "<prefix>-<pod-name>" otherwise it is the name of the backing `Pod`.
```
In the context of this PR just having MemberNamePrefix makes sense but when reading through codebase at large, it's ideal to give full spec path as these two fields are not siblings, so it could be hard to read otherwise.

api/core/v1alpha1/etcd.go
```diff
@@ -314,7 +314,11 @@ type SchedulingConstraints struct {
...
 // +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
 // +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
 type EtcdSpec struct {
+	// MemberNamePrefix defines the prefix for the name of each etcd member. When set, the member name is "<prefix>-<pod-name>", otherwise it defaults to the pod name.
+	// +optional
+	MemberNamePrefix *string `json:"memberNamePrefix,omitempty"`
 	// selector is a label query over pods that should match the replica count.
```
anveshreddy18:
```suggestion
// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
// +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
type EtcdSpec struct {
	// MemberNamePrefix defines the prefix for the name of each etcd member. When set, the member name is "<prefix>-<pod-name>", otherwise it defaults to the pod name.
	// +optional
    // +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field",rule="self == oldSelf"
	MemberNamePrefix *string `json:"memberNamePrefix,omitempty"`
	// selector is a label query over pods that should match the replica count.
```
Right now the immutability constraint is not enforced through CEL as there are layers to enforcing CEL expressions, different layers have different responsibilities. The struct level rule typically checks for the continuous existence of the field ( your change does that), but it doesn't directly enforce immutability. The way to do that is by adding field level marker like shown.

api/core/v1alpha1/etcd.go
```diff
@@ -314,7 +314,11 @@ type SchedulingConstraints struct {
 // +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate field cannot be added or removed dynamically.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
 // +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
 type EtcdSpec struct {
+	// MemberNamePrefix defines the prefix for the name of each etcd member. When set, the member name is "<prefix>-<pod-name>", otherwise it defaults to the pod name.
+	// +optional
+	MemberNamePrefix *string `json:"memberNamePrefix,omitempty"`
```
anveshreddy18, line 360: Since the member prefix is also being used in the lease names, there should be a character limit on how lengthy the prefix can be because lease name potentially does not exceed the 253 char limit, ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names, ref 2: https://github.com/akka/akka-management/issues/910#issuecomment-848536275. 

Can you verify that once and if there's such a limit then we need to enforce how lengthy the memberNamePrefix can be otherwise the lease creation gets errors internally when it is run.

api/core/v1alpha1/helper.go
```diff
@@ -67,6 +67,16 @@ func GetMemberNameFromAddress(etcdObjMeta metav1.ObjectMeta, memberAddress strin
 	return fmt.Sprintf("%s-%s", etcdObjMeta.Name, memberAddress)
 }
 
+// GetMemberName returns the etcd member name for a given pod name and optional prefix.
+// If memberNamePrefix is non-nil and non-empty, the member name is "<prefix>-<podName>".
+// Otherwise, the member name is the name of the backing `Pod`.
+func GetMemberName(memberNamePrefix *string, podName string) string {
+	if memberNamePrefix != nil && *memberNamePrefix != "" {
```
anveshreddy18:
```suggestion
	if prefix := ptr.Deref(memberNamePrefix, ""); prefix!="" {
```
We use ptr pkg extensively in the codebase and it's already imported in this pkg too, so I'd recommend this to be consistent.

anveshreddy18 (changes requested):
Thanks for the PR @acumino , I have few comments PTAL

internal/health/etcdmember/check_ready.go
```diff
@@ -36,6 +36,11 @@ func (r *readyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) []Resul
 	)
 
 	leaseNames := druidv1alpha1.GetMemberLeaseNames(&etcd)
+	podNames := druidv1alpha1.GetAllPodNames(etcd.ObjectMeta, etcd.Spec.Replicas)
+	leaseToPod := make(map[string]string, len(leaseNames))
+	for i, ln := range leaseNames {
+		leaseToPod[ln] = podNames[i]
+	}
```
anveshreddy18: Imo this looks a bit risky as the mapping relies on the ordering of the contents in both leaseNames & podNames. Any change to that ordering in the calling funcs might corrupt the info in leaseToPod.

Podname is anyway a suffix of the leasename right? can't the mapping be derived from that instead of relying on brittle ordering?

ishan16696:
@acumino please rebase the PR on master. Thanks

acumino:
I have rebased the PR and addressed the reviews. PTAL
@anveshreddy18 @ishan16696

api/core/v1alpha1/etcd.go
anveshreddy18: Thanks for adding the maxLength constraint but the immutability of this field seems to be still not enforced I meant this marker below which is still missing 🤔.
```
// +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field",rule="self == oldSelf"
```

api/core/v1alpha1/etcd.go
anveshreddy18: Closing this as the MaxLength is now enforced to `63` char.

api/core/v1alpha1/helper.go
anveshreddy18: Thanks for fixing this.

internal/health/etcdmember/check_ready.go
```diff
@@ -111,6 +112,13 @@ func (r *readyCheck) Check(ctx context.Context, etcd druidv1alpha1.Etcd) []Resul
 
 const memberLeaseHolderIdentitySeparator = ":"
 
+func podNameFromLeaseName(leaseName string, memberNamePrefix *string) string {
+	if memberNamePrefix != nil && *memberNamePrefix != "" {
```
anveshreddy18:
```suggestion
	if prefix := ptr.Deref(memberNamePrefix, ""); prefix!="" {
```
This is another place where we can resort to using ptr

internal/component/configmap/configmap_test.go
anveshreddy18, line 1: While there are individual tests for memberNamePrefix and additionalAdvertisePeerURLs, I think it'd be meaningful to have a testcase where both of them are tested together so that we catch any changes to these two blocks together. So that if someone in the future modifies the code handling additionalAdvertisePeerURLs while also using memberNamePrefix, we can test this combination.

anveshreddy18:
/hold
as Ishan would want to review it too.

anveshreddy18 (approved):
Thanks for making the requested changes @acumino. LGTM

api/core/v1alpha1/etcd.go
```diff
@@ -350,7 +349,15 @@ type SchedulingConstraints struct {
 // EtcdSpec defines the desired state of Etcd
 // +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
 type EtcdSpec struct {
+	// MemberNamePrefix defines the prefix for the name of each etcd member. When set, the member name is "<prefix>-<pod-name>", otherwise it defaults to the pod name.
+	// The combined length of the prefix, pod name, and separator must not exceed 253 characters (DNS subdomain limit for lease names).
```
ishan16696:
```suggestion
	// MemberNamePrefix defines the prefix for the name of each etcd cluster member. When set, the member name would be "<prefix>-<pod-name>", otherwise it defaults to the "pod-name".
	// The combined length of the member-prefix, pod-name, and separator must not exceed 253 characters (DNS subdomain limit for lease names).
```

api/core/v1alpha1/helper.go
```diff
@@ -44,6 +44,16 @@ func GetOrdinalPodName(etcdObjMeta metav1.ObjectMeta, ordinal int) string {
 	return fmt.Sprintf("%s-%d", etcdObjMeta.Name, ordinal)
 }
 
+// GetMemberName returns the etcd member name for a given pod name and optional prefix.
+// If memberNamePrefix is non-nil and non-empty, the member name is "<prefix>-<podName>".
+// Otherwise, the member name is the name of the backing `Pod`.
```
ishan16696:
```suggestion
// GetMemberName returns the etcd cluster member name for a given member-prefix(optional) and pod-name.
// If memberNamePrefix is non-nil and non-empty, the member name would be "<prefix>-<podName>".
// Otherwise, the member name would be the name of the backing `Pod`.
```

api/core/v1alpha1/etcd.go
```diff
@@ -350,7 +349,15 @@ type SchedulingConstraints struct {
 // EtcdSpec defines the desired state of Etcd
 // +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
```
ishan16696, line 352: I found that etcd member can be changed , so would it make sense to make `spec.memberNamePrefix` mutable ?
WAL dump of change the member name from `infra0` -> `infra4`
```
Start dumping log entries from snapshot.
WAL metadata:
nodeID=bf9071f4639c75cc clusterID=6e9fdbc6edbe620 term=2 commitIndex=9 vote=91bc3c398fb3c146
WAL entries:
lastIndex=10
term	     index	type	data
   1	         1	conf	method=ConfChangeAddNode id=292104365751cc93
   1	         2	conf	method=ConfChangeAddNode id=91bc3c398fb3c146
   1	         3	conf	method=ConfChangeAddNode id=bf9071f4639c75cc
   2	         4	norm
   2	         5	norm	method=PUT path="/0/members/91bc3c398fb3c146/attributes" val="{\"name\":\"infra1\",\"clientURLs\":[\"http://127.0.0.1:22379\"]}"
   2	         6	norm	method=PUT path="/0/members/292104365751cc93/attributes" val="{\"name\":\"infra2\",\"clientURLs\":[\"http://127.0.0.1:23379\"]}"
   2	         7	norm	method=PUT path="/0/members/bf9071f4639c75cc/attributes" val="{\"name\":\"infra0\",\"clientURLs\":[\"http://127.0.0.1:2379\"]}"
   2	         8	norm	method=PUT path="/0/version" val="3.5.0"
   2	         9	norm	header:<ID:14741300156799464452 > put:<key:"sap" value:"labs" >
   2	        10	norm	method=PUT path="/0/members/bf9071f4639c75cc/attributes" val="{\"name\":\"infra4\",\"clientURLs\":[\"http://127.0.0.1:2379\"]}"

Entry types (Normal,ConfigChange) count is : 10
```

acumino: I would prefer to keep it immutable for now, since there was no such thing as a member name change before also, later at some point we can change it if there is demand for such feature.

ishan16696: ok, no issues ... I just wanted to let you know that it's possible. I would like to keep this comment as `unresolved` for future reference to capture that etcd member name can be changed. Hope that's fine.

acumino: Sure 👍

api/core/v1alpha1/helper.go
```diff
@@ -44,6 +44,16 @@ func GetOrdinalPodName(etcdObjMeta metav1.ObjectMeta, ordinal int) string {
 	return fmt.Sprintf("%s-%d", etcdObjMeta.Name, ordinal)
 }
 
+// GetMemberName returns the etcd cluster member name for a given member-prefix(optional) and pod-name.
+// If memberNamePrefix is non-nil and non-empty, the member name would be "<prefix>-<podName>".
+// Otherwise, the member name would be the name of the backing `Pod`.
+func GetMemberName(memberNamePrefix *string, podName string) string {
+	if prefix := ptr.Deref(memberNamePrefix, ""); prefix != "" {
+		return prefix + "-" + podName
```
ishan16696:
```suggestion
		return fmt.Sprintf("%s-%s", prefix, podName)
```

api/core/v1alpha1/helper_test.go
```diff
@@ -87,10 +87,43 @@ func TestGetFullSnapshotLeaseName(t *testing.T) {
...
 
 func TestGetMemberLeaseNames(t *testing.T) {
-	g := NewWithT(t)
-	etcdObjMeta := createEtcdObjectMetadata(uuid.NewUUID(), nil, nil, false)
-	leaseNames := GetMemberLeaseNames(etcdObjMeta, 3)
-	g.Expect(leaseNames).To(Equal([]string{"etcd-test-0", "etcd-test-1", "etcd-test-2"}))
+	tests := []struct {
+		name             string
+		memberNamePrefix *string
+		expectedLeases   func(etcdName string) []string
```
ishan16696:
```suggestion
		expectedMemberLeases   func(etcdName string) []string
```

internal/component/configmap/configmap_test.go
```diff
@@ -240,6 +272,170 @@ func TestPrepareInitialCluster(t *testing.T) {
...
+					URLs:       []string{"http://10.0.0.1:2380"},
+				},
+			},
+			expectedURLs: map[string][]string{
+				"etcd-test-0": {"http://etcd-test-0.etcd-test-peer.test-ns.svc:2380", "http://10.0.0.1:2380"},
+				"etcd-test-1": {"http://etcd-test-1.etcd-test-peer.test-ns.svc:2380"},
+			},
+		},
+		{
+			name:             "should NOT append additional peer URLs for client type",
```
ishan16696:
```suggestion
			name:             "should not append additional peer URLs for client type",
```

ishan16696 (changes requested):
overall looks good to me, just few final nits:

ishan16696 (approved):

PR #1315, opened 2026-03-26
Add support for additional advertise peer URLs

### LINKED ISSUE #1238: Support for additional advertise peer URLs

/area control-plane-migration 
/area control-plane 
/kind enhancement 
/kind api-change

Add support for specifying additional peer URLs that etcd members should advertise beyond the default internal Kubernetes service DNS URLs. 

**Why is this needed**:

This enables cross-cluster etcd peer communication via external endpoints (LoadBalancer IPs or DNS names), which is essential for live cluster migration.

**How it works:**

The additional URLs (LoadBalancer IPs or DNS names) are appended to the default internal service URLs in the etcd configuration:

`initial-advertise-peer-urls=https://etcd-main-0.etcd-main-peer.default.svc:2380,https://52.1.2.3:2380`

seshachalam-yv:
<!-- Please ensure that you do not include company internal information. -->

<!--
Please select area, kind, and priority for this pull request. This helps the community categorizing it.
Replace below TODOs or exchange the existing identifiers with those that fit best in your opinion.
If multiple identifiers make sense you can also state the commands multiple times, e.g.
  /area control-plane
  /area auto-scaling
  ...

"/area" identifiers:     audit-logging|auto-scaling|backup|compliance|control-plane-migration|control-plane|cost|delivery|dev-productivity|disaster-recovery|documentation|high-availability|logging|metering|monitoring|networking|open-source|ops-productivity|os|performance|quality|robustness|scalability|security|storage|testing|usability|user-management
"/kind" identifiers:     api-change|bug|cleanup|discussion|enhancement|epic|flake|impediment|poc|post-mortem|question|regression|task|technical-debt|test

For Druid Enhancement Proposals (DEPs), please ensure that the proposal is written in the following [format](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/00-template.md) before submitting this pull request.
-->
 /area control-plane                                             
 /kind api-change                                                                                                                                                                                                                                                             
 /kind enhancement 

This PR adds support for configuring additional `--initial-advertise-peer-urls` for etcd members, enabling etcd peers to advertise multiple URLs for peer-to-peer communication.

 **API Design:**                                                                                                                                                                                                                                                              
  ```yaml                                                                                                                                                                                                                                                                      
  spec:                                                                                                                                                                                                                                                                        
    etcd:                                                                                                                                                                                                                                                                      
      replicas: 3                                                                                                                                                                                                                                                              
      additionalAdvertisePeerUrls:                                                                                                                                                                                                                                             
      - memberName: etcd-main-0                                                                                                                                                                                                                                                
        urls:                                                                                                                                                                                                                                                                  
        - https://10.0.1.10:2380                                                                                                                                                                                                                                               
        - https://etcd-0.external.example.com:2380                                                                                                                                                                                                                             
```                                                                                                                                                                                                                                                                               
  This generates initial-cluster configuration like:                                                                                                                                                                                                                           
  etcd-main-0=https://etcd-main-0.etcd-main-peer.default.svc:2380,etcd-main-0=https://10.0.1.10:2380,etcd-main-0=https://etcd-0.external.example.com:2380                                                                                                                      
                                                   
 The limit of 10 members is due to CEL validation budget constraints in Kubernetes CRDs.                                                                                                                                                                                    

    The reason:                                                                                                                                                                                                                                                                
    - CEL (Common Expression Language) validations have a cost budget limit that prevents expensive validation expressions from consuming too many API server resources                                                                                                        
    - The complex cross-field validations for additionalAdvertisePeerUrls (checking member names against replicas, validating URLs, TLS config consistency, etc.) consume CEL cost budget                                                                                      
    - With MaxItems=10, the validation stays within the budget                                                                                                                                                                                                                 
    - Beyond 10 items, the CEL validation would exceed the allowed cost and cause CRD generation to fail  
    

Fixes #1238

**Checklist**:
        - [x] Update [Table of Contents](https://github.com/gardener/etcd-druid/blob/master/docs/README.md) for the docs.
    - [x] Unit tests
    - [x] Integration tests
    - [ ] E2E tests

  - CEL Validations implemented:*                                                                                                                                                                                                                                           
    - Member name must match an existing replica (index < replicas count)                                                                                                                                                                                                        
    - Member names must follow etcd naming convention: {name}-{index} where index is 0 to (spec.replicas-1) (e.g., etcd-main-0, etcd-main-1, etcd-main-2 for a 3-replica cluster)
    - URLs validated for proper format (scheme, host, optional port)                                                                                                                                                                                                             
    - TLS configuration consistency checks                                                                                                                                                                                                                                       
    - Mutual exclusivity between inline TLS config and secret references   
 
  - E2E tests: Not included because:
    - Unit tests thoroughly verify ConfigMap generation with correct initial-cluster format                                                                                                                                                                                    
    - Integration tests validate CEL rules at Kubernetes admission time  
    - No LoadBalancer support: Kind doesn't support LoadBalancer service types - they remain in "Pending" state. 

  - CEL validation with `memberNamePrefix`
     - Once PR #1309 (`memberNamePrefix`) lands, we can add a CEL expression that validates member names use the configured prefix. 
      - This was not added in the current PR since #1309 is not yet merged. Whichever PR merges first will update the other to ensure consistency.

**Release note**:
<!--
Write your release note:
1. Enter your release note in the below block.
2. If no release note is required, just write "NONE" within the block.

Format of block header: <category> <target_group>
Possible values:
- category:       breaking|noteworthy|feature|bugfix|doc|other
- target_group:   user|operator|developer|dependency
-->
```feature operator
Add support for `additionalAdvertisePeerUrls` field in Etcd CRD to configure multiple advertise peer URLs for etcd members. Supports up to 10 members with optional per-member TLS configuration.
```


api/core/v1alpha1/etcd.go
```diff
@@ -55,6 +55,8 @@ const (
 // +kubebuilder:printcolumn:name="Cluster Size",type=integer,JSONPath=`.spec.replicas`,priority=1
 // +kubebuilder:printcolumn:name="Current Replicas",type=integer,JSONPath=`.status.currentReplicas`,priority=1
 // +kubebuilder:printcolumn:name="Ready Replicas",type=integer,JSONPath=`.status.readyReplicas`,priority=1
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.etcd.additionalAdvertisePeerUrls) || self.spec.etcd.additionalAdvertisePeerUrls.all(m, m.memberName.startsWith(self.metadata.name + '-'))",message="additionalAdvertisePeerUrls member names must start with the Etcd resource name followed by a dash"
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.etcd.additionalAdvertisePeerUrls) || self.spec.etcd.additionalAdvertisePeerUrls.all(m, m.memberName.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$') && int(m.memberName.substring(m.memberName.lastIndexOf('-')+1)) < self.spec.replicas)",message="additionalAdvertisePeerUrls member name index must be less than replicas"
```
Shreyas-s14: The part of the CEL check here after the `!has(self.additionalAdvertisePeerUrls)` is already covered [here in the field level check](https://github.com/seshachalam-yv/etcd-druid/blob/5b1417ceac838d123b31e5466c959e3c5f40c7bb/api/core/v1alpha1/etcd.go#L311). Do we still need it here?
The check in question here is: 
``` go
self.spec.etcd.additionalAdvertisePeerUrls.all(m, m.memberName.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$')
```

api/core/v1alpha1/etcd.go
```diff
@@ -55,6 +55,8 @@ const (
 // +kubebuilder:printcolumn:name="Cluster Size",type=integer,JSONPath=`.spec.replicas`,priority=1
 // +kubebuilder:printcolumn:name="Current Replicas",type=integer,JSONPath=`.status.currentReplicas`,priority=1
 // +kubebuilder:printcolumn:name="Ready Replicas",type=integer,JSONPath=`.status.readyReplicas`,priority=1
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.etcd.additionalAdvertisePeerUrls) || self.spec.etcd.additionalAdvertisePeerUrls.all(m, m.memberName.startsWith(self.metadata.name + '-'))",message="additionalAdvertisePeerUrls member names must start with the Etcd resource name followed by a dash"
```
Shreyas-s14: This enforcing of etcd name being `etcd-main-x` where x < spec.replicas is restricted to this functionality being specific to gardener cases right? Because for open source setups we don't necessarily enforce this.

api/core/v1alpha1/etcd.go
```diff
@@ -215,6 +217,8 @@ type SnapshotCompactionSpec struct {
 }
 
 // EtcdConfig defines the configuration for the etcd cluster to be deployed.
+// +kubebuilder:validation:XValidation:rule="!has(self.additionalAdvertisePeerUrls) || !has(self.peerUrlTls) || self.additionalAdvertisePeerUrls.all(m, m.urls.all(u, u.startsWith('https://') && !u.startsWith('http://')))",message="when peerUrlTls is enabled, all additional advertise peer URLs must use https://"
+// +kubebuilder:validation:XValidation:rule="!has(self.additionalAdvertisePeerUrls) || has(self.peerUrlTls) || self.additionalAdvertisePeerUrls.all(m, m.urls.all(u, u.startsWith('http://') && !u.startsWith('https://')))",message="when peerUrlTls is not enabled, all additional advertise peer URLs must use http://"
```
Shreyas-s14: In both the rules mentioned, after the `has` check, the second part is already enforced in the first check right?
since validating for `startsWith('http://') already covers `!u.startsWith('https://') and vice versa.

```suggestion
// +kubebuilder:validation:XValidation:rule="!has(self.additionalAdvertisePeerUrls) || !has(self.peerUrlTls) || self.additionalAdvertisePeerUrls.all(m, m.urls.all(u, u.startsWith('https://')))",message="when peerUrlTls is enabled, all additional advertise peer URLs must use https://"
// +kubebuilder:validation:XValidation:rule="!has(self.additionalAdvertisePeerUrls) || has(self.peerUrlTls) || self.additionalAdvertisePeerUrls.all(m, m.urls.all(u, u.startsWith('http://')))",message="when peerUrlTls is not enabled, all additional advertise peer URLs must use http://"
```

Shreyas-s14: Maybe also merge them? 
```go
  !has(self.additionalAdvertisePeerUrls) || self.additionalAdvertisePeerUrls.all(m, m.urls.all(u, has(self.peerUrlTls) ? u.startsWith('https://') : u.startsWith('http://')))
```
But with this the error message is generic. So if you feel its unnecessary, you can resolve this.

api/core/v1alpha1/etcd.go
```diff
@@ -251,6 +255,16 @@ type EtcdConfig struct {
 	// ClientUrlTLS contains the ca, server TLS and client TLS secrets for client communication to ETCD cluster
 	// +optional
 	ClientUrlTLS *TLSConfig `json:"clientUrlTls,omitempty"`
+	// AdditionalAdvertisePeerURLs contains extra per-member peer URLs to append
+	// to initial-advertise-peer-urls. Each entry specifies a member name and its additional URLs.
+	// The member name should match the etcd member name pattern: {etcd-cr-name}-{index}
+	// where index is 0 to (replicas-1). For example, for an Etcd resource named "etcd-main"
+	// with 3 replicas, valid member names are: etcd-main-0, etcd-main-1, etcd-main-2.
+	// Non-matching entries are silently ignored at runtime.
```
Shreyas-s14: It is not silently ignored right? With the CEL check, this is rejected on admission if I am not wrong

api/core/v1alpha1/etcd.go
```diff
@@ -285,6 +299,30 @@ type ClientService struct {
 	TrafficDistribution *string `json:"trafficDistribution,omitempty"`
 }
 
+// AdditionalPeerURL specifies additional peer URLs for a specific etcd member.
+type AdditionalPeerURL struct {
+	// MemberName is the etcd member name.
+	// Should match the etcd member name of the cluster (e.g., etcd-main-0) to take effect.
+	// Non-matching names are silently ignored.
```
Shreyas-s14: https://github.com/gardener/etcd-druid/pull/1315#discussion_r3014157967
Correct me if I misunderstood this.

api/core/v1alpha1/etcd.go
```diff
@@ -251,6 +255,16 @@ type EtcdConfig struct {
...
 	ClientUrlTLS *TLSConfig `json:"clientUrlTls,omitempty"`
+	// AdditionalAdvertisePeerURLs contains extra per-member peer URLs to append
+	// to initial-advertise-peer-urls. Each entry specifies a member name and its additional URLs.
+	// The member name should match the etcd member name pattern: {etcd-cr-name}-{index}
+	// where index is 0 to (replicas-1). For example, for an Etcd resource named "etcd-main"
+	// with 3 replicas, valid member names are: etcd-main-0, etcd-main-1, etcd-main-2.
+	// Non-matching entries are silently ignored at runtime.
+	// Note: Max 10 replicas supported for this feature.
+	// +optional
+	// +kubebuilder:validation:MaxItems=10
```
Shreyas-s14, line 267:
```suggestion
    // AdditionalAdvertisePeerURLs contains extra per-member peer URLs to append
    // to initial-advertise-peer-urls. Each entry maps a member name to its
    // additional URLs. The member name must follow the pattern {etcd-name}-{index}
    // where index is 0 to (replicas-1) (e.g., etcd-main-0, etcd-main-1 etc).
    // +optional
    // +kubebuilder:validation:MaxItems=10
```

api/core/v1alpha1/etcd.go
```diff
@@ -285,6 +299,30 @@ type ClientService struct {
...
 }
 
+// AdditionalPeerURL specifies additional peer URLs for a specific etcd member.
+type AdditionalPeerURL struct {
+	// MemberName is the etcd member name.
+	// Should match the etcd member name of the cluster (e.g., etcd-main-0) to take effect.
+	// Non-matching names are silently ignored.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
Shreyas-s14: Do you think this check is needed since the regex already enforces atleast 1 character to be present.

api/core/v1alpha1/etcd.go
```diff
@@ -285,6 +299,30 @@ type ClientService struct {
...
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$`
+	MemberName string `json:"memberName"`
+
+	// URLs is a list of additional peer URLs for this member.
+	// These will be appended to the default internal service URL.
+	// A maximum of 5 URLs can be specified per member.
+	// Must be valid HTTP(S) URLs with scheme, host, and port (e.g., https://10.0.0.1:2380).
```
Shreyas-s14: Maybe we can utilise [the kubernetes CEL Url check](https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-url-library) ?

internal/component/configmap/etcdconfig.go
```diff
@@ -121,10 +121,27 @@ func prepareInitialCluster(etcd *druidv1alpha1.Etcd, peerScheme string) string {
...
+		for _, memberURLs := range etcd.Spec.Etcd.AdditionalAdvertisePeerURLs {
+			if memberURLs.MemberName == podName {
+				peerURLs = append(peerURLs, memberURLs.URLs...)
+				break
+			}
+		}
+
+		// Output member=url for EACH URL (not member=url1,url2)
+		for _, url := range peerURLs {
+			builder.WriteString(fmt.Sprintf("%s=%s,", podName, url))
```
Shreyas-s14:
```suggestion
			fmt.Fprintf(&builder, "%s=%s,", podName, url)
```
My IDE suggested this due to [check](https://staticcheck.dev/docs/checks/#QF1012)

internal/component/configmap/configmap_test.go
```diff
@@ -422,3 +478,114 @@ func matchPeerTLSRelatedConfiguration(g *WithT, etcd *druidv1alpha1.Etcd, actual
...
+					URLs:       []string{"http://10.0.0.1:2380"},
+				},
+			},
+			expectedURLs: map[string][]string{
+				"etcd-test-0": {"http://etcd-test-0.etcd-test-peer.test-ns.svc:2380", "http://10.0.0.1:2380"},
+				"etcd-test-1": {"http://etcd-test-1.etcd-test-peer.test-ns.svc:2380"},
+			},
+		},
+		{
+			name:             "should NOT append additional peer URLs for client type",
```
Shreyas-s14, line 555: THis test case is to check that the clientURLs list is not updated right? The function implemented does not do that at all by having a check [here](https://github.com/seshachalam-yv/etcd-druid/blob/f5aa808929f1ca0ba27befa53c8873ef82dbc3ec/internal/component/configmap/etcdconfig.go#L167-L168)
So there is no scope of this ever occuring right?

test/it/crdvalidation/etcd/specetcd_additionaladvertisepeerurls_test.go
```diff
@@ -0,0 +1,303 @@
...
+
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
+	"github.com/gardener/etcd-druid/test/utils"
+)
+
+// TestValidateSpecEtcdAdditionalAdvertisePeerUrlsMemberName validates the member name CEL validations:
+// - Member name must start with {metadata.name}- prefix
+// - Member name must match DNS label pattern with numeric suffix
+// - Member name index must be less than replicas count
+func TestValidateSpecEtcdAdditionalAdvertisePeerUrlsMemberName(t *testing.T) {
```
Shreyas-s14: Should these test cases be moved to `test/it/crdvalidation/specetcd_test.go` since we have all the tests for the field `spec.etcd` there and this component belongs under that?

Shreyas-s14 (changes requested):
Thanks for the PR @seshachalam-yv ! Have a few comments PTAL

api/core/v1alpha1/etcd.go
seshachalam-yv: This behavior is not specific to Gardener. The `{etcd-cr-name}-{index}` naming convention is inherent to how etcd-druid derives member names for all managed clusters

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-druid/pull/1315/changes/76b5d34f0db4a774c61a8577c33663c2e11a26b2#diff-3f222bc5d6b1a341dae2b7cb1b468e55cb3645800102d76badc22883b658f0ebR59

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as suggested in this PR:
https://github.com/gardener/etcd-druid/pull/1315/changes/76b5d34f0db4a774c61a8577c33663c2e11a26b2#diff-3f222bc5d6b1a341dae2b7cb1b468e55cb3645800102d76badc22883b658f0ebR220-R221

I prefer keeping these as two separate rules, since each produces a clear and actionable error message:

- When peerUrlTls is enabled, all additional advertised peer URLs must use https://
- When peerUrlTls is not enabled, all additional advertised peer URLs must use http://

Merging them into a ternary condition would require a single generic message, which may make it harder for users to understand what needs to be fixed

api/core/v1alpha1/etcd.go
seshachalam-yv: Investigated this — the k8s CEL URL library (url(u)) doesn't expose a getPort() method, so we can't validate that a port is present.

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-druid/pull/1315/changes/382465155428fb268d594fe0f7175dcdb7cc0fc1#diff-3f222bc5d6b1a341dae2b7cb1b468e55cb3645800102d76badc22883b658f0ebR259-R261

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-druid/pull/1315/changes/382465155428fb268d594fe0f7175dcdb7cc0fc1#diff-3f222bc5d6b1a341dae2b7cb1b468e55cb3645800102d76badc22883b658f0ebR259-R261

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed

internal/component/configmap/etcdconfig.go
seshachalam-yv: Fixed

test/it/crdvalidation/etcd/specetcd_additionaladvertisepeerurls_test.go
seshachalam-yv: Moved as part of https://github.com/gardener/etcd-druid/pull/1315/changes/382465155428fb268d594fe0f7175dcdb7cc0fc1#diff-762d587f6a6bf047c8bb7aa0de6d35a2d0c0eefba818f8c0aabd30498ddcce07R113-R402

internal/component/configmap/configmap_test.go
seshachalam-yv: This test is an intentional safeguard. The same getAdvertiseURLs function is used to compute URLs for both peer and client types, and the guard ensures additional URLs are only appended for peer type. The test captures this intent explicitly — if a future refactor modifies getAdvertiseURLs and accidentally breaks that invariant, the test will catch it. Keeping it as a contract test.

api/core/v1alpha1/etcd.go
seshachalam-yv: Removed

internal/component/configmap/configmap_test.go
Shreyas-s14: got it thanks for the explanantion.

Shreyas-s14 (approved):
Thanks for the changes and explanations.
LGTM

Shreyas-s14 (approved):

api/core/v1alpha1/etcd.go
```diff
@@ -251,6 +255,13 @@ type EtcdConfig struct {
 	// ClientUrlTLS contains the ca, server TLS and client TLS secrets for client communication to ETCD cluster
 	// +optional
 	ClientUrlTLS *TLSConfig `json:"clientUrlTls,omitempty"`
+	// AdditionalAdvertisePeerURLs contains extra per-member peer URLs to append
+	// to initial-advertise-peer-urls. Each entry maps a member name to its
+	// additional URLs. The member name must follow the pattern {etcd-name}-{index}
+	// where index is 0 to (replicas-1) (e.g., etcd-main-0, etcd-main-1 etc).
+	// +optional
+	// +kubebuilder:validation:MaxItems=10
+	AdditionalAdvertisePeerURLs []AdditionalPeerURL `json:"additionalAdvertisePeerUrls,omitempty"`
```
anveshreddy18:
```suggestion
	AdditionalAdvertisePeerURLs []AdditionalPeerURL `json:"additionalAdvertisePeerURLs,omitempty"`
```
There's some inconsistencies between the way the field & json tag are named, can you correct this? I prefer URL in full capital, but you may choose Url, anything is fine as long as they're consistent.

api/core/v1alpha1/etcd.go
```diff
@@ -55,6 +55,8 @@ const (
 // +kubebuilder:printcolumn:name="Cluster Size",type=integer,JSONPath=`.spec.replicas`,priority=1
 // +kubebuilder:printcolumn:name="Current Replicas",type=integer,JSONPath=`.status.currentReplicas`,priority=1
 // +kubebuilder:printcolumn:name="Ready Replicas",type=integer,JSONPath=`.status.readyReplicas`,priority=1
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.etcd.additionalAdvertisePeerUrls) || self.spec.etcd.additionalAdvertisePeerUrls.all(m, m.memberName.startsWith(self.metadata.name + '-'))",message="additionalAdvertisePeerUrls member names must start with the Etcd resource name followed by a dash"
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.etcd.additionalAdvertisePeerUrls) || self.spec.etcd.additionalAdvertisePeerUrls.all(m, int(m.memberName.substring(m.memberName.lastIndexOf('-')+1)) < self.spec.replicas)",message="additionalAdvertisePeerUrls member name index must be less than replicas"
```
anveshreddy18:
```suggestion
// +kubebuilder:validation:XValidation:rule="!has(self.spec.etcd.additionalAdvertisePeerURLs) || self.spec.etcd.additionalAdvertisePeerURLs.all(m, m.memberName.startsWith(self.metadata.name + '-'))",message="additionalAdvertisePeerURLs member names must start with the Etcd resource name followed by a dash"
// +kubebuilder:validation:XValidation:rule="!has(self.spec.etcd.additionalAdvertisePeerURLs) || self.spec.etcd.additionalAdvertisePeerURLs.all(m, int(m.memberName.substring(m.memberName.lastIndexOf('-')+1)) < self.spec.replicas)",message="additionalAdvertisePeerURLs member name index must be less than replicas"
```
Same as above, to be consistent with the naming.

api/core/v1alpha1/etcd.go
```diff
@@ -215,6 +217,8 @@ type SnapshotCompactionSpec struct {
 }
 
 // EtcdConfig defines the configuration for the etcd cluster to be deployed.
+// +kubebuilder:validation:XValidation:rule="!has(self.additionalAdvertisePeerUrls) || !has(self.peerUrlTls) || self.additionalAdvertisePeerUrls.all(m, m.urls.all(u, u.startsWith('https://')))",message="when peerUrlTls is enabled, all additional advertise peer URLs must use https://"
+// +kubebuilder:validation:XValidation:rule="!has(self.additionalAdvertisePeerUrls) || has(self.peerUrlTls) || self.additionalAdvertisePeerUrls.all(m, m.urls.all(u, u.startsWith('http://')))",message="when peerUrlTls is not enabled, all additional advertise peer URLs must use http://"
```
anveshreddy18:
```suggestion
// +kubebuilder:validation:XValidation:rule="!has(self.additionalAdvertisePeerURLs) || !has(self.peerUrlTls) || self.additionalAdvertisePeerURLs.all(m, m.urls.all(u, u.startsWith('https://')))",message="when peerUrlTls is enabled, all additional advertise peer URLs must use https://"
// +kubebuilder:validation:XValidation:rule="!has(self.additionalAdvertisePeerURLs) || has(self.peerUrlTls) || self.additionalAdvertisePeerURLs.all(m, m.urls.all(u, u.startsWith('http://')))",message="when peerUrlTls is not enabled, all additional advertise peer URLs must use http://"
```
Same as above.

api/core/v1alpha1/etcd.go
```diff
@@ -285,6 +296,28 @@ type ClientService struct {
...
+	// Must match the etcd member name of the cluster (e.g., etcd-main-0).
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$`
+	MemberName string `json:"memberName"`
+
+	// URLs is a list of additional peer URLs for this member.
+	// These will be appended to the default internal service URL.
+	// A maximum of 5 URLs can be specified per member.
```
anveshreddy18: can you point me to how the `5` urls max decision is arrived at? is this a conscious decision or a magic number?

api/core/v1alpha1/etcd.go
anveshreddy18: can you clarify in the comments that the port is optional? as API docs needs to be accurate?

api/core/v1alpha1/etcd.go
```diff
@@ -285,6 +296,28 @@ type ClientService struct {
...
+	// Must be valid HTTP(S) URLs with scheme, host, and port (e.g., https://10.0.0.1:2380).
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=5
+	// +listType=atomic
+	URLs []string `json:"urls"`
```
anveshreddy18, line 323: I feel it's more idiomatic to accept [URL type from the url pkg of std library](https://pkg.go.dev/net/url#URL) instead of accepting them as strings, that way the validation will be more accurate and we can just rely on their validation than redoing it ourselves.
```suggestion
	URLs []url.URL `json:"urls"`
```

api/core/v1alpha1/etcd.go
anveshreddy18: Also I think the naming needs to be relooked at, because the top level `AdditionalAdvertisePeerURLs` looks like it'll contain the peer urls directly, I recommend `AdditionalAdvertiseMemberPeerURLs` and then the `AdditionalPeerURL` to be changed to `AdditionalPeerURLs`, because the plural form looks like only one url is going to be there, but we allow max 5 urls, right? Also at other place in the codebase there was this line:
```go
for _, memberURLs := range etcd.Spec.Etcd.AdditionalAdvertisePeerURLs {
```
here the `memberURLs` took a plural form, so I think we need to relook into the naming and lock it properly according to what they represent and whether it is singular/plural.

api/core/v1alpha1/etcd.go
```diff
@@ -285,6 +296,28 @@ type ClientService struct {
...
+
+	// URLs is a list of additional peer URLs for this member.
+	// These will be appended to the default internal service URL.
+	// A maximum of 5 URLs can be specified per member.
+	// Must be valid HTTP(S) URLs with scheme, host, and port (e.g., https://10.0.0.1:2380).
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=5
+	// +listType=atomic
```
anveshreddy18, line 322: is it okay to update these fields in live when the cluster is already running with different set of addresses? or does it introduce any problems if they're updated mid-way?

anveshreddy18 (changes requested):
Thank you for the PR @seshachalam-yv, mostly looks good, have few comments PTAL.

The PR description mentions the below:
`Member names must follow etcd naming convention: {name}-{index} where index is 0-9` - I checked and there's no validation anywhere for index to be within `0-9` right? but more importantly does the index needs to be between `0-9` ? I understand that you limit the number of entries to 10 items but you can also use this set of {etcd-main-2, etcd-main-6, etcd-main-11} for example right? it still falls under the 10 items limit while still allowing the user to expose peer addresses for the etcd-main-11 member. So I do not see the link between how the items count limit & `index` play together.

api/core/v1alpha1/etcd.go
seshachalam-yv: Safe to update live `initial-cluster` is only used at bootstrap. At runtime, peer discovery uses etcd's membership API. etcd-druid reconciles the ConfigMap → triggers rolling restart. Each member re-reads config on restart.

api/core/v1alpha1/etcd.go
seshachalam-yv: `url.URL` is a Go struct that doesn't serialize to a JSON scalar — K8s CRDs can't use it. All core K8s API types use string for URLs (e.g. webhookClientConfig.URL). Regex + CEL validation is the idiomatic K8s approach.

api/core/v1alpha1/etcd.go
seshachalam-yv: CEL budget constraint: `5 URLs × 10 members = 50 URL` validations per admission. Beyond this, cross-field CEL cost exceeds budget. Practically 2-3 per member is sufficient; 5 gives headroom.

api/core/v1alpha1/etcd.go
seshachalam-yv: addressed as part of https://github.com/gardener/etcd-druid/pull/1315/commits/c7a6a04db0349fec30912cb9141debc38c5a6404

api/core/v1alpha1/etcd.go
anveshreddy18: Ah got it, thanks for clarifying.

seshachalam-yv:
> "Member names must follow etcd naming convention: {name}-{index} where index is 0-9?"

The 0-9 in the description refers to the naming convention regex pattern (^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$) — it means the suffix must be a numeric digit, not that the index is limited to 0–9. The actual index validation is index < replicas, so any valid ordinal is accepted as long as the member exists.

Updated the PR description also.

anveshreddy18:
/hold
@ishan16696 will be taking a look at the PR.

anveshreddy18 (approved):
Thanks for addressing my comments @seshachalam-yv . LGTM!

ishan16696:
> spec:                                                                                                                                                                                                                                                                        
  etcd:                                                                                                                                                                                                                                                                      
    replicas: 3                                                                                                                                                                                                                                                              
    additionalAdvertisePeerUrls:                                                                                                                                                                                                                                             
    - memberName: etcd-main-0                                                                                                                                                                                                                                                
      urls:                                                                                                                                                                                                                                                                  
      - https://10.0.1.10:2380                                                                                                                                                                                                                                               
      - https://etcd-0.external.example.com:2380                                                                                                                                                                                                                             
     
I have question to API design, currently it has `.spec.additionalAdvertisePeerUrls` which is taking the list of `memberName`, so my question is why it's taking `memberName` via etcd spec ? I'm not in favour of defining the the `memberName` inside API itself because of several reasons:
1. Currently there is other component taking memberName of etcd cluster member directly. 
2. No where we define memberName in api, it has taken from `pod-name` for now, but as you are already aware it will prepended by `memberPrefix` which will be defined in api but not whole member name, memberName will become `<MemberPrefix-podName>`. 
3. If memberName would be like this: `<MemberPrefix-podName>`, how's your CEL expression will work ?
4. Assume, we define `memberName` in this API but backup-restore calculates memberName like this `<MemberPrefix-podName>` then their could be potential mismatch issue.
5. Currently config-map template mentioned like this 
```
    initial-advertise-peer-urls:
      etcd-main-0:
      - https://etcd-main-0.....svc:2380
      etcd-main-1:
      - https://etcd-main-1....svc:2380
     ...
``` 
but druid calculates the `memberName` by itself from `podName`, and in future we can simply prepended this by `memberPrefix`, but here also we aren't defining the whole memberName.
https://github.com/gardener/etcd-druid/blob/3080127a53cfa1549d540b2ad96512c4af023d17/internal/component/configmap/etcdconfig.go#L152-L156

docs/api-reference/etcd-druid-api.md
```diff
@@ -958,6 +959,23 @@ _Appears in:_
...
+MemberPeerURLs specifies additional peer URLs for a specific etcd member.
+
+
+
+_Appears in:_
+- [EtcdConfig](#etcdconfig)
+
+| Field | Description | Default | Validation |
+| --- | --- | --- | --- |
+| `memberName` _string_ | MemberName is the etcd member name.<br />Must match the etcd member name of the cluster (e.g., etcd-main-0). |  | MaxLength: 253 <br />Pattern: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$` <br />Required: \{\} <br /> |
```
ishan16696, line 975: you should only define the memberPrefix, not the memberName as I mentioned in this comment: https://github.com/gardener/etcd-druid/pull/1315#issuecomment-4439447023

seshachalam-yv:
@ishan16696 Thanks for the review questions. Here are my thoughts:

**Why `memberName` is needed in `additionalAdvertisePeerURLs`:**

Without knowing the member name, we cannot associate the additional advertise peer URLs to specific members. The ConfigMap generation needs to know which URLs belong to which member — this is the same pattern used by `externallyManagedMemberAddresses` where each entry is keyed by member identity.

**CEL validation with `memberNamePrefix`:**

Once PR #1309 (`memberNamePrefix`) lands, we can add a CEL expression that validates member names use the configured prefix. For example:

```cel
self.spec.etcd.additionalAdvertisePeerURLs.all(m,
  has(self.spec.memberNamePrefix)
    ? m.memberName.startsWith(self.spec.memberNamePrefix + '-' + self.metadata.name + '-')
    : m.memberName.startsWith(self.metadata.name + '-'))
```

This was not added in the current PR since #1309 is not yet merged. Whichever PR merges first will update the other to ensure consistency.

**ConfigMap generation:**

The ConfigMap uses the same member-keyed pattern as `externallyManagedMemberAddresses` — each member's initial-cluster string includes its additional advertise peer URLs alongside the primary peer URL.

api/core/v1alpha1/etcd.go
```diff
@@ -251,6 +255,15 @@ type EtcdConfig struct {
 	// ClientUrlTLS contains the ca, server TLS and client TLS secrets for client communication to ETCD cluster
 	// +optional
 	ClientUrlTLS *TLSConfig `json:"clientUrlTls,omitempty"`
+	// AdditionalAdvertisePeerURLs contains extra per-member peer URLs to append
+	// to initial-advertise-peer-urls. Each entry maps a member name to its
+	// additional URLs. The member name must follow the pattern {etcd-name}-{index}
+	// where index is 0 to (replicas-1) (e.g., etcd-main-0, etcd-main-1 etc).
```
ishan16696, line 262: can you mention a TODO here, so that once member-prefix PR goes in, this need to be updated.

api/core/v1alpha1/etcd.go
```diff
@@ -285,6 +298,28 @@ type ClientService struct {
...
 }
 
+// MemberPeerURLs specifies additional peer URLs for a specific etcd member.
+type MemberPeerURLs struct {
+	// MemberName is the etcd member name.
+	// Must match the etcd member name of the cluster (e.g., etcd-main-0).
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$`
```
ishan16696, line 311: same as above as CEL expression need to be updated if name format of etcd cluster member get change.

internal/component/configmap/etcdconfig.go
```diff
@@ -125,7 +125,22 @@ func prepareInitialCluster(etcd *druidv1alpha1.Etcd, peerScheme string) string {
...
+
+			// Find and append additional peer URLs if configured for this member
+			for _, memberURLs := range etcd.Spec.Etcd.AdditionalAdvertisePeerURLs {
+				if memberURLs.MemberName == podName {
+					peerURLs = append(peerURLs, memberURLs.URLs...)
+					break
+				}
+			}
+
+			// Output member=url for EACH URL (not member=url1,url2)
```
ishan16696:
```suggestion
			// Output memberName=url for EACH URL (not memberName=url1,url2)
```

internal/component/configmap/etcdconfig.go
```diff
@@ -125,7 +125,22 @@ func prepareInitialCluster(etcd *druidv1alpha1.Etcd, peerScheme string) string {
...
 		for i := range int(etcd.Spec.Replicas) {
 			podName := druidv1alpha1.GetOrdinalPodName(etcd.ObjectMeta, i)
-			builder.WriteString(fmt.Sprintf("%s=%s://%s.%s:%s,", podName, peerScheme, podName, domainName, serverPort))
+
+			// Start with internal service DNS URL
+			peerURLs := []string{fmt.Sprintf("%s://%s.%s:%s", peerScheme, podName, domainName, serverPort)}
+
+			// Find and append additional peer URLs if configured for this member
+			for _, memberURLs := range etcd.Spec.Etcd.AdditionalAdvertisePeerURLs {
+				if memberURLs.MemberName == podName {
```
ishan16696, line 138: can you add a TODO here that this need to be updated with memberName, in another PR of memberPrefix it can be removed.

internal/component/configmap/etcdconfig.go
```diff
@@ -152,7 +167,21 @@ func getAdvertiseURLs(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme, peerSv
...
 			podName := druidv1alpha1.GetOrdinalPodName(etcd.ObjectMeta, i)
-			advUrlsMap[podName] = []string{fmt.Sprintf("%s://%s.%s:%d", scheme, podName, domainName, port)}
+
+			// Start with internal service DNS URL
+			urls := []string{fmt.Sprintf("%s://%s.%s:%d", scheme, podName, domainName, port)}
+
+			// Append additional peer URLs only for peer advertise URLs
+			if advertiseURLType == advertiseURLTypePeer {
+				for _, memberURLs := range etcd.Spec.Etcd.AdditionalAdvertisePeerURLs {
+					if memberURLs.MemberName == podName {
```
ishan16696, line 185: same as above.

ishan16696 (commented):
few nits

internal/component/configmap/etcdconfig.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-druid/pull/1315/commits/db4e6c94e3b5d52e49d93dbc70a499aa269dc57d

internal/component/configmap/etcdconfig.go
seshachalam-yv: Fixed part of https://github.com/gardener/etcd-druid/pull/1315/commits/db4e6c94e3b5d52e49d93dbc70a499aa269dc57d

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-druid/pull/1315/commits/db4e6c94e3b5d52e49d93dbc70a499aa269dc57d

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-druid/pull/1315/commits/db4e6c94e3b5d52e49d93dbc70a499aa269dc57d

seshachalam-yv:
> few nits

Thanks @ishan16696 


**CEL validation with `memberNamePrefix`:**

- Once PR #1309 (`memberNamePrefix`) lands, we can add a CEL expression that validates member names use the configured prefix.
- This was not added in the current PR since #1309 is not yet merged. Whichever PR merges first will update the other to ensure consistency.

Also added this to the **Special notes for your reviewer** section in the PR description.

ishan16696 (approved):
Thanks @seshachalam-yv  for addressing my comments.
LGTM!!

PR #1129, opened 2025-07-03
[DEP-05]: Introduce `EtcdOpsTask` CRD

seshachalam-yv:
**How to categorize this PR?**

/area ops-productivity
/area backup
/kind api-change

**What this PR does / why we need it**:

This PR introduces the `EtcdOpsTask` CRD as proposed in [DEP-05](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/05-etcdopstask.md). It enables declarative management of out-of-band etcd operational tasks, starting with on-demand snapshot creation. This marks a foundational step towards supporting additional task types such as quorum loss recovery and maintenance, as discussed in #1047.

Changes include:
- Added `EtcdOpsTask` API types and validation logic in `api/core/v1alpha1/etcdopstask.go`:
- Defined `Spec`, `Status`, `TaskState`, `OperationState`, `OperationPhase`, and a union-style `Config` field for extensibility.
- Introduced supporting types: `EtcdReference`, `EtcdOpsTaskLastError`, and `EtcdOpsLastOperation`.
- Added Kubebuilder markers for validation, custom print columns, short name (`eot`), and resource categories (`druid`).
- Modified `hack/prepare-chart-resources.sh` to include the new CRD in chart preparation.
- Updated the DEP-05 to use `EtcdOpsTask` instead of `EtcdOperatorTask`

**Which issue(s) this PR fixes**:
Fixes part of #1047

**Special notes for your reviewer**:
- The CRD is designed to be extensible. The union-style `Config` field allows the addition of new task types in the future.
- This PR only includes the CRD definition. Controller logic, tests, and support for additional task types will be implemented in follow-up PRs.
- A short name (`eot`) has been added for ease of use with `kubectl`.

Example usage:

```bash
✗ kubectl get eot -o wide                                                                  
NAME                             STATE     ETCD           AGE   OPERATION STATE   TTL
example-ondemand-full-snapshot   Pending   example-etcd   45s   InProgress        3600
```

```bash
✗ kubectl get etcdopstasks                       
NAME                             STATE     ETCD           AGE
example-ondemand-full-snapshot   Pending   example-etcd   68s
```

**Release note**:
```feature operator
Introduced the `EtcdOpsTask` CRD to enable declarative management of out-of-band `etcd` operational tasks, starting with on-demand snapshot configuration.
```

PR #1060, opened 2025-04-09
[WIP] Refactor e2e tests

shreyas-s-rao:
**How to categorize this PR?**

/area testing quality
/kind enhancement test

**What this PR does / why we need it**:
Refactor e2e tests.

**Which issue(s) this PR fixes**:
Fixes #782 

**Special notes for your reviewer**:

**Release note**:

```noteworthy developer
Refactor e2e tests.
```

PR #1357, opened 2026-05-19
Implement e2e test for EtcdOpsTask

### LINKED ISSUE #1200: Add Documentation for EtcdOpstask.

/area documentation
/kind enhancement

EtcdOpsTask was introduced as a part of #1129 . Adding the below documentation:
- [x] Controller design details to [controllers.md](https://github.com/gardener/etcd-druid/blob/master/docs/development/controllers.md).
- [x] A new document in [usage doc](https://github.com/gardener/etcd-druid/tree/master/docs/usage) to tell users how to use EtcdOpsTask resource itself.
- [x] A guide for developers to add new EtcdOpsTask configs and handlers.

**Why is this needed**:
To make it easier for adopters of etcdopstask to leverage the crd and the logic provided to consume and implement out-of-band tasks for druid managed etcds via EtcdOpsTask.

### LINKED ISSUE #1201: Add e2e tests for EtcdOpsTask.

/area testing
/kind enhancement

Add end-end tests for EtcdOpsTask introduced in #1129 as per the new e2e test format introduced in #1060 once both are merged.

**Why is this needed**:
E2E tests for etcdopstask to test various scenarios. Currently the controller and on-demand snapshot task have been unit tested and integration tests are also present.

Shreyas-s14:

/area testing
/kind enhancement

Implements e2e tests for `EtcdOpsTask`, specifically the `OnDemand Snapshot` task introduced in #1129 with the new test framework introduced in #1060 
The tests introduced in this PR are specifically for on-demand snapshot task types and more tests will have to be introduced as and when new `EtcdOpsTasks` are implemented.

Fixes #1201 #1200 

**Checklist**:
- [ ] Update documentation in the `/docs` folder (if applicable)
    - [ ] Unit tests
    - [ ] Integration tests
    - [ x ] E2E tests

**Release note**:

```other operator
implement e2e tests for `EtcdOpsTask`
```

### DIFF
--- a/test/e2e/controller/etcdopstask/etcdopstask_test.go
+++ b/test/e2e/controller/etcdopstask/etcdopstask_test.go
@@ -74,7 +74,8 @@ func TestAdmit(t *testing.T) {
 			continue
 		}
 		t.Run("duplicate-rejected-same-etcd-"+e2eutils.GetProviderSuffix(provider), testAdmitDuplicateRejected(provider))
-		t.Run("independent-different-etcd-"+e2eutils.GetProviderSuffix(provider), testAdmitIndependentEtcds(provider))
+		t.Run("independent-different-namespace-"+e2eutils.GetProviderSuffix(provider), testAdmitIndependentEtcds(provider, false))
+		t.Run("independent-same-namespace-"+e2eutils.GetProviderSuffix(provider), testAdmitIndependentEtcds(provider, true))
 	}
 }
 
@@ -108,12 +109,19 @@ func testAdmitDuplicateRejected(provider druidv1alpha1.StorageProvider) func(*te
 		logger.Info("successfully created Etcd resource")
 
 		logger.Info("loading data into etcd to ensure snapshot takes measurable time")
-		testEnv.DeployEtcdLoaderJob(g, testNamespace, druidv1alpha1.GetClientServiceName(etcd.ObjectMeta), *etcd.Spec.Etcd.ClientPort, etcd.Spec.Etcd.ClientUrlTLS, 1000, timeoutDeployJob)
+		testEnv.DeployEtcdLoaderJob(g, testNamespace, druidv1alpha1.GetClientServiceName(etcd.ObjectMeta), *etcd.Spec.Etcd.ClientPort, etcd.Spec.Etcd.ClientUrlTLS, 2000, 256*1024, timeoutLargeDeployJob)
 
 		firstTask := newOnDemandSnapshotTask(e2eutils.DefaultEtcdName+"-first-task", testNamespace, e2eutils.DefaultEtcdName, druidv1alpha1.OnDemandSnapshotTypeFull)
 
 		logger.Info("creating first EtcdOpsTask")
 		testEnv.CreateEtcdOpsTask(g, firstTask)
+
+		fetchedFirstTask, err := testEnv.GetEtcdOpsTask(firstTask.Name, firstTask.Namespace)
+		g.Expect(err).NotTo(HaveOccurred())
+		if fetchedFirstTask.Status.State == nil || *fetchedFirstTask.Status.State != druidv1alpha1.TaskStateInProgress {
+			logger.Info("waiting for first task to reach InProgress before creating the duplicate")
+			testEnv.CheckEtcdOpsTaskState(g, firstTask, druidv1alpha1.TaskStateInProgress, timeoutEtcdOpsTaskCompletion)
+		}
 
 		secondTask := newOnDemandSnapshotTask(e2eutils.DefaultEtcdName+"-second-task", testNamespace, e2eutils.DefaultEtcdName, druidv1alpha1.OnDemandSnapshotTypeFull)
 
@@ -128,61 +136,66 @@ func testAdmitDuplicateRejected(provider druidv1alpha1.StorageProvider) func(*te
 		testEnv.CheckEtcdOpsTaskDeleted(g, secondTask, defaultTaskDeletionTimeout)
 		logger.Info("second task deleted after TTL")
 
+		logger.Info("waiting for first task to be deleted after TTL")
+		testEnv.CheckEtcdOpsTaskDeleted(g, firstTask, defaultTaskDeletionTimeout)
+		logger.Info("first task deleted after TTL")
+
 		testSucceeded = true
 	}
 }
 
-func testAdmitIndependentEtcds(provider druidv1alpha1.StorageProvider) func(*testing.T) {
+func testAdmitIndependentEtcds(provider druidv1alpha1.StorageProvider, sameNamespace bool) func(*testing.T) {
 	return func(t *testing.T) {
 		t.Parallel()
 		g := NewWithT(t)
 		log := testr.NewWithOptions(t, testr.Options{LogTimestamp: true})
 		var testSucceeded bool
 
-		tcName := fmt.Sprintf("admit-independent-different-etcd-%s", e2eutils.GetProviderSuffix(provider))
-		ns1 := testutils.GenerateTestNamespaceNameWithTestCaseName(t, testNamespacePrefix, tcName+"-a", 4)
-		ns2 := testutils.GenerateTestNamespaceNameWithTestCaseName(t, testNamespacePrefix, tcName+"-b", 4)
-		logger := log.WithName(tcName)
-		defer func() {
-			e2eutils.CleanupTestArtifacts(retainTestArtifacts, testSucceeded, testEnv, logger, g, ns1)
-			e2eutils.CleanupTestArtifacts(retainTestArtifacts, testSucceeded, testEnv, logger, g, ns2)
-		}()
+		layout := "different-namespace"
+		if sameNamespace {
+			layout = "same-namespace"
+		}
+		tcName := fmt.Sprintf("admit-independent-%s-%s", layout, e2eutils.GetProviderSuffix(provider))
 
 		etcdNameA := "etcd-a"
 		etcdNameB := "etcd-b"
 
-		e2eutils.InitializeTestCase(g, testEnv, logger, ns1, etcdNameA, provider)
-		e2eutils.InitializeTestCase(g, testEnv, logger, ns2, etcdNameB, provider)
+		var nsA, nsB string
+		if sameNamespace {
+			nsA = testutils.GenerateTestNamespaceNameWithTestCaseName(t, testNamespacePrefix, tcName, 4)
+			nsB = nsA
+		} else {
+			nsA = testutils.GenerateTestNamespaceNameWithTestCaseName(t, testNamespacePrefix, tcName+"-a", 4)
+			nsB = testutils.GenerateTestNamespaceNameWithTestCaseName(t, testNamespacePrefix, tcName+"-b", 4)
+		}
+		logger := log.WithName(tcName)
+		defer func() {
+			e2eutils.CleanupTestArtifacts(retainTestArtifacts, testSucceeded, testEnv, logger, g, nsA)
+			if !sameNamespace {
+				e2eutils.CleanupTestArtifacts(retainTestArtifacts, testSucceeded, testEnv, logger, g, nsB)
+			}
+		}()
 
-		logger.Info("creating first Etcd resource", "namespace", ns1, "etcdName", etcdNameA)
-		etcdA := testutils.EtcdBuilderWithoutDefaults(etcdNameA, ns1).
-			WithReplicas(1).
-			WithEtcdClientPort(ptr.To[int32](2379)).
-			WithClientTLS().
-			WithPeerTLS().
-			WithDefaultBackup().
-			WithStorageProvider(provider, fmt.Sprintf("%s/%s", ns1, etcdNameA)).
-			WithBackupRestoreTLS().
-			Build()
+		if sameNamespace {
+			e2eutils.CreateNamespace(g, testEnv, logger, nsA)
+			e2eutils.CreateBackupSecret(g, testEnv, logger, nsA, provider)
+		} else {
+			e2eutils.InitializeTestCase(g, testEnv, logger, nsA, etcdNameA, provider)
+			e2eutils.InitializeTestCase(g, testEnv, logger, nsB, etcdNameB, provider)
+		}
+
+		logger.Info("creating first Etcd resource", "namespace", nsA, "etcdName", etcdNameA)
+		etcdA := buildIndependentEtcd(etcdNameA, nsA, provider, sameNamespace)
 		testEnv.CreateAndCheckEtcd(g, etcdA, timeoutEtcdCreation)
 
-		logger.Info("creating second Etcd resource", "namespace", ns2, "etcdName", etcdNameB)
-		etcdB := testutils.EtcdBuilderWithoutDefaults(etcdNameB, ns2).
-			WithReplicas(1).
-			WithEtcdClientPort(ptr.To[int32](2379)).
-			WithClientTLS().
-			WithPeerTLS().
-			WithDefaultBackup().
-			WithStorageProvider(provider, fmt.Sprintf("%s/%s", ns2, etcdNameB)).
-			WithBackupRestoreTLS().
-			Build()
+		logger.Info("creating second Etcd resource", "namespace", nsB, "etcdName", etcdNameB)
+		etcdB := buildIndependentEtcd(etcdNameB, nsB, provider, sameNamespace)
 		testEnv.CreateAndCheckEtcd(g, etcdB, timeoutEtcdCreation)
 
-		taskA := newOnDemandSnapshotTask(etcdNameA+"-task", ns1, etcdNameA, druidv1alpha1.OnDemandSnapshotTypeFull)
+		taskA := newOnDemandSnapshotTask(etcdNameA+"-task", nsA, etcdNameA, druidv1alpha1.OnDemandSnapshotTypeFull)
+		taskB := newOnDemandSnapshotTask(etcdNameB+"-task", nsB, etcdNameB, druidv1alpha1.OnDemandSnapshotTypeFull)
 
-		taskB := newOnDemandSnapshotTask(etcdNameB+"-task", ns2, etcdNameB, druidv1alpha1.OnDemandSnapshotTypeFull)
-
-		logger.Info("creating both EtcdOpsTask resources concurrently")
+		logger.Info("creating both EtcdOpsTask resources to be admitted and executed concurrently")
 		testEnv.CreateEtcdOpsTask(g, taskA)
 		testEnv.CreateEtcdOpsTask(g, taskB)
 
@@ -198,4 +211,16 @@ func testAdmitIndependentEtcds(provider druidv1alpha1.StorageProvider) func(*tes
 
 		testSucceeded = true
 	}
+}
+
+func buildIndependentEtcd(name, namespace string, provider druidv1alpha1.StorageProvider, sameNamespace bool) *druidv1alpha1.Etcd {
+	builder := testutils.EtcdBuilderWithoutDefaults(name, namespace).
+		WithReplicas(1).
+		WithEtcdClientPort(ptr.To[int32](2379)).
+		WithDefaultBackup().
+		WithStorageProvider(provider, fmt.Sprintf("%s/%s", namespace, name))
+	if !sameNamespace {
+		builder = builder.WithClientTLS().WithPeerTLS().WithBackupRestoreTLS()
+	}
+	return builder.Build()
 }
--- a/test/e2e/controller/etcdopstask/etcdopstask_test.go
+++ b/test/e2e/controller/etcdopstask/etcdopstask_test.go
@@ -1,0 +1,201 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package etcdopstask
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
+	"github.com/gardener/etcd-druid/test/e2e/testenv"
+	e2eutils "github.com/gardener/etcd-druid/test/e2e/utils"
+	testutils "github.com/gardener/etcd-druid/test/utils"
+
+	"github.com/go-logr/logr/testr"
+	"k8s.io/utils/ptr"
+
+	. "github.com/onsi/gomega"
+)
+
+// TestMain sets up the test environment.
+func TestMain(m *testing.M) {
+	kubeconfigPath, err := testutils.GetEnvOrError(e2eutils.EnvKubeconfigPath)
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "KUBECONFIG not provided: %v\n", err)
+		os.Exit(1)
+	}
+	cl, err := e2eutils.GetKubernetesClient(kubeconfigPath)
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Failed to create Kubernetes client: %v\n", err)
+		os.Exit(1)
+	}
+
+	ctx, cancelCtx := context.WithTimeout(context.Background(), timeoutTest)
+
+	testEnv = testenv.NewTestEnvironment(ctx, cancelCtx, cl)
+	if err = testEnv.PrepareScheme(); err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Failed to prepare scheme: %v\n", err)
+		os.Exit(1)
+	}
+
+	val, _ := testutils.GetEnvOrError(e2eutils.EnvRetainTestArtifacts)
+	switch e2eutils.RetainTestArtifactsMode(val) {
+	case e2eutils.RetainTestArtifactsAll:
+		retainTestArtifacts = e2eutils.RetainTestArtifactsAll
+	case e2eutils.RetainTestArtifactsFailed:
+		retainTestArtifacts = e2eutils.RetainTestArtifactsFailed
+	default:
+		retainTestArtifacts = e2eutils.RetainTestArtifactsNone
+	}
+
+	backupProviders := testutils.GetEnvOrDefault(e2eutils.EnvBackupProviders, "none,local")
+	providers, err = e2eutils.ParseBackupProviders(backupProviders)
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Failed to parse backup providers: %v\n", err)
+		os.Exit(1)
+	}
+
+	exitCode := m.Run()
+
+	testEnv.Close()
+	os.Exit(exitCode)
+}
+
+// TestAdmit validates the task-agnostic admit conditions for an EtcdOpsTask.
+func TestAdmit(t *testing.T) {
+	t.Parallel()
+
+	for _, provider := range providers {
+		if provider == "none" {
+			continue
+		}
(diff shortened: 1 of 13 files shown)


test/e2e/controller/etcdopstask/etcdopstask_test.go
shreyas-s-rao, line 1: Any reason to not keep these tests in `test/e2e/controller` directory? Reason I ask is, the testEnv is setup in `TestMain` there, and here you setup a different testEnv, so how does it look to be running the other e2e tests along with etcdopstask e2e tests? Can these be run together, or would they be run one after another? This is not really an issue in the prow jobs since we run each Test suite as a dedicated prow job anyway, but for local development would this cause any hinderance? Also, since you moved all etcdopstask tests into its own directory, would it become necessary to move the other tests into their own separate controller directories? I'm just thinking out loud.

The advantage of having all those test suites in the same dir was that they could all be run together in parallel, especially since controller like secret controller and compaction controller have only one test suite to run, so multiple suites could be run at once. I think in the case of etcdopstask, it comes with its own set of suites for each opstask type, so I guess it's ok to keep these in their own directory, as you've done now.

Anyway, I wanted to know if you chose this approach because of this or any other reason, out of curiosity.

test/e2e/controller/etcdopstask/utils.go
```diff
@@ -0,0 +1,47 @@
...
+	testNamespacePrefix = "eot-e2e"
+
+	timeoutTest                  = 1 * time.Hour
+	timeoutEtcdCreation          = 5 * time.Minute
+	timeoutEtcdOpsTaskCompletion = 3 * time.Minute
+	timeoutEtcdOpsTaskTTLExpiry  = 30 * time.Second
+	timeoutDeployJob             = 2 * time.Minute
+	timeoutFullSnapshot          = 30 * time.Second
+
+	defaultOpsTaskTTL = int32(30)
```
shreyas-s-rao:
```suggestion
	defaultOpsTaskTTLSeconds = int32(30)
```

test/e2e/controller/etcdopstask/etcdopstask_test.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// TestAdmit validates the task-agnostic admit conditions for an EtcdOpsTask.
+func TestAdmit(t *testing.T) {
+	t.Parallel()
+
+	for _, provider := range providers {
+		if provider == "none" {
+			continue
+		}
```
shreyas-s-rao, line 75: Would every opstask be provider-dependent? When on-demand defragmentation is added later which doesn't have a dependency on the backup store provider, or any other such task, this condition will end up skipping those tests when set to provider=none right? Ideally such conditions should be added within the opstask-specific test function.

test/e2e/controller/etcdopstask/etcdopstask_test.go
```diff
@@ -0,0 +1,201 @@
...
+		testEnv.DeployEtcdLoaderJob(g, testNamespace, druidv1alpha1.GetClientServiceName(etcd.ObjectMeta), *etcd.Spec.Etcd.ClientPort, etcd.Spec.Etcd.ClientUrlTLS, 1000, timeoutDeployJob)
+
+		firstTask := newOnDemandSnapshotTask(e2eutils.DefaultEtcdName+"-first-task", testNamespace, e2eutils.DefaultEtcdName, druidv1alpha1.OnDemandSnapshotTypeFull)
+
+		logger.Info("creating first EtcdOpsTask")
+		testEnv.CreateEtcdOpsTask(g, firstTask)
+
+		secondTask := newOnDemandSnapshotTask(e2eutils.DefaultEtcdName+"-second-task", testNamespace, e2eutils.DefaultEtcdName, druidv1alpha1.OnDemandSnapshotTypeFull)
+
+		logger.Info("creating second EtcdOpsTask for the same Etcd (should be rejected)")
```
shreyas-s-rao, line 128: Is it possible that the first task completes before second task gets created, causing the test to fail (since creation of second task will succeed)? Even if the etcd is loaded well, full snapshots are generally quite quick, and even if they take a few seconds, it's possible that the test runner thread may have paused for whatever reason. I see the risk of flakes with this test case.

test/e2e/controller/etcdopstask/etcdopstask_test.go
```diff
@@ -0,0 +1,201 @@
...
+		testEnv.CreateEtcdOpsTask(g, secondTask)
+
+		logger.Info("waiting for second task to be rejected")
+		testEnv.CheckEtcdOpsTaskState(g, secondTask, druidv1alpha1.TaskStateRejected, timeoutEtcdOpsTaskCompletion)
+		logger.Info("second task correctly rejected")
+
+		logger.Info("waiting for second task to be deleted after TTL")
+		testEnv.CheckEtcdOpsTaskDeleted(g, secondTask, defaultTaskDeletionTimeout)
```
shreyas-s-rao, line 136: `CheckEtcdOpsTaskDeleted` needs to be run for first task as well, or else the namespace will try to be deleted while the first task is still running.

test/e2e/controller/etcdopstask/etcdopstask_test.go
```diff
@@ -0,0 +1,201 @@
...
+func testAdmitIndependentEtcds(provider druidv1alpha1.StorageProvider) func(*testing.T) {
+	return func(t *testing.T) {
+		t.Parallel()
+		g := NewWithT(t)
+		log := testr.NewWithOptions(t, testr.Options{LogTimestamp: true})
+		var testSucceeded bool
+
+		tcName := fmt.Sprintf("admit-independent-different-etcd-%s", e2eutils.GetProviderSuffix(provider))
+		ns1 := testutils.GenerateTestNamespaceNameWithTestCaseName(t, testNamespacePrefix, tcName+"-a", 4)
+		ns2 := testutils.GenerateTestNamespaceNameWithTestCaseName(t, testNamespacePrefix, tcName+"-b", 4)
```
shreyas-s-rao: Why create them in two different namespaces? We want to ensure in this test that when two opstasks are created for two different etcds (preferably in the same namespace), the admission succeeds right? If they're in two different namespaces, we obviously know that the creation will succeed since the etcdopstasks are namespace-scoped anyway.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
```diff
@@ -0,0 +1,268 @@
...
+	for _, provider := range providers {
+		for _, tc := range testCases {
+			if !tc.backupEnabled && provider != "none" {
+				continue
+			}
+			if tc.backupEnabled && provider == "none" {
+				continue
+			}
+
+			tcName := fmt.Sprintf("lifecycle-%s-%s", tc.name, e2eutils.GetProviderSuffix(provider))
```
shreyas-s-rao: Needs to have something related to `ondemandsnapshot` in the namespace name, or else it becomes unidentifiable after other opstask tests are added later.

test/e2e/controller/etcdopstask/utils.go
```diff
@@ -0,0 +1,47 @@
...
+	"time"
+
+	druidv1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
+	"github.com/gardener/etcd-druid/test/e2e/testenv"
+	e2eutils "github.com/gardener/etcd-druid/test/e2e/utils"
+	testutils "github.com/gardener/etcd-druid/test/utils"
+)
+
+const (
+	testNamespacePrefix = "eot-e2e"
```
shreyas-s-rao, line 17: Please adapt `make clean-e2e-test-resources` to clean up namespaces with this prefix too.

test/e2e/controller/etcdopstask/etcdopstask_test.go
shreyas-s-rao, line 1: Please adapt `make clean-e2e-test-resources` to clean up PKI resources under this directory too.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	for _, provider := range providers {
+		for _, tc := range testCases {
+			if !tc.backupEnabled && provider != "none" {
+				continue
+			}
+			if tc.backupEnabled && provider == "none" {
+				continue
+			}
```
shreyas-s-rao: I didn't fully understand this. Are you trying to skip cases that are incorrect, like a mismatch between the `PROVIDERS` and the actual Etcd resource? I didn't get this.

The Etcd should be created with whatever provider is defined by `PROVIDERS` env var, so why try to explicitly change this behavior per test case? If the tests are run against an Etcd without any provider (none), then all of these test cases should reject the task, or simply follow what compaction e2e tests does [here](https://github.com/gardener/etcd-druid/blob/d8f18429b50dce538660468c5468595893527e6d/test/e2e/controller/compaction_test.go#L80-L82), because these tests are simply not applicable to Etcds without backup store enabled anyway:
```go
if provider == "none" {
		continue
	}
```

This also means the `backupEnabled` field of the test case should be removed.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
```diff
@@ -0,0 +1,268 @@
...
+			name:          "rejected-etcd-not-ready",
+			purpose:       "test Rejected state of OnDemandSnapshot when Etcd is not in Ready state",
+			snapshotType:  druidv1alpha1.OnDemandSnapshotTypeFull,
+			backupEnabled: true,
+			replicas:      0,
+			expectedState: druidv1alpha1.TaskStateRejected,
+		},
+		{
+			name:          "full-snapshot-fails-etcd-disrupted",
+			purpose:       "test Failed state of full OnDemandSnapshot when etcd is disrupted by scaling down StatefulSet to 0 mid-execution",
```
shreyas-s-rao, line 58: Feels very timing-based, since the disruption needs to happen exactly when the task is in progress, and can lead to flakes.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
```diff
@@ -0,0 +1,268 @@
...
+			purpose:        "test Succeeded state of delta OnDemandSnapshot when the delta snapshot is skipped by backup-restore server",
+			snapshotType:   druidv1alpha1.OnDemandSnapshotTypeDelta,
+			backupEnabled:  true,
+			replicas:       1,
+			needsFullFirst: true,
+			expectedState:  druidv1alpha1.TaskStateSucceeded,
+		},
+		{
+			name:            "mid-flight-deletion-cleans-up",
+			purpose:         "test cleanup when deletion is triggered mid-execution",
```
shreyas-s-rao, line 98: Same as above

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
```diff
@@ -0,0 +1,268 @@
...
+					logger.Info("full snapshot taken successfully")
+				}
+
+				if tc.loadKeys > 0 {
+					logger.Info("loading data into etcd", "keys", tc.loadKeys)
+					testEnv.DeployEtcdLoaderJob(g, testNamespace, druidv1alpha1.GetClientServiceName(etcd.ObjectMeta), *etcd.Spec.Etcd.ClientPort, etcd.Spec.Etcd.ClientUrlTLS, tc.loadKeys, timeoutDeployJob)
+				}
+
+				var fullRevBefore, deltaRevBefore int64
+				if tc.checkLease {
```
shreyas-s-rao, line 179: Do we really need the `checkLease` field, since checking the leases is hardly an expensive operation, and can be run for all test cases, right? It'll help reduce one configurable field in the test case and make it leaner.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
```diff
@@ -0,0 +1,268 @@
...
+				testEnv.CheckEtcdOpsTaskDeleted(g, task, defaultTaskDeletionTimeout)
+				logger.Info("task deleted after TTL expiry")
+
+				testSucceeded = true
+			})
+		}
+	}
+}
+
+func scaleStatefulSetToZero(g *WithT, testEnv *testenv.TestEnvironment, etcd *druidv1alpha1.Etcd) {
```
shreyas-s-rao: This is a general utility that should ideally be in one of the utils files (either utils.go in this dir, or test/utils since this is generic enough to be used by other non-e2e tests too)

test/e2e/utils/setup.go
```diff
@@ -0,0 +1,124 @@
...
+	g.Expect(testEnv.CreateTestNamespace(testNamespace)).To(Succeed())
+	logger.Info("successfully created test namespace")
+}
+
+// GeneratePKIResources generates PKI resources for etcd and returns the directories containing the generated certificates.
+func GeneratePKIResources(g *WithT, logger logr.Logger, testNamespace, etcdName string) (string, string, string) {
+	logger.Info("generating PKI resources")
+	certDir := fmt.Sprintf("%s/%s", PKIResourcesDir, testNamespace)
+	etcdCertsDir := fmt.Sprintf("%s/etcd", certDir)
+	g.Expect(os.MkdirAll(etcdCertsDir, 0755)).To(Succeed()) // #nosec: G301
```
shreyas-s-rao: Missed reasoning string for the nosec (missed in the move from `test/e2e/controller/utils.go`)

test/e2e/utils/setup.go
```diff
@@ -0,0 +1,124 @@
...
+	g.Expect(testutils.GeneratePKIResourcesToDirectory(logger, etcdCertsDir, etcdName, testNamespace)).To(Succeed())
+	etcdPeerCertsDir := fmt.Sprintf("%s/etcd-peer", certDir)
+	g.Expect(os.MkdirAll(etcdPeerCertsDir, 0755)).To(Succeed()) // #nosec: G301
```
shreyas-s-rao: Same as above

test/e2e/utils/setup.go
```diff
@@ -0,0 +1,124 @@
...
+	g.Expect(testutils.GeneratePKIResourcesToDirectory(logger, etcdPeerCertsDir, etcdName, testNamespace)).To(Succeed())
+	etcdbrCertsDir := fmt.Sprintf("%s/etcd-backup-restore", certDir)
+	g.Expect(os.MkdirAll(etcdbrCertsDir, 0755)).To(Succeed()) // #nosec: G301
```
shreyas-s-rao: Same as above

test/e2e/controller/etcdopstask/etcdopstask_test.go
```diff
@@ -0,0 +1,201 @@
...
+			WithStorageProvider(provider, fmt.Sprintf("%s/%s", ns2, etcdNameB)).
+			WithBackupRestoreTLS().
+			Build()
+		testEnv.CreateAndCheckEtcd(g, etcdB, timeoutEtcdCreation)
+
+		taskA := newOnDemandSnapshotTask(etcdNameA+"-task", ns1, etcdNameA, druidv1alpha1.OnDemandSnapshotTypeFull)
+
+		taskB := newOnDemandSnapshotTask(etcdNameB+"-task", ns2, etcdNameB, druidv1alpha1.OnDemandSnapshotTypeFull)
+
+		logger.Info("creating both EtcdOpsTask resources concurrently")
```
shreyas-s-rao: The creation is not really concurrent, is it? the creation still happens sequentially, but the admission, execution and garbage collection of these tasks would happen concurrently. So the log needs to be clear about this.

shreyas-s-rao (changes requested):
@Shreyas-s14 thanks for the PR! I tested the changes locally (specifically the etcdopstask ones) and it worked well in general most of the time, but then I saw a race condition, which led to test failure for `TestAdmit/duplicate-rejected-same-etcd-local`: `failed to get EtcdOpsTask test-second-task: etcdopstasks.druid.gardener.cloud "test-second-task" not found`:
```
Every 1.0s: kubectl get etcd,po,etcdopstask -A | grep -E "eot-e2e|NAMESPACE"
                                                                                                                                        
NAMESPACE                                               NAME                               READY   QUORATE   ALL MEMBERS READY   BACKUP READY   CLUSTER ID MISMATCH   AGE
eot-e2e-admit-duplicate-rejected-same-etcd-local-74bf   etcd.druid.gardener.cloud/test     true    True      True                Unknown        False                 118s
eot-e2e-admit-independent-different-etcd-local-a-32ff   etcd.druid.gardener.cloud/etcd-a   true    True      True                Unknown        False                 115s
eot-e2e-admit-independent-different-etcd-local-b-e12d   etcd.druid.gardener.cloud/etcd-b   true    Unknown   Unknown             Unknown        Unknown               28s

NAMESPACE                                               NAME                                                       READY   STATUS      RESTARTS   AGE
eot-e2e-admit-duplicate-rejected-same-etcd-local-74bf   pod/etcd-loader-7d82a9ae-hvxlv                             0/1     Completed   0          31s
eot-e2e-admit-duplicate-rejected-same-etcd-local-74bf   pod/test-0                                                 2/2     Running     0          117s
eot-e2e-admit-independent-different-etcd-local-a-32ff   pod/etcd-a-0                                               2/2     Running     0          115s
eot-e2e-admit-independent-different-etcd-local-b-e12d   pod/etcd-b-0                                               2/2     Running     0          28s

NAMESPACE                                               NAME                                                STATE       ETCD   AGE
eot-e2e-admit-duplicate-rejected-same-etcd-local-74bf   etcdopstask.druid.gardener.cloud/test-first-task    Rejected    test   11s
eot-e2e-admit-duplicate-rejected-same-etcd-local-74bf   etcdopstask.druid.gardener.cloud/test-second-task   Succeeded   test   11s
```

You can see that `test-second-task` succeeded, but `test-first-task` was rejected in the `eot-e2e-admit-duplicate-rejected-same-etcd-local-74bf` namespace, which is not expected, since the first task should succeed and the second one should be rejected according to the code. As I mentioned in few of my review comments, I see a clear risk of flakiness in some of these test cases, and we need to prevent that.

Also, please add an issue or todo item to add these to the prow jobs config.

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: I chose to do it this way to avoid cluttering the parent e2e repo with files. As of now, we don't  know how many opstasks shall be implemented as a part of etcd-druid. Assuming each task will have its own set of e2e tests which would be better kept in separate files for ex: quorumlossrecovery_test.go, maintenancedefrag_test.go, maintenancecompact_test.go etc
It felt more structured imo.

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: I thought I'd hit 2 cases with one test using this. The test for different etcds and different namespaces. I can have separate tests for these or just one case with 2 etcds in the same namespace. WDYT?

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
Shreyas-s14: This was one way to have a simulated situation where Full Snapshot fails. Is there a better way to showcase the flow for that scenario?

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
Shreyas-s14: None of the other tests really require this check though right? But I can make it a universal check for these tests.

test/e2e/controller/etcdopstask/etcdopstask_test.go
shreyas-s-rao: Yes, I would prefer two separate tests for this.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
shreyas-s-rao: Would have been great to have mocks, but that's not what e2e tests are for anyway, so that's out. Right now, the only thing I can think of is to increase the DB size substantially so that it's almost guaranteed that the full snapshot will take long enough to run subsequent disruption operations without (too many) flakes.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
shreyas-s-rao: Ok, we can keep the check field, since it's only used by this test case. I'd rather have an additional field gating some functionality that run it for all test cases.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
Shreyas-s14: I see one problem with this though:
Namespace has a 63 character limit doesnt it? Since it must be a valid DNS label. In that case do you think something of the sorts: `eot-ond-%s-%s` be fine?
Because as per current suggestion it would look like `lifecycle-ondemandsnapshot-%s-%s` which can exceed this limit.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
Shreyas-s14: The check as done in compaction works for all the tests here except one: `rejected-bacckup-not-enabled` where I specifically want an etcd with no backup for a test where backup is needed. Hence this check.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
Shreyas-s14: Loading with substantially large number of keys reduces the flakiness but doesnt remove it imo. What I'll try instead is to load a certain(high) number of keys, wait for the task to go to the `In Progress` state and then scale down.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
Shreyas-s14: I dont think this test requires a check as per the previous comment. All this test checks is the successful force deletion of the etcdopstask once it's created.

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: So currently only the on demand snapshot is implemented and hence those have to be used as an `EtcdOpsTask` instance for the tests. But the problem is that, the moment there is no backup (set by provider env var), the OpsTask created gets `Rejected` and it'll not be possible to test those other scenarios. 
Anyways these tests are not dependent on the specific behaviour of opstask but the general Admit conditions of each so I think we can keep it as is.

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: Fixed this by loading the etcd with a high enough value so that the snapshot process takes a considerable amount of time. Then, when the first OpsTask is `in Progress`, create a new one for the same etcd which will showcase the rejection

test/e2e/controller/etcdopstask/etcdopstask_test.go
shreyas-s-rao: > the moment there is no backup (set by provider env var), the OpsTask created gets Rejected

Well, this will hit us for other opstasks like quorum loss recovery or on-demand defrag which have nothing to do with backups. This needs to be fixed in the code then. Can you please create an issue so we don't forget this? The backup-check should be within each ops-task controller's actuator, not in the ops-task framework itself.

For now, keep the tests like this, but add a todo item in the issue to also modify this behavior in the test when the issue is worked upon

test/e2e/controller/etcdopstask/etcdopstask_test.go
shreyas-s-rao: Hmmm ok. I still feel this could lead to flakes, but with a reduced probability. Because there's a lot of timing involved in the test process

test/e2e/controller/etcdopstask/etcdopstask_test.go
shreyas-s-rao: Any update for this?

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
shreyas-s-rao: Add this also to the issue about fixing the opstask behavior of not rejecting non-backup etcds

shreyas-s-rao:
@Shreyas-s14 thanks for addressing my comments. I wanted a few more clarifications on the old comments, PTAL thanks.

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: Yeah this is a "limitation" for the kind of OpsTasks we have now. But one thing, these tests, other than this particular checks test a generic functionality of the opstask while the core logic is tested elsewhere. These checks apply regardless of whether it has backup or not. So is it really going to affect us?

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: Yeah but do you see a way where a "duplicate" task can be created while the other one is still In Progress? I assumed throwing a large amount of keys would stall the first one in Progress state which would then deterministically lead the other one to get Rejected.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
Shreyas-s14: Could you explain how this is an issue? Asking because this is specific to a particular OndemandSnapshot test

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: This check has been covered after ensuring the second task is deleted [here](https://github.com/Shreyas-s14/etcd-druid/blob/94cb0b6a1da39a113b7dba1a152d5c5c9084f9f3/test/e2e/controller/etcdopstask/etcdopstask_test.go#L138-L141)

test/e2e/controller/etcdopstask/etcdopstask_test.go
shreyas-s-rao: Assume you're running the e2e tests with provider set to local or none. If none, then the etcdopstask tests will currently never run right?

My point is, we should always run the generic etcdopstask tests, while specific ones like ondemandsnapshot should run based on whether backup is set or not. So this check above should be inside ondemandsnapshot test, not outside it.

test/e2e/controller/etcdopstask/etcdopstask_test.go
shreyas-s-rao: I guess we don't have much control over this now. Document as a possible flake, and if it comes to in the future, we re-think how we test this case.

test/e2e/controller/etcdopstask/ondemandsnapshot_test.go
shreyas-s-rao: Ah my bad, I had included it with the above comment. Resolving this.

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: True. In that case I'll create an issue documenting this since we currently do not have an opstask that doesn't depend on backup being enabled.
The tests however are agnostic to the type of OpsTask and upon implementation of say quorumloss or copybackup etc, we can change the test parameter
WDYT

test/e2e/controller/etcdopstask/etcdopstask_test.go
Shreyas-s14: https://github.com/gardener/etcd-druid/issues/1399
Documented here.

shreyas-s-rao (approved):
@Shreyas-s14 thanks for addressing my comments.
/lgtm
/approve

Issue #1230, opened 2025-12-15
DEP-05QuorumLossRecoveryTask: Automate recovery from quorum loss.


ashwani2k:
**How to categorize this issue?**

/area ops-productivity
/kind enhancement

**What would you like to be added**:
With the implementation of ETCDOpsTask framework with https://github.com/gardener/etcd-druid/pull/1129. 
This sub-issue focusses on implementing one of the  task mentioned [motivation of Dep-05](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/05-etcdopstask.md#motivation)in the QuorumLossRecoveryTask: Automate recovery from quorum loss.
 - [ ] Understand the usage guide as a base for design/implementation
 - [ ] Implement the task
 - [ ] Write Unit Test
 - [ ] Enahance e2e test (if required)
 - [ ] Update usage docs for this new tasks for Operators
 - [ ] Present the task in community meeting

**Why is this needed**:
To ease the pain for operations to recover from permanent quorum loss which currently is a manual activity documented in [Recovery from Permanent Quorum Loss](https://gardener.cloud/docs/other-components/etcd-druid/recovering-etcd-clusters/#recovery).

Shreyas-s14:
Can the PR be held until https://github.com/gardener/etcd-druid/issues/1201 is done so that we have e2e tests.

PR #1358, opened 2026-05-20
Add RecoverFromQuorumLoss handler for EtcdOpsTask

ashwani2k:
## What does this PR do?

Implements a new `EtcdOpsTask` handler — `RecoverFromQuorumLoss` — that automates recovery of a multi-member etcd cluster from permanent quorum loss, replacing the manual steps documented in [Recovering Etcd Clusters](https://github.com/gardener/etcd-druid/blob/master/docs/usage/recovering-etcd-clusters.md)

Relates to #1230.

### New API type
- `RecoverFromQuorumLossConfig` in `api/core/v1alpha1/recoverfromquorumloss.go` with two optional `metav1.Duration` fields:
  - `scaleDownTimeout` (default 60s) — how long to wait for the StatefulSet to reach 0 replicas
  - `podReadyTimeout` (default 3m) — how long to wait for the single-member pod to become Ready

### Handler (`internal/controller/etcdopstask/handler/recoverfromquorumloss/`)
Executes 11 ordered recovery steps:
1. Suspend etcd-druid reconciliation
2. Disable etcd component protection
3. Scale StatefulSet to 0
4. Wait for StatefulSet to reach 0 replicas
5. Delete all PVCs
6. Delete all member leases
7. Patch ConfigMap to single-member configuration
8. Scale StatefulSet to 1
9. Wait for single-member pod to become Ready
10. Re-enable etcd-druid reconciliation (triggers scale-out)
11. Re-enable etcd component protection

**Per-step requeue:** after each step the handler returns `Requeue: true` so the step description is written to `LastOperation.Description`, giving operators visibility into recovery progress.

**Crash-safe resumption:** a step-tracking annotation (`druid.gardener.cloud/quorum-recovery-last-completed-step`) is written to the Etcd CR after each completed step. On any requeue or restart, `Execute()` reads this annotation and skips already-completed steps.

**Admit rejects** the task when:
- The cluster is single-member (`spec.replicas <= 1`)
- No backup store is configured (`spec.backup.store` missing)
- The cluster is already in a Ready state (no quorum loss detected)

**Cleanup** removes the three recovery annotations (suspend, disable-protection, step-tracking) so the operator is never permanently locked out after a mid-execution failure.

### Other changes
- Registered handler in `DefaultTaskHandlerRegistry()` in the reconciler
- Added `delete` verb to PVC RBAC rule in `charts/templates/clusterrole.yaml`
- Added `WithRecoverFromQuorumLossConfig()` builder to `test/utils/etcdopstask.go`
- Updated `docs/usage/using-etcdopstask.md` with full section (prerequisites, config options, example, expected behaviour)
- Added `examples/etcdopstask/recover-from-quorum-loss.yaml`
- Generated: `zz_generated.deepcopy.go`, CRD manifests, API reference

## Test plan

- [ ] `make test-unit` passes locally — all 17 test functions in the handler package pass
- [ ] `TestRecoverFromQuorumLossAdmit` — 5 cases: not found, single-member, no backup, already ready, pass
- [ ] `TestRecoverFromQuorumLossExecute` — table-driven, exercises error paths and happy path via `runUntilDone()`
- [ ] `TestRecoverFromQuorumLossExecuteStepByStep` — drives one step per `Execute()` call, verifies annotation recorded after each step and stale-status side-effect between scale-down and wait steps
- [ ] `TestRecoverFromQuorumLossExecuteWithPVCsAndLeases` — full loop asserting PVCs and leases are deleted
- [ ] `TestRecoverFromQuorumLossCleanup` — 3 cases including step annotation removal
- [ ] 11 unit tests for individual helper methods (scale, delete, patch, ready checks)


Shreyas-s14:
Can you change the PR description link for the doc?
It leads to an incorrect link. 
https://github.com/gardener/etcd-druid/blob/master/docs/usage/recovering-etcd-clusters.md

ashwani2k:
Thanks @Shreyas-s14  for the catch. Fixed the description of the PR with the correct doc reference for recovery from quorum loss.

api/core/v1alpha1/recoverfromquorumloss.go
```diff
@@ -0,0 +1,26 @@
...
+
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// RecoverFromQuorumLossConfig defines the configuration for the RecoverFromQuorumLoss task.
+// This task automates the recovery of a multi-member etcd cluster from a permanent quorum loss.
+type RecoverFromQuorumLossConfig struct {
+	// ScaleDownTimeout is the timeout to wait for the StatefulSet to fully scale down to 0 replicas
+	// during quorum loss recovery. Defaults to 60 seconds.
```
Shreyas-s14, line 13: Can the default values be set here via kubebuilder markers instead of assigning it in the handler? This is the convention followed in ondemand snapshot config and other etcd api fields as well.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+	ErrAdmitRecoverFromQuorumLoss druidapicommon.ErrorCode = "ERR_ADMIT_RECOVER_FROM_QUORUM_LOSS"
+	// ErrExecuteRecoverFromQuorumLoss is the error code for errors during execution.
+	ErrExecuteRecoverFromQuorumLoss druidapicommon.ErrorCode = "ERR_EXECUTE_RECOVER_FROM_QUORUM_LOSS"
+
+	// defaultScaleDownTimeout is the default timeout for waiting for the StatefulSet to scale down to 0.
+	defaultScaleDownTimeout = 300 * time.Second
+	// defaultPodReadyTimeout is the default timeout for waiting for the single-member etcd pod to become ready.
+	defaultPodReadyTimeout = 600 * time.Second
+	// defaultPollInterval is the interval between successive polls when waiting for a condition.
+	defaultPollInterval = 5 * time.Second
```
Shreyas-s14: This variable is unused as of now. Can this be removed.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+)
+
+const (
+	// ErrAdmitRecoverFromQuorumLoss is the error code for errors during the admit phase.
+	ErrAdmitRecoverFromQuorumLoss druidapicommon.ErrorCode = "ERR_ADMIT_RECOVER_FROM_QUORUM_LOSS"
+	// ErrExecuteRecoverFromQuorumLoss is the error code for errors during execution.
+	ErrExecuteRecoverFromQuorumLoss druidapicommon.ErrorCode = "ERR_EXECUTE_RECOVER_FROM_QUORUM_LOSS"
+
+	// defaultScaleDownTimeout is the default timeout for waiting for the StatefulSet to scale down to 0.
+	defaultScaleDownTimeout = 300 * time.Second
```
Shreyas-s14: In the api, the default value for this was mentioned as 60s, but it's 300s here. I'd recommend to have this default in the api as per the previous comment.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+	// defaultPodReadyTimeout is the default timeout for waiting for the single-member etcd pod to become ready.
+	defaultPodReadyTimeout = 600 * time.Second
```
Shreyas-s14: Same for this value. API default: 180s, value here is 600s

api/core/v1alpha1/recoverfromquorumloss.go
```diff
@@ -0,0 +1,26 @@
...
+	// +optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
+	ScaleDownTimeout *metav1.Duration `json:"scaleDownTimeout,omitempty"`
+
+	// PodReadyTimeout is the timeout to wait for the single-member etcd pod (ordinal 0) to become
+	// ready after the StatefulSet is scaled up to 1 during quorum loss recovery.
+	// Defaults to 180 seconds.
```
Shreyas-s14, line 21: Same as https://github.com/gardener/etcd-druid/pull/1358#discussion_r3279863916

api/core/v1alpha1/recoverfromquorumloss.go
Shreyas-s14: Also, do you think a minmum value that's needed for both the fields?

api/core/v1alpha1/recoverfromquorumloss.go
```diff
@@ -0,0 +1,26 @@
...
+	// +optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
+	PodReadyTimeout *metav1.Duration `json:"podReadyTimeout,omitempty"`
```
Shreyas-s14, line 25: Just curious, will the oeprators executing the opstask have an estimated idea for the timeout value? They'll most like;y resort to the defaults right?

docs/usage/using-etcdopstask.md
```diff
@@ -129,7 +127,65 @@ Triggers an on-demand snapshot outside the regular snapshot schedule.
...
+- The target etcd cluster must be a multi-member cluster (`spec.replicas > 1`). For single-member
+  clusters the backup-restore sidecar handles recovery automatically.
+- Backup must be enabled for the target etcd cluster (`spec.backup.store` must be configured).
+  Recovery relies on the backup-restore sidecar restoring from the latest snapshot; without a
+  backup store there is nothing to restore from and the task will be rejected.
+- The cluster must currently **not** be in a ready state (quorum loss has occurred).
+- No other `EtcdOpsTask` should be in progress for the same etcd cluster.
+
+**Configuration options:**
+- `scaleDownTimeout`: Maximum time to wait for the StatefulSet to reach 0 replicas (default: `60s`). Accepts Go duration strings (e.g. `30s`, `2m`).
```
Shreyas-s14, line 169: The default values for these fields once decided can be changed here as well.

docs/usage/using-etcdopstask.md
```diff
@@ -129,7 +127,65 @@ Triggers an on-demand snapshot outside the regular snapshot schedule.
...
+> and cannot recover on its own.
+
+**Prerequisites:**
+- The target etcd cluster must be a multi-member cluster (`spec.replicas > 1`). For single-member
+  clusters the backup-restore sidecar handles recovery automatically.
+- Backup must be enabled for the target etcd cluster (`spec.backup.store` must be configured).
+  Recovery relies on the backup-restore sidecar restoring from the latest snapshot; without a
+  backup store there is nothing to restore from and the task will be rejected.
+- The cluster must currently **not** be in a ready state (quorum loss has occurred).
+- No other `EtcdOpsTask` should be in progress for the same etcd cluster.
```
Shreyas-s14, line 158: Nit: This condition is implicitly set by the reconcile logic right? If you feel its fine to have it here, you can mark this as resolved.

docs/usage/using-etcdopstask.md
```diff
@@ -129,7 +127,65 @@ Triggers an on-demand snapshot outside the regular snapshot schedule.
...
+
+**Configuration options:**
+- `scaleDownTimeout`: Maximum time to wait for the StatefulSet to reach 0 replicas (default: `60s`). Accepts Go duration strings (e.g. `30s`, `2m`).
+- `podReadyTimeout`: Maximum time to wait for the single-member pod (`<etcd-name>-0`) to become ready after scale-up (default: `3m`). Accepts Go duration strings.
```
Shreyas-s14: One thing, if we have these values in seconds, wouldn't validations be easier? For ex: if it was `PodReadyTimeoutSeconds` instead, in the api, markers for default values can be set. Just an opinion

docs/usage/using-etcdopstask.md
```diff
@@ -129,7 +127,65 @@ Triggers an on-demand snapshot outside the regular snapshot schedule.
...
+```
+
+**Expected behaviour:**
+1. The task suspends etcd-druid reconciliation and disables component protection on the target Etcd resource.
+2. The StatefulSet is scaled to 0 and all PVCs and member leases are deleted.
+3. The etcd ConfigMap is patched to configure a single-member cluster using member-0 only.
+4. The StatefulSet is scaled to 1 and the task waits for the pod to become ready.
+5. The backup-restore sidecar restores etcd data from the latest available snapshot.
+6. Once the pod is ready, etcd-druid reconciliation is re-enabled and component protection is restored.
+7. etcd-druid scales the cluster back to the original replica count.
```
Shreyas-s14: Nice section! Maybe we make it a norm to add it for future EtcdOpsTasks (since snapshot is a "simple" task in that regard)

examples/etcdopstask/recover-from-quorum-loss.yaml
```diff
@@ -0,0 +1,39 @@
...
+#   1. Suspend etcd-druid reconciliation.
+#   2. Disable etcd component protection.
+#   3. Scale the StatefulSet down to 0.
+#   4. Delete all PVCs.
+#   5. Delete all member leases.
+#   6. Patch the ConfigMap to configure a single-member cluster (member-0 only).
+#   7. Scale up the StatefulSet to 1.
+#   8. Wait for the single-member etcd pod to become ready.
+#   9. Re-enable etcd-druid reconciliation (triggers scale-out to original replica count).
+#  10. Re-enable etcd component protection.
```
Shreyas-s14: The doc already covers this so imo it can be removed WDYT?
plus, none of the other examples have the license header, so that can be removed to maintain consistency.

examples/etcdopstask/recover-from-quorum-loss.yaml
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  etcdName: etcd-main
+  ttlSecondsAfterFinished: 3600
+  config:
+    recoverFromQuorumLoss:
+      # scaleDownTimeout: how long to wait for the StatefulSet to scale down to 0.
+      # Defaults to 60s. Accepts Go duration strings (e.g. 30s, 2m).
+      scaleDownTimeout: 60s
+      # podReadyTimeout: how long to wait for the single-member etcd pod (ordinal 0) to become ready after scale-up.
+      # Defaults to 180s. Accepts Go duration strings (e.g. 2m, 5m).
```
Shreyas-s14, line 21: Personally I have no opinion on this but in other examples, the fields aren't described. Just wanted to point this out. You can resolve this if needed.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+	for i, s := range stepOrder {
+		if string(s) == val {
+			return i
+		}
+	}
+	return -1
+}
+
+// Handler implements the taskhandler.Handler interface for the RecoverFromQuorumLoss task.
+type Handler struct {
```
Shreyas-s14: All the other methods directly interact with the Hanlder interface from types.go in `internal/controller/etcdopstask/handler` rather than the struct here. So, this need not be exported. (Similar convention followed in on demand snapshot handler)

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+	k8sClient        client.Client
+	etcdReference    types.NamespacedName
+	scaleDownTimeout time.Duration
+	podReadyTimeout  time.Duration
```
Shreyas-s14:
```suggestion
	config druidv1alpha1.RecoverFromQuorumLossConfig
```
It would be cleaner to have the config here and then use the fields from the config instead of having all the fields explicitly set as fields in the struct.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+	cfg := task.Spec.Config.RecoverFromQuorumLoss
+
+	scaleDownTimeout := defaultScaleDownTimeout
+	if cfg != nil && cfg.ScaleDownTimeout != nil {
+		scaleDownTimeout = cfg.ScaleDownTimeout.Duration
+	}
+
+	podReadyTimeout := defaultPodReadyTimeout
+	if cfg != nil && cfg.PodReadyTimeout != nil {
+		podReadyTimeout = cfg.PodReadyTimeout.Duration
```
Shreyas-s14: As per previous comments regarding default.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+func (h *Handler) Admit(ctx context.Context) taskhandler.Result {
+	etcd, errResult := handlerutils.GetEtcd(ctx, h.k8sClient, h.etcdReference, druidv1alpha1.LastOperationTypeAdmit)
+	if errResult != nil {
+		return *errResult
+	}
+
+	if etcd.Spec.Replicas <= 1 {
+		return taskhandler.Result{
+			Description: "RecoverFromQuorumLoss task is not applicable for a single-member etcd cluster. " +
+				"The backup-restore sidecar handles automatic recovery for single-member clusters.",
```
Shreyas-s14:
```suggestion
			Description: "Task is not applicable for a single-member etcd cluster. " +
				"The backup-restore sidecar handles automatic recovery for single-member clusters.",
```
it is already known the task is of type RecoverFromQuorumLoss, so the description can be trimmed.
Also since this is explicitly mentioned in the docs, is there a need for the phrase regarding backup-restore?

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+				string(druidv1alpha1.LastOperationTypeAdmit),
+				"single-member cluster does not require quorum loss recovery",
+			),
+			Requeue: false,
+		}
+	}
+
+	if !etcd.IsBackupStoreEnabled() {
+		return taskhandler.Result{
+			Description: "RecoverFromQuorumLoss requires a configured backup store. " +
```
Shreyas-s14: https://github.com/gardener/etcd-druid/pull/1358#discussion_r3301714649

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+//  3. Scale the StatefulSet down to 0.
+//  4. Wait for the StatefulSet to reach 0 replicas.
+//  5. Delete all PVCs.
+//  6. Delete all member leases.
+//  7. Patch the ConfigMap to configure a single-member cluster (member-0 only).
+//  8. Scale up the StatefulSet to 1.
+//  9. Wait for the single-member etcd pod to become ready.
+//
+// 10. Re-enable etcd-druid reconciliation (triggers scale-out to original replica count).
+// 11. Re-enable etcd component protection.
```
Shreyas-s14: Having to list all the methods here seems unncessary to me. That is the purpose of the docs. But, I'll leave it up to the rest of the reviewers to decide.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+				ErrAdmitRecoverFromQuorumLoss,
+				string(druidv1alpha1.LastOperationTypeAdmit),
+				"etcd cluster is ready; quorum loss not detected",
+			),
+			Requeue: false,
+		}
+	}
+
+	return taskhandler.Result{
+		Description: "Preconditions for RecoverFromQuorumLoss task are satisfied",
```
Shreyas-s14:
```suggestion
		Description: "Admit check passed",
```
To maintain consistency with the OndemandSnapshot task.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+}
+
+// refreshEtcd re-fetches the Etcd object to avoid stale-object patch conflicts.
+func (h *Handler) refreshEtcd(ctx context.Context) (*druidv1alpha1.Etcd, *taskhandler.Result) {
+	etcd, errResult := handlerutils.GetEtcd(ctx, h.k8sClient, h.etcdReference, druidv1alpha1.LastOperationTypeExecution)
+	if errResult != nil {
+		return nil, errResult
+	}
+	return etcd, nil
+}
```
Shreyas-s14: This entire function can be dropped and handled in the Execute method. `handlerutils.GetEtcd` already has the same return type.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+	if pod.Status.Phase != corev1.PodRunning {
+		return false, nil
+	}
+	for _, cond := range pod.Status.Conditions {
+		if cond.Type == corev1.PodReady {
+			return cond.Status == corev1.ConditionTrue, nil
+		}
+	}
+	return false, nil
+}
```
Shreyas-s14:
```suggestion
// isPodReady checks whether the given pod is Running and its Ready condition is True.
func (h *Handler) isPodReady(ctx context.Context, podKey types.NamespacedName) (bool, error) {
      pod := &corev1.Pod{}
      if err := h.k8sClient.Get(ctx, podKey, pod); err != nil {
          if apierrors.IsNotFound(err) {
              return false, nil
          }
          return false, err
      }
      return kubernetes.HasPodReadyConditionTrue(pod), nil
}

```
import from `internal/utils/kubernetes`
Also, the check for Phase != Running is not needed since the `Running` phase is set only when the Pod is in Ready state. Or has there been some case wjere this is not true?

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+		Description: description,
+		Requeue:     true,
+	}
+	return &r
+}
+
+// recordCompletedStep writes the step name into the Etcd CR annotation so it can be
+// skipped on the next reconcile invocation.
+func (h *Handler) recordCompletedStep(ctx context.Context, etcd *druidv1alpha1.Etcd, step recoveryStep) error {
+	patch := client.MergeFrom(etcd.DeepCopy())
```
Shreyas-s14: Curious, why annotate the Etcd instead of the EtcdOpsTask resource? I agree that since the etcd isnt in a functional state, there's no harm in doing so but since the scope of the Task is the Quorum Loss recovery, wouldn't it be better to have it in the EtcdOpstask resource itself? The Operator who triggers the OpsTask will be primarily observing that to see the status, so it would be better to add it there right?

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+// here — every successful step triggers a requeue except the last, which is handled in Execute).
+func (h *Handler) runStep(ctx context.Context, etcd *druidv1alpha1.Etcd, step recoveryStep, description string, fn func() error) *taskhandler.Result {
+	if err := fn(); err != nil {
+		r := taskhandler.Result{
+			Description: fmt.Sprintf("Failed during step %s", step),
+			Error: druiderr.WrapError(err, ErrExecuteRecoverFromQuorumLoss,
+				string(druidv1alpha1.LastOperationTypeExecution),
+				fmt.Sprintf("failed to execute step %s", step)),
+		}
+		return &r
```
Shreyas-s14:
```suggestion
		return &taskhandler.Result{
			Description: fmt.Sprintf("Failed during step %s", step),
			Error: druiderr.WrapError(err, ErrExecuteRecoverFromQuorumLoss,
				string(druidv1alpha1.LastOperationTypeExecution),
				fmt.Sprintf("failed to execute step %s", step)),
		}
```
Similarly for other places.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+						string(druidv1alpha1.LastOperationTypeExecution),
+						"failed to get StatefulSet for scale-down check"),
+				}
+			}
+			if !scaledDown {
+				return taskhandler.Result{
+					Description: "Waiting for StatefulSet to scale down to 0 replicas",
+					Requeue:     true,
+				}
+			}
```
Shreyas-s14, line 402: Shouldn't this step be using the scaledown timeout provided in the api to requeue? Unless I'm missing something.
If my understanding is correct, if the sts takes a value longer than the `ScaleDownTimeout`, there is no consequence which should not be the case

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+						"failed to get pod for readiness check"),
+				}
+			}
+			if !ready {
+				return taskhandler.Result{
+					Description: "Waiting for single-member etcd pod to become ready",
+					Requeue:     true,
+				}
+			}
+			result = h.runStep(ctx, etcd, step, "Single-member etcd pod is ready", func() error { return nil })
```
Shreyas-s14: Same here, the retry should ensure that the timeout as defined in the api config is ensured. I can see an infinite retry here since apart from the New function, the timeout values are not used anywhere.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
```diff
@@ -0,0 +1,629 @@
...
+		if step == stepReenableReconciliation || step == stepReenableComponentProtect {
+			var r *taskhandler.Result
+			etcd, r = h.refreshEtcd(ctx)
+			if r != nil {
+				return *r
+			}
+		}
+
+		switch step {
+		case stepSuspendReconciliation:
```
Shreyas-s14, line 354: Thinking out loud here,
Instead of having a slice and a bunch of helper functions scattered throughout the file, wouldn't it be better to have a slice of structs of say 
```go
type stepDef struct {
      step        recoveryStep
      description string
      fn          func(ctx context.Context, etcd *druidv1alpha1.Etcd) error
  }
```
similar to what we do in reconcile step functions. This can also help reduce the switch case logic and make it "cleaner" imo
So this is then passed into the run function, and will reduce the number of arguments there as well.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
```diff
@@ -0,0 +1,1310 @@
...
+	}
+	return result
+}
+
+// TestRecoverFromQuorumLossExecute tests the Execute method through the full requeue loop.
+func TestRecoverFromQuorumLossExecute(t *testing.T) {
+	g := NewGomegaWithT(t)
+	const (
+		etcdName  = "test-etcd"
+		namespace = "test-ns"
```
Shreyas-s14, line 630: IN a unit test scenario, this is not needed since no k8s environment is created.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
```diff
@@ -0,0 +1,1310 @@
...
+		}
+	}
+
+	// Final step returns success with no requeue.
+	result := taskHandler.Execute(ctx)
+	g.Expect(result.Error).To(BeNil())
+	g.Expect(result.Requeue).To(BeFalse())
+	g.Expect(result.Description).To(Equal("Recovery from quorum loss completed successfully. " +
+		"etcd-druid will reconcile and bring up the remaining cluster members."))
+}
```
Shreyas-s14, line 838: I feel this "flow" is more suitable in the it tests or e2e tests. WDYT? The individual helper or sequence functions are tested anyways and this covers the scope of the unit tests. IT tests would be more suitable since these functions would be applied on a real k8s object rather than a fakeclient response.
WDYT

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
```diff
@@ -0,0 +1,1310 @@
...
+
+// TestUpdateConfigMapForSingleMember tests the updateConfigMapForSingleMember method.
+func TestUpdateConfigMapForSingleMember(t *testing.T) {
```
Shreyas-s14: nit: can the unit tests for the individual recovery functions be ordered based on the sequence in the parent file?
ie
```bash
    stepSuspendReconciliation,
	stepDisableComponentProtect,
	stepScaleDown,
	stepWaitScaleDown,
	stepDeletePVCs,
	stepDeleteMemberLeases,
	stepPatchConfigMap,
	stepScaleUp,
	stepWaitPodReady,
	stepReenableReconciliation,
	stepReenableComponentProtect
```
It'll look more structured and in line with how the functions are defined in the parent file.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
```diff
@@ -0,0 +1,1310 @@
...
+			g.Expect(err).To(BeNil())
+			h := taskHandler.(*Handler)
+			g.Expect(h.scaleDownTimeout.Seconds()).To(Equal(tc.expectedScaleDownTimeout))
+			g.Expect(h.podReadyTimeout.Seconds()).To(Equal(tc.expectedPodReadyTimeout))
+		})
+	}
+}
+
+// TestHelperFunctions tests the serverPortOrDefault and clientPortOrDefault helpers.
+func TestHelperFunctions(t *testing.T) {
```
Shreyas-s14: This naming is a bit vague. It would be better to either have individual tests or rename it to somethhing like `TestServerAndClientPortDefaults`

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
```diff
@@ -0,0 +1,1310 @@
...
+			objs = append(objs, etcd)
+			for _, pvc := range tc.pvcs {
+				objs = append(objs, pvc)
+			}
+			cl := utils.NewTestClientBuilder().WithScheme(kubernetes.Scheme).WithObjects(objs...).Build()
+			task := buildEtcdOpsTask(etcdName, namespace, nil)
+			taskHandler, err := New(cl, task, nil)
+			g.Expect(err).To(BeNil())
+			delErr := taskHandler.(*Handler).deleteAllPVCs(context.Background(), etcd)
+			g.Expect(delErr).To(BeNil())
```
Shreyas-s14, line 1376: Since this is a critical step for the recovery process, the unit test must do a check to ensure deletion. Can that be added?

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
```diff
@@ -0,0 +1,1310 @@
...
+			objs = append(objs, etcd)
+			for _, lease := range tc.leases {
+				objs = append(objs, lease)
+			}
+			cl := utils.NewTestClientBuilder().WithScheme(kubernetes.Scheme).WithObjects(objs...).Build()
+			task := buildEtcdOpsTask(etcdName, namespace, nil)
+			taskHandler, err := New(cl, task, nil)
+			g.Expect(err).To(BeNil())
+			delErr := taskHandler.(*Handler).deleteAllMemberLeases(context.Background(), etcd)
+			g.Expect(delErr).To(BeNil())
```
Shreyas-s14, line 1426: Same as https://github.com/gardener/etcd-druid/pull/1358#discussion_r3303514711

Shreyas-s14 (commented):
Thanks for implementing this manual process via an EtcdOpsTask! While the overall flow is well captured along with the docs, I have a few nits and concerns. PTAL
Also, can integration tests be added in `test/it/crdvalidation` for the spec validations and `test/it/controller/etcdopstask` under a new directory called `recoverfromquorumloss` ?

api/core/v1alpha1/recoverfromquorumloss.go
ashwani2k: I think with metav1.Duration the easier validation is via kubebuilder marker `// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"` and the defaults can be set in the handler. In ondemandsnapshot the timeout are set via *int which is not a k8s idiomatic way of defining time based fields. 

However, the handler's constants (300s/600s) were inconsistent with the API comments (60s/180s) and docs (60s/3m). Aligned all three to 60s / 180s. with a1f86ef9c239a459314e8f97beeefaa3e798f414

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Good catch Yup removed with a1f86ef9c239a459314e8f97beeefaa3e798f414

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Fixed with a1f86ef9c239a459314e8f97beeefaa3e798f414

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Fixed with a1f86ef9c239a459314e8f97beeefaa3e798f414

api/core/v1alpha1/recoverfromquorumloss.go
ashwani2k: Fixed with a1f86ef9c239a459314e8f97beeefaa3e798f414

api/core/v1alpha1/recoverfromquorumloss.go
ashwani2k: Yes, but in case there is a slowness in the cluster then may be a choice to increase the timeout comes in handy. Also we can fine tune the defaults as well to a better heuristic if required. Overall defaults also help in stopping the opstask from running endlessly.

docs/usage/using-etcdopstask.md
ashwani2k: Done with a1f86ef9c239a459314e8f97beeefaa3e798f414

docs/usage/using-etcdopstask.md
ashwani2k: The behavior in the logic is only thrown when you run the task, but here we are explicitly sharing what conditions are kept in mind ensuring that the end user while configuring is aware of them? Do you see any harm here in communicating these or you see this causes more confusion?

docs/usage/using-etcdopstask.md
ashwani2k: Its what metav1.Duration can take so one can give either more readable value of 3m or make it 180s. I think both are valid. I understand in ondemandsnapshot we have made it an integer but then I think duration is a better fit here.

docs/usage/using-etcdopstask.md
ashwani2k: We can leave each task to explain what fits well, as you said some task might be simple in which case it may not add much value wdyt?

examples/etcdopstask/recover-from-quorum-loss.yaml
ashwani2k: Agree, its a redundant information as is already captured in the usage doc. So removed it with a1f86ef9c239a459314e8f97beeefaa3e798f414

examples/etcdopstask/recover-from-quorum-loss.yaml
ashwani2k: I personally feel its always good for an end user to know what values this field takes so that they don't need to revisit the docs to understand them again and there is no ambiguity on the usage.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Good catch, not sure why it did that. I've adapted it with a1f86ef9c239a459314e8f97beeefaa3e798f414

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Agreed ondemand does the same, not sure why this wasn't done here. Adapted as suggested. a1f86ef9c239a459314e8f97beeefaa3e798f414

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Adapted with a1f86ef9c239a459314e8f97beeefaa3e798f414

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Fixed with 86ac85d3e15c0a6e91f6d1da0d8a3d89bd2641f4

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Fixed with 86ac85d3e15c0a6e91f6d1da0d8a3d89bd2641f4

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Fixed with 86ac85d3e15c0a6e91f6d1da0d8a3d89bd2641f4

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: This one is tricky, although I get your point to reduce the switch cases and follow a common pattern by looping through a slice of structe but if you see the code now, it has some variations as it needs to also handle timeout, which is different then the scenarios we have in runtask reconciler as there everything is running in sync and is more or less same for each step. 

I'll say lets watch for other reviewers input on this and take a call, wdyt?

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: I think it makes it easy to understand what Execute will do here, can also help handy in doc generation.

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Agreed, fixed with https://github.com/gardener/etcd-druid/commit/86ac85d3e15c0a6e91f6d1da0d8a3d89bd2641f4

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Yes, this is overengineered fixed with https://github.com/gardener/etcd-druid/commit/86ac85d3e15c0a6e91f6d1da0d8a3d89bd2641f4

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Agreed, fixed with https://github.com/gardener/etcd-druid/commit/86ac85d3e15c0a6e91f6d1da0d8a3d89bd2641f4

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Nice catch, fixed every usage with https://github.com/gardener/etcd-druid/commit/86ac85d3e15c0a6e91f6d1da0d8a3d89bd2641f4

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss.go
ashwani2k: Yup, this was a complete miss in the implementation. Thanks for spotting it, fixed with.
https://github.com/gardener/etcd-druid/commit/86ac85d3e15c0a6e91f6d1da0d8a3d89bd2641f4

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
ashwani2k: Done with https://github.com/gardener/etcd-druid/pull/1358/changes/021fc81f7faef7558f1262acea5b90b664c90f9c#diff-5c71966b51283ff55bfbd0537fe34267eae823120415abc24e5fed90ad02f2a9R360

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
ashwani2k: These are used when creating an ETCD for a specific usage, I see in ondemandsnapshot, instead of constant you pass the name and namespace to every test case where as here it just places them in a constant which can be changed later one place. 
Or am I missing something here?

internal/controller/etcdopstask/handler/recoverfromquorumloss/recoverfromquorumloss_test.go
ashwani2k: Agree makes sense. Fixed with https://github.com/gardener/etcd-druid/pull/1358/changes/7e889586bd41367e38be2064facbf037d64c6d3e

Issue #922, opened 2024-11-13
Provide capability to hibernate and wake-up etcd clusters and handle it completely via etcd-druid


unmarshall:
**How to categorize this issue?**

/area control-plane
/kind enhancement

**What would you like to be added**:

In `Gardener` use case shoots are hibernated and as part of hibernating a shoot the following is done:
1. A full snapshot is taken. This ensures that when the etcd-cluster wakes up it does not lose any data and can reliably restore from the last known state.
2. etcd-cluster is scaled down to 0.

This issue proposes that hibernation and wake-up from hibernation be offered as functionalities directly in etcd-druid which can be used in and outside of gardener context. The consumer can mark an etcd-cluster to be hibernated and watch the `Etcd.Status` to know the status of hibernation. The steps that are done in order to hibernate a cluster are determined based on the `Etcd` resource itself.

Following points can be considered when defining the API:

- API provision should be made to signal request to hibernate and request to wake-up.
- An etcd cluster can be created with backup disabled. For these clusters hibernation will not really include taking a full snapshot as there is no safety net configured to backup full and delta snapshots.
- https://github.com/gardener/etcd-druid/issues/859 talks about how to optimize costs when hibernating etcd clusters. This could be made configurable and could be used for clusters which do not wish to have a backup but would want to preserve the data only via network attached disks that are attached to the node (PV) and used by etcd pods. (This is optional).
- Taking and uploading a full snapshot could error out. One could offer two modes to take backups before scaling down - `preferred` | `required`. If it is `required` then consider providing a timeout beyond which etcd-druid will no longer retry and will report failure. Manual intervention is then required to correct the issue blocking the taking of full snapshot and uploading it to a bucket and then the operation can be re-triggered.

**Why is this needed**:

Hibernation and wake up of etcd clusters are already supported in gardener via reconcile loops running in gardenlet. However it is not exposed as a functionality to non-gardener users. Having a clear and well-defined API to signal hibernation and wake-up of an etcd cluster would ease consumption. It also semantically makes sense for etcd-druid (an etcd operator) to abstract all activities that are performed as part of hibernation and wake-up reducing the burden on the consumers to understand the intricacies/details.

PR #1360, opened 2026-05-20
Take full snapshot before any container image or replica change

### LINKED ISSUE #1359: Take a full snapshot before updating etcd clusters with any container images or replica changes

/area backup
/kind enhancement

In the case of a single-node cluster, I would like to trigger a full snapshot before any image upgrade happens to any of the containers in the `STS` as that leads to rolling of pods and in case of a single member that could potentially be a problem if the rolling causes issues.

- Performance/Cost:

The release cadence we follow is to release once a month, but exceptions are expected due to some security patches or quick integrations - but overall the full snapshot will not be taken very frequently and also the fact that most operators operate it with HA means that the operation is not destructive in any sense to performance or to the cost of storing these blobs.

~~- Why only for non-HA:~~

~~In the case of an HA cluster a rolling typically takes place by killing the pods in the higher ordinal to lower and everytime that happens a member is added as a learner again and it learns from the rest of the members which is not the case in case of a non-HA cluster.~~

We've decided that to make it simpler and also safe-guard multi-node clusters, we'll be going to trigger full snapshot for HA updates also. So now any updates to image/replica irrespective of non-HA or HA involves triggering a full snapshot through opsTask.

**Why is this needed**:

This is to ensure that we have a safe state to return always if the upgrade results in any issues with data corruption or data loss, etc.

**Scope**:

We'll be scoping this just to the image upgrades and not all upgrade to the `podTemplateSpec` even though the latter also rolls the pod under sts because it's not trivial to predict whether the incoming changes are going to change the hash of pod template spec even before applying it, any hash algorithm assumption there and pre-computing it from our side would not be ideal as the sts hashing implementation is internal and is subjected to change at any time.

For this reason, we'll keep it simple and deterministic by choosing to trigger the full snapshot only when key configuration like image update happens.

anveshreddy18:
  /area backup
  /area disaster-recovery
  /kind enhancement

  Triggers a full snapshot via `EtcdOpsTask` before the etcd `StatefulSet` is rolled due to a container image change or a replica count change, irrespective of HA / non-HA. Builds on the snapshot-before-rolling mechanism introduced in #1300, which previously covered only hibernation and the `UpgradeEtcdVersion` gated etcd-version-upgrade flow.

The pre-sync `EtcdOpsTask` name prefix has been renamed from `presync-snapshot-upgrade-` to `presync-snapshot-update-` to reflect the broader trigger. The `UpgradeEtcdVersion` feature gate itself is unchanged; it simply no longer gates the snapshot trigger.
  
Scope is intentionally limited to image and replica fields. Arbitrary `podTemplateSpec` mutations are out of scope because the StatefulSet pod-template-hash is computed by an internal, version-dependent algorithm in Kubernetes; pre-computing whether a roll will happen is not feasible without coupling druid to a specific Kubernetes version.

Fixes #1359

**Checklist**:
    - [x] Unit tests
    - [x] Integration tests
    - [ ] E2E tests

During a druid version upgrade, an in-flight `presync-snapshot-upgrade-N` task created by the older operator will be ignored by the new code. The `etcdopstask` controller's FIFO + one-active-task-per-cluster guarantee means at most one extra full snapshot is taken across the upgrade boundary.

**Release note**:

```other operator
Take a full snapshot via `EtcdOpsTask` before the etcd `StatefulSet` is rolled due to any container image change or replica count change, in both HA and non-HA setups
```

```noteworthy operator
The pre-sync `EtcdOpsTask` name prefix is renamed from `presync-snapshot-upgrade-` to `presync-snapshot-update-`; in-flight tasks created by an older operator may result in at most one additional snapshot across the upgrade boundary.
```

### DIFF
--- a/api/core/v1alpha1/constants.go
+++ b/api/core/v1alpha1/constants.go
@@ -39,12 +39,11 @@ const (
 	// This value will only be effective if etcd-druid is not configured with auto-reconciliation of Etcd resource specification via
 	// --enable-etcd-spec-auto-reconcile CLI flag.
 	DruidOperationReconcile = "reconcile"
-	// SkipNextUpdateSnapshotAnnotation, when present on an Etcd resource, instructs etcd-druid to skip the pre-sync
-	// full snapshot for the immediately pending StatefulSet update triggered by a container image or replica change.
-	// It's a one-shot annotation: etcd-druid removes it as soon as the skip is applied, so it only affects the next
-	// immediate pending update. Its presence alone is sufficient; the annotation value is ignored. It does not affect the
-	// snapshot taken before hibernation.
-	SkipNextUpdateSnapshotAnnotation = "druid.gardener.cloud/skip-next-update-snapshot"
+	// SkipSpecUpdateSnapshotAnnotation is an annotation when present on an Etcd resource, instructs the etcd-druid to skip the pre-sync full
+	// snapshot which would otherwise be taken before the etcd StatefulSet is rolled due to a container image or replica change.
+	// Its presence alone is sufficient; the annotation value is ignored. While the annotation is present the snapshot is always skipped, and etcd-druid does not remove the annotation.
+	// It does not affect the full snapshot taken before hibernation.
+	SkipSpecUpdateSnapshotAnnotation = "druid.gardener.cloud/skip-spec-update-snapshot"
 )
 
 // Compaction Job/Pod reasons that are used to set the reason for a pod condition in the status of an Etcd resource.
--- a/api/core/v1alpha1/constants.go
+++ b/api/core/v1alpha1/constants.go
@@ -39,6 +39,12 @@ const (
 	// This value will only be effective if etcd-druid is not configured with auto-reconciliation of Etcd resource specification via
 	// --enable-etcd-spec-auto-reconcile CLI flag.
 	DruidOperationReconcile = "reconcile"
+	// SkipNextUpdateSnapshotAnnotation, when present on an Etcd resource, instructs etcd-druid to skip the pre-sync
+	// full snapshot for the immediately pending StatefulSet update triggered by a container image or replica change.
+	// It's a one-shot annotation: etcd-druid removes it as soon as the skip is applied, so it only affects the next
+	// immediate pending update. Its presence alone is sufficient; the annotation value is ignored. It does not affect the
+	// snapshot taken before hibernation.
+	SkipNextUpdateSnapshotAnnotation = "druid.gardener.cloud/skip-next-update-snapshot"
 )
 
 // Compaction Job/Pod reasons that are used to set the reason for a pod condition in the status of an Etcd resource.
--- a/docs/usage/managing-etcd-clusters.md
+++ b/docs/usage/managing-etcd-clusters.md
@@ -13,6 +13,7 @@ In order to track the progress of creation of etcd cluster resources you can do 
 * `status.lastOperation` can be monitored to check the status of reconciliation.
 
 * Additional printer columns have been defined for `Etcd` custom resource. You can execute the following command to know if an `Etcd` cluster is ready/quorate.
+
 ```bash
 kubectl get etcd <etcd-name> -n <namespace> -owide
   # you will see additional columns which will indicate the state of an etcd cluster
@@ -23,11 +24,13 @@ kubectl get etcd <etcd-name> -n <namespace> -owide
 * You can additional monitor [all etcd cluster resources](../concepts/etcd-cluster-components.md) that are created for every etcd cluster. 
 
   For etcd-druid version <v0.23.0 use the following command:
+
 ```bash
 kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=instance=<etcd-name>
 ```
 
   For etcd-druid version >=v0.23.0 use the following command:
+
 ```bash
 kubectl get all,cm,role,rolebinding,lease,sa -n <namespace> --selector=app.kubernetes.io/managed-by=etcd-druid,app.kubernetes.io/part-of=<etcd-name>
 ```
@@ -89,6 +92,7 @@ If `etcd-druid` has been deployed with auto-reconciliation then any change done 
 > For a complete list of CLI args you can see [this](../deployment/configure-etcd-druid.md) document.
 
 #### Explicit reconciliation
+
 If `--enable-etcd-spec-auto-reconcile` is set to false or not set at all, then any change to an `Etcd` resource will not be automatically reconciled. To trigger a reconcile you must set the following annotation on the `Etcd` resource:
 
 ```bash
@@ -99,32 +103,35 @@ This option is sometimes recommeded as you would like avoid auto-reconciliation 
 
 ### Full snapshot before a StatefulSet update
 
-Before `etcd-druid` rolls the etcd `StatefulSet` due to a container image change or a replica count change (in both HA and non-HA setups), it triggers a full snapshot via an `EtcdOpsTask`. This ensures there is a recent, safe state to restore from should the rollout run into data corruption or data loss. The snapshot is attempted a bounded number of times; if all attempts fail, `etcd-druid` proceeds with the update regardless (so that a persistently failing snapshot does not block updates indefinitely).
+Before `etcd-druid` rolls the etcd `StatefulSet` due to a container image or replica count change, it triggers a full snapshot via an [`EtcdOpsTask`](using-etcdopstask.md). This provides a recent recovery point in case the rollout results in data corruption or data loss. This behavior applies to both HA and non-HA setups.
+
+The snapshot is attempted a bounded number of times. If all attempts fail, `etcd-druid` proceeds with the update without a fresh snapshot, so that a persistently failing snapshot does not block updates indefinitely.
 
 #### Skip the snapshot before updates
 
 If snapshots are known to be failing, or you need changes to roll out quickly without waiting for a snapshot, you can instruct `etcd-druid` to skip the pre-update snapshot by annotating the `Etcd` resource with `druid.gardener.cloud/skip-spec-update-snapshot`:
 
 ```bash
-   kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/skip-spec-update-snapshot=
+kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/skip-spec-update-snapshot=
 ```
 
 This annotation is:
 
-- **Presence-only**: the annotation value is ignored; simply having the key present enables the skip.
-- **Persistent**: while the annotation is present, the pre-update snapshot is always skipped. `etcd-druid` does not remove it — remove the annotation yourself to re-enable pre-update snapshots.
-- **Scoped to image/replica updates**: it does not affect the full snapshot taken before hibernation (scaling the cluster to zero replicas).
+* **Presence-only**: the annotation value is ignored; simply having the key present enables the skip.
+* **Persistent**: while the annotation is present, the pre-update snapshot is always skipped. `etcd-druid` does not remove it — remove the annotation yourself to re-enable pre-update snapshots.
+* **Scoped to image/replica updates**: it does not affect the full snapshot taken before hibernation (scaling the cluster to zero replicas).
+* **Limited to the pre-update snapshot**: the regular scheduled full and delta snapshots, as well as out-of-band snapshots, continue to be taken as usual while the annotation is present.
 
 #### Surfacing snapshot failures
 
-When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
+If the pre-update snapshot exhausts its retries, `etcd-druid` proceeds with the update without a fresh snapshot and records a `Warning` event with the reason `PreSyncSnapshotFailed` on the `Etcd` resource. You can view the event using:
 
 ```bash
-   kubectl describe etcd <etcd-name> -n <namespace>
+kubectl describe etcd <etcd-name> -n <namespace>
 ```
 
-!!! note
-    A `PreSyncSnapshotFailed` event means the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` `EtcdOpsTask` resources and the backup configuration to find out why the snapshot failed.
+> [!NOTE]
+> A `PreSyncSnapshotFailed` event means that the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` [`EtcdOpsTask`](using-etcdopstask.md) resources and the backup configuration to determine why the snapshot failed.
 
 ## Overwrite Container OCI Images
 
@@ -163,6 +170,7 @@ Edit the `etcd-druid` `Deployment` with:
 * Set `IMAGEVECTOR_OVERWRITE` environment variable whose value must be the path you choose to mount the `ConfigMap`.
 
 To illustrate the changes you can see the following `etcd-druid` Deployment YAML:
+
 ```yaml
 apiVersion: apps/v1
 kind: Deployment
@@ -195,22 +203,21 @@ We provide a generic way to suspend etcd cluster reconciliation via etcd-druid, 
 
 ## Manually modify individual etcd cluster resources
 
-`etcd` cluster resources are managed by `etcd-druid` and since v0.23.0 version of `etcd-druid` any changes to these managed resources are protected via a validating webhook. You can find more information about this webhook [here](../concepts/etcd-cluster-resource-protection.md). To be able to manually modify etcd cluster managed resources two things needs to be done:
+`etcd` cluster resources are managed by `etcd-druid` and since v0.23.0 version of `etcd-druid` any changes to these managed resources are protected via a validating webhook. You can find more information about this webhook in the [concept](../concepts/etcd-cluster-resource-protection.md). To be able to manually modify etcd cluster managed resources two things needs to be done:
 
 1. Annotate the target `Etcd` resource suspending any reconciliation by `etcd-druid`. You can do this by invoking the following command:
 
 ```bash
-   kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/suspend-etcd-spec-reconcile=
+kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/suspend-etcd-spec-reconcile=
 ```
 
 2. Add another annotation to the target `Etcd` resource disabling managed resource protection via the webhook. You can do this by invoking the following command:
 
 ```bash
-   kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/disable-etcd-component-protection=
+kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/disable-etcd-component-protection=
 ```
 
 Now you are free to make changes to any managed etcd cluster resource.
 
 !!! note
     As long as the above two annotations are there, no reconciliation will be done for this etcd cluster by `etcd-druid`. Therefore it is essential that you remove this annotations eventually.
-
--- a/docs/usage/managing-etcd-clusters.md
+++ b/docs/usage/managing-etcd-clusters.md
@@ -133,6 +133,9 @@ kubectl describe etcd <etcd-name> -n <namespace>
 > [!NOTE]
 > A `PreSyncSnapshotFailed` event means that the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` [`EtcdOpsTask`](using-etcdopstask.md) resources and the backup configuration to determine why the snapshot failed.
 
+> [!WARNING]
+> `etcd-druid` detects a container image change by comparing the images it expects (derived from the `Etcd` resource and the configured image vector) against the images currently set on the `StatefulSet`. Always customize container images declaratively via the `Etcd` spec (see [Overwrite Container OCI Images](#overwrite-container-oci-images)) so that `etcd-druid` applies them as part of its own reconciliation. Do **not** mutate the images on the managed `StatefulSet` out-of-band through a mutating admission webhook, etc since such mutations are not reconciled by `etcd-druid`, so every reconciliation will detect an image difference and repeatedly trigger a pre-update snapshot.
+
 ## Overwrite Container OCI Images
 
 To find out image versions of `etcd-backup-restore` and `etcd-wrapper` used by a specific version of `etcd-druid` one way is look for the image versions in [images.yaml](https://github.com/gardener/etcd-druid/blob/master/internal/images/images.yaml). There are times that you might wish to override these images that come bundled with `etcd-druid`. There are two ways in which you can do that:
--- a/docs/usage/managing-etcd-clusters.md
+++ b/docs/usage/managing-etcd-clusters.md
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
 
 This option is sometimes recommeded as you would like avoid auto-reconciliation of accidental changes to `Etcd` resources outside the maintenance time window, thus preventing a potential transient quorum loss due to misconfiguration, attach-detach issues of persistent volumes etc.
 
+### Full snapshot before a StatefulSet update
+
+Before `etcd-druid` rolls the etcd `StatefulSet` due to a container image change or a replica count change (in both HA and non-HA setups), it triggers a full snapshot via an `EtcdOpsTask`. This ensures there is a recent, safe state to restore from should the rollout run into data corruption or data loss. The snapshot is attempted a bounded number of times; if all attempts fail, `etcd-druid` proceeds with the update regardless (so that a persistently failing snapshot does not block updates indefinitely).
+
+#### Skip the snapshot before updates
+
+If snapshots are known to be failing, or you need changes to roll out quickly without waiting for a snapshot, you can instruct `etcd-druid` to skip the pre-update snapshot by annotating the `Etcd` resource with `druid.gardener.cloud/skip-spec-update-snapshot`:
+
+```bash
+   kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/skip-spec-update-snapshot=
+```
+
+This annotation is:
+
+- **Presence-only**: the annotation value is ignored; simply having the key present enables the skip.
+- **Persistent**: while the annotation is present, the pre-update snapshot is always skipped. `etcd-druid` does not remove it — remove the annotation yourself to re-enable pre-update snapshots.
+- **Scoped to image/replica updates**: it does not affect the full snapshot taken before hibernation (scaling the cluster to zero replicas).
+
+#### Surfacing snapshot failures
+
+When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
+
+```bash
+   kubectl describe etcd <etcd-name> -n <namespace>
+```
+
+!!! note
+    A `PreSyncSnapshotFailed` event means the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` `EtcdOpsTask` resources and the backup configuration to find out why the snapshot failed.
+
 ## Overwrite Container OCI Images
 
 To find out image versions of `etcd-backup-restore` and `etcd-wrapper` used by a specific version of `etcd-druid` one way is look for the image versions in [images.yaml](https://github.com/gardener/etcd-druid/blob/master/internal/images/images.yaml). There are times that you might wish to override these images that come bundled with `etcd-druid`. There are two ways in which you can do that:
@@ -185,31 +214,3 @@ Now you are free to make changes to any managed etcd cluster resource.
 !!! note
     As long as the above two annotations are there, no reconciliation will be done for this etcd cluster by `etcd-druid`. Therefore it is essential that you remove this annotations eventually.
 
-## Full snapshot before a StatefulSet update
-
-Before `etcd-druid` rolls the etcd `StatefulSet` due to a container image change or a replica count change (in both HA and non-HA setups), it triggers a full snapshot via an `EtcdOpsTask`. This ensures there is a recent, safe state to restore from should the rollout run into data corruption or data loss. The snapshot is attempted a bounded number of times; if all attempts fail, `etcd-druid` proceeds with the update regardless (so that a persistently failing snapshot does not block updates indefinitely).
-
-### Skip the snapshot for the next update
-
-If snapshots are known to be failing, or you need a change to roll out quickly without waiting for a snapshot, you can instruct `etcd-druid` to skip the pre-update snapshot for the immediately pending update by annotating the `Etcd` resource:
-
-```bash
-   kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/skip-next-update-snapshot=
-```
-
-This annotation is:
-
-- **One-shot**: `etcd-druid` removes it as soon as the skip is applied, so it only affects the next pending update. Any subsequent image or replica change will again trigger a full snapshot.
-- **Presence-only**: the annotation value is ignored; simply having the key present enables the skip.
-- **Scoped to image/replica updates**: it does not affect the full snapshot taken before hibernation (scaling the cluster to zero replicas).
-
-### Surfacing snapshot failures
-
-When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
-
-```bash
-   kubectl describe etcd <etcd-name> -n <namespace>
-```
-
-!!! note
-    A `PreSyncSnapshotFailed` event means the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` `EtcdOpsTask` resources and the backup configuration to find out why the snapshot failed.
--- a/docs/usage/managing-etcd-clusters.md
+++ b/docs/usage/managing-etcd-clusters.md
@@ -184,3 +184,32 @@ Now you are free to make changes to any managed etcd cluster resource.
 
 !!! note
     As long as the above two annotations are there, no reconciliation will be done for this etcd cluster by `etcd-druid`. Therefore it is essential that you remove this annotations eventually.
+
+## Full snapshot before a StatefulSet update
+
+Before `etcd-druid` rolls the etcd `StatefulSet` due to a container image change or a replica count change (in both HA and non-HA setups), it triggers a full snapshot via an `EtcdOpsTask`. This ensures there is a recent, safe state to restore from should the rollout run into data corruption or data loss. The snapshot is attempted a bounded number of times; if all attempts fail, `etcd-druid` proceeds with the update regardless (so that a persistently failing snapshot does not block updates indefinitely).
+
+### Skip the snapshot for the next update
+
+If snapshots are known to be failing, or you need a change to roll out quickly without waiting for a snapshot, you can instruct `etcd-druid` to skip the pre-update snapshot for the immediately pending update by annotating the `Etcd` resource:
+
+```bash
+   kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/skip-next-update-snapshot=
+```
+
+This annotation is:
+
+- **One-shot**: `etcd-druid` removes it as soon as the skip is applied, so it only affects the next pending update. Any subsequent image or replica change will again trigger a full snapshot.
+- **Presence-only**: the annotation value is ignored; simply having the key present enables the skip.
+- **Scoped to image/replica updates**: it does not affect the full snapshot taken before hibernation (scaling the cluster to zero replicas).
+
+### Surfacing snapshot failures
+
+When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
+
+```bash
+   kubectl describe etcd <etcd-name> -n <namespace>
+```
+
+!!! note
+    A `PreSyncSnapshotFailed` event means the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` `EtcdOpsTask` resources and the backup configuration to find out why the snapshot failed.
--- a/internal/common/constants.go
+++ b/internal/common/constants.go
@@ -9,10 +9,10 @@ const (
 	// place an annotation on the StatefulSet pods. The value contains the check-sum of the latest configmap that
 	// should be reflected on the pods.
 	CheckSumKeyConfigMap = "checksum/etcd-configmap"
-	// KeyPreSyncSnapshotExhausted is a key set in OperatorContext.Data by the StatefulSet component's PreSync when a
+	// KeyPreSyncSnapshotFailed is a key set in OperatorContext.Data by the StatefulSet component's PreSync when a
 	// pre-sync full snapshot has exhausted its retries and etcd-druid is proceeding with the update without a fresh
 	// snapshot. The etcd controller reads this key after pre-sync to surface the failure via a warning event.
-	KeyPreSyncSnapshotExhausted = "presync-snapshot-exhausted"
+	KeyPreSyncSnapshotFailed = "presync-snapshot-failed"
 )
 
 // LeaseAnnotationKeyPeerURLTLSEnabled is the annotation key present on the member lease.
--- a/internal/common/constants.go
+++ b/internal/common/constants.go
@@ -9,6 +9,10 @@ const (
 	// place an annotation on the StatefulSet pods. The value contains the check-sum of the latest configmap that
 	// should be reflected on the pods.
 	CheckSumKeyConfigMap = "checksum/etcd-configmap"
+	// KeyPreSyncSnapshotExhausted is a key set in OperatorContext.Data by the StatefulSet component's PreSync when a
+	// pre-sync full snapshot has exhausted its retries and etcd-druid is proceeding with the update without a fresh
+	// snapshot. The etcd controller reads this key after pre-sync to surface the failure via a warning event.
+	KeyPreSyncSnapshotExhausted = "presync-snapshot-exhausted"
 )
 
 // LeaseAnnotationKeyPeerURLTLSEnabled is the annotation key present on the member lease.
--- a/internal/component/statefulset/statefulset.go
+++ b/internal/component/statefulset/statefulset.go
@@ -47,7 +47,7 @@ const (
 	// Pre-sync snapshot task constants
 	preSyncTaskPrefixHibernation = "presync-snapshot-hibernation-"
 	preSyncTaskPrefixUpdate      = "presync-snapshot-update-"
-	// maxPreSyncRetries defines the maximum number of pre-sync snapshot attempts before giving up and proceeding with the upgrade.
+	// maxPreSyncRetries defines the maximum number of pre-sync snapshot attempts before giving up and proceeding with the sync.
 	maxPreSyncRetries = 3
 )
 
--- a/internal/component/statefulset/statefulset.go
+++ b/internal/component/statefulset/statefulset.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 
 	druidapicommon "github.com/gardener/etcd-druid/api/common"
-	druidconfigv1alpha1 "github.com/gardener/etcd-druid/api/config/v1alpha1"
 	druidv1alpha1 "github.com/gardener/etcd-druid/api/core/v1alpha1"
 	"github.com/gardener/etcd-druid/internal/common"
 	"github.com/gardener/etcd-druid/internal/component"
@@ -47,7 +46,7 @@ const (
 
 	// Pre-sync snapshot task constants
 	preSyncTaskPrefixHibernation = "presync-snapshot-hibernation-"
-	preSyncTaskPrefixUpgrade     = "presync-snapshot-upgrade-"
+	preSyncTaskPrefixUpdate      = "presync-snapshot-update-"
 	// maxPreSyncRetries defines the maximum number of pre-sync snapshot attempts before giving up and proceeding with the upgrade.
 	maxPreSyncRetries = 3
 )
@@ -107,29 +106,51 @@ func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Et
 		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskPrefixHibernation)
 	}
 
-	if !druidconfigv1alpha1.DefaultFeatureGates.IsEnabled(druidconfigv1alpha1.UpgradeEtcdVersion) {
-		return nil
-	}
-
-	etcdWrapperImageFromImageVector, _, _, err := utils.GetEtcdImages(etcd, r.imageVector)
+	changed, err := r.hasImageOrReplicaChanged(etcd, existingSts)
 	if err != nil {
 		return druiderr.WrapError(err, ErrGetEtcdWrapperImage, component.OperationPreSync,
 			fmt.Sprintf("Error getting etcd images for etcd: %v", client.ObjectKeyFromObject(etcd)))
 	}
-
-	var existingWrapperImageFromSts string
-	for _, c := range existingSts.Spec.Template.Spec.Containers {
-		if c.Name == common.ContainerNameEtcd {
-			existingWrapperImageFromSts = c.Image
-			break
-		}
-	}
-
-	if existingWrapperImageFromSts != "" && etcdWrapperImageFromImageVector != existingWrapperImageFromSts {
-		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskPrefixUpgrade)
+	if changed {
+		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskPrefixUpdate)
 	}
 
 	return nil
+}
+
+// collectExpectedImages returns the expected images keyed by container name for the etcd pod template.
+func (r _resource) collectExpectedImages(etcd *druidv1alpha1.Etcd) (map[string]string, error) {
+	wrapperImg, brImg, initImg, err := utils.GetEtcdImages(etcd, r.imageVector)
+	if err != nil {
+		return nil, err
+	}
+	return map[string]string{
+		common.ContainerNameEtcd:                              wrapperImg,
+		common.ContainerNameEtcdBackupRestore:                 brImg,
+		common.InitContainerNameChangeBackupBucketPermissions: initImg,
+	}, nil
+}
+
+// hasImageOrReplicaChanged reports whether the etcd spec replicas differ from the STS, or any tracked container/init-container image differs from the expected image vector.
+func (r _resource) hasImageOrReplicaChanged(etcd *druidv1alpha1.Etcd, sts *appsv1.StatefulSet) (bool, error) {
+	if etcd.Spec.Replicas != ptr.Deref(sts.Spec.Replicas, 0) {
+		return true, nil
+	}
+	expected, err := r.collectExpectedImages(etcd)
+	if err != nil {
+		return false, err
+	}
+	for _, c := range sts.Spec.Template.Spec.InitContainers {
+		if want, ok := expected[c.Name]; ok && want != c.Image {
+			return true, nil
+		}
+	}
+	for _, c := range sts.Spec.Template.Spec.Containers {
+		if want, ok := expected[c.Name]; ok && want != c.Image {
+			return true, nil
+		}
+	}
+	return false, nil
 }
 
 // ensurePreSyncSnapshot ensures a pre-sync snapshot is taken via an etcdopstask with retry logic up to maxPreSyncRetries attempts.
--- a/internal/component/statefulset/statefulset.go
+++ b/internal/component/statefulset/statefulset.go
@@ -45,8 +45,8 @@ const (
 	ErrGetEtcdWrapperImage druidapicommon.ErrorCode = "ERR_GET_ETCD_WRAPPER_IMAGE"
 
 	// Pre-sync snapshot task constants
-	preSyncTaskPrefixHibernation = "presync-snapshot-hibernation-"
-	preSyncTaskPrefixUpdate      = "presync-snapshot-update-"
+	preSyncTaskHibernationPrefix = "presync-snapshot-hibernation-"
+	preSyncTaskUpdatePrefix      = "presync-snapshot-update-"
 	// maxPreSyncRetries defines the maximum number of pre-sync snapshot attempts before giving up and proceeding with the sync.
 	maxPreSyncRetries = 3
 )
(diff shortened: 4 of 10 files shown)


anveshreddy18:
@seshachalam-yv and I had discussed the following about this PR:

- Currently we will go ahead with just the image & replica triggers but going forward we can expand the trigger list to more known configuration that affects the cluster rolling the most. We can also explore if we can deterministically pre-compute the pod template spec in the [statefulset builder](https://github.com/gardener/etcd-druid/blob/master/internal/component/statefulset/builder.go) and compare that against the existing statefulset's pod template spec to determine if such a change will roll the pods or not and decide to take a full snapshot through opsTask if it does roll.

- We need a way to skip the fullsnapshot retries if there's a known issue affecting the snapshots or if there is a requirement to get the changes rolled fast to the pods, etc to not block the update process always. So provisioning a way would to skip would be good functionally. We agreed on using annotation as a way to skip as at this point we don't want to provision a way to enable/disable through the API until we are confident that this is something people are interested in configuring.

- Also if the current `N=3` retries fail due to any issue, the etcd controller silently proceeds to the `Sync` method to update the sts, and the failed retries are ignored. We need a way to propagate this failure to the Etcd CR so that the issues plaguing full snapshot retries is known to the operators and can be looked into before it cascades into a bigger issue. We can make them as part of the `LastErrors` so that it is presented somewhere to be seen.

anveshreddy18:
> @seshachalam-yv and I had discussed the following about this PR:
> 
> * Currently we will go ahead with just the image & replica triggers but going forward we can expand the trigger list to more known configuration that affects the cluster rolling the most. We can also explore if we can deterministically pre-compute the pod template spec in the [statefulset builder](https://github.com/gardener/etcd-druid/blob/master/internal/component/statefulset/builder.go) and compare that against the existing statefulset's pod template spec to determine if such a change will roll the pods or not and decide to take a full snapshot through opsTask if it does roll.
> * We need a way to skip the fullsnapshot retries if there's a known issue affecting the snapshots or if there is a requirement to get the changes rolled fast to the pods, etc to not block the update process always. So provisioning a way would to skip would be good functionally. We agreed on using annotation as a way to skip as at this point we don't want to provision a way to enable/disable through the API until we are confident that this is something people are interested in configuring.
> * Also if the current `N=3` retries fail due to any issue, the etcd controller silently proceeds to the `Sync` method to update the sts, and the failed retries are ignored. We need a way to propagate this failure to the Etcd CR so that the issues plaguing full snapshot retries is known to the operators and can be looked into before it cascades into a bigger issue. We can make them as part of the `LastErrors` so that it is presented somewhere to be seen.

@seshachalam-yv I've now updated the PR with the following:

- first one we've concluded previously that we'll go with just the image & replica triggers for now, so have not touched this part.
- I've now introduced an annotation as decided previously to skip the full snapshot incase of upgrades.
- To report the failures in taking full snapshot, I'm now generating k8s event object for that instead of propagating it in `LastErrors` as the former is easier to monitor than the lastErrors which is ephemeral and practically only stays for a few seconds to mins due to the limit on the buffer size we set for how many errors we retain.

api/core/v1alpha1/constants.go
```diff
@@ -39,6 +39,12 @@ const (
 	// This value will only be effective if etcd-druid is not configured with auto-reconciliation of Etcd resource specification via
 	// --enable-etcd-spec-auto-reconcile CLI flag.
 	DruidOperationReconcile = "reconcile"
+	// SkipNextUpdateSnapshotAnnotation, when present on an Etcd resource, instructs etcd-druid to skip the pre-sync
+	// full snapshot for the immediately pending StatefulSet update triggered by a container image or replica change.
+	// It's a one-shot annotation: etcd-druid removes it as soon as the skip is applied, so it only affects the next
+	// immediate pending update. Its presence alone is sufficient; the annotation value is ignored. It does not affect the
+	// snapshot taken before hibernation.
+	SkipNextUpdateSnapshotAnnotation = "druid.gardener.cloud/skip-next-update-snapshot"
```
seshachalam-yv:
```suggestion
	SkipNextUpdateSnapshotAnnotation = "druid.gardener.cloud/skip-pre-update-snapshot"
```

The name `skip-next-update-snapshot` is a bit ambiguous — does "next" mean just the next one (one-shot) or all upcoming snapshots? Since this is strictly one-shot, the name should make that clear upfront.

docs/usage/managing-etcd-clusters.md
```diff
@@ -184,3 +184,32 @@ Now you are free to make changes to any managed etcd cluster resource.
...
+### Surfacing snapshot failures
+
+When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
+
+```bash
+   kubectl describe etcd <etcd-name> -n <namespace>
+```
+
+!!! note
+    A `PreSyncSnapshotFailed` event means the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` `EtcdOpsTask` resources and the backup configuration to find out why the snapshot failed.
```
seshachalam-yv: 1. Section placement — the new section sits outside the existing ## Update & Reconcile an Etcd Cluster hierarchy but describes behaviour that's part of update/reconcile. It should be a ### subsection under that ##, not a new top-level ##.

seshachalam-yv: This is wrong convention, refer https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

seshachalam-yv (changes requested):

internal/component/statefulset/statefulset.go
```diff
@@ -107,29 +109,67 @@ func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Et
...
+	if changed {
+		if druidv1alpha1.HasSkipNextUpdateSnapshotAnnotation(etcd.ObjectMeta) {
+			r.logger.Info("Skipping pre-sync snapshot for pending update due to presence of annotation", "annotation", druidv1alpha1.SkipNextUpdateSnapshotAnnotation)
+			if err := r.removeSkipNextUpdateSnapshotAnnotation(ctx, etcd); err != nil {
+				return druiderr.WrapError(err, ErrRemoveSkipAnnotation, component.OperationPreSync,
+					fmt.Sprintf("Failed to remove %s annotation for etcd: %v", druidv1alpha1.SkipNextUpdateSnapshotAnnotation, client.ObjectKeyFromObject(etcd)))
+			}
+			return nil
+		}
+		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskPrefixUpdate)
```
seshachalam-yv, line 125: The check order feels inverted here. The annotation is an operator-level override — if it's present, there's no need to call hasImageOrReplicaChanged at all. Checking change detection first does unnecessary work.

Prefer moving these checks inside ensurePreSyncSnapshot

api/core/v1alpha1/constants.go
anveshreddy18: Yes this is strictly one-shot as we want to skip the snapshot for that specific upgrade the user is opting to skip for, instead of skip the snapshot for all future upgrades.

anveshreddy18: Putting out the response here to add more context on the reason for it.

> Why someone adds annotation to skip next snapshot only ?

So usually someone would want to skip snapshot if they're doing something urgently, like they're stuck with a bad version and they want to upgrade to a new version immediately, so they can choose to skip taking the full snapshot to speed up the process.

Now adding the annotation to skip all further snapshots is not required because by default we want them to take full snapshot and if they have a problem with a specific upgrade, then they can add this annotation to skip taking the snapshot for that specific upgrade.

internal/component/statefulset/statefulset.go
```diff
@@ -107,29 +109,67 @@ func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Et
...
-	if existingWrapperImageFromSts != "" && etcdWrapperImageFromImageVector != existingWrapperImageFromSts {
-		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskPrefixUpgrade)
+	for _, c := range sts.Spec.Template.Spec.Containers {
+		if want, ok := expected[c.Name]; ok && want != c.Image {
+			return true, nil
+		}
 	}
-
-	return nil
+	return false, nil
```
seshachalam-yv, line 159:
```suggestion
// hasImageOrReplicaChanged reports whether the StatefulSet differs from the
// expected etcd configuration in replica count or tracked container images.
func (r _resource) hasImageOrReplicaChanged(
	etcd *druidv1alpha1.Etcd,
	sts *appsv1.StatefulSet,
) (bool, error) {
	if etcd.Spec.Replicas != ptr.Deref(sts.Spec.Replicas, 0) {
		return true, nil
	}

	expectedImages, err := r.collectExpectedImages(etcd)
	if err != nil {
		return false, err
	}

	if imagesChanged(sts.Spec.Template.Spec.InitContainers, expectedImages) {
		return true, nil
	}

	if imagesChanged(sts.Spec.Template.Spec.Containers, expectedImages) {
		return true, nil
	}

	return false, nil
}

func imagesChanged(
	containers []corev1.Container,
	expectedImages map[string]string,
) bool {
	for _, container := range containers {
		expectedImage, ok := expectedImages[container.Name]
		if ok && expectedImage != container.Image {
			return true
		}
	}

	return false
}
```

seshachalam-yv (changes requested):

api/core/v1alpha1/constants.go
anveshreddy18: @seshachalam-yv 
After some thought, I went ahead with your suggestion of having the annotation always set, and the user has to remove it themselves because "once" was getting a bit confusing if the update doesn't have any image/replica change, then it was not very clear whether to remove the annotation or not in that case. So I chose to make it simpler.

internal/component/statefulset/statefulset.go
anveshreddy18: Sure, I've now reversed the checks. PTAL

internal/component/statefulset/statefulset.go
anveshreddy18: I've simplified it a bit now, PTAL. Thanks

internal/component/statefulset/statefulset.go
```diff
@@ -107,29 +107,64 @@ func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Et
 		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskPrefixHibernation)
 	}
```
seshachalam-yv, line 111: We are already checking whether `etcd.Spec.Replicas == 0`.
Isn't this a duplicate check, since we are already checking it in `hasImageOrReplicaChanged`?

seshachalam-yv (changes requested):

internal/component/statefulset/statefulset.go
anveshreddy18: This particular check `if etcd.Spec.Replicas == 0` is for Hibernation only. Right now we do not have a native way in the ETCD CR to determine if a cluster is being hibernated or not and this is the only way until what Priya [is working on](https://github.com/gardener/etcd-druid/issues/922) comes in, then we'll adjust this to use that.

I'll mention this as a comment in the code to not forget about it later.

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+
+#### Surfacing snapshot failures
+
+When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
+
+```bash
+   kubectl describe etcd <etcd-name> -n <namespace>
+```
+
+!!! note
```
seshachalam-yv: syntax error 
[!Note]

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
 
 This option is sometimes recommeded as you would like avoid auto-reconciliation of accidental changes to `Etcd` resources outside the maintenance time window, thus preventing a potential transient quorum loss due to misconfiguration, attach-detach issues of persistent volumes etc.
 
+### Full snapshot before a StatefulSet update
+
+Before `etcd-druid` rolls the etcd `StatefulSet` due to a container image change or a replica count change (in both HA and non-HA setups), it triggers a full snapshot via an `EtcdOpsTask`. This ensures there is a recent, safe state to restore from should the rollout run into data corruption or data loss. The snapshot is attempted a bounded number of times; if all attempts fail, `etcd-druid` proceeds with the update regardless (so that a persistently failing snapshot does not block updates indefinitely).
```
seshachalam-yv, line 102:
```suggestion
Before `etcd-druid` rolls the etcd `StatefulSet` due to a container image or replica count change, it triggers a full snapshot via an [`EtcdOpsTask`](using-etcdopstask.md). This provides a recent recovery point in case the rollout results in data corruption or data loss. This behavior applies to both HA and non-HA setups.

The snapshot is attempted a bounded number of times. If all attempts fail, `etcd-druid` proceeds with the update without a fresh snapshot, so that a persistently failing snapshot does not block updates indefinitely.
```

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+
+#### Skip the snapshot before updates
+
+If snapshots are known to be failing, or you need changes to roll out quickly without waiting for a snapshot, you can instruct `etcd-druid` to skip the pre-update snapshot by annotating the `Etcd` resource with `druid.gardener.cloud/skip-spec-update-snapshot`:
+
+```bash
+   kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/skip-spec-update-snapshot=
```
seshachalam-yv, line 109:
```suggestion
kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/skip-spec-update-snapshot=
```

internal/component/statefulset/statefulset.go
seshachalam-yv: Do you want the annotation to be honoured in this case also ?

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+#### Surfacing snapshot failures
+
+When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
+
+```bash
+   kubectl describe etcd <etcd-name> -n <namespace>
+```
+
+!!! note
+    A `PreSyncSnapshotFailed` event means the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` `EtcdOpsTask` resources and the backup configuration to find out why the snapshot failed.
```
seshachalam-yv, line 127:
```suggestion
> [!NOTE]
> A `PreSyncSnapshotFailed` event means that the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` [`EtcdOpsTask`](using-etcdopstask.md) resources and the backup configuration to determine why the snapshot failed.
```

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+
+This annotation is:
+
+- **Presence-only**: the annotation value is ignored; simply having the key present enables the skip.
+- **Persistent**: while the annotation is present, the pre-update snapshot is always skipped. `etcd-druid` does not remove it — remove the annotation yourself to re-enable pre-update snapshots.
+- **Scoped to image/replica updates**: it does not affect the full snapshot taken before hibernation (scaling the cluster to zero replicas).
+
+#### Surfacing snapshot failures
+
+When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
```
seshachalam-yv, line 120:
```suggestion
If the pre-update snapshot exhausts its retries, `etcd-druid` proceeds with the update without a fresh snapshot and records a `Warning` event with the reason `PreSyncSnapshotFailed` on the `Etcd` resource. You can view the event using:
```

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+
+```bash
+   kubectl describe etcd <etcd-name> -n <namespace>
```
seshachalam-yv, line 123:
```suggestion
kubectl describe etcd <etcd-name> -n <namespace>
```

internal/component/statefulset/statefulset.go
anveshreddy18: No, Hibernation should always take a full snapshot. If someone is hibernating the cluster, then they don't usually don't care about hibernating it faster, so I don't think there is a reason to gate this with annotation.

seshachalam-yv (changes requested):
Can you add the e2e test to verify this ?

internal/component/statefulset/statefulset.go
seshachalam-yv: I think we should consider providing the same skip mechanism for hibernation as well for consistency.

If the pre-update snapshot is failing and we provide an annotation to allow image/replica updates to proceed without taking the snapshot, the same snapshot failure could also prevent a user from hibernating the cluster. In that situation, the user should have a consistent way to explicitly skip the snapshot when hibernating as well.

seshachalam-yv: I agree with the argument for providing a skip mechanism, but with the current behavior, aren't we already proceeding when the snapshot fails after the maximum number of attempts?

For example, we record `PreSyncSnapshotFailed` and continue with the StatefulSet update without a fresh snapshot:

```go
r.recorder.Eventf(etcd, corev1.EventTypeWarning, "PreSyncSnapshotFailed",
    "pre-sync full snapshot for %s/%s failed after max attempts; proceeding with StatefulSet update without a fresh snapshot",
)
```

If this is the intended behavior, then the same consideration should apply to hibernation as well. Otherwise, we would allow an update to proceed after a failed snapshot but potentially block hibernation on the same snapshot failure.

Would it make more sense to have a consistent behavior for both paths—i.e. after exhausting the snapshot retries, proceed with the operation and surface a `Warning` event—rather than introducing a skip annotation only for the StatefulSet update path?

api/core/v1alpha1/constants.go
```diff
@@ -39,6 +39,11 @@ const (
 	// This value will only be effective if etcd-druid is not configured with auto-reconciliation of Etcd resource specification via
 	// --enable-etcd-spec-auto-reconcile CLI flag.
 	DruidOperationReconcile = "reconcile"
+	// SkipSpecUpdateSnapshotAnnotation is an annotation when present on an Etcd resource, instructs the etcd-druid to skip the pre-sync full
+	// snapshot which would otherwise be taken before the etcd StatefulSet is rolled due to a container image or replica change.
+	// Its presence alone is sufficient; the annotation value is ignored. While the annotation is present the snapshot is always skipped, and etcd-druid does not remove the annotation.
+	// It does not affect the full snapshot taken before hibernation.
+	SkipSpecUpdateSnapshotAnnotation = "druid.gardener.cloud/skip-spec-update-snapshot"
```
Shreyas-s14, line 46: Do you think this annotation should be removed once the statefulset is rolled? Or is it upon the operator to do so.

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
 
 This option is sometimes recommeded as you would like avoid auto-reconciliation of accidental changes to `Etcd` resources outside the maintenance time window, thus preventing a potential transient quorum loss due to misconfiguration, attach-detach issues of persistent volumes etc.
 
+### Full snapshot before a StatefulSet update
+
+Before `etcd-druid` rolls the etcd `StatefulSet` due to a container image change or a replica count change (in both HA and non-HA setups), it triggers a full snapshot via an `EtcdOpsTask`. This ensures there is a recent, safe state to restore from should the rollout run into data corruption or data loss. The snapshot is attempted a bounded number of times; if all attempts fail, `etcd-druid` proceeds with the update regardless (so that a persistently failing snapshot does not block updates indefinitely).
+
+#### Skip the snapshot before updates
+
+If snapshots are known to be failing, or you need changes to roll out quickly without waiting for a snapshot, you can instruct `etcd-druid` to skip the pre-update snapshot by annotating the `Etcd` resource with `druid.gardener.cloud/skip-spec-update-snapshot`:
```
Shreyas-s14, line 106: Not a change for this PR, but in future could we leverage the `druidctl` feature you introduced for all this? Feel free to resolve this comment since it is not a change.

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+
+```bash
+   kubectl annotate etcd <etcd-name> -n <namespace> druid.gardener.cloud/skip-spec-update-snapshot=
+```
+
+This annotation is:
+
+- **Presence-only**: the annotation value is ignored; simply having the key present enables the skip.
+- **Persistent**: while the annotation is present, the pre-update snapshot is always skipped. `etcd-druid` does not remove it — remove the annotation yourself to re-enable pre-update snapshots.
+- **Scoped to image/replica updates**: it does not affect the full snapshot taken before hibernation (scaling the cluster to zero replicas).
```
Shreyas-s14, line 116: Should the doc also mention that the regular scheduled and out of band full snapshots will still be functional even if this annotation is present?

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +97,35 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+
+#### Surfacing snapshot failures
+
+When the pre-update snapshot exhausts its retries and `etcd-druid` proceeds with the update without a fresh snapshot, it records a `Warning` event with reason `PreSyncSnapshotFailed` on the `Etcd` resource so operators are aware. You can see it via:
+
+```bash
+   kubectl describe etcd <etcd-name> -n <namespace>
+```
```
Shreyas-s14, line 124: The status.LastErrors intention was to store temporary errors right? Is the reason to go with this is because this situation leads to an erroneous etcd wherein the LastErrors will be continuously re written?

internal/component/statefulset/statefulset.go
```diff
@@ -103,33 +103,71 @@ func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Et
...
+	// TODO: currently replicas == 0 is used as a proxy for "the cluster is being hibernated". Once native
+	// hibernation support [gardener/etcd-druid#922](https://github.com/gardener/etcd-druid/issues/922) is implemented,
+	// we need to switch to the dedicated hibernation signal on the Etcd resource instead of inferring it from the replica count.
 	if etcd.Spec.Replicas == 0 {
 		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskPrefixHibernation)
 	}
 
-	if !druidconfigv1alpha1.DefaultFeatureGates.IsEnabled(druidconfigv1alpha1.UpgradeEtcdVersion) {
+	if druidv1alpha1.HasSkipSpecUpdateSnapshotAnnotation(etcd.ObjectMeta) {
+		r.logger.Info("Skipping pre-sync snapshot for update due to presence of annotation",
```
Shreyas-s14, line 114: Maybe I am missing some context, but could you explain why this is skipped for hibernation case? There can be a scenario where snapshot halts in that case as well right?

internal/component/statefulset/statefulset_test.go
```diff
@@ -119,166 +123,199 @@ func TestPreSync(t *testing.T) {
...
-			expectedErrCode:    ptr.To(druidapicommon.ErrorCode(druiderr.ErrRequeueAfter)),
+			name:          "hibernation prefix wins when replicas go to 0 even if images also changed",
+			backupEnabled: true,
+			stsExists:     true,
+			stsReplicas:   3,
+			etcdReplicas:  0,
+			stsImages:     map[string]string{common.ContainerNameEtcd: oldWrapperImage},
+			existingTasks: []*druidv1alpha1.EtcdOpsTask{buildPreSyncTask(preSyncTaskPrefixHibernation, 0, ptr.To(druidv1alpha1.TaskStateSucceeded))},
+		},
+		// ---------------- Skip-annotation rows ----------------
```
Shreyas-s14, line 259: I feel this segregation based on comments is not needed. This is not followed in other test suites also. WDYT

test/it/controller/etcd/reconciler_test.go
```diff
@@ -585,6 +585,201 @@ func testPreSyncHibernationProceedsAfterMaxRetries(t *testing.T, testNs string,
 	t.Log("StatefulSet replicas successfully scaled to 0 after max retries exceeded")
 }
 
+// ------------------------------ presync image change tests ------------------------------
```
Shreyas-s14, line 588: same for this set of comments. Feels unnecessary imo

Shreyas-s14 (commented):
Thanks for the PR. Have a few queries, PTAL!

api/core/v1alpha1/constants.go
anveshreddy18: > Do you think this annotation should be removed once the statefulset is rolled?

That was my initial proposal as well but Sesha's view was that it's upto the user what they want to do, that we should not meddle and assume something, this was discussed in the  [comment thread](https://github.com/gardener/etcd-druid/pull/1360#discussion_r3913132429) if you want to take a look.

docs/usage/managing-etcd-clusters.md
anveshreddy18: Yeah sure definitely, that's a good suggestion.

docs/usage/managing-etcd-clusters.md
anveshreddy18: done

docs/usage/managing-etcd-clusters.md
anveshreddy18: yes I initially thought of using status.LastErrors but as you rightly pointed out that it gets continuously rewritten and gets lost in the sea, if in future we want to monitor the update related snapshotting, then having the errors surface as events is more appropriate than having them in the lastErrors, for that reason I went with this approach.

internal/component/statefulset/statefulset.go
anveshreddy18: thanks for the point, the same is being discussed in the [comment thread](https://github.com/gardener/etcd-druid/pull/1360#discussion_r3923669245), pls feel free to add to that. We can shift the discussion there.

internal/component/statefulset/statefulset_test.go
anveshreddy18: removed at all places.

internal/component/statefulset/statefulset.go
anveshreddy18: > Otherwise, we would allow an update to proceed after a failed snapshot but potentially block hibernation on the same snapshot failure.

That's not the case, we do not block hibernation at all, if the 3 retries fail then we surface the failure event and proceed with the `Sync` method.


> have a consistent behavior for both paths—i.e. after exhausting the snapshot retries, proceed with the operation and surface a Warning event

The same func `ensurePreSyncSnapshot` is called in both the hibernation and the update case,  and in both the cases there is a retry-mechanism of 3 retries, if they all fail then the func surfaces the failure event and returns `nil` so that the reconciler goes to the `Sync` method next. So the behaviour is already consistent between them, correct?

anveshreddy18: The only real point is whether we should make the annotation skip triggering a snapshot before Hibernation.

As a refresher, the main reason why we even added the skip annotation for the updates is because sometimes the retries can take time which might make the update process slower, if an urgent update has to happen then an operator can set this skip annotation to proceed fast. Considering this is the only reason, few other points below:

- Taking a full snapshot before hibernating have huge benefits and something we have also discussed multiple times in the past about the need for it [ref](https://github.com/gardener/etcd-druid/issues/922), so imo we should try our best to take a full snapshot anytime before hibernating so that we have the latest snapshot of the database before we tear down the cluster.

- When someone is hibernating, it means they don't need the cluster running anymore - so it doesn't really matter if it takes a few more mins in taking the snapshot, because they it is not blocking any important operation.


So imo I don't think there is a need to skip snapshot while hibernating because of the reasons mentioned above. I see the only reason being mentioned for the ask to skip snapshot is about consistency concerns with the update snapshot process, but imo these two are different processes and should be treated differently. Is there any other reason why we're trying to club hibernation to that of a regular update process?

anveshreddy18:
> Can you add the e2e test to verify this ?

@seshachalam-yv I've thought about this but the functionality is already covered in IT tests, and we have unit tests for the individual functions. In addition, the #1300 PR which introduced the snapshot during upgrade also didn't add e2e tests for the same reason.

Since we have IT tests for this & e2e tests for the OpsTask already ( which is the actual functionality being used here, so the e2e tests for this are the actually needed ones and we have them already so nothing new is needed ). 

The IT tests will cover all the logical cases of when to trigger and when not to trigger a full snapshot, and it is rightly validated. Now whether a full snapshot is actually taken or not is the responsibility of OpsTask, which is already well tested. 

For these reasons, I'd like to avoid duplicating the testing in both the IT & e2e tests. But if you strongly feel that we need to have e2e tests, maybe that can be done separately outside of this PR to avoid any scope creep for this PR. Hope that's fine.

internal/component/statefulset/statefulset.go
seshachalam-yv: I still have a concern regarding the symmetry here, as mentioned in my earlier comment

However, I see the reasoning, and since the other reviewer is also fine with this approach, I'm okay with proceeding.

seshachalam-yv (approved):
Overall, looks good to me, apart from this comment: https://github.com/gardener/etcd-druid/pull/1360#discussion_r3946866238

I'll leave this to the other reviewer to decide. I'm fine with proceeding either way.

/lgtm

internal/component/statefulset/statefulset.go
Shreyas-s14: I too agree with the current approach for hibernation case. Until we have the official support path, this is the "best effort" scenario we can take.

Shreyas-s14:
> > Can you add the e2e test to verify this ?
> 
> @seshachalam-yv I've thought about this but the functionality is already covered in IT tests, and we have unit tests for the individual functions. In addition, the #1300 PR which introduced the snapshot during upgrade also didn't add e2e tests for the same reason.
> 
> Since we have IT tests for this & e2e tests for the OpsTask already ( which is the actual functionality being used here, so the e2e tests for this are the actually needed ones and we have them already so nothing new is needed ).
> 
> The IT tests will cover all the logical cases of when to trigger and when not to trigger a full snapshot, and it is rightly validated. Now whether a full snapshot is actually taken or not is the responsibility of OpsTask, which is already well tested.
> 
> For these reasons, I'd like to avoid duplicating the testing in both the IT & e2e tests. But if you strongly feel that we need to have e2e tests, maybe that can be done separately outside of this PR to avoid any scope creep for this PR. Hope that's fine.

That was the reason the e2e tests weren't done in the initial PR as well. Unless @seshachalam-yv feels it's a requirement to add the e2e tests here, I feel the IT tests here cover all scenarios. Anyways the only way to confirm whether a snapshot is taken in this path is to monitor a snapshot lease which is already covered as a part of the EtcdOpsTask test case.

api/core/v1alpha1/constants.go
Shreyas-s14: Thanks for pointing it out.

docs/usage/managing-etcd-clusters.md
Shreyas-s14: Is there an issue thread for the druidctl component? So that when we implement new commands, we know what to do

anveshreddy18: yes, there is. I've now added this to the druidctl issue tracker here: https://github.com/gardener/etcd-druid/issues/1171 in the ad-hoc section, that can be implemented if the team is okay with it whenever someone picks it up.

Shreyas-s14 (approved):
Thank you for addressing the review comments. LGTM

api/core/v1alpha1/constants.go
shreyas-s-rao: I believe this is annotation sets a `behavior` on the Etcd rather than a one-time operation intent, so the user is telling druid "upon any spec update on this Etcd, please skip the full snapshot", so druid should not be removing this annotation by itself.

shreyas-s-rao: Also, since this affects the general behavior of the Etcd cluster, why was the decision taken to keep this as an annotation and not a spec field? Any reason?

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +101,38 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+This annotation is:
+
+* **Presence-only**: the annotation value is ignored; simply having the key present enables the skip.
+* **Persistent**: while the annotation is present, the pre-update snapshot is always skipped. `etcd-druid` does not remove it — remove the annotation yourself to re-enable pre-update snapshots.
+* **Scoped to image/replica updates**: it does not affect the full snapshot taken before hibernation (scaling the cluster to zero replicas).
+* **Limited to the pre-update snapshot**: the regular scheduled full and delta snapshots, as well as out-of-band snapshots, continue to be taken as usual while the annotation is present.
+
+#### Surfacing snapshot failures
+
+If the pre-update snapshot exhausts its retries, `etcd-druid` proceeds with the update without a fresh snapshot and records a `Warning` event with the reason `PreSyncSnapshotFailed` on the `Etcd` resource. You can view the event using:
```
shreyas-s-rao, line 127: Nice touch.

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +101,38 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+
+```bash
+kubectl describe etcd <etcd-name> -n <namespace>
+```
+
+> [!NOTE]
+> A `PreSyncSnapshotFailed` event means that the update proceeded without a fresh snapshot. Inspect the `presync-snapshot-update-*` [`EtcdOpsTask`](using-etcdopstask.md) resources and the backup configuration to determine why the snapshot failed.
```
shreyas-s-rao, line 134: is it possible that druid creates multiple of these ops-task resources and they just get orphaned later?

internal/common/constants.go
```diff
@@ -9,6 +9,10 @@ const (
 	// place an annotation on the StatefulSet pods. The value contains the check-sum of the latest configmap that
 	// should be reflected on the pods.
 	CheckSumKeyConfigMap = "checksum/etcd-configmap"
+	// KeyPreSyncSnapshotExhausted is a key set in OperatorContext.Data by the StatefulSet component's PreSync when a
+	// pre-sync full snapshot has exhausted its retries and etcd-druid is proceeding with the update without a fresh
+	// snapshot. The etcd controller reads this key after pre-sync to surface the failure via a warning event.
+	KeyPreSyncSnapshotExhausted = "presync-snapshot-exhausted"
```
shreyas-s-rao: I see an issue with naming consistency. Here you call it `KeyPreSyncSnapshotExhausted`, while the event type becomes `PreSyncSnapshotFailed`, while the annotation to skip this pre-sync snapshot is called `skip-spec-update-snapshot` (varies between `sync` and `update`). I feel there's a degree of naming inconsistency, and will end up creating confusion later. Can you please suggest alternatives for this?

internal/component/statefulset/statefulset.go
```diff
@@ -47,8 +46,8 @@ const (
 
 	// Pre-sync snapshot task constants
 	preSyncTaskPrefixHibernation = "presync-snapshot-hibernation-"
-	preSyncTaskPrefixUpgrade     = "presync-snapshot-upgrade-"
-	// maxPreSyncRetries defines the maximum number of pre-sync snapshot attempts before giving up and proceeding with the upgrade.
+	preSyncTaskPrefixUpdate      = "presync-snapshot-update-"
```
shreyas-s-rao: generally, it's easier to name the constants `prefixPreSyncTaskHibernation` and `prefixPreSyncTaskUpdate` so that the prefix of the constant name is as common as possible, or else another approach is `preSyncTaskHibernationPrefix` and `preSyncTaskUpdatePrefix` to keep the suffix and prefix common to a large degree. Please consider renaming these accordingly.

internal/component/statefulset/statefulset.go
```diff
@@ -103,33 +103,71 @@ func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Et
...
+	if err != nil {
+		return false, err
+	}
+	if imagesChanged(sts.Spec.Template.Spec.InitContainers, expectedImages) {
+		return true, nil
+	}
+	if imagesChanged(sts.Spec.Template.Spec.Containers, expectedImages) {
+		return true, nil
 	}
+	return false, nil
```
shreyas-s-rao, line 159: This logic assumes that no actor in the cluster mutates the Statefulset image itself during sts creation/updation, right? If at all such a mutation occurs, then this logic will consider every statefulset component sync as an image change and will end up running the full snapshot ops-task. I'm wondering if there's a way to safeguard against this.

internal/component/statefulset/statefulset.go
```diff
@@ -103,33 +103,71 @@ func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Et
...
+		common.ContainerNameEtcdBackupRestore:                 brImg,
+		common.InitContainerNameChangeBackupBucketPermissions: initImg,
+	}, nil
+}
 
-	if existingWrapperImageFromSts != "" && etcdWrapperImageFromImageVector != existingWrapperImageFromSts {
-		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskPrefixUpgrade)
+// hasImageOrReplicaChanged reports whether the etcd spec replicas differ from the STS, or any tracked container/init-container image differs from the expected image vector.
+func (r _resource) hasImageOrReplicaChanged(etcd *druidv1alpha1.Etcd, sts *appsv1.StatefulSet) (bool, error) {
+	if etcd.Spec.Replicas != ptr.Deref(sts.Spec.Replicas, 0) {
```
shreyas-s-rao, line 146: Similar comment as above for this as well. AFAIK for the self-hosted shoot usecase in gardener, the sts replicas is set to `0` while keeping everything else intact, so that the config is generated for actual replica count but the pods are not created (or this was the behavior before). @CaptainIRS please correct me if I'm wrong.

So how would this logic work in such cases?

internal/controller/etcd/reconcile_spec.go
```diff
@@ -68,6 +69,13 @@ func (r *Reconciler) preSyncEtcdResources(ctx component.OperatorContext, etcd *d
 			return ctrlutils.ReconcileWithError(err)
 		}
 	}
+	// If a pre-sync snapshot exhausted its retries, the component proceeded with the update without a fresh snapshot.
+	// Surface this via a warning event so operators are aware the update rolled without a safety snapshot.
+	if _, exhausted := ctx.Data[common.KeyPreSyncSnapshotExhausted]; exhausted {
+		r.recorder.Eventf(etcd, corev1.EventTypeWarning, "PreSyncSnapshotFailed",
```
shreyas-s-rao: This is an example of the naming inconsistency I was talking about earlier. We already have two words here `exhausted` and `failed`, so why not have a common term for it?

test/it/controller/etcd/reconciler_test.go
```diff
@@ -585,6 +585,201 @@ func testPreSyncHibernationProceedsAfterMaxRetries(t *testing.T, testNs string,
...
+	// etcd-druid must NOT remove the skip annotation; it persists on the resource.
+	g.Consistently(func() bool {
+		updated := &druidv1alpha1.Etcd{}
+		if err := cl.Get(ctx, client.ObjectKeyFromObject(etcdInstance), updated); err != nil {
+			return false
+		}
+		return druidv1alpha1.HasSkipSpecUpdateSnapshotAnnotation(updated.ObjectMeta)
+	}).Within(consistentlyDuration).WithPolling(pollingInterval).Should(BeTrue(), "expected skip-spec-update-snapshot annotation to be retained by druid")
+	t.Log("skip-spec-update-snapshot annotation retained by druid (persistent)")
+}
```
shreyas-s-rao: Why do we have this in the test logic? I'm assuming the previous code had druid removing the annotation, and once the annotation-removal logic was removed, this test was added to safeguard against it, which seems unnecessary. Can you please remove this, or justify why this should be there at all?

shreyas-s-rao (changes requested):
@anveshreddy18 thanks for your PR. I have a few comments, PTAL thanks.

api/core/v1alpha1/constants.go
anveshreddy18: @shreyas-s-rao 

> I believe this is annotation sets a behavior on the Etcd rather than a one-time operation intent

Yes, it depends on the name of the annotation imo, rn the name reflects that it should skip the snapshot for next update, and has to be removed by the user. Until then it continues to honour the annotation.

I've previously used a different name that reflected that it's a one-time thing. So the contention was actually on out of these two behaviours what do we want to offer for the end user for their convenience, and managing it. I was of the opinion that we can name it appropriately and offer it as a one-time skip, where druid removes the annotation after that. And Sesha's view was that it should be left to the user. We eventually agreed on the latter to not complicate things.

> Also, since this affects the general behavior of the Etcd cluster, why was the decision taken to keep this as an annotation and not a spec field? Any reason?

We have discussed this previously with Madhav and he was against making any API changes which are not justified to be API changes i.e unless we are confident that it is something people are interested in configuring, to reduce the no.of incompatible changes when we plan to promote the API to beta. So the alternative was to go with the annotation way.

docs/usage/managing-etcd-clusters.md
anveshreddy18: It creates upto 3 ops-task for snapshotting, and each of them have a TTL of 1hr, so they get auto cleaned after that. This should be fine.

internal/component/statefulset/statefulset.go
anveshreddy18: Nice point, but that shouldn't be a problem. Here's why:

- Users don't mutate the image accidentally as we have the component protection webhook to deny such changes.

- If they do change it deliberately by disabling the webhook
   - If they don't run the reconciliation, then this doesn't get triggered ( which is most of the times because they're trying to observe something with a different image and reconciliation would bring the sts back in sync with Etcd ) 
   - If they run a reconciliation, then yes the first time it does trigger a full snapshot but that's only once and after that once the sts gets in sync, the next loops don't trigger any unless someone keeps on changing the sts image coupled with reconciliation trigger which imo points to something being wrong with the operator ( human or machine alike 😅 )
   
In overall, it's not frequent to change the image in sts, and even then there's only one full snapshot, which imo is okay to take since we're messing with the cluster, so a fresh backup wouldn't hurt )

internal/component/statefulset/statefulset.go
anveshreddy18: We have a logic here 

https://github.com/anveshreddy18/etcd-druid/blob/86f47b318ded97c8b075baadca75c15d280da7d6/internal/component/statefulset/statefulset.go#L102 

that checks whether there's an sts, and if it's replicas are 0 ... in both these cases the presync is a no-op, so it wouldn't come to this stage.

internal/common/constants.go
anveshreddy18: I get your concern, the reason for using `exhausted` in the internal code is because we have 3 retries, so once these retries are "exhaused", we surface the warning as a `failure`. But you're right these two don't have to be different.

test/it/controller/etcd/reconciler_test.go
anveshreddy18: Maybe it's the comment that looks a bit off but this basically ensures that the annotation is not removed by druid after the presync snapshot process. Because that's one of the important decisions we took to keep the annotation persistent, this part just checks that.

But I have no strong opinion if you strongly suggest to remove this.

internal/controller/etcd/reconcile_spec.go
anveshreddy18: done, changed all variable namings from from `exhausted` to `failed` or similar.

internal/component/statefulset/statefulset.go
anveshreddy18: Thanks, updated. PTAL

internal/common/constants.go
anveshreddy18: I've now pushed the changes, PTAL.

api/core/v1alpha1/constants.go
shreyas-s-rao: Understood, thanks for the context.

docs/usage/managing-etcd-clusters.md
shreyas-s-rao: Ok, but is this logic part of the PreSync step or the ops-task controller itself? Just want to confirm.

internal/component/statefulset/statefulset.go
shreyas-s-rao: > Users don't mutate the image accidentally as we have the component protection webhook to deny such changes

Mutation is not guarded by the component protection webhook, since mutation is an in-flight update, not a regular `UPDATE` action/verb on the resource. So it is definitely possible to mutate druid-managed resources without triggering the webhook.

So your next points about reconciliation are also nullified, since any reconciliation, irrespective of whether the webhook is enabled or disabled, will trigger a sync of the sts, upon which the mutation is (or can be) applied.

I'll give you an example: a user creates mutating webhook or a MutatingAdmissionPolicy to update the image to custom sanitized images (ex: gardener CI images), which may not be applied on the pods, but the statefulset itself. While currently the approach is to use the imagevector approach (so that druid applies the specified images as part of the sync itself), this may not always be the case in the future, so I'm worried whether we are safeguarded against this. If at all the behavior changes, or any user of druid chooses to use mutating webhook or MAP to update the component images in-flight, then this logic will fail. So either the logic needs to handle this, or else a warning to users (somewhere in the `usage` docs) documenting this nuance that users should not mutate the component images, and should rather use the Etcd spec to specify any custom images.

test/it/controller/etcd/reconciler_test.go
shreyas-s-rao: The annotation itself was added by this PR and wasn't present before. This PR adds no logic for druid to remove the annotation after sync. So this test isn't really required right? What is the real purpose of this test? If at some point the code in master branch had the logic for druid to remove the annotation, then sure. Or else, it just seems like we added a test to prevent the execution of logic that we never added in the first place.

internal/common/constants.go
shreyas-s-rao: Thanks. What about the annotation itself? It uses the term `update`, while the code uses the term `sync`. Is this drift justified?

docs/usage/managing-etcd-clusters.md
anveshreddy18: It's part of the ops-task controller itself, PreSync just creates them. This was originally implemented in the #1300 PR to take snapshots during etcd upgrades.

internal/common/constants.go
anveshreddy18: I've used the term `update` for the annotation because it is user facing and they might not know about our internal details of `presync`, `sync`, etc. And used `sync` in the code as the code is structured with that terminology. I fear that if we use `sync` in the annotation that might not make sense for the users as what they're really doing is just an update, or if we use `update` internally in the code it doesn't fully acknowledge the internal structure we've already established. I Just tried to use relevant naming based on their usage but don't have a strong opinion..open to change it if you recommend :)

test/it/controller/etcd/reconciler_test.go
anveshreddy18: Yeah what you're saying makes sense. I was just trying to be extra protective by testing it against any future changes that might remove the annotation. But it's not really needed. Will be removing it.

Thanks

internal/component/statefulset/statefulset.go
anveshreddy18: Ahh I see. Thanks for explaining it in detail.

I've now added a warning in the usage doc and if this becomes a problem in the future if someone is using a mutating webhook or policy, we can handle it in the code.

test/it/controller/etcd/reconciler_test.go
anveshreddy18: Removed

shreyas-s-rao (commented):
@anveshreddy18 thanks for addressing all my comments. 
/lgtm

Shreyas-s14 (approved):

docs/usage/managing-etcd-clusters.md
```diff
@@ -97,6 +101,41 @@ kubectl annotate etcd <etcd-name> gardener.cloud/operation=reconcile -n <namespa
...
+
+> [!WARNING]
+> `etcd-druid` detects a container image change by comparing the images it expects (derived from the `Etcd` resource and the configured image vector) against the images currently set on the `StatefulSet`. Always customize container images declaratively via the `Etcd` spec (see [Overwrite Container OCI Images](#overwrite-container-oci-images)) so that `etcd-druid` applies them as part of its own reconciliation. Do **not** mutate the images on the managed `StatefulSet` out-of-band through a mutating admission webhook, etc since such mutations are not reconciled by `etcd-druid`, so every reconciliation will detect an image difference and repeatedly trigger a pre-update snapshot.
+
```
priyadharshini-151193, line 138: What happens if a full snapshot from another source is already in progress when PreSync fires — e.g. a user-triggered on-demand EtcdOpsTask, the scheduled full snapshot, or a still-running pre-sync task from a previous generation (since the task name embeds etcd.Generation)?

internal/component/statefulset/statefulset.go
priyadharshini-151193: Sharing some context from the native hibernation work (#922), since this `replicas == 0` check is the current stand-in for it.

Once druid-driven hibernation lands, hibernation will be driven by `spec.hibernation.enabled` (set by etcd consumers like Gardener), and this branch would change: inside `replicas == 0` we'd check the hibernation state (behind a feature gate if needed).

- **If hibernated:** this becomes a no-op, because the pre-hibernation snapshot is already taken as part of the hibernation flow before the STS is scaled down.
- **If not hibernated:** it follows the current pre-sync snapshot path with the 3 max retries.

Given that, the skip-snapshot annotation would mostly be relevant for the image/replica update path once #922 is in — the hibernation flow would own the snapshot decision itself. So the annotation is really only relevant for the image/replica update path, which aligns with the decision already taken here to not honour it for hibernation.

Nothing to change in this PR — just flagging the direction so we can take an informed decision

internal/component/statefulset/statefulset.go
```diff
@@ -107,7 +107,7 @@ func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Et
 	// hibernation support [gardener/etcd-druid#922](https://github.com/gardener/etcd-druid/issues/922) is implemented,
 	// we need to switch to the dedicated hibernation signal on the Etcd resource instead of inferring it from the replica count.
 	if etcd.Spec.Replicas == 0 {
-		return r.ensurePreSyncSnapshot(ctx, etcd, preSyncTaskHibernationPrefix)
+		return r.ensurePreSyncSnapshot(ctx, etcd, fmt.Sprintf("%s%d-", preSyncTaskHibernationPrefix, etcd.Generation))
```
shreyas-s-rao, line 110: As discussed with @anveshreddy18 in an out-of-band discussion, we decided that simply adding the etcd generation to the etcdopstask name can cause a race between multiple different opstasks for the same full snapshot (especially during quick succession Etcd spec image updates where the first update's reconciliation PreSync step is not completed yet).

Instead, the sts PreSync logic should also check for existing opstask associated with the same Etcd cluster, and latch onto it and wait for its completion before proceeding to Sync step.

shreyas-s-rao (changes requested):

internal/component/statefulset/statefulset.go
anveshreddy18: Made the changes in the last 2 commits.

docs/usage/managing-etcd-clusters.md
anveshreddy18: If an ondemand full snapshot task from other source ( manual, etc) is triggered targetting the same Etcd, then the presync's tasks get rejected and after 3 such failed retries the controller proceeds to Sync by raising a warning event `PreSyncSnapshotFailed`. 

If a still running pre-sync from prev generation is detected, the new controller reconciliation will latch on the previous running one and sees it to completion ( success **or** failure after remaining retries ).

internal/component/statefulset/statefulset.go
```diff
@@ -205,32 +207,46 @@ func (r _resource) ensurePreSyncSnapshot(ctx component.OperatorContext, etcd *dr
...
 	if err := r.client.List(ctx, taskList, client.InNamespace(etcd.Namespace)); err != nil {
 		return nil, nil, err
 	}
 
 	var latestIndex *int
 	var latestTask *druidv1alpha1.EtcdOpsTask
-	for _, task := range taskList.Items {
-		indexStr, found := strings.CutPrefix(task.Name, prefix)
+	for i := range taskList.Items {
+		task := &taskList.Items[i]
```
shreyas-s-rao: Previously we used `for _, task := range taskList.Items` for iteration, but now you change it to `for i := range taskList.Items` - any specific reason?

internal/component/statefulset/statefulset.go
```diff
@@ -205,32 +207,46 @@ func (r _resource) ensurePreSyncSnapshot(ctx component.OperatorContext, etcd *dr
...
-			latestIndex = &idx
-			latestTask = &task
+		lastDash := strings.LastIndex(suffix, "-")
+		if lastDash < 0 {
+			continue
+		}
+		idx, err := strconv.Atoi(suffix[lastDash+1:])
+		if err != nil {
+			continue
+		}
```
shreyas-s-rao: Why not simplify this using `strings.Split()` and get the last element out as the `indexStr`? Looks like this `lastDash` business just adds more lines without any real benefits, unless I'm mistaken.

shreyas-s-rao (changes requested):
@anveshreddy18 thanks for the changes. Just couple of nits, but the logic looks sound.

internal/component/statefulset/statefulset.go
anveshreddy18: That's just a simplification change, previously we used to take the pointer to the task in the middle like `latestTask = &task` since we need to return a pointer at the end, but the current change just reads it as a pointer at the start, this change is for that.

internal/component/statefulset/statefulset.go
anveshreddy18: Thanks for the rec, it sure adds 2 new lines so removed it now :p

internal/component/statefulset/statefulset.go
anveshreddy18: I've pushed the change along with the other one. If it was just this, should've been fine imo.

PR #1361, opened 2026-05-22
Update doc with readiness and liveness probes configuration.

ishan16696:
/area documentation

This PR documents the current readiness and liveness probes configuration set on etcd statefulset deployed by etcd-druid.

Fixes #

**Checklist**:

I'm capturing this info in this doc `docs/concepts/etcd-cluster-components.md` but open for change.

**Release note**:

```other operator
None
```

### DIFF
--- a/docs/concepts/etcd-cluster-components.md
+++ b/docs/concepts/etcd-cluster-components.md
@@ -49,7 +49,7 @@ To ensure that etcd pods are not evicted more than its failure tolerance, `etcd-
 
 ## Role & RoleBinding
 
-`etcd-backup-restore` container running as a side-car in every etcd-member, requires permissions to access resources like `Lease`, `StatefulSet` etc. A dedicated [Role]() and [RoleBinding]() is created and linked to the [ServiceAccount](https://kubernetes.io/docs/concepts/security/service-accounts/) created per `Etcd` cluster.
+`etcd-backup-restore` container running as a side-car in every etcd-member, requires permissions to access resources like `Lease`, `StatefulSet` etc. A dedicated [Role](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole) and [RoleBinding](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding) is created and linked to the [ServiceAccount](https://kubernetes.io/docs/concepts/security/service-accounts/) created per `Etcd` cluster.
 
 **Code reference:** [Role-Component](https://github.com/gardener/etcd-druid/tree/3383e0219a6c21c6ef1d5610db964cc3524807c8/internal/component/role) & [RoleBinding-Component](https://github.com/gardener/etcd-druid/tree/master/internal/component/rolebinding)
 
@@ -77,3 +77,19 @@ One of the responsibilities of `etcd-backup-restore` container is to take period
 > In future these leases will be replaced by [EtcdMember resource](https://github.com/gardener/etcd-druid/blob/3383e0219a6c21c6ef1d5610db964cc3524807c8/docs/proposals/04-etcd-member-custom-resource.md).
 
 **Code reference:** [Snapshot-Lease-Component](https://github.com/gardener/etcd-druid/tree/3383e0219a6c21c6ef1d5610db964cc3524807c8/internal/component/snapshotlease)
+
+## Probes
+
+### ReadinessProbe
+
+Currently a [/readyz](https://github.com/gardener/etcd-wrapper/blob/cf3be92fc03c78941a3275f6ac4b37f54fd9b59a/internal/app/readycheck.go#L143) endpoint provided by [etcd-wrapper](https://github.com/gardener/etcd-wrapper) is being used as readinessProbe. This endpoint performs a linearizable `GET` request, which effectively represents the readiness of the etcd cluster.
+
+### LivenessProbe
+
+Currently `etcd-druid` doesn't configure any [liveness probe](https://kubernetes.io/docs/concepts/workloads/pods/probes/#liveness-probe) for a etcd statefulset. Based on our observations, the introduction of livenessProbe can do more harm then being useful as liveness probe failure simply causes the kubelet to restart the etcd container, which is not helpful in most of the scenarios. Some examples of such scenarios:
+
+1. **Etcd cluster bootstrap**: Determining an appropriate value for `livenessProbe.initialDelaySeconds` is difficult as etcd member(s)may take longer than expected to start for various reasons. If the startup time exceeds initialDelaySeconds, the kubelet will prematurely kills and restart the etcd container.
+
+2. Even if `initialDelaySeconds` is set to some large value (say >200sec) or a startupProbe is introduced, still it won't solve all of the problems, for example: Consider a scenario when etcd cluster member are busy due to high load because of this if any etcd member unable to respond to livenessProbe within the configured time causing livenessProbe to fail, due to that kubelet will kills the etcd container in order to restart the container, this restarting of a member will only worsen the overall situation.
+
+more details can be found here: https://github.com/gardener/etcd-druid/issues/280#issuecomment-4505733020
--- a/docs/concepts/etcd-cluster-components.md
+++ b/docs/concepts/etcd-cluster-components.md
@@ -12,12 +12,12 @@ For every `Etcd` cluster that is provisioned by `etcd-druid` it deploys a set of
   * `etcd-wrapper` : This is the main container which runs an etcd process.
   
   * `etcd-backup-restore` : This is a side-container which does the following:
-    
+
     * Orchestrates the initialization of etcd. This includes validation of any existing etcd data directory, restoration in case of corrupt etcd data directory files for a single-member etcd cluster.
     * Periodically renews member lease.
     * Optionally takes schedule and threshold based delta and full snapshots and pushes them to a configured object store.
     * Orchestrates scheduled etcd-db defragmentation.
-    
+
     > NOTE: This is not a complete list of functionalities offered out of `etcd-backup-restore`. 
 
 **Code reference:** [StatefulSet-Component](https://github.com/gardener/etcd-druid/tree/480213808813c5282b19aff5f3fd6868529e779c/internal/component/statefulset)
@@ -34,7 +34,7 @@ Every `etcd` member requires [configuration](https://etcd.io/docs/v3.4/op-guide/
 
 An etcd cluster requires quorum for all write operations. Clients can additionally configure quorum based reads as well to ensure [linearizable](https://jepsen.io/consistency/models/linearizable) reads (kube-apiserver's etcd client is configured for linearizable reads and writes). In a cluster of size 3, only 1 member failure is tolerated. [Failure tolerance](https://etcd.io/docs/v3.3/faq/#what-is-failure-tolerance) for an etcd cluster with replicas `n` is computed as `(n-1)/2`.
 
-To ensure that etcd pods are not evicted more than its failure tolerance, `etcd-druid` creates a [PodDisruptionBudget](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets). 
+To ensure that etcd pods are not evicted more than its failure tolerance, `etcd-druid` creates a [PodDisruptionBudget](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets).
 
 !!! note
     For a single node etcd cluster a `PodDisruptionBudget` will be created, however `pdb.spec.minavailable` is set to 0 effectively disabling it.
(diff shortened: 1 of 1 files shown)


docs/concepts/etcd-cluster-components.md
```diff
@@ -49,7 +49,7 @@ To ensure that etcd pods are not evicted more than its failure tolerance, `etcd-
 
 ## Role & RoleBinding
 
-`etcd-backup-restore` container running as a side-car in every etcd-member, requires permissions to access resources like `Lease`, `StatefulSet` etc. A dedicated [Role]() and [RoleBinding]() is created and linked to the [ServiceAccount](https://kubernetes.io/docs/concepts/security/service-accounts/) created per `Etcd` cluster.
+`etcd-backup-restore` container running as a side-car in every etcd-member, requires permissions to access resources like `Lease`, `StatefulSet` etc. A dedicated [Role](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole) and [RoleBinding](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding) is created and linked to the [ServiceAccount](https://kubernetes.io/docs/concepts/security/service-accounts/) created per `Etcd` cluster.
```
anveshreddy18: can we provide the similar doc links to k8s resources for the `Lease`, `Statefulset`, etc in this line?

docs/concepts/etcd-cluster-components.md
```diff
@@ -77,3 +77,19 @@ One of the responsibilities of `etcd-backup-restore` container is to take period
 > In future these leases will be replaced by [EtcdMember resource](https://github.com/gardener/etcd-druid/blob/3383e0219a6c21c6ef1d5610db964cc3524807c8/docs/proposals/04-etcd-member-custom-resource.md).
 
 **Code reference:** [Snapshot-Lease-Component](https://github.com/gardener/etcd-druid/tree/3383e0219a6c21c6ef1d5610db964cc3524807c8/internal/component/snapshotlease)
+
+## Probes
+
+### ReadinessProbe
+
+Currently a [/readyz](https://github.com/gardener/etcd-wrapper/blob/cf3be92fc03c78941a3275f6ac4b37f54fd9b59a/internal/app/readycheck.go#L143) endpoint provided by [etcd-wrapper](https://github.com/gardener/etcd-wrapper) is being used as readinessProbe. This endpoint performs a linearizable `GET` request, which effectively represents the readiness of the etcd cluster.
```
anveshreddy18:
```suggestion
Currently a [/readyz](https://github.com/gardener/etcd-wrapper/blob/cf3be92fc03c78941a3275f6ac4b37f54fd9b59a/internal/app/readycheck.go#L143) endpoint provided by [etcd-wrapper](https://github.com/gardener/etcd-wrapper) is being used as readinessProbe. This endpoint performs a linearizable `GET` request, which effectively represents the readiness of the etcd cluster as a linearizable call expects strong consistency which requires quorum among the members.
```

docs/concepts/etcd-cluster-components.md
```diff
@@ -77,3 +77,19 @@ One of the responsibilities of `etcd-backup-restore` container is to take period
...
+
+### LivenessProbe
+
+Currently `etcd-druid` doesn't configure any [liveness probe](https://kubernetes.io/docs/concepts/workloads/pods/probes/#liveness-probe) for a etcd statefulset. Based on our observations, the introduction of livenessProbe can do more harm then being useful as liveness probe failure simply causes the kubelet to restart the etcd container, which is not helpful in most of the scenarios. Some examples of such scenarios:
+
+1. **Etcd cluster bootstrap**: Determining an appropriate value for `livenessProbe.initialDelaySeconds` is difficult as etcd member(s)may take longer than expected to start for various reasons. If the startup time exceeds initialDelaySeconds, the kubelet will prematurely kills and restart the etcd container.
```
anveshreddy18:
```suggestion
1. **Etcd cluster bootstrap**: Determining an appropriate value for `livenessProbe.initialDelaySeconds` is difficult as etcd member(s) may take longer than expected to start for various reasons such as performance or network related among others. If the startup time exceeds initialDelaySeconds, the Kubelet will prematurely kill and restart the etcd container which could cause more delays and consequently increases the chances of the cluster operating with fewer members.
```

docs/concepts/etcd-cluster-components.md
```diff
@@ -77,3 +77,19 @@ One of the responsibilities of `etcd-backup-restore` container is to take period
...
+
+2. Even if `initialDelaySeconds` is set to some large value (say >200sec) or a startupProbe is introduced, still it won't solve all of the problems, for example: Consider a scenario when etcd cluster member are busy due to high load because of this if any etcd member unable to respond to livenessProbe within the configured time causing livenessProbe to fail, due to that kubelet will kills the etcd container in order to restart the container, this restarting of a member will only worsen the overall situation.
```
anveshreddy18:
```suggestion
2. Even if `livenessProbe.initialDelaySeconds` is set to some large value (say >200sec) or a startupProbe is introduced, still it won't solve all of the problems, for example: Consider a scenario where etcd cluster members are busy due to high load, now because of this if any etcd member is unable to respond to livenessProbe within the configured time causing livenessProbe to fail, due to that kubelet will kill the etcd container in order to restart the container, this restarting of a member will only worsen the overall situation.
```

anveshreddy18 (commented):
Thanks for clarifying the probes related info in the docs @ishan16696. Have posted few nits, PTAL.

docs/concepts/etcd-cluster-components.md
ishan16696: > such as performance or network related among others

tbh there can be other reasons as well like bootstrapping the etcd cluster, that's why I choose not to mention reasons by name.

ishan16696: > which could cause more delays and consequently increases the chances of the cluster operating with fewer members.

that's also not true always, suppose a bootstrap of cluster itself fails then there is no point of "operating with few members".

anveshreddy18: > tbh there can be other reasons as well like bootstrapping the etcd cluster, that's why I choose not to mention reasons by name.

yeah there can be many other reasons for that like you said, I just wanted to hint at the most common ones so it's intuitive to read this as then the reader will imagine those issues and understand it better. But yeah it's not a strong opinion, so fine with me.

> that's also not true always, suppose a bootstrap of cluster itself fails then there is no point of "operating with few members".

yeah same as the other one, this only lists a possibility for intuition, there can be other possibilities as well.


But yeah since I don't have any strong opinions, you can close this if you think those are not needed. Thanks

anveshreddy18 (approved):
Thanks for updating the doc. LGTM!

docs/concepts/etcd-cluster-components.md
```diff
@@ -77,3 +77,19 @@ One of the responsibilities of `etcd-backup-restore` container is to take period
...
+
+1. **Etcd cluster bootstrap**: Determining an appropriate value for `livenessProbe.initialDelaySeconds` is difficult as etcd member(s)may take longer than expected to start for various reasons such as network related issue among others etc. If the startup time exceeds initialDelaySeconds, the kubelet will prematurely kills and restart the etcd container.
+
+2. Even if `livenessProbe.initialDelaySeconds` is set to some large value (say >200sec) or a startupProbe is introduced, still it won't solve all of the problems, for example: Consider a scenario where etcd cluster member are busy due to high load, now because of this if any etcd member is unable to respond to livenessProbe within the configured time causing livenessProbe to fail, due to that kubelet will kill the etcd container in order to restart the container, this restarting of a member will only worsen the overall situation.
```
Shreyas-s14: Just a few grammatical changes and a reference link for `startupProbe`:
```suggestion
Even if livenessProbe.initialDelaySeconds is set to some large value (say >200sec) or a [startupProbe](https://kubernetes.io/docs/concepts/workloads/pods/probes/#startup-probe) is introduced, it still won't solve all of the problems. For example, consider a scenario where etcd cluster members are busy due to high load. Because of this, if any etcd member is unable to respond to the livenessProbe within the configured time, the livenessProbe fails, due to which kubelet will kill the etcd container in order to restart it. This restarting of a member will only worsen the overall situation.
```

docs/concepts/etcd-cluster-components.md
```diff
@@ -77,3 +77,19 @@ One of the responsibilities of `etcd-backup-restore` container is to take period
...
+
+## Probes
+
+### ReadinessProbe
+
+Currently a [/readyz](https://github.com/gardener/etcd-wrapper/blob/cf3be92fc03c78941a3275f6ac4b37f54fd9b59a/internal/app/readycheck.go#L143) endpoint provided by [etcd-wrapper](https://github.com/gardener/etcd-wrapper) is being used as readinessProbe. This endpoint performs a linearizable `GET` request, which effectively represents the readiness of the etcd cluster as linearizable call expects strong consistency which requires quorum among the members.
+
+### LivenessProbe
+
+Currently `etcd-druid` doesn't configure any [liveness probe](https://kubernetes.io/docs/concepts/workloads/pods/probes/#liveness-probe) for a etcd statefulset. Based on our observations, the introduction of livenessProbe can do more harm then being useful as liveness probe failure simply causes the kubelet to restart the etcd container, which is not helpful in most of the scenarios. Some examples of such scenarios:
```
Shreyas-s14:
```suggestion
Currently `etcd-druid` doesn't configure any [liveness probe](https://kubernetes.io/docs/concepts/workloads/pods/probes/#liveness-probe) for a etcd statefulset. Based on our observations, the introduction of livenessProbe can do more harm than being useful as liveness probe failure simply causes the kubelet to restart the etcd container, which is not helpful in most of the scenarios. Some examples of such scenarios:
```
minor change.

Shreyas-s14 (changes requested):
Thanks for updating the doc and explaining why `livenessProbe` isn't useful for etcd-druid setup.
Have a few nits PTAL.

seshachalam-yv (commented):
/lgtm
/approve

Shreyas-s14 (approved):
LGTM

docs/concepts/etcd-cluster-components.md
```diff
@@ -77,3 +77,19 @@ One of the responsibilities of `etcd-backup-restore` container is to take period
...
+
+1. **Etcd cluster bootstrap**: Determining an appropriate value for `livenessProbe.initialDelaySeconds` is difficult as etcd member(s)may take longer than expected to start for various reasons such as network related issue among others etc. If the startup time exceeds initialDelaySeconds, the kubelet will prematurely kills and restart the etcd container.
+
+2. Even if `livenessProbe.initialDelaySeconds` is set to some large value (say >200sec) or a [startupProbe](https://kubernetes.io/docs/concepts/workloads/pods/probes/#startup-probe) is introduced, it still won't solve all of the problems. For example, consider a scenario where etcd cluster member(s) are busy due to high load. Because of this, if any etcd member is unable to respond to the livenessProbe within the configured time, the livenessProbe fails, due to which kubelet will kill the etcd container in order to restart the container. This restarting of a member will only worsen the overall situation.
```
unmarshall, line 93: If there is no liveliness probe then almost always an operator will have to get involved to check why a etcd pod/member is not starting.
`/livez` makes a serializable read to the local process. What is the delay that you have observed when using `/livez` as liveness probe? 
If you set a proper combination of `initialDelaySeconds` and `failureThreshold` then you should be able to easily avoid spurious restarts by kubelet and at the same time ensure that eventually it gets restarted after a significant delay.

docs/concepts/etcd-cluster-components.md
```diff
@@ -77,3 +77,19 @@ One of the responsibilities of `etcd-backup-restore` container is to take period
...
+
+### LivenessProbe
+
+Currently `etcd-druid` doesn't configure any [liveness probe](https://kubernetes.io/docs/concepts/workloads/pods/probes/#liveness-probe) for a etcd statefulset. Based on our observations, the introduction of livenessProbe can do more harm than being useful as liveness probe failure simply causes the kubelet to restart the etcd container, which is not helpful in most of the scenarios. Some examples of such scenarios:
+
+1. **Etcd cluster bootstrap**: Determining an appropriate value for `livenessProbe.initialDelaySeconds` is difficult as etcd member(s)may take longer than expected to start for various reasons such as network related issue among others etc. If the startup time exceeds initialDelaySeconds, the kubelet will prematurely kills and restart the etcd container.
```
unmarshall, line 91: >take longer than expected to start for various reasons such as network related

Why do you think the reason will be network related when the call is a serializable call that is only made to the local etcd process?

ishan16696: > Why do you think the reason will be network related when the call is a serializable call that is only made to the local etcd process?

it's only applicable during the bootstrap case because if there is a networking issues during the bootstrap then member will never become healthy hence, it's livenessProbe will always failed.
May be I should have mentioned `bootstrap` word explicitly.

docs/concepts/etcd-cluster-components.md
ishan16696: > /livez makes a serializable read to the local process. What is the delay that you have observed when using /livez as liveness probe?

please refer to this comment for details https://github.com/gardener/etcd-druid/issues/280#issuecomment-4505733020 . Moreover, there can be other issues possible like `volumeAttachment` issues etc. But main concern and point is "how a restarting of etcd container will solve any of such issues", all it can do is hide some deadlock or memory-leaks issue, which is not a nice thing otherwise we would never be able to find such memory-leak issues: https://github.com/gardener/etcd-backup-restore/issues/928

unmarshall (approved):

PR #1370, opened 2026-06-15
component/statefulset: make secretRef optional for cloud backup providers

yeganeahmadnejad:
## What does this PR do?

When a pod runs on a managed Kubernetes cluster with pod-level identity (GKE Workload Identity, EKS IRSA, AKS Workload Identity), the cloud SDK's Application Default Credentials (ADC) chain can mint short-lived tokens without a mounted Kubernetes Secret.

Previously etcd-druid required a non-nil `secretRef` for all cloud backup providers (GCS, S3, ABS, OSS, Swift, OCS), returning an error when it was absent. This forced users on managed Kubernetes to create and rotate a Secret even when the pod identity already grants the necessary permissions.

## Changes

- `getBackupVolume`: return `nil, nil` instead of an error when `store.SecretRef == nil` for cloud providers
- `getEtcdBackupVolumeMount`: return `nil` when `store.SecretRef == nil`

When `secretRef` is absent, no credential volume or mount is injected into the StatefulSet. The backup-restore container's GCS/S3/etc. client falls back to ADC and picks up the pod-level identity automatically.

## How was this tested?

Validated on GKE with Workload Identity:
- Removed `secretRef` from an `Etcd` CR backed by GCS
- etcd-druid successfully created the StatefulSet without a credential volume
- `etcd-backup-restore` authenticated to GCS via the GKE metadata server (ADC) and took snapshots successfully

## Note

`secretRef` remains fully supported — existing deployments that use it are unaffected.


yeganeahmadnejad:
Closing in favour of a rebased PR from master with corrected commit message.

PR #1374, opened 2026-06-29
Add `spec.etcd.backendBboltFreelistType` to Etcd API for etcd version 3.5.x or greater.

### LINKED ISSUE #1373: Make etcd flag: `--backend-bbolt-freelist-type` configurable.
/area storage
/kind enhancement

Make this etcd flag: `--backend-bbolt-freelist-type` configurable and set the value to `array` instead of `map`. 

```
  --backend-bbolt-freelist-type 'map'
    BackendFreelistType specifies the type of freelist that boltdb backend uses(array and map are supported types).
```

**Why is this needed**:

Etcd `v3.5.x` has changed the default value of `--backend-bbolt-freelist-type` from `array` to `map` and because of this bolt-db is using map to maintain the freelist but it has been observed that because of this change after the etcd upgrade, the difference b/w total database size and total database size inuse has been increased significantly.

- There are no spikes or behavioural change in `total database size inuse` before and after the etcd upgrade.
- But there are spikes or constant increase in `total database size` after the etcd upgrade.

Image

ishan16696:
/area storage
/kind api-change

Etcd `v3.5.x` has changed the default value of [--backend-bbolt-freelist-type](https://etcd.io/docs/v3.5/op-guide/configuration/) from `array` to `map` via this PR: https://github.com/etcd-io/etcd/pull/11262 and because of this bolt-db is using hashmap to maintain the freelist.
This PR add `spec.etcd.backendBboltFreelistType` to the Etcd API to make `BblotFreeListType` configurable via etcd resource with default value set to `array` instead of `map`.
We have changed the default value of `--backend-bbolt-freelist-type` because it has been observed that after the etcd upgrade to etcd `v3.5.x`, the total database size has been significantly increased, so it's better to use the `array` to maintain freelist for smaller databases which use few GBs.

Fixes #1373 

**Checklist**:
    - [x] Unit tests
    - [ ] Integration tests
    - [ ] E2E tests

**Release note**:

```noteworthy operator
Added `spec.etcd.backendBboltFreelistType` to the Etcd API to make BblotFreeListType configurable via etcd resource with default value set to `array` instead of `map`.
```


ishan16696:
locally `make check-generate` passed
```
> make check-generate         
> Generate
>> make generate
>> make tidy
```

api/core/v1alpha1/etcd.go
```diff
@@ -320,6 +331,11 @@ type EtcdConfig struct {
 	// ClientService defines the parameters of the client service that a user can specify
 	// +optional
 	ClientService *ClientService `json:"clientService,omitempty"`
+	// BackendBboltFreelistType specifies the freelist type used by the bbolt backend storage engine of etcd.
+	// Supported values are 'array' (default) and 'map'.
+	// It corresponds to the etcd flag --backend-bbolt-freelist-type which available only from etcd version 3.5.x
+	// +optional
+	BackendBboltFreelistType *BboltFreelistType `json:"backendBboltFreelistType,omitempty"`
```
Shreyas-s14, line 340: Do you think we should mention that the official type for 3.5.x versions is map for upstream etcd setup?

api/core/v1alpha1/etcd.go
```diff
@@ -320,6 +331,11 @@ type EtcdConfig struct {
 	// ClientService defines the parameters of the client service that a user can specify
 	// +optional
 	ClientService *ClientService `json:"clientService,omitempty"`
+	// BackendBboltFreelistType specifies the freelist type used by the bbolt backend storage engine of etcd.
+	// Supported values are 'array' (default) and 'map'.
+	// It corresponds to the etcd flag --backend-bbolt-freelist-type which available only from etcd version 3.5.x
+	// +optional
```
Shreyas-s14, line 339: in the PR could you add it tests in `test/it/crdvalidation/etcd/specetcd_test.go` ?
i think only 3-4 cases are needed

1) unset -> default
2) set -> map/array
3) set -> incorrect value

Shreyas-s14 (commented):
Thanks for the PR for adding this flag + making it configurable. Have 2 nits, PTAL

api/core/v1alpha1/etcd.go
ishan16696: Currently the problem is if feature-gate isn't enabled then druid deploys etcd `v3.4.x`, so Once we move the etcdUpgrade feature-gate to GA then we can remove this as we will be no longer supporting the etcd `v3.4.x`. wdyt ?

api/core/v1alpha1/etcd.go
ishan16696: done, PTAL

api/core/v1alpha1/etcd.go
```diff
@@ -320,6 +331,11 @@ type EtcdConfig struct {
 	// ClientService defines the parameters of the client service that a user can specify
 	// +optional
 	ClientService *ClientService `json:"clientService,omitempty"`
+	// BackendBboltFreelistType specifies the freelist type used by the bbolt backend storage engine of etcd.
+	// Supported values are 'array' (default) and 'map'.
```
anveshreddy18, line 333: I feel this line can be misunderstood because upstream default is `map` and our default is `array`. Would you like to add a small mention about the upstream default also?

api/core/v1alpha1/etcd.go
```diff
@@ -94,6 +94,17 @@ type CompressionPolicy string
...
+// BboltFreelistType specifies the freelist type used by the bbolt backend storage engine of etcd.
+// +kubebuilder:validation:Enum=array;map
+type BboltFreelistType string
+
+const (
+	// BboltFreelistArray is the array-based freelist type for bbolt backend.
+	BboltFreelistArray BboltFreelistType = "array"
+	// BboltFreelistMap is the hashmap-based freelist type for bbolt backend.
+	BboltFreelistMap BboltFreelistType = "map"
+)
```
anveshreddy18: Nit: Would you like to move the consts to the existing const block at the start of the file?

anveshreddy18 (commented):
Thanks for identifying and fixing the issue. Just have 2 minor comments. PTAL.

api/core/v1alpha1/etcd.go
ishan16696: I think I did add it somewhere but I guess it got somehow removed while adding tests, anyway thanks for pointing this out. I have now added it. PTAL.

api/core/v1alpha1/etcd.go
Shreyas-s14: as in? Even with 3.5 we still will be having this field right? I'm not sure if they'll revert to `array` post 3.6 or so since from what we've seen, noone has reported a problem for this.

ishan16696: ok, I have mentioned this in new commit https://github.com/gardener/etcd-druid/pull/1374/changes/7c8e2941501408113ee5ad7d872bfd0d2eaf9604, can you take a look

internal/component/configmap/etcdconfig.go
```diff
@@ -25,6 +26,12 @@ const (
 	defaultSnapshotCount   = int64(10000)
 	advertiseURLTypePeer   = "peer"
 	advertiseURLTypeClient = "client"
+
+	// Note: Although etcd v3.5.x defaults `--backend-bbolt-freelist-type` to "map", etcd-druid default to "array"
+	// here because "map" has been observed to cause significant increase in total database size.
+	// The "array" freelist type is more space-efficient for the small databases (a few GBs) clusters.
+	// Please refer to this issue for more issue: https://github.com/gardener/etcd-druid/issues/1373
```
anveshreddy18: Thanks for adding this Ishan, however I feel this information is more useful if we mention this in the api/core/v1alpha1/etcd.go file on top of the API field
```go
// <somewhere here>
BackendBboltFreelistType *BboltFreelistType `json:"backendBboltFreelistType,omitempty"`
```

Because this comment talks about the reasons for why we are choosing to default to `array` from our side, I feel having it in the API will allow the users to know what to choose and why to choose that. In this etcdconfig.go file it doesn't serve much purpose. Wdyt?

ishan16696: ok, make sense ... let me add appropriate comment here as well.

ishan16696: done.

anveshreddy18 (approved):
LGTM. Thanks

Shreyas-s14 (approved):
Thanks for the changes
LGTM

Shreyas-s14:
I tried this out with a local gardener setup.
* I had a script that simulated a standard k8s like workload kept feeding keys into the etcd by initially using the benchmark tool for static floor followed by overwrites with frequently -> the db was never stagnant
* Ran with the current `master` image of etcd-druid which uses the `map` freelist type as default since it isnt configurable (that is introduced in this PR).
* For the workload, it was kept running for about 25-30 minutes and then etcd pod was restarted since the map is in memory and followed by a defrag (as can be seen in the graph where there's a dip in the db size)
* Even after this, the db size increased as soon as the etcd was back after defrag, and the script kept running.
* Ran this again for another 30 minutes and so no improvements even though defrag was done.
* Then bumped the `etcd-druid` image to the one built from this PR branch where the default was explicitly set to `array`.
* Once the local gardener-setup was restarted to pick up the changes, it was followed up with a `Reconcile` of the etcd resource to generate the configmap so as to include this field (This further was verified by attaching a debug container and checking the config.yaml in the backup restore process)
* Defrag was run and the script was continued.
* Now the db size overall was much closer to what was actually used by the etcd db thus proving that `freelist type array` is consuming less overall space than `freelist type map` even though the metric value for `etcd_mvcc_db_size_in_use_in_bytes` was the same for both
image

PR #1388, opened 2026-07-21
Introduce the `local` build tag to gate the `change-backup-bucket-permissions` initcontainer

renormalize:
**How to categorize this PR?**

/area quality robustness
/kind cleanup

**What this PR does / why we need it**:

* `change-backup-bucket-permissions` is only included during compilation when the `local` build tag is used. This initcontainer performs a `chmod` on local directories used for etcd backups when the `Local` provider is configured. This init container only needs to run during e2e tests on the kind cluster, therefore this code can completely be gated behind this build tag.

* `local` tag is passed directly in `skaffold.yaml` to enable this feature locally.

**Which issue(s) this PR fixes**:
Fixes #

**Checklist**:
- [ ] Update documentation in the `/docs` folder (if applicable)
    - If new files added to docs, or docs structure modified:
        - [ ] Update [mkdocs.yml](https://github.com/gardener/etcd-druid/blob/master/mkdocs.yml). Follow [Updating Documentation](https://github.com/gardener/etcd-druid/blob/master/docs/development/updating-documentation.md) guide to test the changes locally.
        - [ ] Update [Table of Contents](https://github.com/gardener/etcd-druid/blob/master/docs/README.md) for the docs.
- [ ] Add tests that cover your changes (if applicable)
    - [ ] Unit tests
    - [ ] Integration tests
    - [ ] E2E tests

**Special notes for your reviewer**:

We can confirm that this chmod init container will not be baked in by default by running:

```go
go tool nm ./bin/etcd-druid | rg "statefulset.init"
```

**Release note**:

```other operator
N/A
```

PR #1390, opened 2026-07-22
Validate backup store container name and harden chown init containers

shreyas-s-rao:

/area security
/area backup
/kind bug

The `spec.backup.store.container` field (`StoreSpec.Container`, shared by `Etcd` and `EtcdCopyBackupsTask`) was string-formatted, without validation, into a `sh -c` command in the root-privileged `change-backup-bucket-permissions` init container. For `provider: Local`, a user able to create/update these resources could inject arbitrary shell that runs as root in the generated pod/job (CWE-77, OS command injection).

This PR closes the issue with two layered fixes:

1. **Input validation (primary).** Adds `Pattern` (`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`) and `MaxLength=63` validation to `StoreSpec.Container`, so shell metacharacters, spaces, path separators, and hostile leading characters are rejected at admission. A `pattern` marker is used rather than a CEL rule because this is a single-scalar character-set constraint; a CEL rule would be the same regex wrapped in `self.matches(...)` with no added value. Being part of the structural schema, the pattern also applies to the CEL-less CRD variant.

2. **Defense in depth.** Rewrites both `chown` init containers (statefulset builder and EtcdCopyBackupsTask reconciler) from `sh -c "chown ..."` to the exec/argv form `["chown","-R","65532:65532",<path>...]`. Neither relied on any shell feature, so this is behaviour-preserving while removing the injection primitive entirely.

The init container continues to run as root (uid 0) intentionally: it fixes ownership of the root-created Local backup hostPath so the non-root (uid 65532) etcd and backup-restore containers can write to it. Removing root is not viable and root itself is not the vulnerability — the shell plus unvalidated input was.

Addresses a privately reported security submission (no public issue).

**Checklist**:
    - [x] Unit tests
    - [x] Integration tests

- `StoreSpec` is shared, so the single field marker protects `Etcd.spec.backup.store.container` and `EtcdCopyBackupsTask.spec.{sourceStore,targetStore}.container`.
- `maxLength: 63` is the common maximum bucket/container-name length across AWS S3, GCS, Azure Blob, and Alibaba OSS. It keeps druid's validation aligned with the downstream object-store limits so a value that druid accepts is not later rejected by the provider.
- The leading-alphanumeric rule matches the bucket-naming rules of S3, GCS, Azure, and OSS (all require a letter or digit first); only OpenStack Swift is more permissive, and no etcd-druid usage relies on a leading `.`/`-`/`_`.
- Verified manually on a KIND cluster: injection payloads are rejected at admission, valid bucket names are accepted, and the `chown` init container (argv form) still correctly chowns the Local hostPath so backup-restore writes full snapshots as uid 65532.

/cc @renormalize

**Release note**:
```other operator
The `spec.backup.store.container` field is now validated (allowed characters `[a-zA-Z0-9._-]`, must start and end with an alphanumeric character, min length 3, max length 63). The init container that adjusts permissions for local backups no longer uses a shell, closing a potential command-injection vector.
```

### DIFF
--- a/api/core/v1alpha1/store.go
+++ b/api/core/v1alpha1/store.go
@@ -13,6 +13,8 @@ type StorageProvider string
 type StoreSpec struct {
 	// Container is the name of the container the backup is stored at.
 	// +optional
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9][a-zA-Z0-9._-]{1,61}[a-zA-Z0-9]$`
 	Container *string `json:"container,omitempty"`
 	// EndpointOverride denotes the storage endpoint that will be used to override the storage provider's default endpoint.
 	// +optional
--- a/api/core/v1alpha1/crds/druid.gardener.cloud_etcdcopybackupstasks.yaml
+++ b/api/core/v1alpha1/crds/druid.gardener.cloud_etcdcopybackupstasks.yaml
@@ -71,6 +71,8 @@ spec:
                   container:
                     description: Container is the name of the container the backup
                       is stored at.
+                    maxLength: 63
+                    pattern: ^[a-zA-Z0-9][a-zA-Z0-9._-]{1,61}[a-zA-Z0-9]$
                     type: string
                   endpointOverride:
                     description: EndpointOverride denotes the storage endpoint that
@@ -109,6 +111,8 @@ spec:
                   container:
                     description: Container is the name of the container the backup
                       is stored at.
+                    maxLength: 63
+                    pattern: ^[a-zA-Z0-9][a-zA-Z0-9._-]{1,61}[a-zA-Z0-9]$
                     type: string
                   endpointOverride:
                     description: EndpointOverride denotes the storage endpoint that
--- a/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
+++ b/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
@@ -314,6 +314,8 @@ spec:
                       container:
                         description: Container is the name of the container the backup
                           is stored at.
+                        maxLength: 63
+                        pattern: ^[a-zA-Z0-9][a-zA-Z0-9._-]{1,61}[a-zA-Z0-9]$
                         type: string
                       endpointOverride:
                         description: EndpointOverride denotes the storage endpoint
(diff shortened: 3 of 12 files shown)


api/core/v1alpha1/store.go
```diff
@@ -13,6 +13,8 @@ type StorageProvider string
 type StoreSpec struct {
 	// Container is the name of the container the backup is stored at.
 	// +optional
+	// +kubebuilder:validation:MaxLength=63
```
renormalize, line 16: Nit:

https://docs.cloud.google.com/storage/docs/buckets says:
    
> Bucket names must contain 3-63 characters. Names containing dots can contain up to *222 characters*, but each dot-separated component can be no longer than 63 characters.

Bucket names being longer than 63 seems very unlikely, but the API for GCS allows it. Do we want the length validation, or do we skip it?

api/core/v1alpha1/store.go
```diff
@@ -13,6 +13,8 @@ type StorageProvider string
...
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`
```
renormalize: The general naming rules followed by the hyperscalers seems to be that the name must always start and end with a letter or number. So perhaps we can add another `[a-zA-Z0-9]` at the end?

From https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html:

> Bucket names must begin and end with a letter or number.

From https://docs.cloud.google.com/storage/docs/buckets:

> Bucket names must start and end with a number or letter.

From https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata:

> Container names must start or end with a letter or number, and can contain only letters, numbers, and the hyphen/minus (-) character.

renormalize (approved):
I think this is a better approach as compared to #1388. We would always need the initcontainer code if users intend to use the local provider in production.

api/core/v1alpha1/store.go
renormalize: On second thought, I think we can stick to the 63 character limit. It's not a constraint that should be problematic to consumers.

shreyas-s-rao: Most providers have this 63-character limit, hence I set it the same way.

api/core/v1alpha1/store.go
shreyas-s-rao: Ah nice catch, will make this change

shreyas-s-rao: Made the changes now. Also enforced character length to 3-63 as supported by most providers.

renormalize (approved):
thanks for the quick PR!

PR #1391, opened 2026-07-23
[hotfix-v0.37] Validate backup store container name and harden chown init containers

shreyas-s-rao:
Cherry-pick of #1390 

/area security
/area backup
/kind bug

The `spec.backup.store.container` field (`StoreSpec.Container`, shared by `Etcd` and `EtcdCopyBackupsTask`) was string-formatted, without validation, into a `sh -c` command in the root-privileged `change-backup-bucket-permissions` init container. For `provider: Local`, a user able to create/update these resources could inject arbitrary shell that runs as root in the generated pod/job (CWE-77, OS command injection).

This PR closes the issue with two layered fixes:

1. **Input validation (primary).** Adds `Pattern` (`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`) and `MaxLength=63` validation to `StoreSpec.Container`, so shell metacharacters, spaces, path separators, and hostile leading characters are rejected at admission. A `pattern` marker is used rather than a CEL rule because this is a single-scalar character-set constraint; a CEL rule would be the same regex wrapped in `self.matches(...)` with no added value. Being part of the structural schema, the pattern also applies to the CEL-less CRD variant.

2. **Defense in depth.** Rewrites both `chown` init containers (statefulset builder and EtcdCopyBackupsTask reconciler) from `sh -c "chown ..."` to the exec/argv form `["chown","-R","65532:65532",<path>...]`. Neither relied on any shell feature, so this is behaviour-preserving while removing the injection primitive entirely.

The init container continues to run as root (uid 0) intentionally: it fixes ownership of the root-created Local backup hostPath so the non-root (uid 65532) etcd and backup-restore containers can write to it. Removing root is not viable and root itself is not the vulnerability — the shell plus unvalidated input was.

Addresses a privately reported security submission (no public issue).

**Checklist**:
    - [x] Unit tests
    - [x] Integration tests

- `StoreSpec` is shared, so the single field marker protects `Etcd.spec.backup.store.container` and `EtcdCopyBackupsTask.spec.{sourceStore,targetStore}.container`.
- `maxLength: 63` is the common maximum bucket/container-name length across AWS S3, GCS, Azure Blob, and Alibaba OSS. It keeps druid's validation aligned with the downstream object-store limits so a value that druid accepts is not later rejected by the provider.
- The leading-alphanumeric rule matches the bucket-naming rules of S3, GCS, Azure, and OSS (all require a letter or digit first); only OpenStack Swift is more permissive, and no etcd-druid usage relies on a leading `.`/`-`/`_`.
- Verified manually on a KIND cluster: injection payloads are rejected at admission, valid bucket names are accepted, and the `chown` init container (argv form) still correctly chowns the Local hostPath so backup-restore writes full snapshots as uid 65532.

/cc @renormalize

**Release note**:
```other operator
The `spec.backup.store.container` field is now validated (allowed characters `[a-zA-Z0-9._-]`, must start and end with an alphanumeric character, min length 3, max length 63). The init container that adjusts permissions for local backups no longer uses a shell, closing a potential command-injection vector.
```

### DIFF
--- a/api/core/v1alpha1/crds/druid.gardener.cloud_etcdcopybackupstasks.yaml
+++ b/api/core/v1alpha1/crds/druid.gardener.cloud_etcdcopybackupstasks.yaml
@@ -71,6 +71,8 @@ spec:
                   container:
                     description: Container is the name of the container the backup
                       is stored at.
+                    maxLength: 63
+                    pattern: ^[a-zA-Z0-9][a-zA-Z0-9._-]{1,61}[a-zA-Z0-9]$
                     type: string
                   endpointOverride:
                     description: EndpointOverride denotes the storage endpoint that
@@ -109,6 +111,8 @@ spec:
                   container:
                     description: Container is the name of the container the backup
                       is stored at.
+                    maxLength: 63
+                    pattern: ^[a-zA-Z0-9][a-zA-Z0-9._-]{1,61}[a-zA-Z0-9]$
                     type: string
                   endpointOverride:
                     description: EndpointOverride denotes the storage endpoint that
--- a/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
+++ b/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
@@ -314,6 +314,8 @@ spec:
                       container:
                         description: Container is the name of the container the backup
                           is stored at.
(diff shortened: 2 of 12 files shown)


renormalize (approved):
Thank you!

Issue #899, opened 2024-10-28
Use Etcd druid to set tolerations, node affinity and TSC policies for HA etcd clusters


unmarshall:
**How to categorize this issue?**

/area control-plane
/area high-availability
/kind enhancement

**What would you like to be added**:
Add capability in `etcd-druid` to determine and add NodeAffinity, TSC and Tolerations to etcd StatefulSet pods for a HA etcd cluster.

**Why is this needed**:

When etcd-druid is used in gardener today then gardener-resource-manager HA webhook does the following:

- Mutates the replicas (see [code](https://github.com/gardener/gardener/blob/73b3d774efa71c400a6fe1d1f83b3cf534363fdc/pkg/resourcemanager/webhook/highavailabilityconfig/handler.go#L297))
- Mutates the node affinity (see [code](https://github.com/gardener/gardener/blob/73b3d774efa71c400a6fe1d1f83b3cf534363fdc/pkg/resourcemanager/webhook/highavailabilityconfig/handler.go#L373))
- Mutates TSC (see [code](https://github.com/gardener/gardener/blob/73b3d774efa71c400a6fe1d1f83b3cf534363fdc/pkg/resourcemanager/webhook/highavailabilityconfig/handler.go#L412))
- Mutates Toleration seconds (see [code](https://github.com/gardener/gardener/blob/73b3d774efa71c400a6fe1d1f83b3cf534363fdc/pkg/resourcemanager/webhook/highavailabilityconfig/handler.go#L455)) - This is done to allow faster recovery for HA clusters in a single zone.

Unfortunately while mutating the TSC policies the `LabelSelector` is set taking the labels from PodTemplateSpec.Labels (of the respective STS). See [code](https://github.com/gardener/gardener/blob/73b3d774efa71c400a6fe1d1f83b3cf534363fdc/pkg/resourcemanager/webhook/highavailabilityconfig/handler.go#L437). This is problematic as new labels can be added during an upgrade and all labels are not used to uniquely identify the pods belonging to a StatefulSet (i.e. an etcd cluster).

So imagine the following scenario:
**Starting State:** 
There is an non-HA etcd-cluster (replicas=1). Lets assume that the pods of a StatefulSet provisioned for the etcd cluster has the following labels:
```
app.kubernetes.io/name: etcd-test
app.kubernetes.io/managed-by: etcd-druid
app.kubernetes.io/part-of: etcd-test
app.kubernetes.io/component: statefulset
```
[...]

PR #1392, opened 2026-07-29
Add tolerations support to Etcd scheduling constraints

yeganeahmadnejad:
/area high-availability
/kind api-change

Extends `spec.schedulingConstraints` on the `Etcd` resource with a `tolerations` field, alongside the existing `affinity` and `topologySpreadConstraints`. This lets `Etcd` pods be scheduled onto nodes with matching taints (e.g. dedicated/tainted node pools for stateful workloads) without needing a `MutatingWebhook` to inject tolerations after the fact.

The field is optional and additive; when unset, behaviour is unchanged. It is passed straight through to the etcd `StatefulSet` pod spec in the same way `affinity` and `topologySpreadConstraints` already are.

Fixes# 
[issue-899](https://github.com/gardener/etcd-druid/issues/899)

**Checklist**:
    - [x] Unit tests
    - [ ] Integration tests
    - [ ] E2E tests

- Generated artifacts (`zz_generated.deepcopy.go`, both `druid.gardener.cloud_etcds*.yaml` CRDs, the copy under `charts/crds/`, and `docs/api-reference/etcd-druid-api.md`) were produced via `make --directory=api generate` and manually cross-checked for the `_without_cel.yaml` variant, since the generation script's md5-based change-detection relies on `md5sum`, which isn't present on macOS by default.
- `internal/component/statefulset/stsmatcher.go` previously had a `NOTE` explicitly opting out of asserting `affinity`/`topologySpreadConstraints` in tests. I added an explicit `Tolerations` assertion there rather than leaving it uncovered, since `MatchFields(IgnoreExtras, ...)` would otherwise silently pass regardless of the field's value.
- No changes were needed in `internal/webhook/**` (component-protection webhook doesn't inspect the spec), `api/validation/**` (covers backup-store only), or `client/**` (type-level codegen, regenerates identically).

**Release note**:
```feature operator
`spec.schedulingConstraints.tolerations` can now be set on the `Etcd` resource to schedule etcd pods onto tainted nodes.
```

### DIFF
--- a/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
+++ b/api/core/v1alpha1/crds/druid.gardener.cloud_etcds.yaml
@@ -904,7 +904,7 @@ spec:
                 description: |-
                   SchedulingConstraints defines the different scheduling constraints that must be applied to the
                   pod spec in the etcd statefulset.
-                  Currently supported constraints are Affinity and TopologySpreadConstraints.
+                  Currently supported constraints are Affinity, TopologySpreadConstraints and Tolerations.
                 properties:
                   affinity:
                     description: |-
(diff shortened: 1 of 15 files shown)


CaptainIRS (approved):
Thanks for the PR! I also wished tolerations were part of the API.

This should close https://github.com/gardener/etcd-druid/issues/899 when merged (the issue was auto-closed by the bot previously).

/ok-to-test

anveshreddy18 (approved):
Thanks for the PR. LGTM!

Holding it just so that @unmarshall can take a look before we merge it as he raised the linked issue.

unmarshall (approved):
Thanks for your contribution!
Looks good.

PR #1397, opened 2026-08-04
Add support for externally managed members

CaptainIRS:
/area control-plane
/area high-availability
/kind enhancement
/kind api-change

This pull request introduces a new mode of operation in `etcd-druid` to support etcd clusters whose members are managed by an external actor, rather than by `etcd-druid` via the `StatefulSet` controller. This is primarily to enable use cases like Gardener's [self-hosted shoot clusters (GEP-28)](https://github.com/gardener/gardener/blob/master/docs/proposals/28-self-hosted-shoot-clusters.md), where a tool like `gardenadm` is responsible for deploying and managing etcd members as static pods on the control plane nodes.

To enable this, a new field, `spec.externallyManagedMemberAddresses`, is introduced in the `Etcd` API. When this field is populated with a list of member IP addresses, `etcd-druid`'s behavior changes as follows:

*   **Decoupled Pod Management**: `etcd-druid` no longer manages the lifecycle of etcd pods. It creates the `StatefulSet` with `replicas=0` to serve as a template but does not create any pods itself.
    * The template now sets the `POD_NAME` to be `<etcd-name>-<ip-address>` for the external tool to spin up `etcd` pods with the `etcd-backup-restore` sidecar using the template.
*   **IP-based Communication**: All etcd configuration is generated to use the provided IP addresses for peer communication. This removes the dependency on Kubernetes DNS and headless services.
    *   The `initial-cluster` and `advertised-<peer|client>-urls` configuration in the `ConfigMap` is populated using the IPs from `spec.externallyManagedMemberAddresses`.
    *   Member identities are based on their IP addresses (e.g., `etcd-main-192.168.0.1`) instead of StatefulSet ordinals.
*   **Disabled Component Creation**: The creation of several components that are tied to pod lifecycle management is disabled:
    *   Client and Peer `Service`s.
    *   `PodDisruptionBudget`.
*   **Continued Management Role**: While pod management is delegated, `etcd-druid` continues to provide value by:
    *   Generating and reconciling the etcd `ConfigMap`.
    *   Providing status reporting and facilitating maintenance operations for the externally managed cluster (enabling all functionality of `etcd-backup-restore` full/delta snapshots, compaction, de-fragmentation, etc.).

This approach provides a generic mechanism to integrate `etcd-druid` with diverse deployment strategies beyond the default `StatefulSet`-based model, making it more flexible.

This work supersedes the previous approach in PR #1117, opting for a more explicit API field (`spec.externallyManagedMemberAddresses`) over an annotation, along with design changes that overcome the limitations of the previous approach.

Part of #1071
Supersedes #1117

This is just https://github.com/gardener/etcd-druid/pull/1214 reopened after rebasing on upstream master.

The core logic is triggered by the presence of the `spec.externallyManagedMemberAddresses` field. Please pay close attention to the validation rules for this new field, as they are crucial for preventing unsupported transitions:

*   The field must contain a list of valid IPv4 address strings.
*   If specified, its length must be equal to `spec.replicas`.
*   The field can be specified during the creation of a new `Etcd` resource.
*   **Mode Conversion is Not Supported**:
    *   The field **cannot** be added to an existing, running cluster or a hibernating cluster that is currently managed by `etcd-druid`
    *   The field **cannot** be removed from a running, externally managed cluster to convert it back to a `etcd-druid`-managed one.
    *   Externally managed clusters cannot be hibernated, i.e it cannot be scaled-in to zero replicas.

When `externallyManagedMemberAddresses` is set, `etcd-druid` effectively transitions from a pod lifecycle manager to a configuration and maintenance provider for an externally managed cluster.

**Release note**:
```feature operator
A new field `spec.externallyManagedMemberAddresses` has been added to the `Etcd` API. When this field is specified with a list of IP addresses, `etcd-druid` enters a new operational mode where it does not manage the etcd member pods directly. In this mode, `etcd-druid` will not create pods, services, or `PodDisruptionBudget`s. Instead, it will generate the etcd `ConfigMap` with a configuration tailored for the provided member IP addresses, enabling peer communication without relying on Kubernetes services.

This feature allows external actors, such as `gardenadm`, to manage the lifecycle of etcd members (e.g., as static pods) while still leveraging `etcd-druid` for configuration generation, status reporting, and other management tasks.
```

### DIFF
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -512,9 +512,11 @@ type SchedulingConstraints struct {
 }
 
 // EtcdSpec defines the desired state of Etcd
-// +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
-// +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.storageClass field cannot be added or removed dynamically.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate field cannot be added or removed dynamically.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
 type EtcdSpec struct {
 	// MemberNamePrefix defines the prefix for the name of each etcd cluster member. When set, the member name would be `<prefix>-<pod-name>`, otherwise it defaults to the `pod-name`.
 	// The combined length of the member-prefix, pod-name, and separator must not exceed 253 characters (DNS subdomain limit for lease names).
@@ -568,6 +570,12 @@ type EtcdSpec struct {
 	// run as root. By default, they run as non-root with user 'nobody'.
 	// +optional
 	RunAsRoot *bool `json:"runAsRoot,omitempty"`
+	// ExternallyManagedMemberAddresses defines the list of addresses of externally managed etcd members. Specifying this
+	// will disable components that are involved in management of etcd members like Pods, Services and PDBs.
+	// Allowed values include: IPv4/IPv6 addresses and hostnames. Protocol or port shall not be specified.
+	// +optional
+	// +listType=set
+	ExternallyManagedMemberAddresses []string `json:"externallyManagedMemberAddresses,omitempty"`
 }
 
 // CrossVersionObjectReference contains enough information to let you identify the referred resource.
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -515,7 +515,7 @@ type SchedulingConstraints struct {
 // +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.storageClass field cannot be added or removed dynamically.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate field cannot be added or removed dynamically.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
-// +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses length must equal to etcd.spec.replicas.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
 // +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
 type EtcdSpec struct {
 	// MemberNamePrefix defines the prefix for the name of each etcd cluster member. When set, the member name would be `<prefix>-<pod-name>`, otherwise it defaults to the `pod-name`.
--- a/api/core/v1alpha1/etcd.go
+++ b/api/core/v1alpha1/etcd.go
@@ -516,7 +516,7 @@ type SchedulingConstraints struct {
 // +kubebuilder:validation:XValidation:message="etcd.spec.storageClass field cannot be added or removed dynamically.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate field cannot be added or removed dynamically.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses length must equal to etcd.spec.replicas.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
-// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field must only be set during creation.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
 type EtcdSpec struct {
 	// MemberNamePrefix defines the prefix for the name of each etcd cluster member. When set, the member name would be `<prefix>-<pod-name>`, otherwise it defaults to the `pod-name`.
 	// The combined length of the member-prefix, pod-name, and separator must not exceed 253 characters (DNS subdomain limit for lease names).
--- a/api/core/v1alpha1/helper.go
+++ b/api/core/v1alpha1/helper.go
@@ -5,7 +5,9 @@
 package v1alpha1
 
 import (
+	"crypto/rand"
 	"fmt"
+	"math/big"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -22,6 +24,22 @@ func GetPeerServiceName(etcdObjMeta metav1.ObjectMeta) string {
 // GetClientServiceName returns the client service name for the Etcd cluster reachable by external clients.
 func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
 	return fmt.Sprintf("%s-client", etcdObjMeta.Name)
+}
+
+// GetClientHostname returns the hostname of the client endpoint for the Etcd cluster. This is the client service hostname when the Etcd members
+// are managed by etcd-druid, else it is a randomly selected member address out of the externally managed member addresses.
+func GetClientHostname(etcd *Etcd) string {
+	if ArePodsManagedByEtcdDruid(etcd) {
+		return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.Namespace)
+	} else {
+		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(etcd.Spec.ExternallyManagedMemberAddresses))))
+		if err != nil {
+			// Fallback to first member address in case of an error
+			return etcd.Spec.ExternallyManagedMemberAddresses[0]
+		}
+		randomIndex := int(n.Int64())
+		return etcd.Spec.ExternallyManagedMemberAddresses[randomIndex]
+	}
 }
 
 // GetServiceAccountName returns the service account name for the Etcd.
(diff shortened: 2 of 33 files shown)


CaptainIRS:
/cc @seshachalam-yv
to check if live CPM related functionality is working correctly after rebase.

api/core/v1alpha1/etcd.go
```diff
@@ -508,9 +508,11 @@ type SchedulingConstraints struct {
...
 // +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.storageClass field cannot be added or removed dynamically.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate field cannot be added or removed dynamically.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
+// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
```
seshachalam-yv:
```suggestion
// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses field cannot be added after the Etcd resource has been created.",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
```

seshachalam-yv: Can it be unset?
Is the intent "set only at creation"?

api/core/v1alpha1/etcd.go
```diff
@@ -508,9 +508,11 @@ type SchedulingConstraints struct {
...
-// +kubebuilder:validation:XValidation:message="etcd.spec.storageClass is an immutable field.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
-// +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate is an immutable field.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
 // +kubebuilder:validation:XValidation:message="etcd.spec.memberNamePrefix is an immutable field.",rule="has(oldSelf.memberNamePrefix) == has(self.memberNamePrefix)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.storageClass field cannot be added or removed dynamically.",rule="has(oldSelf.storageClass) ==  has(self.storageClass)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.volumeClaimTemplate field cannot be added or removed dynamically.",rule="has(oldSelf.volumeClaimTemplate) == has(self.volumeClaimTemplate)"
+// +kubebuilder:validation:XValidation:message="etcd.spec.replicas must be equal to length of etcd.spec.externallyManagedMemberAddresses when etcd.spec.externallyManagedMemberAddresses is specified.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
```
seshachalam-yv:
```suggestion
// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses length must equal etcd.spec.replicas",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
```

api/core/v1alpha1/etcd.go
```diff
@@ -564,6 +566,12 @@ type EtcdSpec struct {
 	// run as root. By default, they run as non-root with user 'nobody'.
 	// +optional
 	RunAsRoot *bool `json:"runAsRoot,omitempty"`
+	// ExternallyManagedMemberAddresses defines the list of addresses of externally managed etcd members. Specifying this
+	// will disable components that are involved in management of etcd members like Pods, Services and PDBs.
+	// Allowed values include: IPv4/IPv6 addresses and hostnames. Protocol or port shall not be specified.
+	// +optional
+	// +listType=set
+	ExternallyManagedMemberAddresses []string `json:"externallyManagedMemberAddresses,omitempty"`
```
seshachalam-yv, line 578: Can you add IP validation ? 
ref: https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-ip-address-library ?

api/core/v1alpha1/helper.go
```diff
@@ -24,6 +26,22 @@ func GetClientServiceName(etcdObjMeta metav1.ObjectMeta) string {
...
+		n, err := rand.Int(rand.Reader, big.NewInt(int64(len(etcd.Spec.ExternallyManagedMemberAddresses))))
+		if err != nil {
+			// Fallback to first member address in case of an error
+			return etcd.Spec.ExternallyManagedMemberAddresses[0]
+		}
+		randomIndex := int(n.Int64())
+		return etcd.Spec.ExternallyManagedMemberAddresses[randomIndex]
+	}
+}
+
```
seshachalam-yv, line 44:
```suggestion
// GetClientHostname returns the host to reach the Etcd cluster's client endpoint.
// For etcd-druid managed clusters it is the client Service DNS name. For externally
// managed clusters it is one of the member addresses, chosen at random on each
// call.

func GetClientHostname(etcd *Etcd) string {
      if ArePodsManagedByEtcdDruid(etcd) {
              return fmt.Sprintf("%s.%s.svc", GetClientServiceName(etcd.ObjectMeta), etcd.Namespace)
      }
      addresses := etcd.Spec.ExternallyManagedMemberAddresses
      return addresses[rand.IntN(len(addresses))]
}

```

Use `math/rand/v2`

api/core/v1alpha1/helper.go
```diff
@@ -65,12 +88,21 @@ func GetAllPodNames(etcdObjMeta metav1.ObjectMeta, replicas int32) []string {
...
+			leaseNames[i] = GetMemberName(etcd.Spec.MemberNamePrefix, podName)
+		}
+		return leaseNames
+	} else {
+		memberAddresses := etcd.Spec.ExternallyManagedMemberAddresses
+		leaseNames := make([]string, len(memberAddresses))
+		for i, memberAddress := range memberAddresses {
+			leaseNames[i] = GetMemberNameFromAddress(etcd, memberAddress)
+		}
+		return leaseNames
```
seshachalam-yv, line 104:
```suggestion
 if ArePodsManagedByEtcdDruid(etcd) {
              leaseNames := make([]string, etcd.Spec.Replicas)
              for i := range int(etcd.Spec.Replicas) {
                      podName := GetOrdinalPodName(etcd.ObjectMeta, i)
                      leaseNames[i] = GetMemberName(etcd.Spec.MemberNamePrefix, podName)
              }
              return leaseNames
      }

      memberAddresses := etcd.Spec.ExternallyManagedMemberAddresses
      leaseNames := make([]string, len(memberAddresses))
      for i, memberAddress := range memberAddresses {
              leaseNames[i] = GetMemberNameFromAddress(etcd, memberAddress)
      }
      return leaseNames
```

internal/component/memberlease/memberlease.go
```diff
@@ -87,6 +89,33 @@ func (r _resource) Sync(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd)
...
+			}
+			if errorList := utils.RunConcurrently(ctx, deleteTasks); len(errorList) > 0 {
+				for _, err := range errorList {
+					errs = multierror.Append(errs, err)
+				}
+			}
+		} else {
+			errs = multierror.Append(errs, err)
+		}
+	}
```
seshachalam-yv, line 126:
```suggestion

if !druidv1alpha1.ArePodsManagedByEtcdDruid(etcd) {
      if err := r.deleteStaleMemberLeases(ctx, etcd); err != nil {
              errs = multierror.Append(errs, err)
      }
}
```

Move the removal logic into a separate function so that the code is more readable.

internal/utils/envvar.go
```diff
@@ -49,10 +51,16 @@ func GetEnvVarFromSecret(name, secretName, secretKey string, optional bool) core
 }
 
 // GetBackupRestoreContainerEnvVars returns non-provider-specific environment variables for the backup-restore container.
-func GetBackupRestoreContainerEnvVars(store *druidv1alpha1.StoreSpec) ([]corev1.EnvVar, error) {
+func GetBackupRestoreContainerEnvVars(etcd *druidv1alpha1.Etcd, store *druidv1alpha1.StoreSpec) ([]corev1.EnvVar, error) {
```
seshachalam-yv: why do you want subset store if we are passing etcd itself ?

docs/concepts/externally-managed-members.md
```diff
@@ -0,0 +1,42 @@
...
+```yaml
+apiVersion: druid.gardener.cloud/v1alpha1
+kind: Etcd
+metadata:
+  name: etcd-main
+spec:
+  externallyManagedMemberAddresses:
+    - 192.168.0.1
+    - 192.168.0.2
+    - 192.168.0.3
```
seshachalam-yv, line 15: Can you add it to examples ?

seshachalam-yv (changes requested):

seshachalam-yv:
> /cc @seshachalam-yv to check if live CPM related functionality is working correctly after rebase.

Yes, it is working as expected.

api/core/v1alpha1/etcd.go
CaptainIRS: No, it cannot be unset. So, the intent is not just "set only at creation", but rather also includes the case of the field being removed after being set.

CaptainIRS: Addressed in https://github.com/gardener/etcd-druid/pull/1397#discussion_r3729891728

api/core/v1alpha1/etcd.go
CaptainIRS: The statement is true only when `etcd.spec.externallyManagedMemberAddresses` is set, so I think the original comment is correct.

api/core/v1alpha1/etcd.go
CaptainIRS: The IP validation is not done intentionally. The reasoning is in https://github.com/gardener/etcd-druid/pull/1214#discussion_r3063761846.

api/core/v1alpha1/helper.go
CaptainIRS: I had tried this previously, but this triggers SAST lint errors because of G404: https://github.com/securego/gosec/blob/master/RULES.md#g4xx-crypto-and-protocol-security.

api/core/v1alpha1/helper.go
CaptainIRS: The helpers in this file are written with the else block explicitly to separate logic between druid-managed and externally-managed modes (this was addressed previously in https://github.com/gardener/etcd-druid/pull/1214#discussion_r3063859206).

internal/component/memberlease/memberlease.go
CaptainIRS: Addressed in 65512afa.

internal/utils/envvar.go
CaptainIRS: Addressed in 65512afa.

docs/concepts/externally-managed-members.md
CaptainIRS: Do you mean a file in the examples directory? I'm not sure that would be useful though, since the examples are generally valid without any modifications, whereas in this case the correct IPs have to be provided by the user.

api/core/v1alpha1/etcd.go
seshachalam-yv:
```suggestion
// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses is immutable and may only be set during creation",rule="has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)"
```
The current message, `"cannot be added or removed dynamically"`, is a bit confusing. If the field can't be added after creation, then saying "cannot be added" without clarifying "after creation" can be interpreted as meaning it can never be added at all.

>[!IMPORTANT]
> Also, this validation only references `self.externallyManagedMemberAddresses` and doesn't depend on any other field in the spec, so it doesn't need to be defined at the spec level. It would be better to move it to the field level.

seshachalam-yv: see : https://github.com/gardener/etcd-druid/pull/1397#discussion_r3733943391

api/core/v1alpha1/etcd.go
seshachalam-yv: Agreed the rule only fires when `externallyManagedMemberAddresses` is set. Since a user hits this error while editing that field, making it the subject reads as more actionable — and repeating "when etcd.spec.externallyManagedMemberAddresses is specified" is redundant. So more precisely: *this field's length must equal replicas.*


```suggestion
// +kubebuilder:validation:XValidation:message="etcd.spec.externallyManagedMemberAddresses length must equal to etcd.spec.replicas.",rule="has(self.externallyManagedMemberAddresses) ? self.replicas == self.externallyManagedMemberAddresses.size() : true"
```

docs/concepts/externally-managed-members.md
seshachalam-yv: Yes — I meant a commented-out placeholder in examples/etcd/*.yaml, not a live value. That file already documents optional/user-specific fields as comments so the example still applies cleanly by default — e.g. the store.secretRef/prefix and the peerUrlTls blocks are all commented with a note.

CaptainIRS: Addressed in 3eac7441.

api/core/v1alpha1/etcd.go
CaptainIRS: Addressed in 3eac7441.

seshachalam-yv (approved):
Thanks @CaptainIRS for addressing my review comments ❤️  
/lgtm

api/core/v1alpha1/etcd.go
CaptainIRS: The field is mutable, so the suggested comment would be incorrect. The existing comment `"etcd.spec.externallyManagedMemberAddresses field cannot be added or removed dynamically."` is written from a usage perspective - if someone tries to add the field or remove the field after creation (which would be the only cases where this rule is triggered), the validation error will state that the field cannot be added/removed dynamically.

Also regarding
> Also, this validation only references self.externallyManagedMemberAddresses and doesn't depend on any other field in the spec, so it doesn't need to be defined at the spec level. It would be better to move it to the field level.

If it is moved to the field level, the validation would become `oldSelf == self`, which is different from `has(oldSelf.externallyManagedMemberAddresses) == has(self.externallyManagedMemberAddresses)` in terms of what is being checked - the existing one is preventing set/unset, the suggested one is enforcing immutability (which we don't want as the field is mutable).

seshachalam-yv (changes requested):
/lgtm

seshachalam-yv (approved):

anveshreddy18 (approved):

PR #1408, opened 2026-08-24
Add `AdditionalAdvertisePeerURLs.OverrideDefaultURL` to suppress default peer service URL

seshachalam-yv:

/area high-availability
/area disaster-recovery
/kind api-change

Adds the `overrideDefaultURL` flag to `spec.etcd.additionalAdvertisePeerURLs`, enabling etcd members to advertise **only** their additional peer URLs while suppressing the default internal peer service URL. This is required for cross-cluster bootstrap scenarios where two clusters share the same `Etcd` resource name and their internal peer service DNS names would otherwise collide.

Fixes #

**Checklist**:
- [ x] Add tests that cover your changes (if applicable)
    - [x] Unit tests
    - [x] Integration tests
    - [ ] E2E tests

API Breaking Change : The field structure changed from flat array to object:
- Old: `spec.etcd.additionalAdvertisePeerURLs: [{memberName: "...", urls: [...]}]`
- New: `spec.etcd.additionalAdvertisePeerURLs: {overrideDefaultURL: false, members: [{name: "...", urls: [...]}]}`

- CEL validation rules updated to iterate `.members.all(m, ...)` and reference `m.name`

**Release note**:

```breaking operator
AdditionalAdvertisePeerURLs field structure changed from array to object with OverrideDefaultURL flag. Existing manifests must be updated: migrate `additionalAdvertisePeerURLs: [{memberName, urls}]` → `additionalAdvertisePeerURLs: {overrideDefaultURL: false, members: [{name, urls}]}`. This enables cross-cluster bootstrap where members advertise only external URLs without internal peer service DNS.
```


api/core/v1alpha1/etcd.go
```diff
@@ -368,9 +368,31 @@ type ClientService struct {
...
+// whether the default internal peer service URL is suppressed for the configured members.
+type AdditionalPeerURLsSpec struct {
+	// OverrideDefaultURL, when true, makes each configured member advertise only
+	// its listed URLs and omit the default internal peer service URL. This is
+	// needed when two clusters share an Etcd resource name and their internal
+	// peer service DNS would otherwise collide.
+	// Defaults to false, meaning the listed URLs are appended to the internal service URL.
+	// +optional
+	// +kubebuilder:default=false
+	OverrideDefaultURL *bool `json:"overrideDefaultURL,omitempty"`
```
Shreyas-s14, line 377: Just a question, why don't we have this by default? Even if the name is different we could have the advertised url be the only source of truth right? Is there any case I'm missing here?

api/core/v1alpha1/etcd.go
```diff
@@ -309,17 +309,17 @@ type EtcdConfig struct {
 	// ClientUrlTLS contains the ca, server TLS and client TLS secrets for client communication to ETCD cluster
 	// +optional
 	ClientUrlTLS *TLSConfig `json:"clientUrlTls,omitempty"`
-	// AdditionalAdvertisePeerURLs contains extra per-member peer URLs to append
-	// to initial-advertise-peer-urls. Each entry maps a member name to its
+	// AdditionalAdvertisePeerURLs holds extra per-member peer URLs added to
+	// initial-advertise-peer-urls, and a flag controlling whether the default
+	// internal peer service URL is suppressed. Each entry maps a member name to its
 	// additional URLs. The member name must follow the pattern {etcd-name}-{index}
```
Shreyas-s14: Does this need to be mentioned in the docstring for this as well since it is covered in line 384 for the member name field.

api/core/v1alpha1/etcd.go
```diff
@@ -379,10 +401,12 @@ type MemberPeerURLs struct {
...
 	// +kubebuilder:validation:MaxLength=253
 	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$`
-	MemberName string `json:"memberName"`
+	Name string `json:"name"`
 
 	// URLs is a list of additional peer URLs for this member.
-	// These will be appended to the default internal service URL.
+	// These are appended to the member's internal peer service URL, unless
+	// AdditionalPeerURLsSpec.OverrideDefaultURL is true, in which case only
+	// these URLs are advertised.
```
Shreyas-s14, line 406: Are we allowing the case for this url to be mutable?

api/core/v1alpha1/etcd.go
```diff
@@ -368,9 +368,31 @@ type ClientService struct {
...
+	// +kubebuilder:default=false
+	OverrideDefaultURL *bool `json:"overrideDefaultURL,omitempty"`
+
+	// Members contains per-member additional peer URLs.
+	// Member names must follow the pattern {etcd-name}-{index}, where index is
+	// 0 to (replicas-1) (e.g. etcd-main-0). When spec.memberNamePrefix is set,
+	// member names become `<memberNamePrefix>-<podName>`.
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=10
```
Shreyas-s14, line 385: The member names for bootstrapWithExistingCluster has a validation for each entry to be atomic. Should it not be enforced here as well?
```go
	// +listType=atomic
```

api/core/v1alpha1/helper.go
```diff
@@ -77,6 +77,36 @@ func GetMemberNameFromAddress(etcd *Etcd, memberAddress string) string {
...
+	return etcd.Spec.Etcd.AdditionalAdvertisePeerURLs != nil &&
+		len(etcd.Spec.Etcd.AdditionalAdvertisePeerURLs.Members) > 0
+}
+
+// IsOverrideDefaultURLEnabled returns true if additionalAdvertisePeerURLs is set
+// with its OverrideDefaultURL flag explicitly true.
+func IsOverrideDefaultURLEnabled(etcd *Etcd) bool {
+	return etcd.Spec.Etcd.AdditionalAdvertisePeerURLs != nil &&
+		ptr.Deref(etcd.Spec.Etcd.AdditionalAdvertisePeerURLs.OverrideDefaultURL, false)
+}
```
Shreyas-s14, line 92: These helpers are not being used anywhere in the repo right? So are they needed?

docs/usage/using-additional-advertise-peer-urls.md
```diff
@@ -212,13 +224,40 @@ You do not need to configure additional URLs for every member. Only members that
...
+            - http://10.0.0.1:2380
+        - name: etcd-main-1
+          urls:
+            - http://10.0.0.2:2380
+        - name: etcd-main-2
+          urls:
+            - http://10.0.0.3:2380
+```
+
+With the above, `etcd-main-0`'s `initial-advertise-peer-urls` contains only `http://10.0.0.1:2380` — the internal `http://etcd-main-0.etcd-main-peer.default.svc:2380` URL is dropped.
```
Shreyas-s14, line 256: Just curious. What would happen if the initial cluster has the override flag and the joining cluster defines the urls but does not have this set?

internal/component/configmap/etcdconfig.go
```diff
@@ -155,7 +155,11 @@ func prepareInitialCluster(etcd *druidv1alpha1.Etcd, peerScheme string) string {
 		for i := range int(etcd.Spec.Replicas) {
 			podName := druidv1alpha1.GetOrdinalPodName(etcd.ObjectMeta, i)
 			memberName := druidv1alpha1.GetMemberName(etcd.Spec.MemberNamePrefix, podName)
-			fmt.Fprintf(&builder, "%s=%s://%s.%s:%s,", memberName, peerScheme, podName, domainName, serverPort)
+			_, overrideDefault := druidv1alpha1.GetAdditionalAdvertisePeerURLs(etcd, podName)
+			// Emit the default peer service URL unless the member overrides it.
+			if !overrideDefault {
```
Shreyas-s14: Here the helper function for isOverride set (which i adked to be removed in my other comment) can be used.

internal/component/configmap/etcdconfig.go
```diff
@@ -167,13 +171,9 @@ func prepareInitialCluster(etcd *druidv1alpha1.Etcd, peerScheme string) string {
...
 		podName := druidv1alpha1.GetOrdinalPodName(etcd.ObjectMeta, i)
 		memberName := druidv1alpha1.GetMemberName(etcd.Spec.MemberNamePrefix, podName)
-		for _, memberURLs := range etcd.Spec.Etcd.AdditionalAdvertisePeerURLs {
-			if memberURLs.MemberName == podName {
-				for _, url := range memberURLs.URLs {
-					fmt.Fprintf(&builder, "%s=%s,", memberName, url)
-				}
-				break
-			}
+		additionalURLs, _ := druidv1alpha1.GetAdditionalAdvertisePeerURLs(etcd, podName)
```
Shreyas-s14: The function `GetAdditionalAdvertisePeerURLs` is called twice. Can you move the call before the `for` blocks and use the same returned values for these?

internal/component/configmap/configmap_test.go
```diff
@@ -939,6 +1009,99 @@ func TestPeerSkipClientSANVerification(t *testing.T) {
...
+			for i, u := range member0URLs {
+				urlStrings[i] = u.(string)
+			}
+
+			if tc.wantDefaultURL {
+				g.Expect(urlStrings).To(ContainElement(peerSvcURL), "default internal service URL must be present")
+			} else {
+				g.Expect(urlStrings).NotTo(ContainElement(peerSvcURL), "default internal service URL must be absent")
+			}
+			if tc.wantAdditionalURL {
```
Shreyas-s14, line 1101: This is set as true in all cases. Is there a case where this will be false?

Shreyas-s14 (commented):
Thanks for fixing the bug. Have a few comments. PTAL

api/core/v1alpha1/etcd.go
seshachalam-yv: I think the default should remain `false` for backward compatibility. `AdditionalAdvertisePeerURLs` already exists in the API, and existing users may be relying on it to append additional URLs to the default internal peer service URL (for example, for cross-zone reachability). If we default this to `true`, upgrading would silently change the advertised peer URLs by suppressing the internal peer service URL, which could break existing clusters.

Also, since `OverrideDefaultURL` is an optional field and changes the existing behavior, I think it should be explicitly opt-in. Users who need to suppress the default internal peer service URL can explicitly set `overrideDefaultURL: true`; otherwise, the existing behavior should be preserved.

api/core/v1alpha1/helper.go
seshachalam-yv: IsAdditionalPeerURLConfigured (line 82) is called at helper.go:98 inside GetAdditionalAdvertisePeerURLs. IsOverrideDefaultURLEnabled (line 89) is called at helper.go:104 inside the same function.

api/core/v1alpha1/etcd.go
seshachalam-yv: Mutability is intentional.

docs/usage/using-additional-advertise-peer-urls.md
seshachalam-yv: It will not join the cluster with error saying "not matching initial cluster urls "

internal/component/configmap/configmap_test.go
seshachalam-yv: Fixed as part of [90e2a83](https://github.com/gardener/etcd-druid/pull/1408/commits/90e2a838ed154c3c0b7b8bb4c409f21dc74a7267)

internal/component/configmap/etcdconfig.go
seshachalam-yv: Fixed as part of [90e2a83](https://github.com/gardener/etcd-druid/pull/1408/commits/90e2a838ed154c3c0b7b8bb4c409f21dc74a7267)

internal/component/configmap/etcdconfig.go
seshachalam-yv: Fixed as part of [90e2a83](https://github.com/gardener/etcd-druid/pull/1408/commits/90e2a838ed154c3c0b7b8bb4c409f21dc74a7267)

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as part of [90e2a83](https://github.com/gardener/etcd-druid/pull/1408/commits/90e2a838ed154c3c0b7b8bb4c409f21dc74a7267)

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as part of [90e2a83](https://github.com/gardener/etcd-druid/pull/1408/commits/90e2a838ed154c3c0b7b8bb4c409f21dc74a7267)

docs/usage/using-additional-advertise-peer-urls.md
seshachalam-yv: When the source cluster has overrideDefaultURL: true, its members advertise only the configured peer URLs — the internal k8s DNS peer service URL is suppressed. When configuring bootstrapWithExistingCluster on the destination, you must provide the source members' configured peer URLs (e.g., the LoadBalancer IPs) in members[*].peerUrls, not their internal DNS names. If you mistakenly provide the suppressed internal DNS URLs, the destination will fail to reach the source and the join will not succeed. The overrideDefaultURL flag on the destination is a separate concern and does not affect this.

Shreyas-s14 (approved):
Thank you for addressing the review comments and explaining the scenarios. 
LGTM

api/core/v1alpha1/etcd.go
```diff
@@ -368,9 +364,32 @@ type ClientService struct {
...
+type AdditionalPeerURLsSpec struct {
+	// OverrideDefaultURL, when true, makes each configured member advertise only
+	// its listed URLs and omit the default internal peer service URL. This is
+	// needed when two clusters share an Etcd resource name and their internal
+	// peer service DNS would otherwise collide.
+	// Defaults to false, meaning the listed URLs are appended to the internal service URL.
+	// +optional
+	// +kubebuilder:default=false
+	OverrideDefaultURL *bool `json:"overrideDefaultURL,omitempty"`
+
```
anveshreddy18, line 378: can we have a note mentioning that the flag is only honored for members that have URLs configured, like the one in the #1425 PR
like:

```
// Note: this flag is only honored for members that have URLs configured under `members`.
// Members without a matching entry always use the default internal URL, regardless of
// this flag.
```

internal/component/configmap/etcdconfig.go
```diff
@@ -204,27 +197,28 @@ func getAdvertiseURLs(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme, peerSv
...
 			podName := druidv1alpha1.GetOrdinalPodName(etcd.ObjectMeta, i)
 			memberName := druidv1alpha1.GetMemberName(etcd.Spec.MemberNamePrefix, podName)
-			advUrlsMap[memberName] = []string{fmt.Sprintf("%s://%s.%s:%d", scheme, podName, domainName, port)}
+			defaultURL := fmt.Sprintf("%s://%s.%s:%d", scheme, podName, domainName, port)
+
+			// Add additional peer URLs only for peer advertise URLs. When the
+			// member sets OverrideDefaultURL, advertise only those URLs and drop
+			// the default peer service URL.
+			var urls []string
+			if advertiseURLType == advertiseURLTypePeer {
```
anveshreddy18, line 206: Lot of common code to that of #1425 PR, this needs a rebase once the other one gets merged.

anveshreddy18 (commented):
Thanks for the PR, looks good but just some consistency issues I see with the other #1425 PR. PTAL

internal/component/configmap/etcdconfig.go
seshachalam-yv: I will do it 👍

seshachalam-yv:
Closing this PR to consolidate the changes in #1438, as suggested in the review.

The API has now been unified for advertising both client and peer URLs, including the `overrideDefaultURL` handling, in a single implementation.

ishan16696:
This PR can be closed now in favour of this PR https://github.com/gardener/etcd-druid/pull/1438, refer to this comment: https://github.com/gardener/etcd-druid/pull/1408#issuecomment-5678647910
/close

PR #1409, opened 2026-08-24
feat(e2e): enable `UpgradeEtcdVersion` feature gate by default in local e2e testing

seshachalam-yv:

/area testing
/kind enhancement

Enables the `UpgradeEtcdVersion` feature gate by default when running local e2e tests via `make ci-e2e-kind`.

Fixes #

**Checklist**:
- [ ] Update documentation in the `/docs` folder (if applicable)
    - [ ] Unit tests
    - [ ] Integration tests
    - [ ] E2E tests

**Release note**:

```other developer
Enable `UpgradeEtcdVersion` feature gate by default in local e2e testing via `make ci-e2e-kind`.
```

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -63,8 +63,14 @@ profiles:
       - env: DRUID_E2E_TEST=true
     patches:
       - op: add
+        path: /deploy/helm/releases/0/setValues/featureGates
+        value:
+          UpgradeEtcdVersion: true
+      - op: add
         path: /deploy/helm/releases/0/setValues/operatorConfig
         value:
+          featureGates:
+            UpgradeEtcdVersion: true
           controllers.etcd.concurrentSyncs: 10
           controllers.etcd.etcdStatusSyncPeriod: 5s
           controllers.compaction.eventsThreshold: "15"
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -63,10 +63,6 @@ profiles:
       - env: DRUID_E2E_TEST=true
     patches:
       - op: add
-        path: /deploy/helm/releases/0/setValues/featureGates
-        value:
-          UpgradeEtcdVersion: true
-      - op: add
         path: /deploy/helm/releases/0/setValues/operatorConfig
         value:
           featureGates:


skaffold.yaml
CaptainIRS, line 64: Could we instead add the values here? That way it is useful for testing before raising a PR, without additional steps.

seshachalam-yv: Fixed as part of [8dec8b8](https://github.com/gardener/etcd-druid/pull/1409/commits/8dec8b85dbd7c0ba428369d9804b0ebfc7e54c48)

skaffold.yaml
```diff
@@ -89,6 +93,7 @@ profiles:
         path: /deploy/helm/releases/0/setValues/operatorConfig
         value:
           webhooks.etcdComponentProtection.enabled: true
+          featureGates.UpgradeEtcdVersion: true
```
CaptainIRS: I'm not sure if this should be added to the `ENABLE_ETCD_COMPONENT_PROTECTION_WEBHOOK` activation (sounds a bit hacky). I think the problem could be solved by setting the `add` patches in `e2e-test` to be `/deploy/helm/releases/0/setValues/operatorConfig/controllers` and patches in `enable-etcdcomponentprotection-webhook` to be `/deploy/helm/releases/0/setValues/operatorConfig/webhooks` instead of the parent field `/deploy/helm/releases/0/setValues/operatorConfig`. This way we can avoid the overwrite.

CaptainIRS (changes requested):

skaffold.yaml
seshachalam-yv: This change is not required. I've reverted it.

skaffold.yaml
```diff
@@ -62,6 +62,10 @@ profiles:
     activation:
       - env: DRUID_E2E_TEST=true
     patches:
+      - op: add
+        path: /deploy/helm/releases/0/setValues/featureGates
+        value:
+          UpgradeEtcdVersion: true
       - op: add
```
CaptainIRS: The druid logs say `"featureGates":{"UpgradeEtcdVersion":false}` after I run `make deploy DRUID_E2E_TEST=true`, so I don't think this works (since this only changes the helm chart values which affect CLI flags, whereas the deployment uses operator config). Could you please check if you're able to reproduce the issue?

CaptainIRS: I think the way around this would be https://github.com/gardener/etcd-druid/pull/1409#discussion_r3841955535.

CaptainIRS (changes requested):

skaffold.yaml
seshachalam-yv: I've addressed this in 6a48d20 (https://github.com/seshachalam-yv/etcd-druid/commit/6a48d2067) by setting the feature gate via operatorConfig

seshachalam-yv: Validated 
```bash
➜  e2e-upgrade-version-default git:(feat/e2e-upgrade-version-default)  DRUID_E2E_TEST=true ENABLE_ETCD_COMPONENT_PROTECTION_WEBHOOK=true $SKAFFOLD render -m etcd-druid -n etcd-druid-e2e 2>&1 | grep UpgradeEtcdVersion -B3
        exemptServiceAccounts:
          - system:serviceaccount:kube-system:generic-garbage-collector
    featureGates:
      UpgradeEtcdVersion: true

```

CaptainIRS (approved):

Issue #1423, opened 2026-08-31
☂️ Support for Multi-node Scale-In


seshachalam-yv:
**How to categorize this issue?**

/area control-plane
/area high-availability
/kind epic

**What would you like to be added**:

Umbrella issue tracking all changes needed to support quorum-safe scale-in of multi-node etcd clusters managed by `etcd-druid`

## Motivation
Today `etcd-druid` blocks any decrease of `etcd.spec.replicas` other than to zero, so operators have no declarative way to shrink a multi-node etcd cluster.
DEP-08 introduces a declarative, quorum-safe scale-in path. 

Three concrete use cases are blocked today:

1. **Shrinking an over-provisioned HA cluster** — operators cannot reduce `etcd.spec.replicas` (e.g. `5 → 3`) after temporary scale-out for maintenance or cost reduction.
2. **Live control-plane migration** — once target members have joined via `bootstrapWithExistingCluster`, the source members cannot be decommissioned declaratively. This unblocks the final decommission step of [GEP-0039 Live Control Plane Migration](https://github.com/gardener/enhancements/tree/main/geps/0039-live-control-plane-migration).
3. **Zero-downtime single-node maintenance** — the `1 → 3 → 1` pattern (scale out to absorb disruption, scale back in) is blocked at the `3 → 1` step.

**Tasks**
- [x] https://github.com/gardener/etcd-druid/pull/1369
- [ ] https://github.com/gardener/etcd-backup-restore/pull/1058
- [ ] https://github.com/gardener/etcd-druid/pull/1425
- [ ] https://github.com/gardener/etcd-druid/pull/1426

PR #1425, opened 2026-08-31
[DEP-08] feat(api): add additionalAdvertiseClientURLs support

### LINKED ISSUE #1424: (DEP-08) feat: add additionalAdvertiseClientURLs support

/area control-plane
/area control-plane-migration
/kind api-change

Add `spec.etcd.additionalAdvertiseClientURLs` to the `Etcd` API, allowing operators to configure per-member extra client URLs that are advertised alongside (or instead of) the default internal Kubernetes service URL.

### API

```yaml
spec:
  etcd:
    additionalAdvertiseClientURLs:
      overrideDefaultURL: false   # when true, suppresses the default internal svc DNS URL
      members:
        - memberName: etcd-main-0
          urls:
            - https://10.0.0.1:2379
        - memberName: etcd-main-1
          urls:
            - https://10.0.0.2:2379
```

**Why is this needed**:
This is a prerequisite for DEP-08 scale-in support.

During a live control-plane migration, the target etcd-druid decommissions source members one per reconcile cycle using the quorum-safe BootstrapMembersRemoval flow. Before removing each source member, the controller must:

1. Check health and identify the leader — to order removals correctly (learners first, non-leader voters next, leader last) and verify the cluster can safely lose one more member without breaking quorum.
2. Call MemberRemove on the source cluster's etcd client endpoint.

Both require the target etcd-druid to reach source cluster members via a routable address. Source members only advertise their internal Kubernetes cluster DNS names by default — these are unreachable from the target cluster's network.

`additionalAdvertiseClientURLs` on the source Etcd resource lets operators configure LoadBalancer or NodePort URLs for each source member before the removal sequence starts. The target etcd-druid reads these advertised URLs from etcd.status (via advertise-client-urls in the etcd ConfigMap) to connect to the source cluster, inspect membership health, determine the leader, and perform each MemberRemove in quorum-safe order.

Without this, the target controller has no routable path to the source cluster and cannot safely drive the decommission sequence.

seshachalam-yv:

/area control-plane-migration
/area control-plane
/kind api-change

Adds `spec.etcd.additionalAdvertiseClientURLs` to the `Etcd` API, allowing operators to configure per-member extra client URLs that are advertised alongside (or instead of) the default internal Kubernetes service DNS URL.

Fixes #1424
Part of #1237 #1423

**Checklist**:
    - [x] Unit tests
    - [ ] Integration tests
    - [ ] E2E tests

**Release note**:

```feature operator
Add `spec.etcd.additionalAdvertiseClientURLs` to the `Etcd` API. Operators can now configure per-member external client URLs (e.g. LoadBalancer IPs) to be advertised alongside or instead of the default internal Kubernetes service URL.
The `overrideDefaultURL` flag suppresses the default URL when external-only advertisement is required (e.g. during live control-plane migration source-member decommission).
```


api/core/v1alpha1/etcd.go
```diff
@@ -395,6 +409,57 @@ type MemberPeerURLs struct {
...
+	// MemberName is the etcd member name.
+	// Must match the etcd member name of the cluster (e.g., etcd-main-0).
+	// When spec.memberNamePrefix is set, the member name becomes
+	// `<memberNamePrefix>-<podName>`. The top-level CEL rules on
+	// Etcd already incorporate the prefix when validating these names.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?-[0-9]+$`
+	MemberName string `json:"memberName"`
```
anveshreddy18: Can we make the naming of this field consistent across both the `additionalAdvertiseClientURLs.MemberClientURLs` and `additionalAdvertisePeerURLs.MemberPeerURLs` because the [PR#1408](https://github.com/gardener/etcd-druid/pull/1408) seems to be diverging from this convention as it uses `Name` and this PR uses `MemberName`

api/core/v1alpha1/helper.go
```diff
@@ -77,6 +77,36 @@ func GetMemberNameFromAddress(etcd *Etcd, memberAddress string) string {
...
+}
+
+// GetAdditionalAdvertiseClientURLs returns the additional client URLs for the member
+// backing podName, and whether that member suppresses the default internal client
+// service URL. The lookup accounts for spec.memberNamePrefix.
+func GetAdditionalAdvertiseClientURLs(etcd *Etcd, podName string) ([]string, bool) {
+	if !IsAdditionalClientURLConfigured(etcd) {
+		return nil, false
+	}
+	memberName := GetMemberName(etcd.Spec.MemberNamePrefix, podName)
```
anveshreddy18, line 101: The current peer code uses raw pod name for the check like `memberURLs.MemberName == podName` but I see that it is now being changed to the same convention as this in the https://github.com/gardener/etcd-druid/pull/1408 PR. Just want to confirm whether these two PRs are planned to be going together or not, if not that might be problematic.

internal/component/configmap/etcdconfig.go
```diff
@@ -204,7 +204,19 @@ func getAdvertiseURLs(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme, peerSv
 		for i := range int(etcd.Spec.Replicas) {
 			podName := druidv1alpha1.GetOrdinalPodName(etcd.ObjectMeta, i)
 			memberName := druidv1alpha1.GetMemberName(etcd.Spec.MemberNamePrefix, podName)
-			advUrlsMap[memberName] = []string{fmt.Sprintf("%s://%s.%s:%d", scheme, podName, domainName, port)}
+			defaultURL := fmt.Sprintf("%s://%s.%s:%d", scheme, podName, domainName, port)
+
+			var urls []string
+			if advertiseURLType == advertiseURLTypeClient {
+				additionalURLs, overrideDefault := druidv1alpha1.GetAdditionalAdvertiseClientURLs(etcd, podName)
```
anveshreddy18, line 211: can this behaviour of overrideDefault being set to `false` when there are no additionalURLs configured be documented well in the doc & in the `overrideDefault` API surface (that it'll only be honored when additionalURLs are set, if not, then the overrideDefault is infered as false in the code ).

api/core/v1alpha1/etcd.go
```diff
@@ -395,6 +409,57 @@ type MemberPeerURLs struct {
...
+// whether the default internal client service URL is suppressed for the configured members.
+type AdditionalClientURLsSpec struct {
+	// OverrideDefaultURL, when true, makes each configured member advertise only
+	// its listed URLs and omit the default internal client service URL. This is
+	// needed when two clusters share an Etcd resource name and their internal
+	// client service DNS would otherwise collide.
+	// Defaults to false, meaning the listed URLs are appended to the internal service URL.
+	// +optional
+	// +kubebuilder:default=false
+	OverrideDefaultURL *bool `json:"overrideDefaultURL,omitempty"`
```
anveshreddy18, line 470: This is where I would like you to explain the above behaviour.

internal/component/configmap/etcdconfig.go
```diff
@@ -204,7 +204,19 @@ func getAdvertiseURLs(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme, peerSv
...
+				if !overrideDefault {
+					urls = append(urls, defaultURL)
+				}
+				urls = append(urls, additionalURLs...)
+			} else {
```
anveshreddy18, line 216: I see that the changes are made to only the `ArePodsManagedByEtcdDruid` branch, what happens if the pods aren't managed by etcd-druid? shouldn't we still be setting the config map correctly as it might be used by the static pod translator or some other actor in the self-hosted story? Just curious to know whether self-hosted story is considered for this.

anveshreddy18 (changes requested):
Thanks for the PR. Can you also rebase it on the latest master before further reviews.

api/core/v1alpha1/helper.go
seshachalam-yv: No, they don't need to go together.
Both independent PRs when configured AdditionalAdvertiseClientURLs or AdditionalAdvertisePeerURLs

internal/component/configmap/etcdconfig.go
seshachalam-yv: The behavior is already handled in GetAdditionalAdvertiseClientURLs — when no member entry is found for the pod, it returns (nil, false) regardless of the OverrideDefaultURL flag, so the default URL is always included for unconfigured members. Added a unit test to explicitly cover this case: overrideDefaultURL=true with no matching member entry still returns override=false.

internal/component/configmap/etcdconfig.go
seshachalam-yv: Yes, this is intentional. Externally managed members are configured via member address IPs (set in spec.externallyManagedMemberAddresses), not pod names. The additionalAdvertiseClientURLs field identifies members by name (e.g. etcd-main-0), which is only meaningful for druid-managed pods. For externally managed members, the addresses provided in spec.externallyManagedMemberAddresses already represent the actual reachable endpoints, so there is no default internal DNS URL to override or supplement. Supporting additionalAdvertiseClientURLs for the externally managed case would require a different member identity scheme and is not in scope for this PR.

api/core/v1alpha1/etcd.go
seshachalam-yv: Fixed as part of https://github.com/gardener/etcd-druid/pull/1425/commits/d00204e7c07b68579c41327a950cc1ef1b90ef29

internal/component/configmap/etcdconfig.go
anveshreddy18: Thanks for adding the unit-test.

What I meant originally was to document in the API surface that when there are no additional advertise urls, then even if they set the `overrideDefault` to `true`, the default URL will still be used. 

I see that now the API has it, can you also add it in the doc here: https://github.com/seshachalam-yv/etcd-druid/blob/feat/additional-advertise-client-urls/docs/usage/using-additional-advertise-client-urls.md#overridedefaulturl in the `overrideDefault` section.

internal/component/configmap/etcdconfig.go
anveshreddy18: Got it, thanks for the clarification.

Can we also document this in the API & doc properly so that in future if anyone questions whether externally managed members is considered, you can show that and having it in the [API here](https://github.com/seshachalam-yv/etcd-druid/blob/8a2d6910f829c5456d1e4e1ca39b11a2551a682d/api/core/v1alpha1/etcd.go#L327-L336) also clears up confusion in the future. Wdyt?

docs/usage/using-additional-advertise-client-urls.md
anveshreddy18, line 1: This doc needs to be added to the mkdocs.yml file under the usage section as a sibling to the point `- Using Additional Advertise Peer URLs: usage/using-additional-advertise-peer-urls.md`

anveshreddy18 (changes requested):

docs/usage/using-additional-advertise-client-urls.md
seshachalam-yv: Fixed as part of [169e8a0](https://github.com/gardener/etcd-druid/pull/1425/commits/169e8a042cb5e8dc7841883556cde4cd1cb4dca4)

api/core/v1alpha1/etcd.go
```diff
@@ -395,6 +412,59 @@ type MemberPeerURLs struct {
...
+	// this flag.
+	// +optional
+	// +kubebuilder:default=false
+	OverrideDefaultURL *bool `json:"overrideDefaultURL,omitempty"`
+
+	// Members contains per-member additional client URLs.
+	// Member names must follow the pattern {etcd-name}-{index}, where index is
+	// 0 to (replicas-1) (e.g. etcd-main-0). When spec.memberNamePrefix is set,
+	// member names become `<memberNamePrefix>-<podName>`.
+	// +required
```
anveshreddy18, line 476: Shouldn't we have the `+listType=atomic` field on the `Members` field similar to how it's done in #1408 ?

api/core/v1alpha1/helper_test.go
```diff
@@ -565,6 +565,116 @@ func TestIsPodManagementEnabled(t *testing.T) {
 	}
 }
 
+func TestGetAdditionalAdvertiseClientURLs(t *testing.T) {
```
anveshreddy18, line 661: Can we add the tests `TestIsOverrideDefaultURLEnabled` and `TestIsAdditionalClientURLConfigured` which you've implemented in the #1408 PR?

seshachalam-yv: Fixed as part of a2c9cfe9754815687292d97acb410f61f541f4dd

api/core/v1alpha1/etcd.go
seshachalam-yv: Yes added

anveshreddy18:
/hold

To not let the bot auto-merge until we have one more reviewer's approval.

anveshreddy18 (approved):
Thanks for addressing my comments. LGTM 🚀

api/core/v1alpha1/etcd.go
```diff
@@ -341,6 +345,19 @@ type EtcdConfig struct {
...
+	// additional URLs. The member name must follow the pattern {etcd-name}-{index}
+	// where index is 0 to (replicas-1) (e.g., etcd-main-0, etcd-main-1 etc).
+	// When spec.memberNamePrefix is set, member names become `<memberNamePrefix>-<podName>`.
+	// Updating this field on a running cluster triggers a ConfigMap update
+	// and a rolling restart of the StatefulSet.
+	// Note: this field only applies to pods managed by etcd-druid. For externally
+	// managed members (spec.externallyManagedMemberAddresses), the provided addresses
+	// are used directly and this field has no effect.
+	// +optional
+	AdditionalAdvertiseClientURLs *AdditionalClientURLsSpec `json:"additionalAdvertiseClientURLs,omitempty"`
```
Shreyas-s14, line 360: Just a suggestion, correct me if I am wrong. Can the api be restructured as per below
```go
additonalUrlsAdvertised {
      //optional
      OverrideDefaultUrls: true/false
      // optional
      clientUrls: {}
      //optional 
      peerUrls: {}
}
```
Is it mutually exclusive to have override between peer and client? i.e if peer url is overriden wont client url also be overriden?
This way we can have a more concise api imo
Since these fields are optional, the individual urls(client/peer) can also be advertised.

api/core/v1alpha1/etcd.go
```diff
@@ -341,6 +345,19 @@ type EtcdConfig struct {
...
+	// client service URL is suppressed. Each entry maps a member name to its
+	// additional URLs. The member name must follow the pattern {etcd-name}-{index}
+	// where index is 0 to (replicas-1) (e.g., etcd-main-0, etcd-main-1 etc).
+	// When spec.memberNamePrefix is set, member names become `<memberNamePrefix>-<podName>`.
+	// Updating this field on a running cluster triggers a ConfigMap update
+	// and a rolling restart of the StatefulSet.
+	// Note: this field only applies to pods managed by etcd-druid. For externally
+	// managed members (spec.externallyManagedMemberAddresses), the provided addresses
+	// are used directly and this field has no effect.
+	// +optional
```
Shreyas-s14, line 359: The member-name pattern ({etcd-name}-{index}, and the memberNamePrefix case) is already documented on MemberClientURLs.Name. Can we drop it from this field docstring to avoid duplication? This keeps it consistent with how the peer field is documented in #1408, where the pattern lives only on the member field.

api/core/v1alpha1/helper.go
```diff
@@ -77,6 +77,36 @@ func GetMemberNameFromAddress(etcd *Etcd, memberAddress string) string {
...
+	for _, memberURLs := range etcd.Spec.Etcd.AdditionalAdvertiseClientURLs.Members {
+		if memberURLs.Name == memberName {
+			return memberURLs.URLs, IsOverrideDefaultClientURLEnabled(etcd)
+		}
+	}
+	return nil, false
+}
```
Shreyas-s14, line 108: The same set of functions are present for the PeerUrls field as well. The logic is the same. IMO, the urltype can be added as an argument and then one set of functions with the same functionality as what is done can remain. WDYT

internal/component/configmap/etcdconfig.go
```diff
@@ -204,9 +204,25 @@ func getAdvertiseURLs(etcd *druidv1alpha1.Etcd, advertiseURLType, scheme, peerSv
...
+				urls = append(urls, defaultURL)
+			}
+			advUrlsMap[memberName] = urls
```
Shreyas-s14, line 219: This will cause a merge conflict wrt 1408 right? If so since the overall code block is very similar except variable changes, imo, adding up on https://github.com/gardener/etcd-druid/pull/1425#discussion_r3978104099, this can also be rephrased right?

Shreyas-s14 (commented):
Thanks for the PR. Have a few suggestions. PTAL

api/core/v1alpha1/etcd.go
anveshreddy18: Yeah I too agree with this with a slight modification of having overrides individually instead as they can be configured independently.

```go
additonalUrlsAdvertised {
      // optional
      clientUrls: {
         //optional
         OverrideDefaultUrl: true/false
         ...
      }
      //optional
      peerUrls: {
        //optional
        OverrideDefaultUrl: true/false
        ...
      }
}
```

This makes the structure better and as @Shreyas-s14 mentioned can consolidate the helper funcs for both peer & client urls.

seshachalam-yv:
Closing this PR to consolidate the changes in #1438, as suggested in the review.

The API has now been unified for advertising both client and peer URLs, including the `overrideDefaultURL` handling, in a single implementation.
/close

PR #1426, opened 2026-08-31
DEP-08:  multi node scale-in support

seshachalam-yv:

/area high-availability
/area scalability
/area control-plane-migration
/kind enhancement
/kind api-change

Implements [DEP-08](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md): quorum-safe scale-in of multi-node etcd clusters, and declarative decommissioning of source members joined via `bootstrapWithExistingCluster`.

Previously, reducing `spec.replicas` would immediately resize the underlying StatefulSet without first verifying quorum safety, and surplus PVCs were silently leaked. This PR replaces that with a one-member-at-a-time removal loop: `StatefulSet.PreSync` removes one etcd member per reconcile cycle (only while the cluster stays quorate), then `Sync` shrinks the StatefulSet and deletes surplus PVCs in the same pass. A PVC-deletion-before-shrink deadlock present in earlier versions of this work (where the PVC deletes prematurely requeued, preventing the StatefulSet from ever shrinking, which prevented the `pvc-protection` finalizer from releasing) has been fixed.

Scale operations are now tracked by a new `ScaleOperationComplete` status condition with reasons `ScalingIn`, `ScalingOut`, `BootstrapMembersRemoval`, and `NoScaleOperation`. CEL admission rules on the CRD use this condition to reject an opposite-direction concurrent scale change while an operation is in flight.

Source members joined via `spec.etcd.bootstrapWithExistingCluster` can be decommissioned declaratively by removing their entries from the spec (or unsetting the field entirely). etcd-druid removes them one per reconcile cycle using the same quorum-safe removal path, tracked by reason `BootstrapMembersRemoval`.

Fixes #
Part of #1423 #1237 

**Checklist**:
        - [x] Update [Table of Contents](https://github.com/gardener/etcd-druid/blob/master/docs/README.md) for the docs.
    - [x] Unit tests
    - [x] Integration tests
    - [x] E2E tests

- https://github.com/gardener/etcd-backup-restore/pull/1058
- https://github.com/gardener/etcd-druid/pull/1425

> **Condition renamed and polarity inverted from the DEP-08 draft.**
> The proposal originally defined this condition as `ScaleOperationInProgress` with `True` meaning an operation is in flight and `False` meaning idle. The implementation uses **`ScaleOperationComplete`** with **inverted polarity**: `True` (reason `NoScaleOperation`) is the converged/idle state, `False` (with a reason naming the operation) marks an in-flight operation. This brings it in line with positive-polarity conventions used by all other `Etcd` status conditions (`Ready`, `AllMembersReady`, `BackupReady`). **The DEP document has been updated in this PR to reflect the new name and polarity throughout.**

> | Type | Status | Reason | Meaning |
> |---|---|---|---|
> | `ScaleOperationComplete` | `True` | `NoScaleOperation` | Idle — no scale work in progress |
> | `ScaleOperationComplete` | `False` | `ScalingIn` | Scale-in in progress |
> | `ScaleOperationComplete` | `False` | `ScalingOut` | Scale-out in progress |
> | `ScaleOperationComplete` | `False` | `BootstrapMembersRemoval` | Removing source members joined via `bootstrapWithExistingCluster` |

**Release note**:

```feature operator
Operators can now safely scale in HA etcd clusters by reducing `spec.replicas`. During scale-in, etcd-druid removes surplus etcd members **one at a time**, performing a quorum-safety check before each removal. The StatefulSet is shrunk only after the corresponding member has been safely decommissioned, and surplus PVCs are cleaned up as part of the same operation, preventing both quorum loss and storage leaks.

A new `ScaleOperationComplete` status condition provides visibility into scaling operations, indicating whether a scale operation is in progress (`False/ScalingIn`, `False/ScalingOut`, or `False/BootstrapMembersRemoval`) or the cluster is idle (`True/NoScaleOperation`). CRD-level CEL validation also prevents conflicting scale operations while another membership operation is in progress.

Members from the source cluster configured through `spec.etcd.bootstrapWithExistingCluster` can now be decommissioned declaratively by removing them from the specification or clearing the field. These members are removed through the same quorum-safe member removal flow under the `BootstrapMembersRemoval` reason.

See [DEP-08](https://github.com/gardener/etcd-druid/blob/master/docs/proposals/08-scale-in.md) for more details.
```


internal/client/etcd/fake/fake.go
```diff
@@ -0,0 +1,134 @@
+// SPDX-FileCopyrightText: Contributors to the Gardener project
```
CaptainIRS, line 1: Could we use mocks instead of fakes? Since I think it will be useful to see if the API surface has changed during etcd version upgrades. I suspect it will also simplify the codebase a bit (excluding the parts auto-generated with mockgen).

internal/component/memberlease/memberlease.go
```diff
@@ -158,6 +163,61 @@ func (r _resource) doDelete(ctx component.OperatorContext, objectKey client.Obje
 	return nil
 }
 
+// deleteSurplusLeases skips deletion when replicas=0 to avoid wiping
+// member lease state needed to wake a hibernated cluster.
+func (r _resource) deleteSurplusLeases(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, objectKeys []client.ObjectKey) error {
```
CaptainIRS, line 168: Could this be consolidated with the deleteStaleMemberLeases function above?

internal/component/memberlease/memberlease.go
```diff
@@ -158,6 +163,61 @@ func (r _resource) doDelete(ctx component.OperatorContext, objectKey client.Obje
...
+	}
+	var errs error
+	for _, err := range utils.RunConcurrently(ctx, tasks) {
+		errs = multierror.Append(errs, err)
+	}
+	return errs
+}
+
+// doDeleteLease deletes a single surplus member lease and logs on success.
+func (r _resource) doDeleteLease(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd, objKey client.ObjectKey) error {
```
CaptainIRS, line 201: It seems to be the same as doDelete above, could we remove this helper?

api/core/v1alpha1/etcd.go
```diff
@@ -66,6 +66,11 @@ const (
...
+// Scale coordination rules: reject conflicting opposite-direction membership changes while a scale operation is in flight. See docs/proposals/08-scale-in.md.
+// +kubebuilder:validation:XValidation:message="Cannot scale out while a scale-in or bootstrap members removal is in progress.",rule="(self.spec.replicas > oldSelf.spec.replicas && self.spec.replicas > 0 && oldSelf.spec.replicas > 0) ? (!has(self.status) || !has(self.status.conditions) || !self.status.conditions.exists(c, c.type == 'ScaleOperationComplete' && c.status == 'False' && (c.reason == 'ScalingIn' || c.reason == 'BootstrapMembersRemoval'))) : true"
+// +kubebuilder:validation:XValidation:message="Cannot scale in while a scale-out or bootstrap members removal operation is in progress.",rule="(self.spec.replicas < oldSelf.spec.replicas && self.spec.replicas > 0 && oldSelf.spec.replicas > 0) ? (!has(self.status) || !has(self.status.conditions) || !self.status.conditions.exists(c, c.type == 'ScaleOperationComplete' && c.status == 'False' && (c.reason == 'ScalingOut' || c.reason == 'BootstrapMembersRemoval'))) : true"
```
CaptainIRS, line 71: I think this CEL validation is not effective as etcd.Spec.Replicas == stsReplicas is used to set the condition to True (default case in the switch statement). So, as soon as the sts replicas becomes n+1, this CEL validation would admit the request - which I don't think is what we intend. I think we should reconsider when the condition is being set to True during scale-out (maybe on AllMembersReady perhaps?) or drop this validation and handle it in controller logic. Also, allowing scale-in while bootstrap members removal is in progress seems reasonable as the controller can handle it gracefully (from what I understand), so is there a reason we disallow it?

api/core/v1alpha1/etcd.go
```diff
@@ -66,6 +66,11 @@ const (
...
+// +kubebuilder:validation:XValidation:message="Cannot remove bootstrap members while scale-in or scale-out is in progress.",rule="has(oldSelf.spec.etcd.bootstrapWithExistingCluster) && has(self.spec.etcd.bootstrapWithExistingCluster) && self.spec.etcd.bootstrapWithExistingCluster.members != oldSelf.spec.etcd.bootstrapWithExistingCluster.members ? (!has(self.status) || !has(self.status.conditions) || !self.status.conditions.exists(c, c.type == 'ScaleOperationComplete' && c.status == 'False' && (c.reason == 'ScalingIn' || c.reason == 'ScalingOut'))) : true"
```
CaptainIRS, line 72: This rule seems redundant given the above 2 rules. Is there a specific case that this rule is covering that isn't already covered by the above?

api/core/v1alpha1/helper.go
```diff
@@ -105,6 +105,41 @@ func GetMemberLeaseNames(etcd *Etcd) []string {
 	}
 }
 
+// GetBootstrapMemberNames returns the set of member names declared in
+// spec.etcd.bootstrapWithExistingCluster.members (empty when unset).
+func GetBootstrapMemberNames(etcd *Etcd) map[string]bool {
```
CaptainIRS, line 110: The function name suggests that a slice is being returned, but we're returning a map[string]bool. Could we instead return a slice and do slices.Contains in the call site?

api/core/v1alpha1/etcd.go
```diff
@@ -614,6 +618,23 @@ const (
 	// to True once the target has successfully joined the existing cluster and remains
 	// sticky-True thereafter, surviving transient member outages.
 	ConditionTypeBootstrappedWithExistingCluster ConditionType = "BootstrappedWithExistingCluster"
+	// ConditionTypeScaleOperationComplete indicates whether the etcd cluster is free of an in-flight
+	// scale operation (scale-in, scale-out, or bootstrap-member removal). Positive polarity: True when
+	// no operation is in progress, False while one is in flight (with the reason naming it).
```
CaptainIRS, line 623: Is there a specific reason we want positive polarity as opposed to the earlier proposed ScaleOperationInProgress, since we do already have conditions with negative polarity like the clusterID mismatch condition?

internal/controller/etcd/reconcile_scale.go
```diff
@@ -0,0 +1,238 @@
...
+// The mapping (proposal step "Detection (Step 3)"):
+//
+//   - spec.replicas < StatefulSet.spec.replicas          -> False, ScalingIn
+//   - spec.replicas > StatefulSet.spec.replicas          -> False, ScalingOut
+//   - bootstrap members removed / bootstrap config unset -> False, BootstrapMembersRemoval
+//   - no scale signal                                    -> True,  NoScaleOperation
+//
+// The condition is set back to True (complete) on successful completion by
+// recordReconcileSuccessOperation.
+func (r *Reconciler) detectAndRecordScaleOperation(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd) ctrlutils.ReconcileStepResult {
```
CaptainIRS, line 39: Should we detect this case mentioned in the DEP:

> The backstop here is the `etcd-druid` controller: scale-in detection re-derives the target set on every reconcile and removes the surplus member again under the per-cycle quorum-safety check, so a member restarting with a wiped data directory cannot persist in the cluster after a scale-in.

as well when recording a scale operation? Since, if this happens, currently scale-in is not recorded, but preSync does member removal.

internal/controller/etcd/reconcile_scale.go
```diff
@@ -0,0 +1,238 @@
...
+	if etcd.Spec.Replicas == 0 || stsReplicas == 0 {
+		return druidv1alpha1.ConditionTrue, druidv1alpha1.ScaleOperationReasonNoScaleOperation
+	}
+
+	switch {
+	case etcd.Spec.Replicas < stsReplicas:
+		return druidv1alpha1.ConditionFalse, druidv1alpha1.ScaleOperationReasonScalingIn
+	case etcd.Spec.Replicas > stsReplicas:
+		return druidv1alpha1.ConditionFalse, druidv1alpha1.ScaleOperationReasonScalingOut
+	default:
```
CaptainIRS, line 100: I think we should use ready replicas/AllMembersReady for this check since otherwise even before the member fully comes up, it shows that the scale operation is complete. (CEL validations will not be effective in that case)

internal/component/memberlease/memberlease.go
```diff
@@ -67,27 +67,32 @@ func (r _resource) GetExistingResourceNames(ctx component.OperatorContext, etcdO
 	return resourceNames, nil
 }
 
-// PreSync is a no-op for the member lease component.
 func (r _resource) PreSync(_ component.OperatorContext, _ *druidv1alpha1.Etcd) error { return nil }
 
-// Sync creates or updates the member leases for the given Etcd.
+// Sync creates or updates member leases and removes surplus leases after a scale-in.
```
CaptainIRS, line 72: Could you please add unit test for the new cases handled?

internal/utils/kubernetes/tls.go
```diff
@@ -0,0 +1,202 @@
...
+		cfg.Certificates = []tls.Certificate{cert}
+	}
+
+	return "https", cfg, nil
+}
+
+// BuildBackupRestoreCATLSConfig builds a CA-only *tls.Config for the
+// backup-restore HTTP server. The CA is resolved from the Secret mounted under
+// the "backup-restore-ca" volume. Returns (nil, nil) when the volume is absent.
+func BuildBackupRestoreCATLSConfig(ctx context.Context, cl client.Client, sts *appsv1.StatefulSet, namespace, caDataKey string) (*tls.Config, error) {
```
CaptainIRS, line 175: This seems unused anywhere, so could it be removed?

internal/utils/kubernetes/tls.go
```diff
@@ -0,0 +1,202 @@
...
+		return tls.Certificate{}, false, fmt.Errorf("failed to build client key pair from secret %s/%s: %w", namespace, secretName, err)
+	}
+	return cert, true, nil
+}
+
+// BuildEtcdClientTLSConfig builds an mTLS *tls.Config for the etcd client service.
+// The CA is resolved from the Secret under the "etcd-ca" volume (caDataKey defaults
+// to "ca.crt"). A client keypair is attached when the "etcd-client-tls" volume is
+// present. Returns (nil, nil) when the CA volume is absent.
+func BuildEtcdClientTLSConfig(ctx context.Context, cl client.Client, sts *appsv1.StatefulSet, namespace, caDataKey string) (*tls.Config, error) {
```
CaptainIRS, line 90: This seems unused other than tests, so could be removed.

internal/utils/kubernetes/tls.go
```diff
@@ -0,0 +1,202 @@
...
+	if !caPool.AppendCertsFromPEM(caData) {
+		return nil, fmt.Errorf("%w %s/%s", ErrAppendCACerts, namespace, secretName)
+	}
+	return caPool, nil
+}
+
+// LoadClientKeyPairFromVolumeSecret resolves the Secret mounted under volumeName on sts,
+// fetches it, and builds a tls.Certificate from certDataKey and keyDataKey.
+// Returns (tls.Certificate{}, false, nil) when sts is nil or the volume is absent.
+func LoadClientKeyPairFromVolumeSecret(ctx context.Context, cl client.Client, sts *appsv1.StatefulSet, namespace, volumeName, certDataKey, keyDataKey string) (tls.Certificate, bool, error) {
```
CaptainIRS, line 66: This seems to be used in a function that isn't used anywhere, so could be removed.

internal/utils/kubernetes/tls.go
```diff
@@ -0,0 +1,202 @@
...
+	// ErrMissingDataKey indicates the requested data key is absent from the Secret.
+	ErrMissingDataKey = errors.New("data key not found in secret")
+	// ErrAppendCACerts indicates the CA bundle could not be parsed into a cert pool.
+	ErrAppendCACerts = errors.New("failed to append CA certs from secret")
+)
+
+// LoadCAPoolFromVolumeSecret resolves the Secret mounted under volumeName on sts,
+// fetches it, and builds an x509.CertPool from the bytes stored under caDataKey.
+// Returns (nil, nil) when sts is nil or the volume is absent.
+func LoadCAPoolFromVolumeSecret(ctx context.Context, cl client.Client, sts *appsv1.StatefulSet, namespace, volumeName, caDataKey string) (*x509.CertPool, error) {
```
CaptainIRS, line 47: This seems to be used in a function that isn't used anywhere, so could be removed.

internal/utils/kubernetes/statefulset.go
```diff
@@ -60,6 +61,27 @@ func GetStatefulSet(ctx context.Context, cl client.Client, etcd *druidv1alpha1.E
...
+func ComputeScaleInReplicaDelta(ctx context.Context, cl client.Client, etcd *druidv1alpha1.Etcd) (int32, error) {
+	if etcd.Spec.Replicas == 0 {
+		return 0, nil
+	}
+	sts, err := GetStatefulSet(ctx, cl, etcd)
+	if err != nil {
+		return 0, err
+	}
+	if sts == nil {
+		return 0, nil
```
CaptainIRS, line 76: Should we return an error here?

internal/utils/kubernetes/statefulset.go
```diff
@@ -60,6 +61,27 @@ func GetStatefulSet(ctx context.Context, cl client.Client, etcd *druidv1alpha1.E
 	return nil, nil
 }
 
+// ComputeScaleInReplicaDelta returns the number of StatefulSet replicas that
+// exceed the desired Etcd replica count. Returns 0 when the StatefulSet is
+// absent, not yet larger than desired, or when replicas==0 (hibernation).
+func ComputeScaleInReplicaDelta(ctx context.Context, cl client.Client, etcd *druidv1alpha1.Etcd) (int32, error) {
```
CaptainIRS, line 67: Unit tests could be added for this.

internal/utils/kubernetes/lease.go
```diff
@@ -34,8 +34,11 @@ func isPeerURLTLSEnabledForMembers(ctx context.Context, cl client.Client, logger
 	if err != nil {
 		return false, err
 	}
-	targetMembers := leaseObjMetaSlice[:replicas]
-	for _, leaseObjMeta := range targetMembers {
+	// Use min(len, replicas) to guard against a stale STS replica count that
+	// exceeds the number of leases that currently exist (e.g. during scale-up
+	// before all member leases have been created).
+	n := min(int(replicas), len(leaseObjMetaSlice))
```
CaptainIRS, line 40: I wonder if this is practically applicable, since we sync member lease before StatefulSet.

internal/utils/kubernetes/lease.go
```diff
@@ -72,8 +79,6 @@ func ListAllMemberLeaseObjectMeta(ctx context.Context, cl client.Client, etcd *d
 	); err != nil {
 		return nil, err
 	}
-	// This OK to do as we do not support downscaling an etcd cluster.
-	// If and when we do that by then we should have already stabilised the labels and therefore this code itself will not be there.
 	allPossibleMemberNames := druidv1alpha1.GetMemberLeaseNames(etcd)
 	leasesObjMeta := make([]metav1.PartialObjectMetadata, 0, len(objMetaList.Items))
 	for _, lease := range objMetaList.Items {
```
CaptainIRS, line 84: I think we can remove this code as stated by the earlier comment, WDYT?

test/e2e/controller/etcd_test.go
```diff
@@ -368,12 +368,28 @@ func TestScaleOut(t *testing.T) {
...
 				logger.Info("successfully created Etcd")
+				// A single-member cluster starts with exactly one member PVC.
+				testEnv.CheckEtcdPVCCount(g, etcd, 1, timeoutEtcdCreation)
 
 				logger.Info("scaling out Etcd to 3 replicas")
 				etcd.Spec.Replicas = 3
 				updateEtcdTLSAndLabels(etcd, true, tc.peerTLSEnabledAfterScaleOut, true, tc.additionalLabelsAfterScaleOut)
 				testEnv.UpdateAndCheckEtcd(g, etcd, timeoutEtcdUpdation)
 				logger.Info("successfully scaled out Etcd to 3 replicas")
+				// A scaled-out cluster must have one PVC per member.
```
CaptainIRS, line 379: This should be a separate test since this test is to specifically test scale-out. Also, I think te verification should be a bit robust - checking the condition, probing the member list with etcd client, etc. so that we check if any newly introduced logic results in incorrect member removal, ghost members, etc. The checks should be done right after the condition is true, to check the working of the condition as well.

internal/utils/kubernetes/tls_test.go
```diff
@@ -0,0 +1,355 @@
...
+// TestBuildEtcdClientTLSConfig verifies the mTLS config assembly for the etcd client:
+// the CA is resolved from the etcd-ca volume, an optional client keypair from the
+// etcd-client-tls volume, and absence of the CA volume/StatefulSet yields (nil, nil).
+func TestBuildEtcdClientTLSConfig(t *testing.T) {
+	t.Parallel()
+	caPEM, err := testutils.GenerateCACert("etcd-main")
+	if err != nil {
+		t.Fatalf("failed to generate CA cert: %v", err)
+	}
+	clientCertPEM, clientKeyPEM, err := testutils.GenerateClientCert("etcd-client")
```
CaptainIRS: Just fake strings could be used here. So, the helper can be removed as well.

internal/controller/etcd/reconcile_scale.go
```diff
@@ -0,0 +1,238 @@
...
+		return druidv1alpha1.ConditionTrue, druidv1alpha1.ScaleOperationReasonNoScaleOperation
+	}
+}
+
+// isBootstrapMembersRemoval reports whether the operator has requested removal
+// of source members the target already joined via bootstrapWithExistingCluster.
+// It is true when the target has recorded joined members in status and one or
+// more of those members is no longer present in the spec (or the spec's
+// bootstrapWithExistingCluster has been unset entirely).
+func isBootstrapMembersRemoval(etcd *druidv1alpha1.Etcd) bool {
```
CaptainIRS, line 110: nit:
```suggestion
func hasBootstrapMembersToBeRemoved(etcd *druidv1alpha1.Etcd) bool {
```

test/it/crdvalidation/etcd/specupdate_test.go
```diff
@@ -101,11 +101,13 @@ func TestValidateUpdateSpecReplicas(t *testing.T) {
 			expectErr:       false,
 		},
 		{
-			name:            "Invalid update to replicas #1",
-			etcdName:        "etcd-invalid-dec",
+			// DEP-08: a decrease of a multi-node cluster is now allowed when no
+			// conflicting scale operation is in progress.
```
CaptainIRS, line 105: We can remove this comment as this is just a normal supported operation in druid after this PR, hence we don't need a specific reference.

test/it/crdvalidation/etcd/specscale_test.go
```diff
@@ -0,0 +1,166 @@
...
+		{
+			name:            "Valid: scale-in while ScalingIn in progress (same direction)",
+			initialReplicas: 5,
+			updatedReplicas: 3,
+			conditionStatus: druidv1alpha1.ConditionFalse,
+			conditionReason: druidv1alpha1.ScaleOperationReasonScalingIn,
+			expectErr:       false,
+		},
+		{
+			name:            "Valid: scale-in to zero is never blocked",
```
CaptainIRS, line 91: Just wondering, should we disallow hibernation when scale-in/scale-out is in progress? Since that could lead to inconsistent state.

test/it/crdvalidation/etcd/specscale_test.go
```diff
@@ -0,0 +1,166 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and Gardener contributors
```
CaptainIRS, line 1: Header has to be updated.

internal/client/etcd/memberclient.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	return &memberClient{api: cli, closer: cli.Close}, nil
+}
+
+// memberAPI is the narrow subset of clientv3 that memberClient uses.
+// *clientv3.Client satisfies this interface.
+type memberAPI interface {
+	MemberList(ctx context.Context) (*clientv3.MemberListResponse, error)
+	MemberRemove(ctx context.Context, id uint64) (*clientv3.MemberRemoveResponse, error)
+	Status(ctx context.Context, endpoint string) (*clientv3.StatusResponse, error)
```
CaptainIRS, line 92: I'm not sure about putting these 3 methods in memberAPI, since from etcd perspective, MemberList and MemberRemove are part of the Cluster client and Status is part of the Maintenance client. (If we use mocks instead of fakes, I think this will be a lot clearer and aligned with etcd APIs)

internal/client/etcd/memberclient.go
```diff
@@ -0,0 +1,254 @@
...
+	defer cancel()
+
+	resp, err := c.api.MemberList(listCtx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list etcd members: %w", err)
+	}
+
+	members := c.buildMembersFromList(resp.Members)
+	healthyURLs := c.probeAndSetHealth(ctx, resp.Members, members)
+	c.resolveAndSetLeader(ctx, members, healthyURLs)
```
CaptainIRS, line 118: Is there a reason we have 2 separate steps for it? I think this could just be just consolidated into one step safely, I don't see any issues with that, WDYT?

internal/client/etcd/memberclient.go
```diff
@@ -0,0 +1,254 @@
...
+// defaultStatusTimeout, derived from the caller's ctx so cancellation still
+// propagates. Returns true on the first success; false if all URLs error
+// (fail-closed).
+func (c *memberClient) probeMemberHealthy(ctx context.Context, urls []string) bool {
+	for _, url := range urls {
+		if url == "" {
+			continue
+		}
+		statusCtx, statusCancel := context.WithTimeout(ctx, defaultStatusTimeout)
+		_, err := c.api.Status(statusCtx, url)
```
CaptainIRS, line 196: I think this could be problematic as the member can respond to status API calls even though it is down (maybe a get call is a better alternative?) - for example when an NOSPACE/CORRUPT alarm is active, or even when there is no quorum in the cluster.

internal/client/etcd/memberclient_test.go
```diff
@@ -0,0 +1,369 @@
...
+		{
+			name:      "typed member-not-found is treated as success",
+			removeErr: rpctypes.ErrMemberNotFound,
+			wantErr:   false,
+		},
+		{
+			name:      "gRPC NotFound status is treated as success",
+			removeErr: grpcstatus.Error(codes.NotFound, "etcdserver: member not found"),
+			wantErr:   false,
+		},
```
CaptainIRS, line 318: Could this case happen in etcd? My understanding is that the gRPC NotFound is for cases where the RPC endpoint itself is not found.

internal/component/statefulset/memberremoval.go
```diff
@@ -0,0 +1,150 @@
...
+			fmt.Sprintf("member %s not removed: removal would break quorum for etcd %v; waiting for the cluster to become healthy",
+				candidate.Name, client.ObjectKeyFromObject(etcd)))
+	}
+
+	return r.removeAndRequeue(ctx, etcd, &candidate, mc)
+}
+
+// surplusMemberNames returns the union of scale-in surplus (ordinals >= spec.replicas
+// when STS > desired) and bootstrap source members to decommission.
+func (r _resource) surplusMemberNames(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd) (map[string]bool, error) {
```
CaptainIRS, line 81: Due to how the anti-rejoin guard works, we can't guarantee that the information from the CR and StatefulSet is sufficient to determine surplus. I think the fool-proof way would be to calculate expected members from CR/sts and get the actual members from the etcd cluster to calculate the diff.

internal/component/statefulset/statefulset_test.go
```diff
@@ -399,6 +401,74 @@ func TestSyncWhenNoSTSExists(t *testing.T) {
 	}
 }
 
+// TestSyncScaleInShrinksStatefulSet is the regression guard for the DEP-08
+// scale-in deadlock. A scale-in Sync must, in a single pass, delete the surplus
+// PVCs (ordinals >= spec.replicas) AND shrink the StatefulSet to spec.replicas.
+// The earlier bug requeued after issuing the PVC deletes, so the StatefulSet was
+// never shrunk: the surplus pods that held the pvc-protection finalizer were
+// never removed, the PVCs stayed Terminating, and the reconcile requeued forever.
```
CaptainIRS, line 409: Could you clarify on what is the earlier bug being referenced here?

internal/controller/etcd/reconcile_spec.go
```diff
@@ -76,9 +88,12 @@ func (r *Reconciler) syncEtcdResources(ctx component.OperatorContext, etcd *drui
 	for _, kind := range resourceOperators {
 		op := r.operatorRegistry.GetOperator(kind)
 		if err := op.Sync(ctx, etcd); err != nil {
-			if derr := druiderr.AsDruidError(err); derr != nil && derr.Code == druiderr.ErrRequeueAfter {
-				ctx.Logger.Info("retrying sync of component", "kind", kind, "syncRetryInterval", syncRetryInterval.String(), "reason", derr.Message)
-				return ctrlutils.ReconcileAfter(syncRetryInterval, fmt.Sprintf("retrying sync of component %s after %s", kind, syncRetryInterval.String()))
+			if derr := druiderr.AsDruidError(err); derr != nil {
+				switch derr.Code {
+				case druiderr.ErrRequeueAfter:
```
CaptainIRS, line 93: This could be reverted.

internal/controller/etcd/reconcile_scale.go
```diff
@@ -0,0 +1,238 @@
...
+
+	sts, err := kubernetes.GetStatefulSet(ctx, r.client, etcd)
+	if err != nil {
+		ctx.Logger.Error(err, "failed to get StatefulSet while detecting scale operation; preserving existing scale operation condition")
+		return existingScaleOperationConditionOrDefault(etcd)
+	}
+	// No StatefulSet yet (fresh cluster) means there is no observed size to
+	// compare against, so there is no scale operation in progress.
+	if sts == nil || sts.Spec.Replicas == nil {
+		return druidv1alpha1.ConditionTrue, druidv1alpha1.ScaleOperationReasonNoScaleOperation
```
CaptainIRS, line 85: I think we should consider cluster creation also as scale-out. Otherwise the CEL rules will be ineffective till the cluster is fully bootstrapped.

internal/controller/etcd/reconcile_spec.go
```diff
@@ -60,9 +63,18 @@ func (r *Reconciler) preSyncEtcdResources(ctx component.OperatorContext, etcd *d
...
+					ctx.Logger.Info("retrying pre-sync of component", "kind", kind, "syncRetryInterval", syncRetryInterval.String(), "reason", derr.Message)
+					return ctrlutils.ReconcileAfter(syncRetryInterval, fmt.Sprintf("requeueing pre-sync of component %s to be retried after %s", kind, syncRetryInterval.String()))
+				case statefulset.ErrQuorumUnsafeMemberRemoval:
+					// ErrQuorumUnsafeMemberRemoval originates only from the StatefulSet
+					// PreSync member-removal branch (removeOneSurplusMember); this is its
+					// sole handler. Requeue with the error recorded in status.lastErrors.
+					ctx.Logger.Info("holding pre-sync of component", "kind", kind, "syncRetryInterval", syncRetryInterval.String(), "reason", derr.Message)
+					return ctrlutils.ReconcileWithErrorAfter(syncRetryInterval, err)
+				}
```
CaptainIRS, line 77: From what I understand, this is already handled in the existing code, so wondering if we need this special case handling here.

internal/component/statefulset/statefulset.go
```diff
@@ -89,6 +94,10 @@ func (r _resource) GetExistingResourceNames(ctx component.OperatorContext, etcdO
 
 // PreSync performs pre-sync operations for the statefulset component.
 func (r _resource) PreSync(ctx component.OperatorContext, etcd *druidv1alpha1.Etcd) error {
+	if err := r.removeOneSurplusMember(ctx, etcd); err != nil {
```
CaptainIRS, line 97: Could we name this something like `ensureSurplusMembersAreRemoved` so that it is a bit clear from the flow here? That way the err return signifies that we haven't ensured that all surplus members are removed yet, so we want to requeue.

internal/component/statefulset/memberremoval.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if !etcdmember.QuorumSafeToRemove(members, candidate.ID) {
+		ctx.Logger.Info("holding scale-in: removing the selected member would break quorum",
+			"member", candidate.Name, "memberID", hexID(candidate.ID))
+		return druiderr.New(ErrQuorumUnsafeMemberRemoval, component.OperationPreSync,
+			fmt.Sprintf("member %s not removed: removal would break quorum for etcd %v; waiting for the cluster to become healthy",
+				candidate.Name, client.ObjectKeyFromObject(etcd)))
+	}
+
+	return r.removeAndRequeue(ctx, etcd, &candidate, mc)
```
CaptainIRS, line 76: I think if we figure out that the candidate is a leader (i.e there is no other option), we could potentially have someting like r.moveLeaderAndRequeue which voluntarily hands over the leader to the other replicas if the resulting cluster size is more than 1. WDYT?

test/it/crdvalidation/etcd/specbootstrap_test.go
```diff
@@ -175,6 +175,84 @@ func TestValidateUpdateSpecBootstrapWithExistingClusterFreezeWhileInProgress(t *
...
+			{
+				Type:               druidv1alpha1.ConditionTypeScaleOperationComplete,
+				Status:             status,
+				LastTransitionTime: metav1.Now(),
+				LastUpdateTime:     metav1.Now(),
+				Reason:             reason,
+				Message:            "test message",
+			},
+		}
+		g.Expect(cl.Status().Update(ctx, etcd)).To(Succeed())
```
CaptainIRS, line 201: This could be placed within Eventually to avoid flakes due to update conflicts.

go.mod
```diff
@@ -5,16 +5,19 @@ go 1.26.0
 require (
 	github.com/Masterminds/semver/v3 v3.5.0
 	github.com/gardener/etcd-druid/api v0.0.0
-	github.com/gardener/etcd-druid/client v0.0.0
+	github.com/gardener/etcd-druid/client v0.0.0-00010101000000-000000000000
```
CaptainIRS, line 8: Is there a reason why this is changed?

go.mod
```diff
@@ -96,3 +107,9 @@ tool (
 	golang.org/x/tools/cmd/goimports
 	sigs.k8s.io/controller-runtime/tools/setup-envtest
 )
+
+replace go.etcd.io/etcd/api/v3 => go.etcd.io/etcd/api/v3 v3.5.27
+
+replace go.etcd.io/etcd/client/v3 => go.etcd.io/etcd/client/v3 v3.5.27
+
+replace go.etcd.io/etcd/client/pkg/v3 => go.etcd.io/etcd/client/pkg/v3 v3.5.27
```
CaptainIRS, line 115: Could we just pin the dependencies directly without replace directives? Would that cause problems?

api/core/v1alpha1/etcd.go
```diff
@@ -66,6 +66,11 @@ const (
...
+// +kubebuilder:validation:XValidation:message="Cannot remove bootstrap members while scale-in or scale-out is in progress.",rule="has(oldSelf.spec.etcd.bootstrapWithExistingCluster) && has(self.spec.etcd.bootstrapWithExistingCluster) && self.spec.etcd.bootstrapWithExistingCluster.members != oldSelf.spec.etcd.bootstrapWithExistingCluster.members ? (!has(self.status) || !has(self.status.conditions) || !self.status.conditions.exists(c, c.type == 'ScaleOperationComplete' && c.status == 'False' && (c.reason == 'ScalingIn' || c.reason == 'ScalingOut'))) : true"
+// +kubebuilder:validation:XValidation:message="Cannot unset bootstrapWithExistingCluster while scale-in or scale-out is in progress.",rule="has(oldSelf.spec.etcd.bootstrapWithExistingCluster) && !has(self.spec.etcd.bootstrapWithExistingCluster) ? (!has(self.status) || !has(self.status.conditions) || !self.status.conditions.exists(c, c.type == 'ScaleOperationComplete' && c.status == 'False' && (c.reason == 'ScalingIn' || c.reason == 'ScalingOut'))) : true"
```
CaptainIRS, line 73: I think we should also prevent removal of entries from bootstrapWithExistingCluster like we disallow unsetting during scale-in or scale-out. WDYT?

internal/component/statefulset/statefulset_test.go
```diff
@@ -399,6 +401,74 @@ func TestSyncWhenNoSTSExists(t *testing.T) {
...
+	// (handleTLSChanges is a no-op when the STS already matches the etcd spec).
+	etcd := testutils.EtcdBuilderWithDefaults(testutils.TestEtcdName, testutils.TestNamespace).
+		WithReplicas(initialReplicas).
+		Build()
+	cl := testutils.CreateTestFakeClientForObjects(nil, nil, nil, nil, []client.Object{buildBackupSecret()}, getObjectKey(etcd.ObjectMeta))
+	operator := New(cl, iv, &etcdfake.MemberClientFactory{Client: &etcdfake.MemberClient{}})
+	opCtx := component.NewOperatorContext(context.Background(), logr.Discard(), uuid.NewString())
+	opCtx.Data[common.CheckSumKeyConfigMap] = testutils.TestConfigMapCheckSum
+
+	g.Expect(operator.Sync(opCtx, etcd)).ToNot(HaveOccurred())
```
CaptainIRS, line 431:
```suggestion
	g.Expect(operator.Sync(opCtx, etcd)).To(Succeed())
```

internal/component/statefulset/statefulset_test.go
```diff
@@ -399,6 +401,74 @@ func TestSyncWhenNoSTSExists(t *testing.T) {
...
+	// Request a scale-in and mark the in-flight condition the detector would have set.
+	etcd.Spec.Replicas = targetReplicas
+	etcd.Status.Conditions = []druidv1alpha1.Condition{{
+		Type:   druidv1alpha1.ConditionTypeScaleOperationComplete,
+		Status: druidv1alpha1.ConditionFalse,
+		Reason: druidv1alpha1.ScaleOperationReasonScalingIn,
+	}}
+
+	// A single Sync must shrink the StatefulSet AND delete the surplus PVCs.
+	g.Expect(operator.Sync(opCtx, etcd)).ToNot(HaveOccurred())
```
CaptainIRS, line 451:
```suggestion
	g.Expect(operator.Sync(opCtx, etcd)).To(Succeed())
```

internal/component/statefulset/statefulset_test.go
```diff
@@ -399,6 +401,74 @@ func TestSyncWhenNoSTSExists(t *testing.T) {
...
+	pvcExists := func(ordinal int32) bool {
+		pvcName := fmt.Sprintf("%s-%s-%d", vctName, stsName, ordinal)
+		getErr := cl.Get(context.Background(), client.ObjectKey{Namespace: etcd.Namespace, Name: pvcName}, &corev1.PersistentVolumeClaim{})
+		return getErr == nil
+	}
+	for _, ordinal := range []int32{3, 4} {
+		g.Expect(pvcExists(ordinal)).To(BeFalse(), "surplus PVC for ordinal %d must be deleted during scale-in", ordinal)
+	}
+	for _, ordinal := range []int32{0, 1, 2} {
+		g.Expect(pvcExists(ordinal)).To(BeTrue(), "retained PVC for ordinal %d must not be deleted", ordinal)
```
CaptainIRS, line 468: `Expect(err).To(MatchError(apierrors.IsNotFound, "IsNotFound"))` could be done to ensure other errors don't get accepted.

internal/etcd/selection_test.go
```diff
@@ -0,0 +1,219 @@
...
+			wantIDs: []uint64{20, 30},
+		},
+		{
+			name: "multiple learners ordered by ascending member ID within the learner tier",
+			candidates: []Member{
+				{ID: 30, Name: "learner-2", Role: MemberRoleLearner, Health: MemberHealthHealthy},
+				{ID: 10, Name: "learner-0", Role: MemberRoleLearner, Health: MemberHealthHealthy},
+				{ID: 20, Name: "learner-1", Role: MemberRoleLearner, Health: MemberHealthHealthy},
+			},
+			wantIDs: []uint64{10, 20, 30},
```
CaptainIRS, line 193: I think this could be removed as more than 1 learner is not supported.

internal/etcd/selection_test.go
```diff
@@ -0,0 +1,219 @@
...
+package etcd
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+// voter builds a healthy voting member.
+func voter(id uint64, name string) Member {
```
CaptainIRS, line 14:
```suggestion
func healthyVoter(id uint64, name string) Member {
```

CaptainIRS (changes requested):
@seshachalam-yv , thanks for the PR. I've added some comments inline, also I'd like to discuss some points:
- Currently the configmap checksum also includes the peer/client URLs in the calculation, which might lead to additional statefulsets than necessary. In case of scale-in from 3->2, this is more problematic as it leads to quorum loss for the duration of the statefulset rollout. Same is the case when removing bootstrap members -  So, I think we should remove the peer/client URLs from the checksum calculation. WDYT?
- Regarding e2e tests
    - I see that the scale-in is added as just a sub-step of the scale-out steps, but I think we should separate that out into a separate set of tests. (with + without TLS)
    - For bootstrap with existing members, I think we should also include the removal of bootstrap members in the e2e test.
- Regarding etcd client in etcd-druid, we create an etcd client per reconciliation. But I wonder if we can cache the client that is created for a particular etcd instance? This is to avoid future issues in etcd-druid due to excess amount of etcd clients being created like https://github.com/rancher/rke2/issues/10059.
- Regarding the upgrade feature flag - older versions of backup-restore will not have the anti-rejoin guard. So, should we consider disallowing scale-in in such cases?

internal/client/etcd/fake/fake.go
seshachalam-yv: The convention is documented at  ["Unit Tests" section](docs/development/testing.md) prefer the fake client over mocking, because "Mocks decrease maintainability because they expect the tested component to follow a certain way to reach the desired goal."

api/core/v1alpha1/helper.go
seshachalam-yv: The serialisation here is intentional — BootstrapMembersRemoval is a distinct operation with different semantics from ScalingIn/ScalingOut, even though both remove members. Rule 3 keeps them mutually exclusive for now to avoid complicating the state machine. If there's a concrete need to allow them concurrently in future, we can revisit at that point.

api/core/v1alpha1/etcd.go
seshachalam-yv: The other Etcd conditions (`Ready`, `AllMembersReady`) all use `True` as the healthy state, and `ScaleOperationComplete` follows the same pattern.

Scale operations are infrequent — `True` is the normal steady state, `False` is the exception. From a readability standpoint, False immediately draws attention, same as Ready=False meaning "something needs attention". Keeping the same polarity avoids confusion for anyone already familiar with the other conditions.

internal/client/etcd/memberclient.go
seshachalam-yv: The two steps serve different authorities and cannot safely be merged:

1. **K8s re-Get + surplus recompute** — checks the *desired* state: has `spec.replicas` been bumped back up (or `bootstrapWithExistingCluster` cleared) by a concurrent edit since we first read it? If yes, the candidate is no longer surplus and we abort. This is the TOCTOU guard — without it, a user who cancels a scale-in mid-flight by editing the spec could still have the member removed.

2. **`MemberList` re-list** — checks the *actual* etcd state: is the candidate still a live member? It may have already been removed (e.g. by a concurrent reconcile on another controller replica) since our initial list at the top of `ensureMemberRemoval`.

These two checks consult different sources (Kubernetes API server vs. etcd). A single `MemberList` can only answer question 2 — it cannot tell us whether the spec has changed. So they must stay as two separate calls.

go.mod
seshachalam-yv: `k8s.io/apiextensions-apiserver` `v0.35.5` is already a direct dependency of etcd-druid, and its `go.mod ` declares [`go.etcd.io/etcd/client/v3 v3.6.5`](https://github.com/kubernetes/apiextensions-apiserver/blob/v0.35.5/go.mod#L19-L20). So therefore  Go by default resolves `client/v3 to v3.6.5` when running go mod tidy, overriding our preference for `v3.5.27`.

internal/utils/kubernetes/lease.go
seshachalam-yv: That is the case this handles. Surplus member leases are deleted before the StatefulSet is synced, so during scale-in sts.replicas can be larger than the number of remaining leases. The min covers that gap: it keeps us checking only the members whose leases still exist, and stops us slicing past the leases we have.

internal/utils/kubernetes/lease.go
seshachalam-yv: https://github.com/gardener/etcd-druid/pull/1426#discussion_r4045037615

internal/component/memberlease/memberlease.go
seshachalam-yv: Yes I will consolidate.

internal/component/memberlease/memberlease.go
seshachalam-yv: Yes I will reuse.

api/core/v1alpha1/etcd.go
seshachalam-yv: Now  `determineScaleOperationInProgress` only sets the condition to False. True is set in recordScaleOperationComplete, which runs as step 7, after `preSyncEtcdResources` and `syncEtcdResources` have completed.

api/core/v1alpha1/etcd.go
seshachalam-yv: Rules 1 and 2 only fire on `spec.replicas changes`. An operator can remove source members by editing .`bootstrapWithExistingCluster.members` without touching `spec.replicas`, so neither rule would catch that. Rule 3 blocks that combination: editing .members while a `ScalingIn` or `ScalingOut` is in progress. We intentionally disallow running bootstrap member removal concurrently with a `spec.replicas` scale operation.

internal/etcd/selection_test.go
seshachalam-yv: Fixed.

test/it/crdvalidation/etcd/specscale_test.go
seshachalam-yv: Now I've added the CEL validation to prevent the scaling to zero replicas

internal/controller/etcd/reconcile_scale.go
seshachalam-yv: Now, I've added new [recordScaleOperationComplete](https://github.com/gardener/etcd-druid/pull/1426/changes/091484705362bd4c949aa1bc2b7e766e4123aa1b#diff-73e93276c994b8fde02d63293b745eb57ccec0c8152e63ae49a1946af694597aR114) which prevents this case by ensuring all the members are removed before setting the scalingoperationcompelete condition to true

internal/client/etcd/memberclient.go
seshachalam-yv: Whole etcd client package is refactored as part of this  f0c9e36d55d2598753f7dc1fec17ed624581c907.

internal/component/statefulset/memberremoval.go
seshachalam-yv: MemberRemove takes care of leader move internally

seshachalam-yv:
Thanks for the detailed review @CaptainIRS. Addressed all the points below.

Summary of changes

- Added recordScaleOperationComplete — sets ScaleOperationComplete=True at the end of reconciliation. For scale-in, it first queries the live member list and only flips the condition once all surplus members are gone.
- Refactored the etcd client package to mirror the upstream interface with a minimal surface. Health is checked per-member via a Get call rather than a status 
- Reused the existing memberlease cleanup to remove surplus leases during scale-in.

---

▎ Configmap checksum / StatefulSet rollout during scale-in

Fixed in 5e71f2f82 — peer and client URLs derived from spec.replicas are excluded from the checksum, so changing the replica count no longer triggers a StatefulSet rollout. The fix covers both scale-in and bootstrap members removal.

---

▎ e2e tests

Scale-in tests (with and without TLS) and bootstrap members removal end-to-end coverage are in d4601258f.
---

▎ etcd client caching

The etcd client is only created during scale-in, not on every reconciliation, so caching isn't needed here.

PTAL

api/core/v1alpha1/etcd.go
CaptainIRS: I'm not sure I understand. this. Even if I remove the rule and run the CRD validation tests, everything passes. Could you add some test that this case specifically covers in that case?

seshachalam-yv:
e2e test results — PROVIDERS=none, KIND, commit 5e71f2f82

```
--- PASS: TestBasic (6 subtests)
    --- PASS: TestBasic/basic-no-tls-0-none       (10.72s)
    --- PASS: TestBasic/basic-tls-0-none          (10.19s)
    --- PASS: TestBasic/basic-no-tls-1-none       (64.44s)
    --- PASS: TestBasic/basic-tls-1-none          (63.75s)
    --- PASS: TestBasic/basic-no-tls-3-none       (244.94s)
    --- PASS: TestBasic/basic-tls-3-none          (240.81s)

--- PASS: TestBootstrapWithExistingCluster (1 subtest)
    --- PASS: TestBootstrapWithExistingCluster/bootstrap-existing-none  (437.80s)

--- PASS: TestScaleOut (5 subtests)
    --- PASS: TestScaleOut/scaleout-basic-none                  (132.61s)
    --- PASS: TestScaleOut/scaleout-with-peer-tls-none          (132.32s)
    --- PASS: TestScaleOut/scaleout-with-label-change-none      (156.09s)
    --- PASS: TestScaleOut/scaleout-enable-peer-tls-none        (293.41s)
    --- PASS: TestScaleOut/scaleout-enable-ptls-label-change-none (313.94s)

--- PASS: TestScaleIn (3 subtests)
    --- PASS: TestScaleIn/scalein-3to1-none              (167.09s)
    --- PASS: TestScaleIn/scalein-3to2-none              (164.02s)
    --- PASS: TestScaleIn/scalein-5to3-zerodowntime-none (185.12s)

--- PASS: TestScaleRejectedDuringBootstrapMembersRemoval (1 subtest)
    --- PASS: TestScaleRejectedDuringBootstrapMembersRemoval/scale-reject-bootstrapmembersremoval-none (141.28s)

--- PASS: TestTLSAndLabelUpdates (6 subtests)
    --- PASS: TestTLSAndLabelUpdates/update1-enable-peer-none               (212.09s)
    --- PASS: TestTLSAndLabelUpdates/update1-label-change-none              (193.29s)
    --- PASS: TestTLSAndLabelUpdates/update1-enable-c-p-br-tls-none        (352.16s)
    --- PASS: TestTLSAndLabelUpdates/update1-enable-ptls-label-change-none  (211.43s)
    --- PASS: TestTLSAndLabelUpdates/update1-enable-c-p-br-label-change-none (355.39s)
    --- PASS: TestTLSAndLabelUpdates/update1-disable-c-br-tls-none         (254.33s)

--- PASS: TestRecovery (7 subtests)
    --- PASS: TestRecovery/recovery-1-del-1-pod-none          (69.79s)
    --- PASS: TestRecovery/recovery-1-corrupt-data-none       (76.53s)
    --- PASS: TestRecovery/recovery-3-del-1-pod-none          (164.36s)
    --- PASS: TestRecovery/recovery-3-del-2-pods-none         (252.16s)
    --- PASS: TestRecovery/recovery-3-del-3-pods-none         (241.78s)
    --- PASS: TestRecovery/recovery-3-corrupt-1-mem-none      (195.49s)
    --- PASS: TestRecovery/recovery-3-del-2-pods-corrupt-1-mem-none (264.39s)

--- PASS: TestClusterUpdate (4 subtests)
    --- PASS: TestClusterUpdate/update2-1-no-update-none  (45.72s)
    --- PASS: TestClusterUpdate/update2-1-update-none     (64.33s)
    --- PASS: TestClusterUpdate/update2-3-no-update-none  (140.53s)
    --- PASS: TestClusterUpdate/update2-3-update-none     (203.34s)

ok  github.com/gardener/etcd-druid/test/e2e/controller  1114.674s

40 subtests, 0 failures. Druid image: g5e71f2f82 (HEAD).
```

api/core/v1alpha1/etcd.go
CaptainIRS: Could you please add CRD validation tests for the new rule that was added for this?

api/core/v1alpha1/helper.go
```diff
@@ -148,14 +148,80 @@ func GetMemberLeaseNames(etcd *Etcd) []string {
...
+	for i, memberAddress := range memberAddresses {
+		leaseNames[i] = GetMemberNameFromAddress(etcd, memberAddress)
+	}
+	return leaseNames
+}
+
+// GetMemberLeaseNamesForReplicas returns member lease names for the given replica count,
+// ignoring etcd.Spec.Replicas. For externally-managed members the spec-based list is
+// always returned (the replica count argument is ignored for that path).
+func GetMemberLeaseNamesForReplicas(etcd *Etcd, replicas int32) []string {
```
CaptainIRS: I don't see this being used anywhere, perhaps it could be removed?

api/core/v1alpha1/helper.go
```diff
@@ -148,14 +148,80 @@ func GetMemberLeaseNames(etcd *Etcd) []string {
...
 	}
+	return GetMemberLeaseNames(etcd)
+}
+
+// ExpectedMemberNames returns the set of etcd member names the spec currently
+// expects: managed pod-ordinal members for ordinals [0, spec.replicas) plus
+// any bootstrap members declared in spec.etcd.bootstrapWithExistingCluster.
+// Both memberremoval (statefulset package) and the scale controller use this
+// to determine which live cluster members are surplus.
+func ExpectedMemberNames(etcd *Etcd) map[string]bool {
```
CaptainIRS: Unit tests are missing for this function.

api/core/v1alpha1/helper.go
CaptainIRS: Could you clarify on what the comment means? I don't get the context.

CaptainIRS: Unit tests could be added to this function.

api/core/v1alpha1/helper.go
```diff
@@ -188,6 +254,11 @@ func GetStatefulSetName(etcdObjMeta metav1.ObjectMeta) string {
 	return etcdObjMeta.Name
 }
 
+// GetClientPort returns the etcd client port, defaulting to 2379 when unset.
+func GetClientPort(etcd *Etcd) int32 {
```
CaptainIRS: Not sure if this helper is needed since there is 1 use. Could you move it inline?

api/core/v1alpha1/helper.go
```diff
@@ -266,3 +337,55 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
...
+// GetScaleOperationCompleteCondition returns the status and reason of the
+// ScaleOperationComplete condition, defaulting to ConditionTrue/NoScaleOperation
+// when the condition is not yet present. The default preserves the "no operation
+// in progress" view so a missing condition (or a transient inability to observe
+// the cluster) is never mistaken for an in-flight scale operation.
+func GetScaleOperationCompleteCondition(etcd *Etcd) (ConditionStatus, string) {
+	cond := GetCondition(etcd, ConditionTypeScaleOperationComplete)
+	if cond == nil {
+		return ConditionTrue, ScaleOperationReasonNoScaleOperation
+	}
```
CaptainIRS: This should return nil so that we don't add the condition if the cluster is never scaled anytime.

api/core/v1alpha1/helper.go
```diff
@@ -266,3 +337,55 @@ func RemoveOperationAnnotation(etcdObjMeta metav1.ObjectMeta) {
...
+// IsScaleInInProgress reports whether a scale-in operation is currently recorded
+// in the Etcd status (ScaleOperationComplete=False/ScalingIn).
+func IsScaleInInProgress(etcd *Etcd) bool {
+	cond := GetCondition(etcd, ConditionTypeScaleOperationComplete)
+	return cond != nil && cond.Status == ConditionFalse && cond.Reason == ScaleOperationReasonScalingIn
+}
+
+// HasZeroReplicas reports whether the Etcd is configured with zero replicas
+// (spec.replicas <= 0).
+func HasZeroReplicas(etcd *Etcd) bool {
```
CaptainIRS: This could be moved inline.

